PR #213, opened 2018-07-20
tests: extract common test code to internal package

shantuo:
This will allow the code to be shared with server tests against other platforms.


shantuo:
This PR is moved to #266

PR #218, opened 2018-07-24
Add refactor step to tutorial

zombiezen:
The refactor step further separates setup code from application code to
demonstrate a clear seam between the two and make any future changes to
the supported cloud providers straightforward.


zombiezen:
Fixed the `switch`: it was in the README but not the code.

PR #221, opened 2018-07-24
runtimevar: simplify context usage

oliverpool:
According to the [context documentation](https://golang.org/pkg/context/#Context):
>        // If Done is not yet closed, Err returns nil.
>        // If Done is closed, Err returns a non-nil error explaining why
So using ctx.Err() seems better than selecting the ctx.Done() channel (easier to read and probably a bit faster)


oliverpool:
(I signed it)

PR #240, opened 2018-07-26
blob: add modification time to driver.ObjectAttrs

benhinchley:
At [Sajari](https://github.com/sajari) we have been doing something similar to [blob](https://godoc.org/github.com/google/go-cloud/blob) with our own [storage](https://github.com/sajari/storage) pkg for quite some time now.

I started working on [migrating](https://github.com/sajari/storage/pull/5) our [storage](https://github.com/sajari/storage) pkg to use [go-cloud/blob](https://godoc.org/github.com/google/go-cloud/blob) underneath, and noticed that modification time was not a part of the [`driver.ObjectAttrs`](https://github.com/google/go-cloud/blob/master/blob/driver/driver.go#L64) struct. This is something that we use internally and is a part of our [`storage.File`](https://github.com/sajari/storage/blob/master/storage.go#L15) struct. 

This adds ModTime to [`driver.ObjectAttrs`](https://github.com/google/go-cloud/blob/master/blob/driver/driver.go#L64) and adds it to the current driver implementations.

### DIFF
--- a/blob/driver/driver.go
+++ b/blob/driver/driver.go
@@ -67,7 +67,7 @@ type ObjectAttrs struct {
 	Size int64
 	// ContentType is the MIME type of the blob object. It must not be empty.
 	ContentType string
-	// ModTime is the modified time of the blob object.
+	// ModTime is the modified time of the blob object. Will be time.Time zero value if unknown.
 	ModTime time.Time
 }
 
--- a/blob/driver/driver.go
+++ b/blob/driver/driver.go
@@ -19,6 +19,7 @@ package driver
 import (
 	"context"
 	"io"
+	"time"
 )
 
 // ErrorKind is a code to indicate the kind of failure.
@@ -66,6 +67,8 @@ type ObjectAttrs struct {
 	Size int64
 	// ContentType is the MIME type of the blob object. It must not be empty.
 	ContentType string
+	// ModTime is the modified time of the blob object.
+	ModTime time.Time
 }
 
 // Bucket provides read, write and delete operations on objects within it on the
--- a/blob/fileblob/fileblob.go
+++ b/blob/fileblob/fileblob.go
@@ -30,6 +30,7 @@ import (
 	slashpath "path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/google/go-cloud/blob"
 	"github.com/google/go-cloud/blob/driver"
@@ -98,8 +99,9 @@ func (b *bucket) NewRangeReader(ctx context.Context, key string, offset, length 
 	}
 	if length == 0 {
 		return reader{
-			size: info.Size(),
-			xa:   xa,
+			size:    info.Size(),
+			modTime: info.ModTime(),
+			xa:      xa,
 		}, nil
 	}
 	f, err := os.Open(path)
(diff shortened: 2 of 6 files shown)


benhinchley:
I signed it!

blob/gcsblob/gcsblob.go
```diff
@@ -65,11 +68,18 @@ func (r *reader) Attrs() *driver.ObjectAttrs {
 func (b *bucket) NewRangeReader(ctx context.Context, key string, offset, length int64) (driver.Reader, error) {
 	bkt := b.client.Bucket(b.name)
 	obj := bkt.Object(key)
+	attrs, err := obj.Attrs(ctx)
```
zombiezen: It's unfortunate that this requires a second network call, but it was bound to happen eventually.

My one concern is that there is a race here: `Attrs` could be for a different version than the reader is for. In this case, there's not that much cause for concern, but I'd like to add the constraint now so that as the code evolves, we don't hit the issue later.

blob/driver/driver.go
```diff
@@ -66,6 +67,8 @@ type ObjectAttrs struct {
 	Size int64
 	// ContentType is the MIME type of the blob object. It must not be empty.
 	ContentType string
+	// ModTime is the modified time of the blob object.
```
zombiezen: I can imagine circumstances where implementations might not have this information. Let's add a comment to the effect that `ModTime` might be the zero time if unknown.

blob/fileblob/fileblob.go
```diff
@@ -148,6 +152,7 @@ func (r reader) Attrs() *driver.ObjectAttrs {
 	return &driver.ObjectAttrs{
 		Size:        r.size,
 		ContentType: r.xa.ContentType,
+		ModTime:     r.modTime,
```
zombiezen: Please also add tests for this new functionality for `fileblob`.

blob/gcsblob/gcsblob.go
shantuo: One alternative is to not call `obj.Attrs` inside `NewRangeReader`, instead, call it when user explicitly asks for it the first time. At that point we can check whether this field is non-zero and decide to make the call. WDYT?

blob/driver/driver.go
benhinchley: Updated comment in [c04be8e](https://github.com/google/go-cloud/pull/240/commits/c04be8e0a626e3709d36f16a7466fa4549969636).

blob/fileblob/fileblob.go
benhinchley: Add test in [7e530ee](https://github.com/google/go-cloud/pull/240/commits/7e530ee85c8ebea2c231011465279c1042b402a4)

zombiezen:
@benhinchley Is this ready for @shantuo's review?

benhinchley:
@zombiezen yep changes you suggested have been made

blob/gcsblob/gcsblob.go
shantuo: The [updated field](https://godoc.org/google.golang.org/api/storage/v1#Object.Updated) is included in the object but not carried by the client library we use here. I suggest removing the implementation here for GCS and keep those for S3 and file, add comments to the blob package saying that this ModTime field is optional. We'll try to have it supported by the client library first and add it back here.

benhinchley:
@shantuo I've update the gcsblob impl and added the comment to `blob.ModTime()` about this possibly returning `time.Time` zero value

PR #247, opened 2018-07-26
travis: remove disabled varcheck and add build-tags

### LINKED ISSUE #69: travis: golint-ci fails build due to "unused" provider sets in package main
While testing a PR, I noticed that Travis flagged a Wire provider set as an unused variable ([source](https://travis-ci.com/google/go-cloud/builds/75304110#L661)):

```
samples/guestbook/main.go:269:5: `appSet` is unused (varcheck)
var appSet = wire.NewSet(
    ^
```

This is technically correct, but not an actual issue, and it did cause the build to fail, thus blocking submission.  Based on recommendation from @cflewis, I'm disabling `varcheck` as a temporary workaround, then keeping this open for solving the underlying issue.

sendilkumarn:
cleanup: remove namingRuleURL

closes #69

### DIFF
--- a/blob/gcsblob/gcsblob.go
+++ b/blob/gcsblob/gcsblob.go
@@ -105,8 +105,6 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
 	return err
 }
 
-const namingRuleURL = "https://cloud.google.com/storage/docs/naming"
-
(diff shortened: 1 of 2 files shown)


sendilkumarn:
Thanks @zombiezen 👍

PR #249, opened 2018-07-26
travis: add wire check step

### LINKED ISSUE #30: wire: add check command
gowire should have a `check` command that verifies that all of the provider sets declared in a package are valid (e.g. `NewSet` arguments are well-formed, no cycles).  This wouldn't entail new logic, but is blocked by #29.

When this is added, we should add a step to our Travis builds that runs `gowire check` across all packages.

zombiezen:
Fixes #30

### DIFF
--- a/internal/testing/wirecheck.sh
+++ b/internal/testing/wirecheck.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


internal/testing/wirecheck.sh
```diff
@@ -0,0 +1,31 @@
...
+
+if [[ $# -gt 0 ]]; then
+  echo "usage: wirecheck.sh" 1>&2
+  exit 64
+fi
+
+module="github.com/google/go-cloud"
+vgo mod -vendor || exit 1
+mapfile -t all_pkgs < <( vgo list "$module/..." ) || exit 1
+wire check "${all_pkgs[@]}" || exit 1
```
ijt: This mapfile pattern seems a bit indirect. Why not just write

wire check $(vgo list "$module/...") || exit 1

?

zombiezen: It's to address the hypothetical case of an import path containing spaces. This is the approach `shellcheck` output recommends.

PR #261, opened 2018-07-30
Some docstring/comment wording changes. Also, return nil Reader/Writer on errors. 

vangent:


blob/blob.go
```diff
@@ -26,8 +26,8 @@ import (
 	"github.com/google/go-cloud/blob/driver"
 )
 
-// Reader implements io.ReadCloser to read from blob. It must be closed after
-// reads finished. It also provides the size of the blob.
+// Reader implements io.ReadCloser to read a blob. It must be closed after
+// reads are finished. It also provides the Size and ContentType of the blob.
```
shantuo: Since ContentType and Size are methods and not fields in Reader, I suggest removing the last sentence.

blob/blob.go
```diff
@@ -148,11 +151,15 @@ func (b *Bucket) NewRangeReader(ctx context.Context, key string, offset, length
 	if offset < 0 {
 		return nil, errors.New("new blob range reader: offset must be non-negative")
 	}
-	r, err := b.b.NewRangeReader(ctx, key, offset, length)
-	return &Reader{r: r}, newBlobError(err)
+	var r driver.Reader
```
shantuo: These changes seem unnecessary to me.

blob/blob.go
```diff
@@ -172,8 +179,10 @@ func (b *Bucket) NewWriter(ctx context.Context, key string, opt *WriterOptions)
 				return nil, err
 			}
 			ct := mime.FormatMediaType(t, p)
-			w, err = b.b.NewTypedWriter(ctx, key, ct, dopt)
-			return &Writer{w: w}, err
+			if w, err = b.b.NewTypedWriter(ctx, key, ct, dopt); err != nil {
```
shantuo: Same here.

blob/blob.go

blob/blob.go
vangent: OK, removed.

blob/blob.go

Issue #214, opened 2018-07-20
all: default should be to run only fast tests, opt in for slow and complicated tests


ijt:
At present, running `vgo test ./...` at the root takes 1m20s and outputs lots of failures if its dependencies have not been set up.

The current way to work around that is to add the `-short` flag mentioned in `CONTRIBUTING.md`, but it's easy to forget and seems like the reverse of how things should be arranged. Also, that still doesn't filter out the wire tests, and those take 40s to run on my laptop.

I suggest we [use build tags](https://stackoverflow.com/questions/25965584/separating-unit-tests-and-integration-tests-in-go) to allow opting in to running progressively larger and more complicated tests. We could use the usual [Google test size conventions](https://testing.googleblog.com/2010/12/test-sizes.html) for these. There would probably need to also be an additional, implicit size of `tiny` for the many test files that run in under 1s or so. Running only those tests would then be the default that happens when running `vgo test ./...`.

PR #262, opened 2018-07-30
all: use -record instead of -short to turn on recording in tests

ijt:
This partially addresses issue #214.


blob/gcsblob/gcsblob_integration_test.go
```diff
@@ -24,11 +24,12 @@ import (
 	"github.com/google/go-cloud/gcp"
 )
 
+var record = flag.Bool("record", false, "whether to record")
```
shantuo: Group these two together and the description should be more clear.

ijt: Done.

Issue #164, opened 2018-07-03
wire: error out if wire.Build has any unused arguments


zombiezen:
h/t @neild for the suggestion.

Wire can know during generation if `wire.Build` has any arguments that are unused.  Much like unused imports, including something unused should be made into a `gowire` error to prevent dependency bloat.

PR #268, opened 2018-07-31
wire: wire.Build now returns an error if it has any unused arguments

vangent:


wire/internal/wire/testdata/UnusedInputs/foo/foo.go
```diff
@@ -0,0 +1,54 @@
+// Copyright 2018 Google LLC
```
zombiezen: (For the directory, not this particular file/line:) Wire uses the term "input" to mean arguments passed to the injector. A better name for this test would be `UnusedProviders`.

wire/internal/wire/testdata/UnusedInputs/foo/wire.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/google/go-cloud/wire"
+)
+
+func injectBar() Bar {
+	wire.Build(
+		provideFoo,                            // needed as input for provideBar
+		provideBar,                            // needed for Bar
+		provideUnused,                         // not needed -> error
+		wire.Value("unused"),                  // not needed -> error
+		wire.NewSet(provideUnusedInSet),       // nothing in set is needed -> error
```
zombiezen: Nice, this covers this quite thoroughly! Two more cases to check:

1. A set that is partially used (i.e. contains at least one provider that is used and at least one provider that is not) should not produce an error.
2. A set that is fully used should not produce an error.

wire/internal/wire/analyze.go
```diff
@@ -226,14 +230,71 @@ dfs:
...
+	if len(ec.errors) > 0 {
+		return nil, ec.errors
+	}
 	return calls, nil
 }
 
-// buildProviderMap creates the providerMap field for a given provider set.
-// The given provider set's providerMap field is ignored.
-func buildProviderMap(fset *token.FileSet, hasher typeutil.Hasher, set *ProviderSet) (*typeutil.Map, []error) {
+func verifyArgsUsed(set *ProviderSet, used []*providerSetSrc, ec *errorCollector) {
```
zombiezen: Instead of taking in an `*errorCollector`, return `[]error`.

zombiezen:
(Fixes #164)

wire/internal/wire/analyze.go
zombiezen: Add a doc comment for this function.

wire/internal/wire/testdata/UnusedInputs/foo/foo.go
vangent: Done.

wire/internal/wire/analyze.go
vangent: Done.

vangent: Done.

vangent:
> The srcMap is being built for every provider set, where it only needs to be built for the top-level.

Hmm, any tips on how to detect this? By "top level", do you mean "only provider sets that are created for wire.Build, not for wire.NewSet"? 

Maybe add a boolean "retainSources" or some such to processNewSet?

> ... the salient information is giving the output type of bindings and values, since that will help the user more quickly identify what needs to be removed.

Done.

wire/internal/wire/analyze.go
```diff
@@ -230,14 +230,15 @@ dfs:
...
+	if errs := verifyArgsUsed(set, used); len(errs) > 0 {
+		return nil, errs
 	}
 	return calls, nil
 }
 
-func verifyArgsUsed(set *ProviderSet, used []*providerSetSrc, ec *errorCollector) {
+// verifyArgsUsed ensures that all of the arguments in set were used during solve.
+func verifyArgsUsed(set *ProviderSet, used []*providerSetSrc) []error {
+	var errs []error
```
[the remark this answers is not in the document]
vangent: Ack. Doesn't seem necessary to me.

wire/internal/wire/testdata/UnusedProviders/out.txt
```diff
@@ -0,0 +1,5 @@
+ERROR
+unused provider set
```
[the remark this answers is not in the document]
vangent: Do provider sets have names? imp.Name seems to be the empty string in my testcase. wire.NewSet(...) doesn't have any way to set a name, does it?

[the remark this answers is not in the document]
zombiezen: Provider sets defined as package variables should have this property, but it's possible this is broken.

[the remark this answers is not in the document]
vangent: OK. I will include the name if it's present and look into why it's the empty string in my test separately.

vangent:
Per offline discussion, I think it's better to just always generate srcMap; it may be useful for other things like https://github.com/google/go-cloud/issues/165, and adding making it conditional adds a bunch of conditionals and "this field is sometimes set and sometimes nil" complexity.

wire/internal/wire/testdata/UnusedProviders/out.txt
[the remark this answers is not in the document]
vangent: Done.

zombiezen:
Cool. Thanks for opening #277.

PR #282, opened 2018-08-02
wire: respect -record flag for tests

### LINKED ISSUE #214: all: default should be to run only fast tests, opt in for slow and complicated tests

### LINKED ISSUE #281: wire: turn tests into replay tests
@vangent had an excellent suggestion about how to speed up Wire tests: save a copy of the "golden" code generation and add a record mode.

Currently, the Wire tests are split into `Generate` (runs `go build`) and `Determinism` (runs Wire twice and compares for equality). @vangent observed the similarity between this and our other replay tests, and suggested to merge this into one test: `Generate`, which by default would read from the golden file. It can be passed a `-record` flag to create the golden file, which will be run against `go build` to ensure that the output matches the expected output.

ijt:
Fixes #281 
Fixes #214 (enough for now)

```
ijt:~/gopath/src/github.com/google/go-cloud/wire/internal/wire$ time vgo test -record .
PASS
ok      github.com/google/go-cloud/wire/internal/wire   27.177s

real    0m29.352s
user    0m48.305s
sys     0m23.626s
ijt:~/gopath/src/github.com/google/go-cloud/wire/internal/wire$ time vgo test .
ok      github.com/google/go-cloud/wire/internal/wire   9.175s

real    0m11.971s
user    0m17.902s
sys     0m4.846s
```


zombiezen:
@vangent, want to take first pass?

wire/internal/wire/wire_test.go
```diff
@@ -59,88 +61,109 @@ func TestWire(t *testing.T) {
...
-		if _, err := os.Stat(filepath.Join(build.Default.GOROOT, "bin", "go")); err != nil {
-			t.Skip("go toolchain not available:", err)
-		}
-		for _, test := range tests {
-			test := test
-			t.Run(test.name, func(t *testing.T) {
-				t.Parallel()
+	// The golden directory has a dot prefix so its contents won't be found
+	// by go test.
+	goldenDir := ".golden"
```
vangent: Would it also work to put it inside testdata? E.g., "testdata/golden/...". Or even in the same directory where the testcase is defined, e.g., "testdata/MyTest/golden.go"?

wire/internal/wire/wire_test.go
```diff
@@ -59,88 +61,109 @@ func TestWire(t *testing.T) {
...
-					}
-					if !test.wantError {
-						t.Fatal("Did not expect errors.")
+	// goldenPath constructs the path to the golden output for the test
+	// with the given name.
+	goldenPath := func(name string) string {
+		return filepath.Join(goldenDir, name) + ".go"
+	}
+
+	if *setup.Record {
```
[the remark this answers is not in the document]
vangent: Sorry, forgot a couple of steps:
-- The test always verifies the errors returned by Generaet().
-- When Record is true, it also does the Go build and exec.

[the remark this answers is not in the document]
ijt: Done. PTAL.

wire/internal/wire/wire_test.go
```diff
@@ -59,116 +61,108 @@ func TestWire(t *testing.T) {
...
 	wd := filepath.Join(magicGOPATH(), "src")
 
-	t.Run("Generate", func(t *testing.T) {
-		if _, err := os.Stat(filepath.Join(build.Default.GOROOT, "bin", "go")); err != nil {
-			t.Skip("go toolchain not available:", err)
-		}
-		for _, test := range tests {
-			test := test
-			t.Run(test.name, func(t *testing.T) {
-				t.Parallel()
```
vangent: Why not Parallel() anymore?

wire/internal/wire/wire_test.go
```diff
@@ -59,116 +61,108 @@ func TestWire(t *testing.T) {
...
-				realBuildCtx := &build.Context{
-					GOARCH:      bctx.GOARCH,
-					GOOS:        bctx.GOOS,
-					GOROOT:      bctx.GOROOT,
-					GOPATH:      gopath,
-					CgoEnabled:  bctx.CgoEnabled,
-					Compiler:    bctx.Compiler,
-					BuildTags:   bctx.BuildTags,
-					ReleaseTags: bctx.ReleaseTags,
+					t.Fatalf("Failed to read golden file: %v", err)
```
vangent: Consider adding something like "If this is a new testcase, run with --record to generate the golden file."

wire/internal/wire/wire_test.go
```diff
@@ -59,116 +61,108 @@ func TestWire(t *testing.T) {
...
 				}
-				if err := runGo(realBuildCtx, "build", "-o", testExePath, genPkg.ImportPath); err != nil {
-					t.Fatal("build:", err)
+				if !bytes.Equal(gen, gold) {
+					t.Fatalf("wire output differs from golden file:\n%s", diff(string(gold), string(gen)))
```
vangent: Same here, consider adding something like "If this change is expected, run with --record to update the golden file."

wire/internal/wire/wire_test.go
```diff
@@ -59,116 +61,108 @@ func TestWire(t *testing.T) {
...
-				if err != nil {
-					t.Fatal(err)
+			goldenDir := filepath.Join("testdata", test.name, ".golden")
+			if err := os.MkdirAll(goldenDir, 0777); err != nil {
+				t.Fatalf("Failed to make golden dir: %v", err)
+			}
+			goldenFile := filepath.Join(goldenDir, "wire_gen.go")
+			if *setup.Record {
+				// Record ==> Check wire output with go build, and save golden file if it looks good.
+				goBuildCheck(t, test, wd, bctx, gen)
```
vangent: Shouldn't goBuildCheck return an error, and check it here with t.Fatalf? (and remove t as an argument).

wire/internal/wire/wire_test.go
```diff
@@ -59,116 +61,108 @@ func TestWire(t *testing.T) {
...
+				if err := ioutil.WriteFile(goldenFile, gen, 0666); err != nil {
+					t.Fatalf("Failed to write golden file: %v", err)
 				}
-
-				// Run a `go build` with the generated output.
-				gopath, err := ioutil.TempDir("", "wire_test")
+			} else {
+				// Replay ==> Load golden file and compare to generated result.
```
vangent: Since determinism was a concern in the previous version, maybe add a comment here noting that if Generate isn't deterministic, the replay test will become flaky and catch that (although the --record run won't).

wire/internal/wire/wire_test.go
```diff
@@ -59,116 +61,108 @@ func TestWire(t *testing.T) {
...
+			if len(errs) == 0 && test.wantError {
+				t.Fatal("wirego succeeded; want error")
+			}
 
-				// Find the absolute import path, since test.pkg may be a relative
-				// import path.
-				genPkg, err := bctx.Import(test.pkg, wd, build.FindOnly)
-				if err != nil {
-					t.Fatal(err)
+			goldenDir := filepath.Join("testdata", test.name, ".golden")
```
vangent: WDYT about loading the "golden" file in loadTestData below? And updating the comments for that function to include the new golden file.

vangent: There's also an existing "parseGoldenFile" func that should probably be renamed, as it's now a bit confusing since it's not about the golden file you're adding here.

wire/internal/wire/wire_test.go
ijt: It complicates the output and doesn't provide much of a speedup, at least on my system. I would guess it's because the tests hit the disk a lot.

wire/internal/wire/wire_test.go
ijt: Done.

wire/internal/wire/wire_test.go
ijt: Done.

wire/internal/wire/wire_test.go
ijt: Done.

wire/internal/wire/wire_test.go
ijt: Done.

wire/internal/wire/wire_test.go
vangent: OK.

wire/internal/wire/wire_test.go
ijt: I updated loadTestData as suggested, and moved the out.txt files into the .golden dirs. I didn't rename the `parseGoldenOutput` function since I think the name is accurate, but I renamed some fields for greater clarity.

wire/internal/wire/wire_test.go
```diff
@@ -309,9 +320,15 @@ func loadTestCase(root string, wireGoSrc []byte) (*testCase, error) {
 	if err != nil {
 		return nil, fmt.Errorf("load test case %s: %v", name, err)
 	}
+	var wireGold []byte
 	wantErrorStrings, wantError := parseGoldenOutput(out)
```
vangent: Suggest renamiong "parseGoldenOutput" to "parseWantOutput" or something like that, since "golden" means something else now.

zombiezen:
Taking a look now...

ijt:
Done. PTAL.

wire/internal/wire/wire_test.go
```diff
@@ -54,121 +56,113 @@ func TestWire(t *testing.T) {
...
-				if err != nil {
-					t.Fatal(err)
+				for _, s := range test.wantWireErrorStrings {
+					if !errorListContains(errs, s) {
+						t.Errorf("Errors did not contain %q", s)
+					}
 				}
+				return
+			}
+			if len(errs) == 0 && test.wantWireError {
```
vangent: Nit: this could be "if test.wantWireError"; len(errs) will always be 0 here.

wire/internal/wire/wire_test.go
```diff
@@ -284,40 +279,64 @@ func isIdent(s string) bool {
...
+	var wantWireOutput []byte
+	wireErrb, err := ioutil.ReadFile(filepath.Join(root, "want", "wire_errs.txt"))
+	wantWireError := err == nil
+	var wantWireErrorStrings []string
+	if wantWireError {
+		wantWireErrorStrings = strings.Split(strings.TrimSpace(string(wireErrb)), "\n")
+	} else {
+		wantWireOutput, err = ioutil.ReadFile(filepath.Join(root, "want", "wire_gen.go"))
+		if err != nil {
+			return nil, fmt.Errorf("load test case %s: %v. If this is a new testcase, run with -record to generate the wire_gen.go file.", name, err)
```
zombiezen: Wouldn't this fail everything new under record?

zombiezen:
Overall, direction LGTM! Thanks @ijt and @vangent!

wire/internal/wire/wire_test.go
```diff
@@ -54,121 +56,114 @@ func TestWire(t *testing.T) {
...
-					if !test.wantError {
-						t.Fatal("Did not expect errors.")
-					}
-					for _, s := range test.wantErrorStrings {
-						if !errorListContains(errs, s) {
-							t.Errorf("Errors did not contain %q", s)
-						}
-					}
-					return
+	if _, err := os.Stat(filepath.Join(build.Default.GOROOT, "bin", "go")); err != nil {
```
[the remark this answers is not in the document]
ijt: Done.

wire/internal/wire/wire_test.go
```diff
@@ -54,121 +56,114 @@ func TestWire(t *testing.T) {
...
+		t.Skip("go toolchain not available:", err)
+	}
+	for _, test := range tests {
+		test := test
```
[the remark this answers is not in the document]
ijt: Done.

wire/internal/wire/wire_test.go
```diff
@@ -54,121 +56,114 @@ func TestWire(t *testing.T) {
...
-				if err := test.materialize(gopath); err != nil {
-					t.Fatal(err)
-				}
-				if len(gen) > 0 {
-					genPath := filepath.Join(gopath, "src", filepath.FromSlash(genPkg.ImportPath), "wire_gen.go")
-					if err := ioutil.WriteFile(genPath, gen, 0666); err != nil {
-						t.Fatal(err)
-					}
+			if *setup.Record {
+				// Record ==> Check wire output with go build,
```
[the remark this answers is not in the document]
ijt: Done.

ijt:
Updated. PTAL.

wire/internal/wire/wire_test.go
ijt: Good point. Fixed it.

Issue #216, opened 2018-07-23
all: create GitHub contribution linter


zombiezen:
It would be useful to have a GitHub bot that enforces/auto-suggests some of the non-code things in CONTRIBUTING.md. In particular:

- [ ] Issue title format
- [ ] PR title format
- [ ] Commit message format (one line summary, "Fixes" on last line of description)
- [ ] License headers in new files
- [ ] Branches are in a fork, not on the main repository

This can use the [checks API](https://developer.github.com/changes/2018-05-07-new-checks-api-public-beta/) to present the information as well as some limited support for auto-fixing, like the license headers.

zombiezen:
I'm going to set up the authentication and deployment parts of this, which aren't blocked on go-github.

tbpg:
google/go-github#913 is fixed/merged.

zombiezen:
For this issue, I am planning to create a minimal event worker server that will remove the "In Progress" label from any closed issues. I will open issues for the other desired checks.

PR #283, opened 2018-08-02
internal/contributebot/webhook: add App Engine listener

zombiezen:
Includes Terraform configuration for adding the Pub/Sub topic.

Updates #216

### DIFF
--- a/internal/contributebot/README.md
+++ b/internal/contributebot/README.md
@@ -1,0 +1,5 @@
+# Contribute Bot
+
+Contribute Bot is a small service (written using Go Cloud!) that performs
+automated checks on issues and pull requests to help keep contributions
+organized and easy to triage for maintainers.
--- a/internal/contributebot/README.md
+++ b/internal/contributebot/README.md
@@ -3,3 +3,13 @@
 Contribute Bot is a small service (written using Go Cloud!) that performs
 automated checks on issues and pull requests to help keep contributions
 organized and easy to triage for maintainers.
+
+Contribute Bot has two servers: a webhook endpoint and an event listener. The
+webhook endpoint publishes events to a Cloud Pub/Sub topic that are eventually
+processed by the event listener. GitHub has a [10 second webhook response time
+limit][github-async] combined with a [5000 request/hour API rate
+limit][github-ratelimit], so this adds buffering with the assumption that
+incoming events are bursty.
+
+[github-async]: https://developer.github.com/v3/guides/best-practices-for-integrators/#favor-asynchronous-work-over-synchronous
+[github-ratelimit]: https://developer.github.com/v3/#rate-limiting
--- a/internal/contributebot/webhook/README.md
+++ b/internal/contributebot/webhook/README.md
@@ -3,8 +3,11 @@
 Contribute Bot Webhook Listener is a microservice that listens for
 [GitHub webhook][] events and publishes them to a [Cloud Pub/Sub][] topic.
 
-[GitHub webhook]: https://developer.github.com/webhooks/
+You will need the [Go App Engine SDK][] to develop this server.
+
 [Cloud Pub/Sub]: https://cloud.google.com/pubsub/docs/publisher
+[GitHub webhook]: https://developer.github.com/webhooks/
+[Go App Engine SDK]: https://cloud.google.com/appengine/docs/standard/go/download
 
 ## Running Locally
 
--- a/internal/contributebot/webhook/README.md
+++ b/internal/contributebot/webhook/README.md
@@ -1,0 +1,24 @@
+# Contribute Bot Webhook Listener
+
+Contribute Bot Webhook Listener is a microservice that listens for
+[GitHub webhook][] events and publishes them to a [Cloud Pub/Sub][] topic.
+
+[GitHub webhook]: https://developer.github.com/webhooks/
+[Cloud Pub/Sub]: https://cloud.google.com/pubsub/docs/publisher
+
+## Running Locally
+
+```shell
+cp app.yaml.in app.yaml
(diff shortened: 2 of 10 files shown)


internal/contributebot/webhook/main.go
```diff
@@ -0,0 +1,119 @@
...
+		return
+	}
+	w.Header().Set("Content-Length", fmt.Sprint(len(responseData)))
+	w.Header().Set("Content-Type", "text/html; charset=utf-8")
+	if r.Method == "GET" {
+		io.WriteString(w, responseData)
+	}
+}
+
+func handleWebhook(w http.ResponseWriter, r *http.Request) {
```
ijt: What does this do? Please add a comment to describe this function's behavior.

vangent: +1 to some comments about what's going on.

internal/contributebot/webhook/main.go
```diff
@@ -0,0 +1,119 @@
...
+	if r.Method != "POST" {
+		w.Header().Set("Allow", "POST")
+		http.Error(w, "Only POST allowed", http.StatusMethodNotAllowed)
+		return
+	}
+	ctx := appengine.NewContext(r)
+	payload, err := github.ValidatePayload(r, []byte(os.Getenv("CONTRIBUTEBOT_WEBHOOK_SECRET")))
+	if err != nil {
+		log.Errorf(ctx, "%v", err)
+		http.Error(w, "Could not connect to PubSub", http.StatusInternalServerError)
```
vangent: This error message doesn't look right. The log.Errorf should also have some more detail. Maybe a helper function that logs an error and also calls http.Error so you don't have to repeat the messages?

ijt: Alternatively, the handler could return an error and there could be a wrapper function that logs and calls http.Error.

internal/contributebot/webhook/README.md
```diff
@@ -0,0 +1,24 @@
+# Contribute Bot Webhook Listener
+
+Contribute Bot Webhook Listener is a microservice that listens for
+[GitHub webhook][] events and publishes them to a [Cloud Pub/Sub][] topic.
+
+[GitHub webhook]: https://developer.github.com/webhooks/
+[Cloud Pub/Sub]: https://cloud.google.com/pubsub/docs/publisher
+
+## Running Locally
```
shantuo: Mention GCP & App Engine SDK is required.

internal/contributebot/webhook/main.go
```diff
@@ -0,0 +1,119 @@
...
+	"context"
+	"encoding/base64"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+
+	"github.com/google/go-github/github"
+	"golang.org/x/oauth2"
+	"google.golang.org/api/pubsub/v1"
```
shantuo: From godoc: This package is DEPRECATED. Use package cloud.google.com/go/pubsub instead.

zombiezen: Except for App Engine Standard because of the networking issues. From https://cloud.google.com/pubsub/docs/reference/libraries:

> However, we recommend using the older Google APIs Client Libraries if running on Google App Engine standard environment.

internal/contributebot/webhook/main.go
zombiezen: Done.

internal/contributebot/webhook/main.go
zombiezen: Extracted the event publishing as a function and fixed the error message.

internal/contributebot/webhook/README.md
zombiezen: Done.

zombiezen:
@shantuo @vangent PTAL, I need approvals before I can merge in.

internal/contributebot/README.md
```diff
@@ -0,0 +1,5 @@
+# Contribute Bot
+
+Contribute Bot is a small service (written using Go Cloud!) that performs
```
vangent: How does it use Go Cloud?

internal/contributebot/webhook/README.md
```diff
@@ -0,0 +1,27 @@
+# Contribute Bot Webhook Listener
+
+Contribute Bot Webhook Listener is a microservice that listens for
+[GitHub webhook][] events and publishes them to a [Cloud Pub/Sub][] topic.
+
```
vangent: Sorry, I'm missing how this fits in with what Contribute Bot is intended to do.
Is there a separate component that will subscribe to the Pub/Sub topic and then react?
Why is the pub/sub indirection needed?

internal/contributebot/README.md
zombiezen: The event worker (subscriber to the topic) will use Go Cloud. It's a forward-looking statement, happy to remove.

internal/contributebot/webhook/README.md
zombiezen: There will be a separate worker that will run through the Pub/Sub topic. GitHub has a [10 second response time limit](https://developer.github.com/v3/guides/best-practices-for-integrators/#favor-asynchronous-work-over-synchronous) combined with a [5000 request/hour rate limit](https://developer.github.com/v3/#rate-limiting), so this adds buffering with the assumption that incoming events are bursty.

vangent: Makes sense, can you add this as a comment somewhere, probably README.md?

zombiezen: Done.

internal/contributebot/webhook/main.go
shantuo: I see, sigh.

PR #285, opened 2018-08-03
gcp: return an error if the user's default credentials can't be found (instead of nil)

vangent:


gcp/gcp.go
```diff
@@ -68,10 +68,9 @@ func DefaultTransport() http.RoundTripper {
 // DefaultCredentials obtains the default GCP credentials with Cloud Platform
 // scope.
 func DefaultCredentials(ctx context.Context) (*google.Credentials, error) {
-	// TODO(light): don't assume that an error always means credentials not found.
```
vangent: Maybe I'm missing something as to why nil was being returned before. I didn't have the environment variable set, and got a lame error like this (when running with --record):

gcp: no credentials available

After this change, I get the much more helpful:

google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.

zombiezen: There was once a concept of Wire optional bindings where this made more sense, but it no longer does.

PR #287, opened 2018-08-06
wire: use cmp.Diff instead of shelling out to "diff" to compare against golden strings in tests

vangent:


vangent:
It does if you pass it a list of strings, thus the cmp.Diff(strings.Split(gotS, "\n"), strings.Split(wantS, "\n")).

zombiezen:
Woah 😮

Issue #276, opened 2018-08-01
all: replay tests have diffs even if interactions haven't changed


zombiezen:
The HTTP/gRPC replay tests currently don't scrub transient attributes like date or request ID, which makes diffs noisy.  This makes it difficult for a reviewer to see whether an interaction has actually changed or whether it's just normal noise.

vangent:
We've decided that this not feasible, for reasons described in the design.md file changes in https://github.com/google/go-cloud/pull/297.

PR #289, opened 2018-08-06
tests: Sort recorded HTTP requests by Method/URL/Body

vangent:
The ordering in the file doesn't matter, as the replay does
matching on the Method/URL/Body to find a matching recorded
response for a request. However, using a stable sort order
will reduce spurious diffs when tests are re-recorded.

#276

### DIFF
--- a/blob/gcsblob/gcsblob_test.go
+++ b/blob/gcsblob/gcsblob_test.go
@@ -36,12 +36,6 @@ const bucketPrefix = "go-cloud"
 
 var projectID = flag.String("project", "", "GCP project ID (string, not project number) to run tests against")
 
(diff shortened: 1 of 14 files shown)


vangent:
BTW, I updated the test .yaml files by calling the new sort function on all of the existing test files, not by running with --record. So, the recorded content in the files is exactly the same, just re-ordered per the sort.

vangent:
Hmm, Travis is not happy. I'll look into it tomorrow.

vangent:
I had to revert the sorting of the AWS HTTP request, it looks like our recording code is enforcing ordering there using a fabricated "X-Gocloud-Seq" header. Tests fail if it's ignored. I'd like to make this consistent across GCP/AWS but will look into that later.

PR #292, opened 2018-08-07
server: Add note about RequestLogger to server.New docs

### LINKED ISSUE #155: server: health check endpoints are not logged
I configure a server like so:

```
opts := &server.Options{
    RequestLogger: requestlog.NewNCSALogger(os.Stdout, func(err error) {
        log.Printf("Whoops! Error: %s", err)
    }),
}
s := server.New(opts)
// ... listen and serve, etc.
```

When I send a request to `/`, I get a nice log message.

When I send a request to `/healthz/liveness` or `/healthz/readiness`, I get no message.

enocom:
A RequestLogger does not log HealthChecks. This commit documents that
fact.

Fixes #155.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -51,8 +51,7 @@ type Options struct {
 	DefaultSamplingPolicy trace.Sampler
 }
 
-// New creates a new server. New(nil) is the same as new(Server). Note: A
-// configured Requestlogger will not log HealthChecks.
+// New creates a new server. New(nil) is the same as new(Server).
 func New(opts *Options) *Server {
 	srv := new(Server)
 	if opts != nil {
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -51,7 +51,8 @@ type Options struct {
 	DefaultSamplingPolicy trace.Sampler
 }
 
-// New creates a new server. New(nil) is the same as new(Server).
+// New creates a new server. New(nil) is the same as new(Server). Note: A
+// configured Requestlogger will not log HealthChecks.
```
zombiezen: This seems like a more appropriate message under `ListenAndServe`. WDYT?

enocom: Both seem fine to me. I'll update.

enocom:
The build failures are a result of `vgo` changes:

```
go: go mod -vendor is now go mod vendor
```

PR #293, opened 2018-08-07
Split record/replay HTML into separate files per test case, instead of per-test

vangent:
#276


vangent:
The build failed with:

go: go mod -vendor is now go mod vendor

I don't think that's related to this CL....?

zombiezen:
It's not related, but we should fix that before going forward. I'll send a fix.

PR #298, opened 2018-08-09
blob: document storing a ctx in blob.Writer

### LINKED ISSUE #274: blob: clearly document cases where blob saves a Context for later use
In at least one case (NewWriter), blob saves a context for later use.
* ctx is passed to NewWriter
* If new opt.ContentType is provided, it saves the ctx in the returned *Writer.
* That ctx  may be used later as part of a call to Write or to Close (which don't take a Context argument).

The [Context documentation](https://golang.org/pkg/context/) says "Do not store Contexts inside a struct type", so this is kind of a no-no, but I'm not sure how to avoid it. There's a [proposal](https://github.com/golang/go/issues/20280) to add Context args to Write and Close, but that's speculative and unlikely to happen soon.

We should at least clearly document the behavior.

shantuo:
Fixes #274

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -59,6 +59,12 @@ type Writer struct {
 
 	// These fields exist only when w is not created in the first place when
 	// NewWriter is called.
+	//
+	// A ctx is stored in the Writer since we need to pass it into NewTypedWriter
+	// when we finished detecting the content type of the object and create the
+	// underlying driver.Writer. This step happens inside Write or Close and
+	// neither of them take a context.Context as an argument. The ctx must be set
+	// to nil after we have done passing it.
 	ctx    context.Context
 	bucket driver.Bucket
 	key    string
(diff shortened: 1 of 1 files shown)


blob/blob.go
```diff
@@ -59,6 +59,12 @@ type Writer struct {
 
 	// These fields exist only when w is not created in the first place when
 	// NewWriter is called.
+	//
+	// A ctx is stored in the Writer since we need to pass it into NewTypedWriter
+	// when we finished detecting the content type of the object and create the
```
vangent: Nit: finished -> finish, "done passing it" -> "passed it".

blob/blob.go
```diff
@@ -59,6 +59,12 @@ type Writer struct {
 
 	// These fields exist only when w is not created in the first place when
 	// NewWriter is called.
+	//
```
vangent: For the bug, I meant that we should document this in the public API. I.e., in NewWriter, which takes a ctx, we should document that the context must remain open until after the returned Writer is closed.

PR #309, opened 2018-08-10
wire/internal/wire: run wire tests in parallel

shantuo:


wire/internal/wire/wire_test.go
```diff
@@ -70,6 +70,8 @@ func TestWire(t *testing.T) {
 	}
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
```
zombiezen: This is racy unless you add `test := test` before `t.Run`

shantuo: Done, thanks for the catch!

Issue #301, opened 2018-08-09
tests: Do less scrubbing of recorded HTTP requests


vangent:
There's a bunch of scrubbing going on, but we've decided not to do any scrubbing other than of the Authorization header.

There will still need to be some string manipulation when matching requests during replay, since we match HTTP bodies and there's some dynamic stuff in there. The scrub function for that should be provider-specific.

vangent:
Fixed by #311.

PR #310, opened 2018-08-10
Get rid of the X-Gocloud-Seq header, and improve logging/comments in replay.NewRecorder.

vangent:
Part of Issue #301 .


vangent:
Note: the headers are still in the recorded golden files, but they don't hurt anything and will be dropped next time someone generates them.

vangent:
Never mind, I'm going to send one giant PR.

PR #327, opened 2018-08-15
blob/s3blob: get Size from ContentRange instead of from ContentLength, if available. ContentLength for partial reads refers to the length of the content being read, not the size of the entire blob.

### LINKED ISSUE #305: s3blob: Size attribute returns different values based on offset
For example, if actual size of the blob is of size 100, and you initialize the RangeReader with offset 20, the Size attribute is returned as 80. I believe it should always be 100. We should make sure this is clearly documented and consistent across providers.

vangent:
Fixed #305.

### DIFF
--- a/blob/s3blob/s3blob.go
+++ b/blob/s3blob/s3blob.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


blob/s3blob/s3blob.go
```diff
@@ -188,10 +189,23 @@ func (b *bucket) NewRangeReader(ctx context.Context, key string, offset, length
 		}
 		return nil, err
 	}
+	// Default size to ContentLength, but that's incorrect for partial-length reads,
```
shantuo: How about wrapping this inside a function?

vangent: Done.

PR #328, opened 2018-08-15
internal/contributebot: basic skeleton

zombiezen:
Handles auth, subscription setup, and basic event handler for removing "in progress" label.

Updates #216


internal/contributebot/README.md
```diff
@@ -13,3 +13,43 @@ incoming events are bursty.
 
 [github-async]: https://developer.github.com/v3/guides/best-practices-for-integrators/#favor-asynchronous-work-over-synchronous
 [github-ratelimit]: https://developer.github.com/v3/#rate-limiting
+
+## Setup
+
+1.  [Create a new GCP project][].
+1.  Set your project using `gcloud config set project PROJECTID`, where
+    `PROJECTID` is the project's ID.
+1.  Enable App Engine with `gcloud app create`.
```
vangent: Consider using 

```bash
```

for this, with environment variables for the variables?

E.g.,

```bash
# Set your project id here.
PROJECTID=
# Set your bucket name for Terrafrom state here.
TERRAFORM_BUCKET_NAME=

gcloud config set project ${PROJECTID}
gcloud app create
gsutil mb -c regional -l us-west2 gs://${TERRAFORM_BUCKET_NAME}
...
```

internal/contributebot/ghauth.go
```diff
@@ -0,0 +1,308 @@
...
+	available   <-chan struct{}
+	cancelWatch context.CancelFunc
+	watchDone   <-chan struct{}
+
+	mu         sync.RWMutex
+	privateKey *rsa.PrivateKey
+}
+
+func newGitHubAppAuth(id int64, privateKey *runtimevar.Variable, rt http.RoundTripper) *gitHubAppAuth {
+	done := make(chan struct{})
```
vangent: Nit: I think it would more clear to call this watchDone throughout here.

internal/contributebot/ghauth.go
```diff
@@ -0,0 +1,308 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	auth := &gitHubAppAuth{
+		id:          id,
+		base:        rt,
+		available:   available,
+		cancelWatch: cancel,
+		watchDone:   done,
+	}
+	go func() {
+		defer close(done)
```
vangent: Maybe a comment like:

// Watches privateKey until Stop is called to cancel ctx.

internal/contributebot/ghauth.go
```diff
@@ -0,0 +1,308 @@
...
+	"strconv"
+	"sync"
+	"time"
+
+	jwt "github.com/dgrijalva/jwt-go"
+	"github.com/google/go-cloud/runtimevar"
+	"github.com/google/go-github/github"
+)
+
+// gitHubAppAuth makes HTTP requests with GitHub application credentials.
```
vangent: Can you explain more about what is going on with watching the PrivateKey?

internal/contributebot/ghauth.go
```diff
@@ -0,0 +1,308 @@
...
+func (auth *gitHubAppAuth) Stop() {
+	auth.cancelWatch()
+	<-auth.watchDone
+	auth.mu.Lock()
+	auth.privateKey = nil
+	auth.mu.Unlock()
+}
+
+// watch watches the private key variable until the context is cancelled.
+// The available channel is closed on the first successful Watch.
```
vangent: ...so that waitPrivateKey is notified?

internal/contributebot/ghauth.go
```diff
@@ -0,0 +1,308 @@
...
+func (auth *gitHubAppAuth) watch(ctx context.Context, available chan<- struct{}, v *runtimevar.Variable) {
+	first := true
+	for ctx.Err() == nil {
+		snap, err := v.Watch(ctx)
+		if err != nil {
+			log.Println("Watching GitHub private key:", err)
```
vangent: Nit: space after : ?

internal/contributebot/ghauth.go
```diff
@@ -0,0 +1,308 @@
...
+			continue
+		}
+		curr := snap.Value.(*rsa.PrivateKey)
```
vangent: Consider logging that a private key was received?

internal/contributebot/ghauth.go
```diff
@@ -0,0 +1,308 @@
...
+		auth.privateKey = curr
+		auth.mu.Unlock()
+		if first {
+			close(available)
+			first = false
+		}
+	}
+}
+
+func (auth *gitHubAppAuth) CheckHealth() error {
```
vangent: docstring please

internal/contributebot/ghauth.go
```diff
@@ -0,0 +1,308 @@
...
+	go func() {
+		defer close(done)
+		auth.watch(ctx, available, privateKey)
+	}()
+	return auth
+}
+
+// Stop stops watching the private key.
+func (auth *gitHubAppAuth) Stop() {
+	auth.cancelWatch()
```
vangent: Should this close available too? Otherwise a caller in waitPrivateKey will block forever, no?

internal/contributebot/ghauth.go
```diff
@@ -0,0 +1,308 @@
...
+		// Pass through without adding credentials.
+		return auth.base.RoundTrip(req)
+	}
+	forwarded := false
+	if req.Body != nil {
+		defer func() {
+			if forwarded {
+				return
+			}
+			if err := req.Body.Close(); err != nil {
```
vangent: Why this? Is the call to RoundTrip always supposed to call req.Body.Close()?

internal/contributebot/main.go
```diff
@@ -0,0 +1,167 @@
...
+			msg.Nack()
+			return
+		}
+		msg.Ack()
+		log.Printf("Processed %s event %s", eventType, id)
+	})
+}
+
+func (w *worker) receiveIssueEvent(ctx context.Context, e *github.IssuesEvent) error {
+	const inProgressLabel = "in progress"
```
vangent: Maybe put this in a func, e.g., "checkForClosedIssueWithInProgress" or something? Presumably we will end up adding more checks.

internal/contributebot/README.md
```diff
@@ -13,3 +13,43 @@ incoming events are bursty.
 
 [github-async]: https://developer.github.com/v3/guides/best-practices-for-integrators/#favor-asynchronous-work-over-synchronous
 [github-ratelimit]: https://developer.github.com/v3/#rate-limiting
+
+## Setup
```
vangent: Presumably nobody should actually do this, right? Maybe document where contributebot is running now?

internal/contributebot/main.go
zombiezen: Added a bit more structure, since there's somewhat of a two-phase test-and-run procedure. Let me know what you think.

PR #331, opened 2018-08-15
wire: make tests work with Go modules

### LINKED ISSUE #280: all: test against Go 1.11
Also we should update our documentation on anything different.

shantuo:
This adds go.mod files to the generated directories and put test commands under example.com module.

Fixes #280


wire/internal/wire/parse.go
```diff
@@ -845,7 +845,7 @@ func (pv ProviderOrValue) Provider() *Provider {
 	return pv.p
 }
 
-// Provider returns pv as a Value pointer. It panics if pv points to a
+// Value returns pv as a Value pointer. It panics if pv points to a
```
vangent: I believe this is already fixed on master, can you merge?

wire/internal/wire/testdata/Chain/pkg
```diff
@@ -1 +1 @@
-foo
+example.com/foo
```
vangent: Can you explain why all of these need to change to example.com?

vangent:
> This branch has conflicts that must be resolved

Please fix.

wire/internal/wire/wire_test.go
```diff
@@ -576,10 +584,28 @@ func (test *testCase) materialize(gopath string) error {
 	return nil
 }
 
-func runGo(bctx *build.Context, args ...string) error {
+// writeGoMod generates go.mod files for the test package and its dependency.
```
vangent: Should these be checked in and verified? Or are they always going to be the same if wire_gen.go is the same as before?

wire/internal/wire/parse.go
shantuo: Done.

wire/internal/wire/testdata/Chain/pkg
shantuo: Added comments in code.

This is the place the test code copied the packages to. I could add the module path when copying the code, is that preferred?

wire/internal/wire/wire_test.go
shantuo: These should stay the same:
1) All test packages should be under "module example.com"
2) We are not going to have random third-party dependencies, and will likely only depend on the wire.go file.

wire/internal/wire/testdata/Chain/pkg
vangent: Right, but why? It's not required for go 1.11 that everything be under a domain, right?

I.e., why not leave it so that the code is just in package "foo" and don't change the test code to copy the packages to example.com/foo?

shantuo: Some of the test cases contain more than one packages (foo and bar ...), it is better we put them in one module.

vangent: Why? It just seems to me like it's making the whole test suite look like it requires modules, which it shouldn't; it should continue to work with go 1.10.

wire/internal/wire/wire_test.go
```diff
@@ -576,10 +584,45 @@ func (test *testCase) materialize(gopath string) error {
...
+//
+//			... (Packages to be built and tested)
+//				any Go files copied recursively
+//
+//		github.com/google/go-cloud/
+//
+//			go.mod
+//
+//			... (Dependency files copied)
+func writeGoMod(gopath string) error {
```
vangent: For example, what happens here if the user has go 1.10, which doesn't support modules?

shantuo: The go.mod files should be ignored by Go 1.10 and lower.

vangent: OK.

vangent:
> I.e., why not leave it so that the code is just in package "foo" and don't change the test code to copy the packages to example.com/foo?

I'm still not sure I understand your answer to this. Couldn't you make this work without putting every test under example.com? Is that an independent change you are making (to modularize the test examples), or is it required by the go 1.11 (I don't think it's the latter).

PR #332, opened 2018-08-15
wire: update Provider.Out to be a slice of provided types, and keep track of the provided concrete type in ProviderSet.providerMap

### LINKED ISSUE #325: wire: unused provider error when only pointer to struct is used
Repro case:

```go
package main

import "github.com/google/go-cloud/wire"

type Foo struct{}

func inject() *Foo {
  wire.Build(Foo{})
  return nil
}
```

This produces the error:

```
foo.go:7:1: inject inject: unused provider "Foo"
wire: generate failed
```

I expect this to succeed.

vangent:
To more clearly model named struct providers (which provide both the struct
type and a pointer to the struct type).

Fixes #325.


zombiezen:
The diff seems to include unrelated diffs.. can you rebase your changes?

vangent:
Done.

wire/internal/wire/analyze.go
```diff
@@ -204,7 +204,7 @@ dfs:
 			})
 		case pv.IsValue():
 			v := pv.Value()
-			if !types.Identical(v.Out, curr.t) {
+			if !types.Identical(pv.ConcreteType(), curr.t) {
```
zombiezen: Keep using `v.Out` for this one since it is unambiguous and to keep consistency with the next few lines.

wire/internal/wire/analyze.go
```diff
@@ -388,36 +391,43 @@ func verifyAcyclic(providerMap *typeutil.Map, hasher typeutil.Hasher) []error {
...
 			visited.Set(head, true)
-			switch x := providerMap.At(head).(type) {
-			case nil:
+			x := providerMap.At(head)
+			if x == nil {
 				// Leaf: input.
-			case *Value:
+				continue
+			}
+			pt, ok := x.(*ProvidedType)
```
zombiezen: If you're going to use the `ok` value just to panic, it is more concise to just type assert without the `ok` value.

wire/internal/wire/parse.go
```diff
@@ -122,8 +116,8 @@ type Provider struct {
 	// elements in Args.
 	Fields []string
 
-	// Out is the type this provider produces.
-	Out types.Type
+	// Out is the slice of types this provider produces.
```
zombiezen: What about "Out is the set of types this provider produces. It will always contain at least one type."

This implies no defined ordering and calls out that it will never be empty.

wire/internal/wire/analyze.go
vangent: Done.

wire/internal/wire/analyze.go
vangent: Done.

wire/internal/wire/parse.go

PR #341, opened 2018-08-17
blob: Fix comments and clean up tempdir from blob/filestore tests

vangent:


blob/fileblob/fileblob_test.go
```diff
@@ -36,7 +36,7 @@ func makeBucket(t *testing.T) (*blob.Bucket, func()) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	return b, func() {}
+	return b, func() { _ = os.RemoveAll(dir) }
```
shantuo: The body can just be `os.RemoveAll(dir)`

vangent: The linter that Travis uses complains about that, unfortunately, something like "Error value not being checked."

Issue #335, opened 2018-08-16
gcsblob: reader.ModTime() returns zero for non-zero-length reads


vangent:
See https://github.com/google/go-cloud/issues/314 for context. gcsblob provides ModTime when you do a 0-length read, but not for non-zero-length reads.

Blocked by GoogleCloudPlatform/google-cloud-go#1091.

PR #343, opened 2018-08-17
blob/gcsblob: support ModTime for non-zero length reader

shantuo:
Fixes: #335


blob/drivertest/drivertest.go
```diff
@@ -253,7 +253,7 @@ func testAttributes(t *testing.T, makeBkt BucketMaker) {
 		b, done := init(t)
 		defer done()
 
-		r, err := b.NewRangeReader(ctx, key, 0, 0)
+		r, err := b.NewRangeReader(ctx, key, 0, 1)
```
vangent: Since the implementations for several providers are different for 0-length vs non, how about testing both? With your suggested change, we're no longer testing the 0-length read.

You could do this either by putting a loop around all of the t.Runs here, something like:

for _, readSize := []int64{0, 1} {
  t.Run(fmt.Sprintf("ReadLength%d", readSize), ... {
     t.Run("Size", ...
     t.Run("Attributes", ...
      ..
  }
}

Or, by adding a "readLength int" argument to testAttributes and calling it twice up above.

blob/gcsblob/gcsblob.go
```diff
@@ -110,11 +110,13 @@ func (b *bucket) NewRangeReader(ctx context.Context, key string, offset, length
 		}
 		return nil, err
 	}
+	// updated is set to zero value when non-nil error returned, no need to check or report.
+	updated, _ := r.LastModified()
```
vangent: Hmm, so this always existed? I thought we were blocked on the upstream bug...?

blob/drivertest/drivertest.go
shantuo: Note that we do a second round of read a few seconds after to test the ModTime change, that one tests zero-length read. I think for this specific case the coverage is enough. WDYT?

blob/gcsblob/gcsblob.go
shantuo: No, that issue is just fixed. I also updated go.mod file to bump the dep version.

blob/drivertest/drivertest.go
vangent: That's true for ModTime, but not the others (Size + ContentLength), so I think it's better to test both explicitly. I'd update the 2nd read for ModTime to be the same as the first one.

blob/gcsblob/gcsblob.go
vangent: Oh, sorry, I missed that diff. Got it.

shantuo:
Removed.

PR #347, opened 2018-08-17
blob: Use the Harness approach for Blob conformance tests

vangent:


blob/drivertest/drivertest.go
```diff
@@ -50,34 +30,40 @@ import (
...
-// Functions should return the bucket along with a "done" function to be called
-// when the test is complete.
-// If bucket creation fails, functions should t.Fatal to stop the test.
-// See the package comments for more details on required semantics.
-type BucketMaker func(t *testing.T) (*blob.Bucket, func())
+// Harness descibes the functionality test harnesses must provide to run
+// conformance tests.
+type Harness interface {
+	// MakeBucket creates a *blob.Bucket to test.
+	MakeBucket(ctx context.Context, t *testing.T) *blob.Bucket
```
zombiezen: Wouldn't `HarnessMaker` already store the `*testing.T` in the `Harness`?

blob/drivertest/drivertest.go
```diff
@@ -50,34 +30,40 @@ import (
...
-// BucketMaker describes functions used to create a bucket for a test.
-// Functions should return the bucket along with a "done" function to be called
-// when the test is complete.
-// If bucket creation fails, functions should t.Fatal to stop the test.
-// See the package comments for more details on required semantics.
-type BucketMaker func(t *testing.T) (*blob.Bucket, func())
+// Harness descibes the functionality test harnesses must provide to run
+// conformance tests.
+type Harness interface {
+	// MakeBucket creates a *blob.Bucket to test.
```
zombiezen: You should probably say something about the assumptions about the bucket's contents.

blob/drivertest/drivertest.go
```diff
@@ -14,26 +14,6 @@
 //
```
zombiezen: Please turn this into a blank line, otherwise the whole license header will be copied into the godoc.

blob/drivertest/drivertest.go
vangent: It could, but I think it's more explicit to pass it (since all Harness implementations would need to store it, and MakeBucket *has* to use it to fail on errors.

I added a comment that it should fail the test on errors to make this more clear.

blob/drivertest/drivertest.go

blob/drivertest/drivertest.go

blob/drivertest/drivertest.go
vangent: I could do that, but it adds a lot of unhelpful boilerplate to the actual test functions, which are certainly not going to do anything but t.Fatal if an error is returned from any Harness function. Do you think that adds clarity, or ...?

vangent: Sorry, missed your previous comment. 

> "which *testing.T should I fail?"

Not sure I get this, isn't there only one?

zombiezen: `HarnessMaker` takes in a `*testing.T` and the `Harness` methods take in a `*testing.T`. These hypothetically could be different values (subtests). I'm thinking defensively, since I have definitely had code where logs have gone to wrong tests because of this sort of issue.

I'm not so worried about boilerplate to our conformance tests, which is less code than the number of implementors. I see it being far too easy for a `Harness` to accidentally call `Error` instead of `Fatal`, which making it explicit return value would avoid. The fact that a bug in a `Harness` can cause improper control flow in the test seems like spooky-action-at-a-distance.

zombiezen: Actually, a more concrete objection: I can also imagine circumstances where the conformance tests might actually *expect* an operation to fail. Hard-coding to `*testing.T` means that the conformance tests would not be able to do this easily.

vangent: OK, done.

PR #349, opened 2018-08-20
internal/contributebot: allow configuration to be used as a module

zombiezen:
Makes it easier to maintain a separate dev and prod setup with different state while still sharing the majority of the code.


internal/contributebot/prod/main.tf
```diff
@@ -0,0 +1,38 @@
...
+
+variable "github_app_key" {
+  default     = ""
+  description = "PEM-encoded GitHub application private key. This defaults to empty for bootstrapping reasons."
+}
+
+module "contributebot" {
+  source         = ".."
+  project        = "go-cloud-contribute-bot"
+  zone           = "us-central1-c"
```
shantuo: Should this be a var?

internal/contributebot/README.md
```diff
@@ -22,30 +22,26 @@ To set up your own instance of Contribute Bot for local testing or deployment:
...
 1.  Enable App Engine with `gcloud app create`.
-1.  Create a bucket to store the [Terraform state][] using
-    `gsutil mb -c regional -l us-west2 gs://some-bucket`, replacing
-    `some-bucket` with a new unique name.
-1.  Run `terraform init -backend-config 'bucket=some-bucket'`, replacing
-    `some-bucket` with the same name you used in the last step,
-1.  Create a [variable file][] for the settings from `variables.tf`. Skip
-    `github_app_key` for now.
+1.  Copy the `prod` directory to a directory called `dev`.
+1.  In `dev/main.tf`, Remove the `backend "gcs"` block and change the project
```
shantuo: s/Remove/remove

zombiezen: Done.

internal/contributebot/prod/main.tf
zombiezen: No, as it is not sensitive and the goal of this PR is to start checking in the actual configuration into version control.

PR #350, opened 2018-08-21
samples/guestbook: use gopkg.in/pipe.v2, the correct import path

h8liu:


h8liu:
Found by [gopkgs.io](https://gopkgs.io):

https://gopkgs.io/builds/20180821-170240-f19fc7f9

PR #358, opened 2018-08-22
runtimevar: Port runtimevar tests to the shared conformance test suite

### LINKED ISSUE #337: runtimevar: Merge all provider-specific tests into a single conformance test suite
Similar to what was done for blob.

vangent:
Also fixed some issues I found during the port; e.g., neither paramstore nor runtimeconfigurator properly deduped errors.

Issue #337.


runtimevar/drivertest/drivertest.go
```diff
@@ -18,16 +18,20 @@ package drivertest
...
 	"github.com/google/go-cloud/runtimevar"
+	"github.com/google/go-cmp/cmp"
 )
 
 // Harness descibes the functionality test harnesses must provide to run
 // conformance tests.
 // All functions should fail the test on error.
 type Harness interface {
 	// MakeVar creates a *runtimevar.Variable to watch the given variable.
+	// For implementations that poll, the poll interval must be at least 50 milliseconds.
```
zombiezen: Remove this guidance (for reasons I'll state down below).

runtimevar/drivertest/drivertest.go
```diff
@@ -48,5 +52,439 @@ type HarnessMaker func(t *testing.T) (Harness, error)
...
+		t.Fatal(err)
+	}
+	if got.Value.(string) != content1 {
+		t.Errorf("got %q want %q", got, content1)
+	}
+
+	// Update in a separate goroutine, waiting a bit to give the Watch
+	// a chance to block.
+	go func() {
+		time.Sleep(10 * time.Millisecond)
```
zombiezen: There is no way to actually test for the property you are trying to test given the way the API is structured. Sleeping will not help with that. The problem is that when a driver does not meet this property, this test will become flaky and it will be tempting for someone (especially an external implementation) to dismiss this as a bad test.

There are two ways to address this:

1.  Delete the test as the property cannot be conclusively tested for.
2.  Make the failure more likely to occur by running it more times (but with no sleep). The test failure message should include a warning to the effect of "If you see this test flaking, it is indicating that there is almost certainly an issue with your driver as this is a probabilistic test."

(h/t @neild for the guidance here.)

runtimevar/paramstore/paramstore.go
```diff
@@ -106,8 +108,19 @@ func (w *watcher) Close() error {
 func (w *watcher) ping(_ context.Context) (*driver.Variable, error) {
 	p, err := readParam(w.sess, w.name, w.lastVersion)
 	if err != nil {
+		errCode := err.Error()
```
zombiezen: Please include in your commit message what was fixed in this package.

runtimevar/runtimeconfigurator/runtimeconfigurator.go
```diff
@@ -157,8 +152,14 @@ func (w *watcher) WatchVariable(ctx context.Context) (driver.Variable, error) {
 func (w *watcher) ping(ctx context.Context) (*driver.Variable, error) {
 	// Use GetVariables RPC and check for deltas based on the response.
 	vpb, err := w.client.GetVariable(ctx, &pb.GetVariableRequest{Name: w.name})
-	w.lastRPCTime = time.Now()
 	if err != nil {
+		errCode := grpc.Code(err)
```
zombiezen: Same here: please include in your commit message what was fixed here.

runtimevar/drivertest/drivertest.go
```diff
@@ -48,5 +52,439 @@ type HarnessMaker func(t *testing.T) (Harness, error)
...
+	}
+	if got.Value.(string) != content1 {
+		t.Errorf("got %q want %q", got, content1)
+	}
+	prev := got
+
+	// Delete in a separate goroutine, waiting a bit to give the Watch
+	// a chance to block.
+	go func() {
+		time.Sleep(10 * time.Millisecond)
```
zombiezen: Same guidance from above goes here. No ~~capes~~ sleeps.

nocapes

runtimevar/drivertest/drivertest.go
vangent: Done.

runtimevar/drivertest/drivertest.go
vangent: Done.

vangent: I left a TODO to run the test more times when not in -record mode.

runtimevar/paramstore/paramstore.go
vangent: Done.

runtimevar/runtimeconfigurator/runtimeconfigurator.go
vangent: Done.

vangent:
Per meatspace discussion, I removed the concurrent tests for now and will file an issue to think of ways to get better test coverage.

PR #362, opened 2018-08-23
runtimevar/tests: Correctly call runtimevar.Close in driver tests

vangent:


runtimevar/drivertest/drivertest.go
```diff
@@ -85,6 +81,7 @@ func testNonExistentVariable(t *testing.T, newHarness HarnessMaker) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer v.Close()
```
zombiezen: Check for error?

vangent: Done.

PR #369, opened 2018-08-27
runtimevar: move dedupe detection and poll-wait into runtimevar

### LINKED ISSUE #365: runtimevar: Move dedupe detection into runtimevar itself
Currently, the driver (specifically, Watcher.WatchVariable) must block until the variable value (or error value) changes. For many provider implementations, this means they have to keep track of the last variable value and error value. See the paramstore and runtimeconfigurator implementations for example.

Instead, I think we should allow WatchVariable to return values whenever it wants, and do the deduping in runtimevar.go.

vangent:
This simplifies the provider implementations significantly and allows better test coverage.

Fixes #365.


runtimevar/runtimevar_test.go
```diff
@@ -18,16 +18,183 @@ package runtimevar_test
...
+				&watchVariableCall{v: var1, version: 1},
+				&watchVariableCall{v: var1, version: 1},
+				&watchVariableCall{v: var1, version: 1},
+				&watchVariableCall{v: var2, version: 2},
+			},
+			want: []*watchResp{
+				&watchResp{snap: snap1},
+				&watchResp{snap: snap2},
+			},
+		},
```
vangent: I could add a test that returns a non-zero waitTime and verify that the next WatchVariable call is at least <waitTime> after the previous one. WDYT?

vangent:
> I wonder if this would be more simply expressed as an adapter
> utility inside the driver package rather than a change to the interface

That's sort of what it was before, no? internal.Pinger was the utility. An adapter

1. Requires providers to maintain the state needed; I suppose this could be done by having an adapater with state that the provider would call at various points to update/check it, but that's definitely not simpler.
2. Requires provider implementations to know about and use the adapter; for example someone writing a new provider implementation wouldn't necessarily know about the adapter.

I also like that this allows better test coverage, which might be harder to ensure with an adapter since correct behavior would also depend on the providers using it correctly.

Isn't the ability to provide common functionality one of the reasons why we use concrete types? If runtimevar.Variable was an interface, then we'd have to provide an adapter, but IMHO this allows it to be simpler.

zombiezen:
The main concern is that if a new backend came around that supports a push model (i.e. long-standing connections), that it would be difficult to shoehorn in if the interface is a pulling model. An alternative could be to make the `runtimevar.Variable` be able to be created with this alternative interface and then do the shimming internally.

vangent:
> The main concern is that if a new backend came around that supports a push model
> ... it would be difficult to shoehorn

I don't think so; the filevar implementation is a push model and works fine. The interface
supports polling by returning a Wait duration if nothing has changed, but implementations
that can block until a change is detected can do so.

vangent:
Gentle ping on this one? It's blocking a bunch of other changes.

zombiezen:
@vangent and I brainstormed the design space offline and we came to the conclusion that this is the right approach generally: this reduces the current (error-prone) repetition in implementers of the `Watcher` interface by reducing the amount of state that it needs to maintain. The two cases that are tricky to design around are (specific examples given but could be applicable to new implementations):

-  The AWS Parameter Store implementation has an optimization where it checks the version as one request then conditionally does another request to fetch the parameter data.
-  The `filevar` implementation necessarily switches between a push-based watch (fsnotify) and a pull-based watch (when the file has been deleted).

I came up with a couple different alternatives, but Robert proved that they were isomorphic to this PR but with more moving parts.

The one orthogonal idea that we'd like to tackle before releasing this interface change is to reduce the parameters using the following types (sketching here):

```go
package driver

// State represents a discrete sample from a Watcher.
type State interface {
  // Value returns the data represented by this state.
  Value() (Variable, error)
  // SameAs compares two states for equality.
  SameAs(State) bool
}

// StateReader samples from the underlying data source. (This is what used to be called Watcher.)
type StateReader interface {
  ReadState(ctx context.Context, prev State) (State, time.Duration, error)
}
```

runtimevar/driver/driver.go
```diff
@@ -39,20 +39,33 @@ type Variable struct {
...
+	// For example, an implementation that can detect changes in the underlying
+	// variable could block until it detects a change (or until ctx is Done).
+	// A polling implementation could poll on every call to WatchVariable,
+	// returning nil v/version/err and a non-zero wait to set the poll interval
+	// if there's no change.
+	//
+	// Version is provider-specific; for example, it could be an actual version
+	// number, it could be the variable value, or it could be raw bytes for the
+	// variable before decoding it.
+	// TODO(rvangent): Consider refactoring to encapsulate State (including
```
zombiezen: Let's open an issue for this (and update the other TODO) to reflect.

runtimevar/driver/driver.go
```diff
@@ -39,20 +39,33 @@ type Variable struct {
 // dictate the type of Variable.Value and a decoding function.  The Watcher provider can use the
 // runtimevar.Decoder to facilitate the decoding logic.
 type Watcher interface {
-	// WatchVariable blocks until the variable changes, the Context's Done channel closes or an
-	// error occurs.
+	// WatchVariable returns one of:
```
zombiezen: I re-read this, and I don't have any concrete suggestions on how to improve clarity in this iteration. Let's focus on this when we have the interface we like.

runtimevar/filevar/filevar.go
```diff
@@ -88,34 +89,31 @@ func NewVariable(file string, decoder *runtimevar.Decoder, opts *WatchOptions) (
...
-	// Check for Context cancellation first before proceeding.
-	select {
-	case <-ctx.Done():
-		return zeroVar, ctx.Err()
-	default:
-		// Continue.
+func (w *watcher) WatchVariable(ctx context.Context, prevVersion interface{}, prevErr error) (*driver.Variable, interface{}, time.Duration, error) {
+
+	// checkSameErr checks to see if err is the same as prevErr, and if so, returns
+	// the "no change" signal with w.waitTime.
```
zombiezen: Let's add a TODO here to revisit when we make the interface change we discussed. Comparing errors in this way is brittle.

vangent:
> type State interface

I added a TODO for this.

runtimevar/paramstore/paramstore.go
```diff
@@ -86,54 +84,56 @@ type watcher struct {
...
+	decoder *runtimevar.Decoder
 }
 
-// WatchVariable begins watching the given parameter and waiting for it to change.
-// The function will block until either the parameter changes, or the context
-// is canceled.
-func (w *watcher) WatchVariable(ctx context.Context) (driver.Variable, error) {
-	return internal.Pinger(ctx, w.ping, w.waitTime)
+func (w *watcher) WatchVariable(ctx context.Context, prevVersion interface{}, prevErr error) (*driver.Variable, interface{}, time.Duration, error) {
+
```
zombiezen: Remove blank line

runtimevar/runtimevar_test.go
zombiezen: I don't understand what the test is. Is it that the amount of time has elapsed? If that's the case, I don't think that's needed. (See my comment above about the duration being a hint.)

runtimevar/driver/driver.go
```diff
@@ -39,20 +39,33 @@ type Variable struct {
...
 	//
-	// If the variable has changed, then WatchVariable must return a Variable object with the
-	// new value.
+	// 1. A new value for the variable in v, along with a provider-specific
+	//    version that will be passed to the next WatchVariable call. wait is
+	//    ignored and err must be nil.
+	// 2. A new error. v, version, and wait are ignored.
+	// 3. A nil v and version and err, indicating that the value of the variable
+	//    or the error returned has not changed. WatchVariable will not be called
+	//    again for wait.
```
zombiezen: Optional: Seems like this should be a hint more than anything else. There might be reasons (especially in the final form) why the caller might not be able (or want) to respect the time.

PR #373, opened 2018-08-28
internal/contributebot: Improve READMEs for running contributebot

vangent:


internal/contributebot/webhook/README.md
```diff
@@ -23,5 +23,6 @@ dev_appserver.py .
 ```shell
 cp app.yaml.in app.yaml
 # Edit app.yaml to fill in configuration.
+# You may have to run `go get` if you're using go 1.11.
```
zombiezen: This isn't version-specific. If anything, on Go 1.11 you need to ensure `GO111MODULE=off`.

vangent: Done.

PR #380, opened 2018-08-30
internal/contributebot: go mod tidy

vangent:


zombiezen:
This LGTM, but in examining it, I discovered that the Go Modules proxy I had set up isn't working as I expect. Please hold off merging until I can investigate next week.

PR #382, opened 2018-08-30
blob/tests: test ContentType with (0, -1) length reads; don't bother with length=1

vangent:


blob/drivertest/drivertest.go
```diff
@@ -224,22 +224,22 @@ func testAttributes(t *testing.T, newHarness HarnessMaker) {
 		}
 	}
 
-	t.Run("ContentType", func(t *testing.T) {
-		b, done := init(t)
-		defer done()
+	for _, rLen := range []int64{0, -1} {
+		t.Run(fmt.Sprintf("ReadLength%d", rLen), func(t *testing.T) {
```
shantuo: nit: keep the name "ReadLength=%d" instead of "ReadLength%d".

vangent: I will use space.

PR #384, opened 2018-08-31
runtimevar/constantvar: Add a runtimevar implementation for constant values (#245)

### LINKED ISSUE #245: runtimevar: add constant variable implementation
For testing, it can be helpful to have a `*runtimevar.Variable` that is set to a constant value. Seems like a good utility function to include.

vangent:
Fixes #245


vangent:
Hmm, sorry, this is incorrect, Watch needs to block forever. Will re-send.

PR #385, opened 2018-08-31
wire: omit the local package identifier if it matches the package name

vangent:


wire/internal/wire/wire.go
```diff
@@ -172,9 +172,12 @@ func (g *gen) frame() []byte {
 		}
 		sort.Strings(imps)
 		for _, path := range imps {
-			// TODO(light): Omit the local package identifier if it matches
-			// the package name.
-			fmt.Fprintf(&buf, "\t%s %q\n", g.imports[path], path)
+			// Omit the local package identifier if it matches the package name.
+			if g.imports[path] == path {
```
zombiezen: While this is commonly correct, the more reliable check is to get the package's identifier from the type checker and check those for equality. The information should be there, just another matter of plumbing it.

vangent: PTAL.

zombiezen: Not quite, `filepath.Base` is commonly correct, but not always. [`Package.Name()`](https://godoc.org/go/types#Package.Name) is what you should trust.

zombiezen:
Is this ready for review?

vangent:
Yes.

wire/internal/wire/wire.go
```diff
@@ -172,9 +172,13 @@ func (g *gen) frame() []byte {
 		}
 		sort.Strings(imps)
 		for _, path := range imps {
-			// TODO(light): Omit the local package identifier if it matches
-			// the package name.
-			fmt.Fprintf(&buf, "\t%s %q\n", g.imports[path], path)
+			// Omit the local package identifier if it matches the package name.
+			name := g.prog.Package(path).Pkg.Name()
```
zombiezen: I think this might be incorrect for vendored directories (I forget when that prefix strip happens).

vangent: Yea, you are right; when I set `export GO111MODULE=off` and run `go test --run TestWire/Vendor`, the previous code crashed because `g.prog.Package(path)` returned `nil`.

Now checking for `nil`, and it works in both `GO111MODUDLE` modes.

zombiezen: What I'm saying is: the package information is there, but under the vendored import path. If you collect that information before the vendor prefix stripping and put it in the map, then you will never have the `== nil` case.

vangent:
Fixes #424

wire/internal/wire/wire.go
vangent: Ah, OK. Done.

PR #386, opened 2018-08-31
wire: use package names to disambiguate variables

### LINKED ISSUE #383: wire: improve the names of the generated variables.
Right now wire names the variables holding provided values based on their types and if there are collisions in those names, it adds a counter to the name. So if there are Clients coming from 3 different packages, we'd have client, client2, client3. 

I think the generated code could be better if the package is also taken into consideration when naming a variable. For instance if we have redis.Client and mongo.Client, the variables should be redisClient and mongoClient instead of client and client2.

Edit.: Looking at the code, there seems to already be a todo for this:
https://github.com/google/go-cloud/blob/master/wire/internal/wire/wire.go#L676

vangent:
Fixes #383.


wire/internal/wire/wire.go
```diff
@@ -661,22 +661,35 @@ func zeroValue(t types.Type, qf types.Qualifier) string {
...
-// guarantees about whether the name is exported or unexported: call
-// export() or unexport() to convert.
-func typeVariableName(t types.Type) string {
+// typeVariableName invents variable names derived from the type name,
+// or returns a slice with defaultName if one could not be found.
+// The first returned name is preferred; others may be used for
+// disambiguation.
+// There are no guarantees about whether the names are exported or unexported:
+// call export() or unexport() to convert.
+func typeVariableName(t types.Type, defaultName string) []string {
```
zombiezen: While I like that this is composable, I think it might push too much complexity on the caller (and thus spreading the places that need to be tested). I think it's time for `typeVariableName` to return the final name directly. Perhaps the signature could be:

```go
func typeVariableName(t types.Type, defaultName string, transform func(string) string, collides func(string) bool) string
```

vangent: Done. Also added a test for typeVariableName.

wire/internal/wire/wire.go
```diff
@@ -234,7 +234,8 @@ func (g *gen) inject(pos token.Pos, name string, sig *types.Signature, set *Prov
 			}
 			if g.values[c.valueExpr] == "" {
 				t := c.valueTypeInfo.TypeOf(c.valueExpr)
-				name := disambiguate("_wire"+export(typeVariableName(t))+"Value", g.nameInFileScope)
+
+				name := typeVariableName(t, "", func(name string) string { return "_wire" + export(name) + "Value" }, g.nameInFileScope)
```
zombiezen: It was surprising to me that this default is the empty string. I realize root cause and remarked below.

wire/internal/wire/wire.go
```diff
@@ -661,22 +657,50 @@ func zeroValue(t types.Type, qf types.Qualifier) string {
...
+		}
+	}
+	if len(names) == 0 {
+		names = append(names, defaultName)
 	}
-	return ""
+
+	// Transform the names.
+	for i, name := range names {
+		names[i] = transform(name)
```
zombiezen: It is surprising to me that the `defaultName` is transformed. I would have expected no processing, just disambiguation.

wire/internal/wire/wire.go
```diff
@@ -712,6 +736,15 @@ func unexport(name string) string {
 	return sbuf.String()
 }
 
+// unexportSlice returns a slice of unexported names.
+func unexportSlice(names []string) []string {
```
zombiezen: Remove: unused.

wire/internal/wire/wire.go
```diff
@@ -729,7 +762,7 @@ func export(name string) string {
 	return sbuf.String()
 }
 
-// disambiguate picks a unique name, preferring name if it is already unique.
+// disambiguate picks a unique name, preferring a name in names if it is already unique.
```
zombiezen: Revert: no longer correct.

zombiezen:
Is this ready for review?

vangent:
Yes; if it's assigned to you then according to me it's ready for review :-).

Issue #78, opened 2018-06-07
wire: support loading type-checking with Go modules


zombiezen:
Wire uses [`golang.org/x/tools/go/loader`](https://godoc.org/golang.org/x/tools/go/loader) to get type information, but this does not use Go modules.  The current workaround is to run `vgo mod -vendor`, but this occasionally requires adding more imports temporarily to pin all the dependencies (especially when using `awscloud.AWS` or `gcpcloud.GCP`).  There isn't a good way to solve this until Go module tooling hits mainstream, but I'm keeping this issue here to track.

zombiezen:
No longer blocked on the go/packages issue, but now blocked on [golang.org/cl/127258](https://golang.org/cl/127258) which allows modifying the list of files before it hits the type checker.

PR #402, opened 2018-09-06
all: upgrade to Go 1.11

zombiezen:
Unblocks #380


shantuo:
Do we want to drop testing for older version (1.10)?

zombiezen:
If I understand you, are you saying we should also run a Travis job for Go 1.10? I'm not sure the complexity in the scripts (vgo versus go invocation) outweighs the benefit (support for a slightly less new Go standard library in a pre-1.0 library).

shantuo:
Yes that would be tricky, originally I was thinking of just use `go` command instead of `vgo` for pre-1.10, but we do use some of the new commands.

shantuo:
Is the Travis failure due to the module proxy?

zombiezen:
No, it was due to a still-unknown issue with Wire + the `go` tool when modules are enabled. Added a TODO, and hoping that it will be addressed by #78.

PR #413, opened 2018-09-07
internal/contributebot: add a check for issue title format

### LINKED ISSUE #316: internal/contributebot: add check for issue title format
See the [design doc](https://github.com/google/go-cloud/blob/master/internal/contributebot/DESIGN.md#issue-title-format-check) for this check.

(Continuation of #216)

vangent:
Fixes #316.


internal/contributebot/process_test.go
```diff
@@ -22,9 +22,15 @@ import (
...
 
 func TestProcessIssueEvent(t *testing.T) {
+	const (
+		defaultTitle = "foo: bar"
+	)
+
 	tests := []struct {
 		description string
 		action      string
+		title           string // defaults to defaultTitle
```
zombiezen: Just to remove branching, use `defaultTitle` explicitly in the test cases.

(Is this gofmt'd?)

internal/contributebot/process_test.go
```diff
@@ -49,6 +55,31 @@ func TestProcessIssueEvent(t *testing.T) {
...
+		},
+		{
+			description:     "edit to invalid issue title -> add comment",
+			action:          "edited",
+			title:           "prev",
+			prevTitle: "foo",
+			want: &issueEdits{
+				AddComments: []string{issueTitleComment},
+			},
+		},
```
zombiezen: Also test adding a comment not triggering the comment for the invalid issue title.

internal/contributebot/process_test.go
```diff
@@ -57,12 +88,27 @@ func TestProcessIssueEvent(t *testing.T) {
...
+			}
+			var chg *github.EditChange
+			if tc.action == "edited" {
+				chg = &github.EditChange{}
+				if tc.prevTitle != "" {
+					title := struct {
+						From *string `json:"from,omitempty"`
+					}{From: github.String(tc.prevTitle)}
+					chg.Title = &title
+				}
```
zombiezen: Optional: I'm a bit wary of this test-specific language of describing `issueData` drifting from the actual data structure over time, since then you have to understand the logic here in order to understand the actual input. However, I see the benefit in making the fields that the test cares about explicit. A few possibilities:

- Write tests for the abstract structure to the specific data structure.
- Just use the GitHub-provided data structure directly in the test cases.

It's not too bad right now, but thought I'd pass along the suggestion.

internal/contributebot/go.sum
```diff
@@ -6,6 +6,7 @@ contrib.go.opencensus.io/exporter/stackdriver v0.0.0-20180421005815-665cf5131b71
 github.com/GoogleCloudPlatform/cloudsql-proxy v0.0.0-20180321230639-1e456b1c68cb/go.mod h1:aJ4qN3TfrelA6NZ6AXsXRfmEVaYin3EDbSPJrKS8OXo=
 github.com/aws/aws-sdk-go v1.13.20/go.mod h1:ZRmQr0FajVIyZ4ZzBYKG5P3ZqPz9IHG41ZoMu1ADI3k=
 github.com/aws/aws-xray-sdk-go v1.0.0-rc.5/go.mod h1:XtMKdBQfpVut+tJEwI7+dJFRxxRdxHDyVNp2tHXRq04=
+github.com/basvanbeek/ocsql v0.0.0-20180627133915-d946a62502e3/go.mod h1:5xGI8UcldrK//AiyiYs6Qzyos4YKCml4mlUw6XuHWPE=
```
zombiezen: Unrelated to PR. Revert?

internal/contributebot/process_test.go
vangent: Done, Done.

internal/contributebot/process_test.go
vangent: Sorry, isn't that the 2nd testcase I added? (edit on an invalid issue title, but the title didn't change)? That's what "adding a comment" would look like.

internal/contributebot/process_test.go
vangent: OK. Let's see how it goes. It may be the case that eventually we are mucking with enough of the fields that it makes more sense to just create the entire struct for each testcase. Can change this later if needed.

internal/contributebot/process_test.go
zombiezen: Ah, I didn't understand that. Perfect.

PR #418, opened 2018-09-10
runtimevar: add a test to check that Watch blocks if the value hasn't changed

vangent:


runtimevar/drivertest/drivertest.go
```diff
@@ -188,6 +189,19 @@ func testString(t *testing.T, newHarness HarnessMaker) {
 	} else if gotS != content {
 		t.Errorf("got %q want %q", got.Value, content)
 	}
+
+	// A second watch should block forever since the value hasn't changed.
+	// A short wait here doesn't guarantee that this is working, but will catch
+	// most problems.
+	cancelCtx, cancel := context.WithCancel(ctx)
```
shantuo: This may be simplified to use `context.WithTimeout`.

But I think this looks flaky. Maybe call `Watch` in a goroutine and cancel the ctx in main thread, then:

```
select {
case <-ctx.Done():
    // pass
case <-errc:
    // fail
}

vangent: Changed to use WithTimeout.

I don't think it is flaky, in that it will always pass if the provider implementation does the right thing. It could pass incorrectly (i.e., if the impl is about to return a value instead of blocking, but takes > 10ms to do so), but I think it's worth having the test; in practice, if an implementation is doing this wrong, the test will fail a high percentage of the time.

I don't think the goroutine you suggested helps the above; no matter how long you wait, the impl could always just be really slow.

Issue #317, opened 2018-08-13
internal/contributebot: add check for pull request title


zombiezen:
See the [design doc](https://github.com/google/go-cloud/blob/master/internal/contributebot/DESIGN.md#issue-title-format-check) for this check.

(Continuation of #216)

PR #421, opened 2018-09-11
internal/contributebot: add a check for pull request title

vangent:


vangent:
Fixes #317

internal/contributebot/process.go
```diff
@@ -27,11 +27,13 @@ import (
 const (
 	inProgressLabel            = "in progress"
 	issueTitleComment          = "Please edit the title of this issue with the name of the affected package, followed by a colon, followed by a short summary of the issue. Example: `blob/gcsblob: not blobby enough`."
+	pullRequestTitleComment    = "Please edit the title of this pull request with the name of the affected package, followed by a colon, followed by a short summary of the change. Example: `blob/gcsblob: improve comments`."
```
shantuo: ... with the name of the affected package, or "all", followed by a colon,...

internal/contributebot/process.go
```diff
@@ -27,11 +27,13 @@ import (
...
 	branchesInForkCloseComment = "Please create pull requests from your own fork instead of from branches in the main repository. Also, please delete this branch."
 )
 
 var (
-	issueTitleRegexp = regexp.MustCompile("^[a-z0-9/]+: .*$")
+	issueTitleRegexp       = regexp.MustCompile("^[a-z0-9/]+: .*$")
+	pullRequestTitleRegexp = issueTitleRegexp
```
shantuo: "-" should be added to the regex, it's a common character in the package name.

vangent: Done.

internal/contributebot/process.go
vangent: Done.

PR #425, opened 2018-09-11
runtimevar/filevar+etcdvar: don't subscribe for changes on the first Watch call

vangent:


runtimevar/filevar/filevar.go
```diff
@@ -109,17 +109,20 @@ func (w *watcher) WatchVariable(ctx context.Context, prevVersion interface{}, pr
...
-	if notifierErr == nil {
-		defer func() {
-			_ = w.notifier.Remove(w.file)
-		}()
+	// We may never use the notifier if we read the file below and detect a
+	// change, but we must subscribe here to avoid race conditions.
+	var notifierErr error
+	if prevVersion != nil {
+		notifierErr := w.notifier.Add(w.file)
+		if notifierErr == nil {
```
ijt: Why isn't the case of notifierErr != nil handled?

runtimevar/etcdvar/etcdvar.go
```diff
@@ -58,7 +58,10 @@ func (w *watcher) WatchVariable(ctx context.Context, prevVersion interface{}, pr
 
 	// Create a watching channel in case the variable hasn't changed.
 	// We must create it now before the Get to avoid race conditions.
-	ch := w.client.Watch(ctx, w.name)
+	var ch clientv3.WatchChan
```
ijt: Shouldn't there be an additional conformance test for the changes in this pull request?

ijt:
What is the motivation for this change?
Performance?
Correctness?
What goes wrong without it?

vangent:
> What is the motivation for this change?
> Performance?
> Correctness?
> What goes wrong without it?

Performance. It doesn't make sense to initialize a subscriber/notifier if you know in advance you're not going to use it.

runtimevar/etcdvar/etcdvar.go
vangent: No, there's no change in behavior, we're just avoiding some unnecessary work.

runtimevar/filevar/filevar.go
vangent: It is handled down below, if we end up needing the notifier, see line 147.

ijt:
So should there be a benchmark to show that it improves performance?

vangent:
> So should there be a benchmark to show that it improves performance?

IMHO that's not necessary. We don't have a benchmark today, and it might be hard to show improvement anyway (how long does it take to call fsnotify?). The etcd one saves an RPC, but since the test runs against a local instance, it probably doesn't take long either.

ijt:
You could add the benchmark in a separate commit, then reorder the commits
(use git rebase -i) to put the benchmark before this change, then run the
benchmark with and without the change. Otherwise we don't know if it makes
the code run any faster.

vangent:
I don't think it will make the code noticeably faster running locally. I think it's still good to avoid unnecessary operations though.

ijt:
In that case, I suggest removing the optimization for the local version.

vangent:
> In that case, I suggest removing the optimization for the local version.
Why? It seems silly to create an fsnotifier that we know we're not going to use, and I think it actually makes the code more understandable since it points out that we're only going to use the notifier if we've already returned a value, and it hasn't changed since then.

Issue #429, opened 2018-09-11
runtimevar/etcd: import go.etcd.io/etcd instead of github.com/coreos/etcd


vangent:
github.com/coreos/etcd has renamed to use package name go.etcd.io/etcd. They are two different packages, and hence some of the packages stops working with go get. could you change to import go.etcd.io/etcd? Thanks.

vangent:
(filing for @h8liu based on a [comment](https://github.com/google/go-cloud/commit/0d586fec90b6c38e17b5294793a240cded9ff4a5#commitcomment-30470483) on a pr)

vangent:
Per @h8liu 

Here is an example of build failure trying to build the etcd repo with github.com/coreos/etcd import path:

https://gopkgs.io/builds/20180911-195307-3d66b81f

vangent:
Yea, I did try to use go.etcd.io/etcd at first, but it keeps pulling in the old coreos/etcd as well, and then I get compatibility errors:

`unknown field 'NoFreelistSync' in struct literal of type bolt.Options`

https://github.com/google/go-cloud/compare/master...vangent:etcd-rename shows the go.mod changes that result.

I think we'd prefer to wait until there's an official etcd release with the new name, then we can switch to use it.

vangent:
Blocked on a version after 3.3.9:
https://github.com/etcd-io/etcd/releases

vangent:
I tried again with 3.3.10, but got the same problems with coreos things getting pulled in.

PatrickHuang888:
I tried go.etcd.io/etcd v3.3.10+incompatible // indirect got  error
```
unknown field 'NoFreelistSync' in struct literal of type bolt.Options
```

PR #430, opened 2018-09-11
using go.etcd.io/etcd for etcd import path

h8liu:
for #429

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,8 +23,7 @@ require (
 	github.com/basvanbeek/ocsql v0.0.0-20180627133915-d946a62502e3
 	github.com/beorn7/perks v0.0.0-20180321164747-3a771d992973 // indirect
 	github.com/census-ecosystem/opencensus-go-exporter-aws v0.0.0-20180411051634-41633bc1ff6b
-	github.com/coreos/bbolt v1.3.1-coreos.6 // indirect
-	github.com/coreos/etcd v3.3.9+incompatible
+	github.com/coreos/etcd v3.3.9+incompatible // indirect
 	github.com/coreos/go-semver v0.2.0 // indirect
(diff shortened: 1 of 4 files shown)


h8liu:
interesting that your ci is using an internal GOPROXY? is it managed by your team?

vangent:
> interesting that your ci is using an internal GOPROXY? is it managed by your team?

Yup.

gyuho:
Hi, we are switching to go module in master branch soon. But, maybe not in release branches. Do you need us use go module for release branch as well?

h8liu:
I do not really use your code, I am just seeing the build breakages from gopkgs.io. :)

it depends on if you want to support `go get` with go 1.10 and earlier in any flavor. their are cloud infra with CI/CD that are still staying behind using older go compiler versions. making master module only now is basically forcing your users to upgrade their entire setup to 1.11 (or vendor some older version).

Here is an example of another widely used library that starts using module but then force to revert back the import path change due to user's complaints:

https://github.com/pierrec/lz4/issues/33

h8liu:
(I am actually not a member of go-cloud)

so quoting from:

https://github.com/etcd-io/etcd/issues/10083

> etcd is a binary program not a library. We have been talking about how to separate the client library versioning from the etcd binary versioning, but have not found a good way to do it.

I got this reply when I asked them to bump the version to 4.0.0 because import path change is not backward compatible. They refused because their version number is for the binary, not for the source code (client api). As a result, using release tags from the etcd git repo in go mod is probably incorrect.

vangent:
Can we close this PR and revisit once etcd has done a new release with the new paths (whatever they end up tagging it)?

h8liu:
sure, whatever.

again, no matter what version they use, it is not for go module import. the version is for the binary not the library.

PR #434, opened 2018-09-12
internal/docs: add documentation about how to do a release

vangent:


internal/docs/release.md
```diff
@@ -0,0 +1,31 @@
...
+
+```bash
+$ TAG=v0.n.0
+$ COMMIT=aaaaaa
+$ git tag -a ${TAG} ${COMMIT}
+# Enter your release notes in your editor.
+$ git push upstream ${TAG}
+```
+
+-   Go to [Releases](https://github.com/google/go-cloud/releases). Click `Draft
```
zombiezen: IIRC you can create the tag during this step through the GitHub UI.

internal/docs/release.md
```diff
@@ -0,0 +1,31 @@
...
+    a new release` and select your new tag. Enter the release notes again.
```
zombiezen: Again?

internal/docs/release.md
```diff
@@ -0,0 +1,31 @@
...
+
+-   Send an email to the external mailing list announcing the release, and
```
zombiezen: s/external mailing list/go-cloud@googlegroups.com/ (and linkify the group to the Google Groups UI)

internal/docs/release.md
```diff
@@ -0,0 +1,31 @@
+# Releases
+
+We aim to do a release after each Sprint (roughly, every 3 weeks).
+
+To do a release:
+
+-   Identify the [commit](https://github.com/google/go-cloud/commits/master) you
```
zombiezen: I've found the slightly easier view is https://github.com/google/go-cloud/compare and then plugging in the latest version tag as the base. You can find the latest release by visiting the magic URL https://github.com/google/go-cloud/releases/latest, but that might be too much magic. 😀

internal/docs/release.md
vangent: You can, but only if the commit you want is the last 5 or 10 commits. Modified.

internal/docs/release.md
vangent: That's what I did. I'm not sure what the UI does if you Draft the release all in one step; I see notes on both the previous tags and the previous releases.

Maybe we can just leave the notes for the Tag empty, and just have them on the release?

vangent: Changed to use "-m ${TAG}" for `git tag`.

internal/docs/release.md
vangent: Added a link to that.

PR #435, opened 2018-09-12
internal: re-run new wire internally

vangent:


vangent:
This is just building `wire` from HEAD and running it against all the places we have `wire_gen.go` in the repo.

PR #441, opened 2018-09-13
internal: run go mod tidy

vangent:
I verified that all of the new dependencies have "nonotice' licenses (either MIT, or Apache 2.0, or BSD 3-clause).


vangent:
(go-cloud-bot fail)

PR #442, opened 2018-09-13
AUTHORS: fix Google LLC entry

FiloSottile:
"The Go Cloud Authors" are the ones specified by this list, so the existing entry was self-referencing. While the paradox might be an interesting way to keep away robot lawyers, the copyright owner for the Googlers contributions is Google LLC.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -11,8 +11,8 @@
 # Please keep the list sorted.
 
 Ben Hinchley <benjaminhinchley@gmail.com>
-The Go Cloud Authors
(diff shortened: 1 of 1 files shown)


zombiezen:
A rogue `sed` call on my part, no doubt. Good catch. 😀

FiloSottile:
> Please create pull requests from your own fork instead of from branches in the main repository. 

This was a branch in my own fork.

vangent:
Yes, sorry, that was a go-cloud-bot fail.

PR #444, opened 2018-09-13
internal/contributebot: update readme's deployment steps

vangent:


vangent:
(contributebot fail)

internal/contributebot/README.md
```diff
@@ -77,18 +77,24 @@ go run . --project=your-project-name --github_app=42 --github_key=/foo.pem
 
 ## Deploying
 
+To deploy an updated Contribute Bot to production, follow these steps.
```
zombiezen: Since it is reasonable to want to stage an infrastructure change, also include some verbage on how to change the process to fit your dev environment?

internal/contributebot/README.md
```diff
@@ -77,18 +77,24 @@ go run . --project=your-project-name --github_app=42 --github_key=/foo.pem
...
 ```shell
 # Build Docker image.
-gcloud builds submit --config cloudbuild.yaml ../..
+gcloud builds submit --config cloudbuild.yaml ../.. --project=go-cloud-contribute-bot
 
 # Edit k8s/contributebot.yaml and replace placeholders surrounded by double
 # curly braces (e.g. {{PROJECT_ID}}) with the appropriate values.
+# -- {{PROJECT_ID}} should be set to "go-cloud-contribute-bot"
+# -- {{BUILD_ID}} should be set to the ID from the output of the "gcloud builds"
+#    command above, something like "e892c5f0-a7d3-4a16-a782-eb621afc9695".
```
zombiezen: Optional, an idea I've been tossing around: perhaps instead of a template YAML file, the YAML file that is checked in actually is the production configuration.

internal/contributebot/README.md
vangent: Done.

internal/contributebot/README.md
vangent: I like it. Done.

PR #445, opened 2018-09-13
internal/contributebot: check whether a PR is from a fork directly

shantuo:
Instead of comparing the repo name.

### DIFF
--- a/internal/contributebot/process.go
+++ b/internal/contributebot/process.go
@@ -161,8 +161,8 @@ func processPullRequestEvent(data *pullRequestData) *pullRequestEdits {
 	defer log.Printf("-> %v", edits)
 	pr := data.PullRequest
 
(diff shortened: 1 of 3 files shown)


internal/contributebot/process.go
```diff
@@ -160,8 +161,9 @@ func processPullRequestEvent(data *pullRequestData) *pullRequestEdits {
 	defer log.Printf("-> %v", edits)
 	pr := data.PullRequest
 
-	// If the pull request is from a branch of the main repo, close it and request that it come from a fork instead.
-	if data.Action == "opened" && pr.GetHead().GetRepo().GetName() == data.Repo {
+	// If the pull request is from a branch of the main repo, close it and request
```
vangent: from a branch of the main repo -> not from a fork 

in the comment.

shantuo: Done

PR #447, opened 2018-09-13
internal/contributebot: skip processing when the PR is closed

### LINKED ISSUE #446: internal/contributebot: skip processing when the PR is closed

shantuo:
Fixes #446


internal/contributebot/process.go
```diff
@@ -161,6 +161,13 @@ func processPullRequestEvent(data *pullRequestData) *pullRequestEdits {
 	defer log.Printf("-> %v", edits)
 	pr := data.PullRequest
 
+	// Skip the process when the PR is closed, we check this here instead of when
```
vangent: Can you add a similar check in processIssueEvent? It will have to be after the in-progress label check, since that is explicitly for closed issues. The rest of the checks in there should be skipped for closed issues though.

internal/contributebot/process.go
```diff
@@ -161,6 +161,13 @@ func processPullRequestEvent(data *pullRequestData) *pullRequestEdits {
...
+	// calling processPullRequest so that it is easier to add any process in future
+	// for closed PR's.
+	if pr.GetState() == "closed" {
+		return edits
```
vangent: Can you add a testcase for this?

internal/contributebot/process.go
shantuo: Done.

internal/contributebot/process.go
shantuo: Done.

Issue #291, opened 2018-08-07
blob: explicit multipart/resumable upload support


nbari:
Could "multipart upload" be implemented for the blob storage?

zombiezen:
Relevant docs:

- GCS: https://cloud.google.com/storage/docs/json_api/v1/how-tos/multipart-upload
- S3: https://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadAbort.html

zombiezen:
Can you clarify what support you want, @nbari? From what I can tell, a multipart upload on GCS means that it adds metadata (which we support for the current attributes). A multipart upload on AWS seems to mean what GCS calls resumable, is that what you are after? (#222 might be related.)

nbari:
I was searching for a way to upload a file and been available to resume the upload if for some reason the connection went down, I found that this could be done by using "multipart upload" but this implies to do a "Multipart Upload Initiation" and then a "Parts Upload".

Following the current example ([Writing the file to the bucket](https://github.com/google/go-cloud/tree/master/samples/tutorial#writing-the-file-to-the-bucket)):

```go
package main

// No new imports.

func main() {
    // ...

    w, err := b.NewWriter(ctx, file, nil)
    if err != nil {
        log.Fatalf("Failed to obtain writer: %s", err)
    }
    _, err = w.Write(data)
    if err != nil {
        log.Fatalf("Failed to write to bucket: %s", err)
    }
    if err := w.Close(); err != nil {
        log.Fatalf("Failed to close: %s", err)
    }
}
```

I found that there is a `NewWriter`,  `Write` and `Close`  but probably could be good to have a `NewMultipartWriter`, `WritePart`  `abortMultipartWriter`.

Also if I am correct when trying to do a "multipart upload" there are some [limits](https://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html), for example:


Item | Specification
-- | --
Maximum number of parts per upload | 10,000
Part numbers | 1 to 10,000 (inclusive)
Part size | 5 MB to 5 GB, last part can be < 5 MB
Maximum number of parts returned for a list parts request | 1000
Maximum number of multipart uploads returned in a list multipart uploads request | 1000


Going furthere with the implementation the idea could be to do internaly the mulipart upload and have the method handle the  logic of spliting  the file, reading from disk the chunks/parts, upload concurrently, etc, probably  something like:

```go
MultipartUpload(file string,  cuncurent_uploads, max_parts, retries int)
```

[...]

PR #463, opened 2018-09-17
s3blob: remove unused bufferSize option

vangent:


vangent:
Issue #291.

vangent:
Wait, I think there is actually an option for this we should be using. Will send a separate PR.

PR #468, opened 2018-09-19
blob/gcsblob/test: use a test bucket from a shared project

vangent:
I created a new project with a bucket, and made go-cloud-eng@ owners. This way we don't have to rewrite all of the golden files whenever we add a new one. This is similar to what we have for S3.

Can you also verify that you can run with "-record" with this bucket name?


vangent:
Just shared among go-cloud-eng@google.com. External contributors will still either need to create their own bucket and update this (plus all goldens), same as before, or ask us to record the goldens.

PR #469, opened 2018-09-19
wire: make Vendor test work with Go module

### LINKED ISSUE #326: wire: investigate how vendor works when module is turned on

shantuo:
1. vendor directory is used only at the top level with Go module turned on and "-mod=vendor" is used.
2. The dummy bar package needs to be removed or `go build` complains ambiguity.

Fixes #326


wire/internal/wire/wire_test.go
```diff
@@ -142,7 +138,7 @@ func goBuildCheck(test *testCase, wd string, bctx *build.Context, gen []byte) er
 	if err != nil {
 		return err
 	}
-	defer os.RemoveAll(gopath)
+	// defer os.RemoveAll(gopath)
```
vangent: Uncomment this?

shantuo: Sorry forgot to uncomment it.

PR #471, opened 2018-09-19
docs: add a section on escape hatches to design.md

vangent:


internal/docs/design.md
```diff
@@ -141,6 +141,94 @@ canonical example is `gcpkms` and `awskms`.
...
+wants to access provider-specific functionality, which might consist of:
+
+1.  **Top-level APIs**. For example, `blob` does not expose a `Copy`, but some
+    provider might.
+1.  **Data fields**. For example, **blob** exposes a few attributes like
+    ContentType and Size, but S3 and GCS both have many more.
+1.  **Options**. Different providers may support different options for
+    functionality.
+
+Escape hatches provide the user a way to escape the Go Cloud abstraction to
```
zombiezen: Since you are introducing a term, maybe emphasize it?

internal/docs/design.md
```diff
@@ -141,6 +141,94 @@ canonical example is `gcpkms` and `awskms`.
...
+difficult to represent them in a generic way (although, we have not tried).
+
+Using an escape hatch implies that the resulting code is no longer portable; the
+escape hatched code will need to be ported in order to switch providers.
+Therefore, they should be avoided if possible.
+
+### Ways To Escape Hatch
+
+Users can always access the provider service directly, by constructing the
+top-level handle and making API calls bypassing Go Cloud.
```
zombiezen: Comma between "calls" and "bypassing".

internal/docs/design.md
```diff
@@ -141,6 +141,94 @@ canonical example is `gcpkms` and `awskms`.
...
+
+*   For top-level APIs, this may be fine, although it might require a bunch of
```
zombiezen: "top-level operations", maybe?

internal/docs/design.md
```diff
@@ -141,6 +141,94 @@ canonical example is `gcpkms` and `awskms`.
...
+approach. The proposal is to extend Go Cloud APIs, at the top-level (e.g.
+`blob.Bucket`) and in data objects, to expose provider-specific handles. For
+example:
+
+```
+// The existing blob.Reader exposes some blob attributes, but not everything
+// that every provider exposes.
+type Reader struct {
+    ...
+    Size() int64
```
zombiezen: Not methods, this is a struct.

internal/docs/design.md

internal/docs/design.md

internal/docs/design.md

internal/docs/design.md

Issue #222, opened 2018-07-24
blob: bucket writer has no 'abandon' or 'discard' functionality


EdSchouten:
When inserting a large blob into a bucket, there is often a desire to stream a value from some source (e.g., a large file on disk). This process may fail halfway, meaning there should be some way to discard a partially written file.

The [`Writer` object](https://godoc.org/github.com/google/go-cloud/blob#Writer) only has a `Write()` and `Close()` function, but no method to cancel the write itself. Should such a function be added? Noteworthy: The Minio and S3 libraries solve this by not having a `Writer`-like object, but instead have a function that takes a `Reader` and consumes the data.

ijt:
I'm thinking the proposed method should attempt to revert the writer object and the bucket back to the state right after calling `bucket.NewWriter()`. With that in mind, it seems like a better name for the method would be `Reset()`.

ijt:
@zombiezen suggested offline that we call the method `Abort()` and have it put the object into a terminal state such that it cannot be used anymore. He pointed out that this is easier to test and would allow associated resources to be freed up. I agree.

ijt:
I tried a straightforward approach of just deleting the blob that has been written so far, but it raised a question that I don't see how to answer without adopting the approach in issue #353. Suppose a write has failed due to a network partition and we want to abort. If the network partition persists then it won't work to delete what has already been written to the cloud. So what's already up there will continue to be in its partially finished state. The question is: how should we deal with this?

If we add an atomicblob package supporting ACID semantics then we can get an `Abort` method that doesn't have this defect. The contents for the key to be written won't be updated until `Close` is called, so the only thing that could be left in an incomplete state would be the blob contents of a temporary key.

ijt:
I'm backing off from this issue to focus on pub/sub. If someone else wants to take this on while I work on that, I'm happy to share my work-in-progress branch and context. Otherwise, I'll return to this later.

zombiezen:
Summarizing from offline conversation with @ijt:

At least on GCS and S3, this ability to stop an in-progress upload already exists natively in the providers (although I am hazy on particular mechanisms. @shantuo, do you know how these work?). For `fileblob`, we can get the atomic semantics by using temporary files and moving them on successful `Close`. However, the blob driver needs to add a new method. At this early stage, it's fine for us to just add this as a required method of a blob driver (a breaking change).

msample:
[...]

PR #473, opened 2018-09-20
blob: cancel the Writer ctx to abort a write

vangent:
The updated dnaeon/go-vcr is to pick up a fix I submitted to it:
https://github.com/dnaeon/go-vcr/pull/34#issuecomment-423033367

### DIFF
--- a/blob/drivertest/drivertest.go
+++ b/blob/drivertest/drivertest.go
@@ -479,9 +479,11 @@ func testCanceledWrite(t *testing.T, newHarness HarnessMaker, pathToTestdata str
 			}
 			// Cancel the context to abort the write.
 			cancel()
-			// Close should return context's Err().
-			if err := w.Close(); false && err != cancelCtx.Err() {
-				t.Errorf("got Close error %v want %v", err, cancelCtx.Err())
+			// Close should return some kind of canceled context error.
+			// We can't verify the kind of error cleanly, so we just verify there's
+			// an error.
+			if err := w.Close(); err == nil {
+				t.Errorf("got Close error %v want canceled ctx error", err)
 			}
 			// A Read of the same key should fail; the write was aborted
 			// so the blob shouldn't exist.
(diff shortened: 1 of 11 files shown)


vangent:
Fixes #222.

vangent:
BTW, the .tar file I added for the chunked test contains 250 copies of test-large.jpg.

blob/drivertest/drivertest.go
```diff
@@ -414,6 +417,81 @@ func testWrite(t *testing.T, newHarness HarnessMaker, pathToTestdata string) {
...
+				t.Fatal(err)
+			}
+			// Write a few bytes.
+			if _, err := w.Write(test.content[:len(test.content)-5]); err != nil {
+				t.Fatal(err)
+			}
+			// Cancel the context to abort the write.
+			cancel()
+			// Close should return context's Err().
+			if err := w.Close(); false && err != cancelCtx.Err() {
```
zombiezen: `false &&` no-ops?

vangent: Ha ha, good catch. Forgot to remove that after debugging.
Unfortunately we can't verify that the error is actually some kind of "canceled context" error (sometimes it's wrapped).
So, removed the "false &&" and now just verifying that it's non-nil.

zombiezen: SGTM!

blob/drivertest/drivertest.go
```diff
@@ -414,6 +417,83 @@ func testWrite(t *testing.T, newHarness HarnessMaker, pathToTestdata string) {
...
+			b, err := h.MakeBucket(ctx)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			// Create a writer with the context that we're going
+			// to cancel.
+			opts := &blob.WriterOptions{
+				ContentType: test.contentType,
+				BufferSize:  8 * 1024 * 1024,
```
shantuo: Add a comment saying this should be the minimum allowed buffer size and varies between providers. So that when another provider is added, they can adjust it to have the expected behavior.

vangent: (in PR #476 )

Issue #475, opened 2018-09-20
blob/test: TestCanceledWrite/ChunkedUpload is flaky on Travis


vangent:
https://travis-ci.com/google/go-cloud/jobs/147073872

--- FAIL: TestConformance (0.63s)
    --- FAIL: TestConformance/TestCanceledWrite (0.19s)
        --- FAIL: TestConformance/TestCanceledWrite/ChunkedUpload (0.13s)
            replay.go:57: Replaying from golden file testdata/TestConformance/TestCanceledWrite/ChunkedUpload
            replay.go:104: matched request #0 (POST https://go-cloud-bucket.s3.us-east-2.amazonaws.com/blob-for-canceled-write?uploads=)
            replay.go:77: mismatched Method at request #1; got "PUT" want "GET"
            replay.go:77: mismatched Method at request #1; got "GET" want "POST"
FAIL
FAIL	github.com/google/go-cloud/blob/s3blob	0.647s

vangent:
Test is disabled now; I'll see if I can figure out if there's a way to enable it without it being flaky. There may not be.

PR #476, opened 2018-09-20
blob/test: disable TestCanceledWrite/ChunkedUpload for now; it is flaky on Travis

vangent:
Issue #475


vangent:
Yes, I know the cause. We may a large write that the provider is going to break up into multiple chunks. Then we cancel the ctx. Sometimes the provider has time to make the HTTP request for the first chunk, sometimes it doesn't.

PR #479, opened 2018-09-21
runtimevar: always obey "wait", to avoid double-get for pollers

### LINKED ISSUE #439: runtimevar: always obey "wait", to avoid double-get for pollers
Polling implementations today do something like this in WatchVariable:

-- Get the variable
-- If no change, return (nil, <poll interval>); else return (<new blob>, 0).

The concrete type only enforces <wait> (the 2nd return value) if there was no change. Note that if the variable changes, and the end user calls Watch() again immediately (as expected), this results in 2 gets in rapid succession.

It would be simpler and more efficient to just always obey <wait>.

vangent:
Fixes #439.

This avoids polling implementations (like paramstore and runtimeconfigurator) from retrieving the variable twice every time it changes.

Previously, it would go something like this:

- User calls Watch, which calls driver.WatchVariable
- WatchVariable fetches value; first time, so it's new -> returned to user
- User calls Watch again, which calls driver.WatchVariable
- WatchVariable fetches value; it's the same -> returns nil; concrete types waits <poll interval> before calling WatchVariable again.

--> Note the two fetches.

Now, the concrete type always obeys the returned wait, and the polling implementations always return it. So it looks like this:

- User calls Watch, which calls driver.WatchVariable
- WatchVariable fetches value; first time, so it's new -> returned to user
- User calls Watch again, which waits for <poll interval> before calling driver.WatchVariable
- WatchVariable fetches value; it's the same -> returns nil; concrete types waits <poll interval> before calling WatchVariable again.

--> Note one fetch of the value per poll interval.


runtimevar/drivertest/drivertest.go
```diff
@@ -26,10 +26,17 @@ import (
 	"github.com/google/go-cmp/cmp"
 )
 
+const (
+	noWait = 1 * time.Microsecond
```
shantuo: These names are confusing, why `noWait` is 1us while `defaultWait` is 0? Sounds like `noWait` should be 0?

runtimevar/etcdvar/etcdvar.go
```diff
@@ -28,17 +28,29 @@ import (
...
+const defaultWait = 30 * time.Second
 
 // New constructs a runtimevar.Variable object that uses client to watch
 // variables in etcd.
 // Provide a decoder to unmarshal updated configurations into similar
 // objects during the Watch call.
-func New(name string, cli *clientv3.Client, decoder *runtimevar.Decoder) (*runtimevar.Variable, error) {
+func New(name string, cli *clientv3.Client, decoder *runtimevar.Decoder, opts *WatchOptions) (*runtimevar.Variable, error) {
+	if opts == nil {
+		opts = &WatchOptions{}
```
shantuo: nit:

```
if opts == nil {
  opts = &WatchOptions{ WaitTime: defaultWait }
}
```

And you can simplify the switch to be a single `if waitTime < 0`

runtimevar/etcdvar/etcdvar.go
```diff
@@ -130,3 +143,9 @@ func (w *watcher) WatchVariable(ctx context.Context, prev driver.State) (driver.
 func (w *watcher) Close() error {
 	return nil
 }
+
+// WatchOptions allows the specification of various options to a watcher.
+type WatchOptions struct {
+	// WaitTime controls the frequency of retries after an error. Defaults to 30 seconds.
```
shantuo: nit: "...Default to defaultWait"?

runtimevar/paramstore/paramstore_test.go
```diff
@@ -12,11 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package paramstore
```
shantuo: This can be removed as it's the same package paramstore

vangent:
> A general question, should defaultWait be a behavior Go Cloud enforces?
> In another word, should it be passed in from the concrete type to the driver

Not sure exactly what you mean. Do you mean that the concrete type would define WatchOptions with a WaitTime, take it in the constructor, and pass the WaitTime to driver.WatchVariable?

Or do you mean that runtimevar should define a DefaultWait = 30 * time.Second, and provider implementations can continue to have their own WatchOptions, but can use runtimevar.DefaultWait instead of defining their own?

runtimevar/drivertest/drivertest.go
vangent: Good point. I renamed `noWait` to `shortWait` and added comments.

defaultWait is 0, which causes driver implementations to use their default (usually 30 seconds).

runtimevar/etcdvar/etcdvar.go
vangent: I copied this from the other implementations, so I'd rather leave it consistent.

I don't think we could remove the "waitTime == 0" from the switch, because someone could pass &WatchOptions{}, which is not nil but also has a 0 WaitTime that we need update the default for. For example, many of the conformance tests do this now.

runtimevar/etcdvar/etcdvar.go
vangent: defaultWait is a private var that's not visible in the godoc.
For example, see the existing godoc [here](https://godoc.org/github.com/google/go-cloud/runtimevar/runtimeconfigurator), the docstring references `defaultWait` but you can't see it anywhere.

runtimevar/paramstore/paramstore_test.go
vangent: Done.

vangent:
Thanks for the quick review! PTAL.

shantuo:
What I meant was that the concrete type defines a WatchOptions with a WaitTime, and a defaultWait, user can choose provide a custom WaitTime or leave it. The concrete type will use the user provided or defaultWait and pass it into drvier.WatchVariable. The driver.WatchVariable can always expect to receive a meaningful WaitTime.

vangent:
> the concrete type defines a WatchOptions with a WaitTime...

Yes, we could do that. It would save a bunch of duplicate code, at the downside of exposing an option that isn't always used (e.g., constantvar wouldn't use WaitTime). I think it's a separate suggestion from what's happening in this issue/PR though. If you like the idea, can you file a new Issue?

PR #487, opened 2018-09-24
wire: report an error if a func with wire.Build in it is an invalid injector

### LINKED ISSUE #428: wire: functions calling wire.Build after assigning a variable are silently erased in output
This came up in the following situation:
```go
// setupGCP is a Wire injector function that sets up the application using GCP.
func setupGCP(ctx context.Context, flags *cliFlags) (*application, func(), error) {
        traceOpt := withNoTraceOptions()
        // This will be filled in by Wire with providers from the provider sets in
        // wire.Build.
        wire.Build(
                gcpcloud.GCP,
                applicationSet,
                gcpBucket,
                gcpMOTDVar,
                gcpSQLParams,
                wire.Value(traceOpt),
        )
        return nil, nil, nil
}

func withNoTraceOptions() ocsql.TraceOption {
        return func(o *ocsql.TraceOptions) {}
}
```
Running `go generate` here resulted in `wire_gen.go` not containing a `setupGCP` function anymore, because of the assignment to traceOpt. 

This assignment was made necessary by Wire's rejection of `wire.Value(withNoTraceOptions())` since it contains a function call and is therefore too complex by Wire's standards. Maybe the error message should suggest that such assignments be moved out of the injector function and placed in package scope.

vangent:
Fixes #428 .


wire/internal/wire/parse.go
```diff
@@ -773,50 +777,57 @@ func processInterfaceValue(fset *token.FileSet, info *types.Info, call *ast.Call
 // isInjector checks whether a given function declaration is an
 // injector template, returning the wire.Build call. It returns nil if
 // the function is not an injector template.
-func isInjector(info *types.Info, fn *ast.FuncDecl) *ast.CallExpr {
+func isInjector(info *types.Info, fn *ast.FuncDecl) (*ast.CallExpr, error) {
```
zombiezen: Perhaps a better name for this function is `findInjectorBuild` or some such.

wire/internal/wire/wire.go
```diff
@@ -113,7 +117,7 @@ func copyNonInjectorDecls(g *gen, files []*ast.File, info *types.Info) {
 		for _, decl := range f.Decls {
 			switch decl := decl.(type) {
 			case *ast.FuncDecl:
-				if isInjector(info, decl) != nil {
+				if buildCall, _ := isInjector(info, decl); buildCall != nil {
```
zombiezen: It's not obvious why the error can be ignored: add a comment that any error cases would already be filtered out.

wire/internal/wire/parse.go

wire/internal/wire/wire.go
vangent: Done.

Issue #490, opened 2018-09-27
wire: document how to bind interfaces to interfaces


oceanful:
Although returning interfaces from functions is not idiomatic in Go, it would be nice to support it in Wire.

Example use case:

```golang
// Consumer
package foo

type Foo interface {
  ...
}

// Producer 1
package one

type FooOne foo.Foo

func New() FooOne {
  ...
}

var Set = wire.NewSet(New)

// Producer 2
package two

type FooTwo foo.Foo

func New() FooTwo {
  ...
}

var Set = wire.NewSet(New)

// Somewhere else
package composite

type FooComposite foo.Foo

func New(fooOne one.FooOne, fooTwo two.FooTwo) FooComposite {
  ...
}

var SuperSet = wire.NewSet(
    one.Set,
    two.Set,
    New,
    // This currently doesn't work; there is no way to bind an interface to another interface.
    wire.Bind(new(foo.Foo), new(FooComposite)))
```

Perhaps this is what `wire.InterfaceValue()` is intended for? (I wasn't able to get it to work.)

vangent:
Hi @oceanful, unless I'm missing something, this should work.

You can either use an injector that returns FooComposite, and leave the `wire.Bind` out of `SuperSet`: 

```
var SuperSet = wire.NewSet(one.Set, two.Set, New)
func injectComposite() FooComposite {
	wire.Build(SuperSet)
	return nil
}
```

Or, you can use an injector that returns `foo.Foo` and add a `*` to your `wire.Bind`. Without the `*`, you are binding a pointer to the interface, not the interface itself. `wire` gives the error `*example.com/composite.FooComposite does not implement *example.com/foo.Foo`; suggestions on how to make that more clear are welcome.

```
var SuperSet2 = wire.NewSet(one.Set, two.Set, New, wire.Bind(new(foo.Foo), *new(FooComposite)))
func injectComposite2() foo.Foo {
	wire.Build(SuperSet2)
	return nil
}
```

You can see the full example here:
https://github.com/vangent/go-cloud/tree/xxx/wire/internal/wire/testdata/ExampleForIssue490/composite

oceanful:
Thanks Robert! Adding the `*` to the implementation parameter does exactly what I need.

In terms of how to make things more clear, perhaps you can add an example in the [Binding Interfaces](https://github.com/google/go-cloud/tree/master/wire#binding-interfaces) section of the README that details an example of binding one interface to another.

Thanks for this awesome library!

[...]

PR #491, opened 2018-09-27
wire: fix error messages for Bind and InterfaceValue when the arg doesn't implement the interface

vangent:
Previously, the error message included the arg type, which is a pointer to an interface. E.g., `string does not implement *io.Reader`, instead of `io.Reader`.

Added a test for this, and a couple of other tests for invalid wire.Bind and wire.InterfaceValue calls.

Added an example of how to bind an interface to another interace.


vangent:
Fixes #490.

PR #496, opened 2018-09-28
blob: Add support for key/value Metadata

### LINKED ISSUE #324: blob: Add support for key/value metadata strings for blobs
GCS has "labels", S3 has "metadata".

vangent:
Fixes #324.

The most controversial thing in this PR is force-lowercasing the Metadata keys during reading and writing. Provider behavior is inconsistent on this. I think both S3 and GCS store keys case-sensitive, but GCS claims it is case-insensitive 

- GCP claims to be case-insensitive (search for "Field names are case-insensitive" [here](https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata)). However, the backend is actually case-sensitive. The JSON and XML-based APIs handle casing, but the HTTP one does not. Further discussion [here](https://github.com/GoogleCloudPlatform/google-cloud-go/issues/1132).

- S3 claims to store them in lowercase (search for "user-defined metadata keys in lowercase" [here](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingMetadata.html#object-metadata)). However, the HTTP API actually returns them capitalized due to Go's header canonicalization; see [this long-standing issue](https://github.com/aws/aws-sdk-go/issues/445) complaining about it. It suggests parsing the headers yourself, but the original case can't be recovered that way so I'm not sure how that helps.

So, I think the best way to make the behavior consistent across providers and simple for users is to force-lowercase the metadata keys during reading and writing. There might still be unexpected behavior if:
1. The provider stores keys case-sensitive
2. The user is writing metadata using some mechanism other than Go Cloud
In this scenario, multiple case-sensitive keys will be collapsed into a single key when reading using Go Cloud. Note that this would happen anyway for S3, at least when using the HTTP client.

### DIFF
--- a/blob/driver/driver.go
+++ b/blob/driver/driver.go
@@ -58,6 +58,8 @@ type WriterOptions struct {
 	// write in a single request, if supported. Larger objects will be split into
 	// multiple requests.
 	BufferSize int
+	// Metadata holds key/value strings to be associated with the blob.
+	Metadata map[string]string
 }
 
 // ReaderAttributes contains a subset of attributes about a blob that are
@@ -76,6 +78,9 @@ type ReaderAttributes struct {
 type Attributes struct {
 	// ContentType is the MIME type of the blob object. It must not be empty.
 	ContentType string
+	// Metadata holds key/value pairs associated with the blob.
+	// Keys are case-insensitive and will always be returned in lowercase.
+	Metadata map[string]string
 	// ModTime is the time the blob object was last modified.
 	ModTime time.Time
 	// Size is the size of the object in bytes.
--- a/blob/driver/driver.go
+++ b/blob/driver/driver.go
@@ -59,6 +59,7 @@ type WriterOptions struct {
 	// multiple requests.
 	BufferSize int
 	// Metadata holds key/value strings to be associated with the blob.
+	// Keys are guaranteed to be non-empty and lowercased.
 	Metadata map[string]string
 }
 
@@ -79,7 +80,10 @@ type Attributes struct {
 	// ContentType is the MIME type of the blob object. It must not be empty.
 	ContentType string
 	// Metadata holds key/value pairs associated with the blob.
-	// Keys are case-insensitive and will always be returned in lowercase.
+	// Keys will be lowercased by the concrete type before being returned
+	// to the user. If there are duplicate case-insensitive keys (e.g.,
+	// "foo" and "FOO"), only one value will be kept, and it is undefined
+	// which one.
 	Metadata map[string]string
 	// ModTime is the time the blob object was last modified.
 	ModTime time.Time
--- a/blob/drivertest/drivertest.go
+++ b/blob/drivertest/drivertest.go
@@ -493,6 +493,12 @@ func testMetadata(t *testing.T, newHarness HarnessMaker) {
 			wantErr:  true,
 		},
 		{
+			name:     "duplicate case-insensitive key fails",
+			content:  hello,
+			metadata: map[string]string{"abc": "foo", "aBc": "bar"},
+			wantErr:  true,
+		},
+		{
 			name:    "valid metadata",
 			content: hello,
 			metadata: map[string]string{
--- a/blob/drivertest/drivertest.go
+++ b/blob/drivertest/drivertest.go
@@ -59,6 +59,9 @@ func RunConformanceTests(t *testing.T, newHarness HarnessMaker, pathToTestdata s
 	t.Run("TestCanceledWrite", func(t *testing.T) {
 		testCanceledWrite(t, newHarness, pathToTestdata)
 	})
+	t.Run("TestMetadata", func(t *testing.T) {
+		testMetadata(t, newHarness)
+	})
 	t.Run("TestDelete", func(t *testing.T) {
 		testDelete(t, newHarness)
 	})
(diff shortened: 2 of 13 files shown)


zombiezen:
Another reasonable restriction might be to keep keys as ASCII-clean for now.

blob/fileblob/fileblob.go
```diff
@@ -107,6 +107,7 @@ func (b *bucket) Attributes(ctx context.Context, key string) (driver.Attributes,
 	}
 	return driver.Attributes{
 		ContentType: xa.ContentType,
+		Metadata:    xa.Metadata,
```
zombiezen: This is missing a pass to lowercase all the keys.

blob/drivertest/drivertest.go
```diff
@@ -465,6 +468,89 @@ func testCanceledWrite(t *testing.T, newHarness HarnessMaker, pathToTestdata str
...
+				"key-a": "value-a",
+				"kEy-B": "value-b",
+				"key-c": "vAlUe-c",
+			},
+			want: map[string]string{
+				"key-a": "value-a",
+				"key-b": "value-b",
+				"key-c": "vAlUe-c",
+			},
+		},
```
zombiezen: Test for two keys that differ only by case. I would expect an error.

blob/driver/driver.go
```diff
@@ -58,6 +58,8 @@ type WriterOptions struct {
 	// write in a single request, if supported. Larger objects will be split into
 	// multiple requests.
 	BufferSize int
+	// Metadata holds key/value strings to be associated with the blob.
```
zombiezen: This doc comment should also specify that the caller is responsible for guaranteeing that all keys are already lowercased.

blob/gcsblob/gcsblob.go
```diff
@@ -120,6 +121,7 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
 	w.ContentType = contentType
 	if opts != nil {
 		w.ChunkSize = bufferSize(opts.BufferSize)
+		w.Metadata = opts.Metadata
```
zombiezen: This should throw an error if the metadata contains a key like `Content-Type`.

blob/s3blob/s3blob.go
```diff
@@ -225,12 +238,20 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
 			u.PartSize = int64(opts.BufferSize)
 		}
 	})
+	var metadata map[string]*string
+	if opts != nil && len(opts.Metadata) > 0 {
+		metadata = make(map[string]*string, len(opts.Metadata))
+		for k, v := range opts.Metadata {
+			metadata[k] = aws.String(v)
```
zombiezen: Same here: this should return an error if the key is like `Content-Type`.

blob/driver/driver.go
```diff
@@ -76,6 +78,9 @@ type ReaderAttributes struct {
 type Attributes struct {
 	// ContentType is the MIME type of the blob object. It must not be empty.
 	ContentType string
+	// Metadata holds key/value pairs associated with the blob.
+	// Keys are case-insensitive and will always be returned in lowercase.
```
zombiezen: nit: s/will/must/

blob/fileblob/fileblob.go
vangent: Hmm, the point is to do all of the lowercasing in the concrete type, so that the individual driver implementations don't have to do it. So, there's a ToLower in blob.Bucket.Attributes for reading, and one in blob.NewWriter for writing.

blob/drivertest/drivertest.go

blob/driver/driver.go
vangent: No, the concrete type will lowercase them before creating the driver.WriterOptions. I added a comment to that effect.

blob/gcsblob/gcsblob.go
vangent: Why? It seems to work fine.... The providers namespace the keys, so they don't conflict with real HTTP headers/ attributes.

For example, S3 does this:

```
Content-Type:
      - text/plain; charset=utf-8
...
X-Amz-Meta-Content-Type:
      - value-a
```

and GCS passes the metadata in the body, not in an HTTP header (at least for the request we're using; for other requests they are headers, but have a prefix similar to what S3 does).

blob/s3blob/s3blob.go
vangent: See above.

blob/driver/driver.go
zombiezen: Sounds like the right way to resolve this. Thanks!

blob/driver/driver.go
```diff
@@ -76,6 +79,12 @@ type ReaderAttributes struct {
 type Attributes struct {
 	// ContentType is the MIME type of the blob object. It must not be empty.
 	ContentType string
+	// Metadata holds key/value pairs associated with the blob.
```
shantuo: Since we enforce lowercase in the concrete type. Should the comment here say that the implementation can expect to receive only lowercase attributes?

vangent: No, these Attributes are going in the other direction -- the driver implementation is creating them and passing them back to the concrete type, which is then returning them to the user. The comment I added in the last PR added more detail about that; I think it's pretty clear now (it was harder before when Attributes was used both in driver and in blob).

PR #497, opened 2018-09-29
runtimevar/etcdvar: avoid double reads of variables when they change

### LINKED ISSUE #440: runtimevar: edge-based implementations should use a long-running subscription
We have 2: filevar and etcdvar. They both do something like this:

Watch() {
  10 Set up subscription
  20 Read blob
  30 If changed, RETURN new value
  40 Block on subscription until change is detected
  50 GOTO 20
}

Note that when the variable changes, they end up reading the blob twice (assuming that the end user re-calls Watch() immediately, as expected). This is inefficient.

It would be better to start a background goroutine on construction that sets up the subscription and stores a copy of the last value in memory; WatchVariable would then just read that copy. This avoids the slow reader problem, and aligns with the Go philosophy of modeling concurrent things with goroutines.

vangent:
Fixes #440.

Also avoids any repeated overhead of creating the etc client.Watch.

This code overlaps somewhat with `filevar`, but it's different enough that I don't think it's a good idea to try to refactor the common bits out.

### DIFF
--- a/runtimevar/etcdvar/etcdvar.go
+++ b/runtimevar/etcdvar/etcdvar.go
@@ -28,30 +28,22 @@ import (
 	"google.golang.org/grpc/codes"
 )
 
-// defaultWait is the default value for WatchOptions.WaitTime.
-const defaultWait = 30 * time.Second
-
 // New constructs a runtimevar.Variable object that uses client to watch
 // variables in etcd.
 // Provide a decoder to unmarshal updated configurations into similar
 // objects during the Watch call.
-func New(name string, cli *clientv3.Client, decoder *runtimevar.Decoder, opts *WatchOptions) (*runtimevar.Variable, error) {
-	if opts == nil {
-		opts = &WatchOptions{}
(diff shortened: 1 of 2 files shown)


runtimevar/etcdvar/etcdvar.go
```diff
@@ -70,84 +62,106 @@ func (s *state) UpdateTime() time.Time {
...
-	name    string
-	client  *clientv3.Client
-	decoder *runtimevar.Decoder
-	wait    time.Duration
+	// The background goroutine writes new *state values to ch.
+	// It is buffered so that the background goroutine can write without
+	// blocking; it always drains the buffer before writing so that the latest
+	// write is buffered. If writes could block, the background goroutine could be
+	// blocked indefinitely from reading etcd's Watch events.
+	ch chan *state
```
shantuo: Can these two channels be combined? When the watch is canceled, it closes ch?

vangent: Yup, good catch. That works; done.

PR #498, opened 2018-09-29
gcp: Allow custom scopes for DefaultCredentials

### LINKED ISSUE #466: gcp: DefaultCredentials does not allow for passing in scopes
Working through the examples I noticed that [`gcp.DefaultCredentials()`](https://github.com/google/go-cloud/blob/master/gcp/gcp.go#L70) always sets the `cloud-platform` scope, which is basically "super duper admin"which feels a little extreme. It doesn't let you pass in scopes, even though the underlying function it calls [`oauth2.google.FindDefaultCredentials`](https://godoc.org/golang.org/x/oauth2/google#FindDefaultCredentials) does take a variadic `scopes` argument.

Given the recommendation in GCP is usually to create service accounts with a limited scope how does this interact when credentials are given that have not been granted that scope? Say for example in the tutorial case, if I were to give it a service account that just had `https://www.googleapis.com/auth/devstorage.read_write`, would it work? Does it automatically downgrade to the scopes of the issued service account, or will this cause an error when requests are made?

If it's the latter case, would a PR that updates `DefaultCredentials` to accept a variadic `scopes` be welcome? It would be nice to not have to do the credential load dance yourself with the oauth2 library just to be able to tweak the scopes.

daenney:
Adds a second variadic argument to DefaultCredentials allowing to pass
in different scopes in case the 'cloud-platform' scope is not what the
caller wishes to use.

Closes #466

### DIFF
--- a/gcp/gcp.go
+++ b/gcp/gcp.go
@@ -65,10 +65,19 @@ func DefaultTransport() http.RoundTripper {
 	return http.DefaultTransport
 }
 
-// DefaultCredentials obtains the default GCP credentials with Cloud Platform
-// scope.
-func DefaultCredentials(ctx context.Context) (*google.Credentials, error) {
-	adc, err := google.FindDefaultCredentials(ctx, "https://www.googleapis.com/auth/cloud-platform")
+// DefaultCredentials obtains the default GCP credentials.
+//
+// When no scopes are specified the cloud-platform scope is assumed.
+func DefaultCredentials(ctx context.Context, scopes ...string) (*google.Credentials, error) {
+	var err error
+	var adc *google.Credentials
+
+	if len(scopes) == 0 {
+		adc, err = google.FindDefaultCredentials(ctx, "https://www.googleapis.com/auth/cloud-platform")
+	} else {
+		adc, err = google.FindDefaultCredentials(ctx, scopes...)
+	}
+
 	if err != nil {
 		return nil, err
 	}


daenney:
The two test builds fail with the following:

```
/home/travis/gopath/src/github.com/google/go-cloud/samples/guestbook/inject_gcp.go:38:1: inject setupGCP: no provider found for []string (required by provider of *github.com/google/go-cloud/vendor/golang.org/x/oauth2/google.Credentials)
/home/travis/gopath/src/github.com/google/go-cloud/tests/gcp/app/inject.go:28:1: inject initialize: no provider found for []string (required by provider of *github.com/google/go-cloud/vendor/golang.org/x/oauth2/google.Credentials)
wire: error loading packages
```

I don't believe this is related to my PR, but I'm not sure. If there's anything I need to do to resolve this I'd be happy for some pointers.

vangent:
Thanks for the PR!

Yes, that test failure is related to your change. Wire is using DefaultCredentials as a provider, [here](https://github.com/google/go-cloud/blob/master/gcp/gcp.go#L30) via [here](https://github.com/google/go-cloud/blob/master/gcp/gcpcloud/gcpcloud.go#L28). It expects to find a way to fill all of the arguments to a provider, and it is interpreting the variadic argument you added as "I need to find a `[]string` to pass here".

I can think of a few possible solutions to this:
- We could add a new function, `DefaultCredentialsWithScope`, instead of modifying the existing one. Users that want to use a non-default scope would have to call it and would not be able to use the `gcpcloud.GCP` provider set; they'd have to duplicate it without `DefaultCredentials`.
- We could require everyone to provide scopes; this would involve leaving your change the way it is, but updating the current uses of the `gcpcloud.GCP` provider set (at least the two test errors, but there might be more in doc examples) to provide a `[]string` for the scopes, either as a `wire.Value` or via a provider (e.g., a new `CloudPlatformScope` function in `gcp`.).
- Same, but add a new type for scopes instead of using `[]string`, to make it more explicit for Wire.

@zombiezen , thoughts on those ideas, or additional ones?

zombiezen:
@daenney, thank you for looking into this. I disagree with this change on the following grounds:

`gcp.DefaultCredentials` is a thin shim for `google.FindDefaultCredentials`, which can already take in a custom scope. The reason `gcp.DefaultCredentials` exists is to provide a convenient default provider to put in most GCP server Wire graphs, as the general recommendation is for service account privilege to be limited by IAM, not by oauth2 scopes. The point of Wire is that if someone needs more customization for their application over what we are providing for common cases, they can just add their own custom provider to their graph.

PR #499, opened 2018-10-02
 mysql/cloudsql: make cloudsql work on AppEngine with local /cloudsql socket

### LINKED ISSUE #494: mysql/cloudsql: ensure package works on Go 1.11 App Engine
In GAE Standard, you have a built-in socket to CloudSQL at /cloudsql.

To use it, set host = '/cloudsql/{db_connection_name}'

Here's a [Python sample](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/michaelawyu-patch-appengine-py37-cloudsql/appengine/standard_python37/cloudsql/main_mysql.py) using it.

In case we need the Go Cloud code to behave differently on GAE Standard, look for the GAE_ENV env var.

Here's some more sample code in Go:

```
func GetDB() *sql.DB {
	user := os.Getenv("DB_USER")
	db := os.Getenv("DB_DATABASE")
	inst := os.Getenv("DB_INSTANCE")
	pw := os.Getenv("DB_PASSWORD")

	// format: user:pw@unix(/cloudsql/<project-id>:<region>:<inst>)/dbname
	// Note that DB_INSTANCE is already in the format of project-id:region:instance-name
	dbURI := fmt.Sprintf("%s:%s@unix(/cloudsql/%s)/%s", user, pw, inst, db)
	// go-sql-driver/mysql has a special cloudsql driver, which has the format
	// user:pw@cloudsql(project-id:region:instance-name)/dbname
	// Except, this opens a connection to cloudsql://..., which GAE TI doesn't support.
	// broken := fmt.Sprintf("%s:%s@cloudsql(%s)/%s", user, pw, inst, db),
	conn, err := sql.Open("mysql", dbURI)
        ...
```

We might need to set the `Addr` field differently [here](https://github.com/google/go-cloud/blob/master/mysql/cloudmysql/cloudmysql.go#L74), but not sure.

ijt:
Fixes #494 

This includes additions to the `mysql` package and the guestbook sample.


ijt:
The new `cloudmysql.OpenGAE` function relies on configuration from environment variables because Go on App Engine does not yet have a way to set flags when the server is launched. If that functionality becomes available in the future it would probably make sense to make `cloudmysql.OpenGAE` take a `params` parameter similar to the `cloudmysql.Open` function.

gcp/gcpcloud/gcpcloud.go
```diff
@@ -37,3 +37,17 @@ var Services = wire.NewSet(
 	gcp.NewHTTPClient,
 	runtimeconfigurator.Set,
 	sdserver.Set)
+
+// GAE is a Wire provider set, similar to the GCP Wire provider
```
vangent: Nit: make the comment less focused on what exactly is different, and more on what the intent is. E.g., 

```
GAE is a Wire provider set, similar to the GCP Wire provider set but intended for use with GAE (Google App Engine).
```

gcp/gcpcloud/gcpcloud.go
```diff
@@ -37,3 +37,17 @@ var Services = wire.NewSet(
...
+// set except that GAEServices is used instead of Services.
+var GAE = wire.NewSet(GAEServices, gcp.DefaultIdentity)
+
+// GAEServices is a Wire provider set, similar to Services
+// except that cloudmysql.OpenGAE is used instead of cloudmysql.Open.
+var GAEServices = wire.NewSet(
```
vangent: I think you could also factor out the common ones, e.g.:

var sharedServices = wire.NewSet(cloudmysql.CertSourceSet, ...)
var Services = wire.NewSet(sharedServices, cloudmysql.Open)
var GAEServices = wire.NewSet(sharedServices, cloudmysql.OpenGAE)

Not sure if this is better though.

mysql/cloudmysql/cloudmysql.go
```diff
@@ -79,6 +80,28 @@ func Open(ctx context.Context, certSource proxy.CertSource, params *Params) (*sq
 	return sql.OpenDB(connector(cfg.FormatDSN())), nil
 }
 
+// OpenGAE opens a Cloud SQL database on Google App Engine (GAE) using the
+// environment variables $DB_USER, $DB_DATABASE, $DB_INSTANCE, and
+// $DB_PASSWORD.
+func OpenGAE(ctx context.Context) (*sql.DB, error) {
```
vangent: ctx is unused?

mysql/cloudmysql/cloudmysql.go
```diff
@@ -79,6 +80,28 @@ func Open(ctx context.Context, certSource proxy.CertSource, params *Params) (*sq
 	return sql.OpenDB(connector(cfg.FormatDSN())), nil
 }
 
+// OpenGAE opens a Cloud SQL database on Google App Engine (GAE) using the
+// environment variables $DB_USER, $DB_DATABASE, $DB_INSTANCE, and
+// $DB_PASSWORD.
```
vangent: Is this how GAE commonly does it, putting everything in environment variables?

samples/guestbook/wire_gen.go
```diff
@@ -6,28 +6,28 @@
...
-	"github.com/google/go-cloud/runtimevar"
-	"github.com/google/go-cloud/runtimevar/filevar"
-	"github.com/google/go-cloud/runtimevar/paramstore"
-	"github.com/google/go-cloud/runtimevar/runtimeconfigurator"
-	"github.com/google/go-cloud/server"
-	"github.com/google/go-cloud/server/sdserver"
-	"github.com/google/go-cloud/server/xrayserver"
-	"go.opencensus.io/trace"
-	"net/http"
+	context "context"
```
vangent: This looks like you might be using an old version of wire, can you regenerate this file with the latest?

mysql/cloudmysql/cloudmysql.go
ijt: It's currently necessary because Go on GAE doesn't yet support an entrypoint field in app.yaml that would allow setting flags on the server.

gcp/gcpcloud/gcpcloud.go
ijt: Done.

gcp/gcpcloud/gcpcloud.go
ijt: Done.

mysql/cloudmysql/cloudmysql.go
ijt: True, although that's also true of the Open function above. Removing it there as well.

samples/guestbook/wire_gen.go
ijt: Done.

mysql/cloudmysql/cloudmysql.go
```diff
@@ -56,7 +58,7 @@ type Params struct {
 }
 
 // Open opens a Cloud SQL database.
-func Open(ctx context.Context, certSource proxy.CertSource, params *Params) (*sql.DB, error) {
+func Open(certSource proxy.CertSource, params *Params) (*sql.DB, error) {
```
vangent: Hmm, it's probably worth checking with whoever added `Open` (@zombiezen or @shantuo probably) about why it took a content, before removing it. Maybe there's a good reason.

zombiezen: When in doubt, it's usually better to plumb a `Context`. Especially if you're doing something that will probably do networky things, it's usually better to include it.

mysql/cloudmysql/cloudmysql.go
```diff
@@ -79,6 +80,28 @@ func Open(ctx context.Context, certSource proxy.CertSource, params *Params) (*sq
 	return sql.OpenDB(connector(cfg.FormatDSN())), nil
 }
 
+// OpenGAE opens a Cloud SQL database on Google App Engine (GAE) using the
+// environment variables $DB_USER, $DB_DATABASE, $DB_INSTANCE, and
+// $DB_PASSWORD.
+func OpenGAE() (*sql.DB, error) {
```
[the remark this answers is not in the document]
zombiezen: I'm similarly put off by the fact that the App Engine codepath would be different and that the application has to do App Engine specific things. Is there a compelling reason to do it this way?

mysql/cloudmysql/cloudmysql.go
ijt: In this case, is that because we might need it later and we don't want to break existing calls to cloudmysql.Open?

ijt: Also, done.

zombiezen: Precisely.

mysql/cloudmysql/cloudmysql.go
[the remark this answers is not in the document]
ijt: I guess there is no compelling reason. It would probably be better to factor out a function like `GetParamsFromEnv()` and have its output passed into `Open`. Trying that now.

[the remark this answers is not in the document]
ijt: Okay, I changed it. I'm not thrilled about how the new code checks $GAE_ENV in three places total for the guestbook sample, but that issue will go away once the `go111` runtime supports command line flags.

ijt:
The guestbook sample has been re-generated and deployed to https://gocloudgae.appspot.com/ with the latest changes.

zombiezen:
I've posted a few existential questions on #494.

ijt:
Just noticed this sitting in our list of pull requests. Should I close this out?

PR #500, opened 2018-10-02
wire: improve error message for provider set conflicts

### LINKED ISSUE #411: wire: error messages not showing enough context in cases of conflict
While implementing a new server type, I added a set to an existing wire.Set. The new set had a new provider for `server.Server`, redundant with one already provided. When I ran `go generate`, `wire` issued the following error message:

```
/home/issactrotts/src/github.com/google/go-cloud/server/jaegerserver/server.go:33:11: multiple bindings for *github.com/google/go-cloud/server.Server (previous binding in "github.com/google/go-cloud/server".Set)
```
This error message doesn't pinpoint the cause of the problem, i.e., the location where the first `server.Set` was transitively included. I think it would help to have something akin to a stack trace for this kind of problem.

vangent:
Fixes #411.


wire/internal/wire/analyze.go
```diff
@@ -433,15 +430,15 @@ func verifyAcyclic(providerMap *typeutil.Map, hasher typeutil.Hasher) []error {
...
+	if set.VarName == "" {
+		fmt.Fprintf(sb, "wire.Build")
 	} else {
-		err = fmt.Errorf("multiple bindings for %s (previous binding in %q.%s)",
-			typString, prevSet.PkgPath, prevSet.VarName)
+		fmt.Fprintf(sb, set.VarName)
 	}
-	return notePosition(fset.Position(pos), err)
+	fmt.Fprintf(sb, " has multiple bindings for %s (", types.TypeString(typ, nil))
+	fmt.Fprintf(sb, "current binding: %s", strings.Join(cur.trace(fset, typ), " <- "))
```
vangent: I considered using multi-line errors here (e.g., replacing " <- " with "\n  " or some such), but don't think the test framework supports that. Maybe a TODO?

wire/internal/wire/parse.go
```diff
@@ -38,6 +38,37 @@ type providerSetSrc struct {
 	Import   *ProviderSet
 }
 
+// trace returns a slice of strings describing the (possibly recursive) source
+// of p, including line numbers.
+func (p *providerSetSrc) trace(fset *token.FileSet, typ types.Type) []string {
```
vangent: I also considered adding a String() func to each of the 4 types used here (Provider, ProviderSet, IfaceBinding, Value) and using %v here, thoughts?

wire/internal/wire/testdata/MultipleBindings/want/wire_errs.txt
```diff
@@ -0,0 +1,6 @@
+/wire_gopath/src/example.com/foo/wire.go:27:8: wire.Build has multiple bindings for example.com/foo.Foo (current binding: provider "provideFooAgain" (/wire_gopath/src/example.com/foo/foo.go:39:6); previous binding: provider "provideFoo" (/wire_gopath/src/example.com/foo/foo.go:35:6)
```
vangent: Do you think it's too fragile to include the line numbers here?
I don't think we've done that before, but they are interspersed with the actual error text here and so hard to avoid.
Just verifying "multiple bindings" isn't useful for this test.

wire/internal/wire/testdata/MultipleBindings/foo/wire.go
```diff
@@ -0,0 +1,53 @@
...
+
+func injectDuplicateValues() Foo {
+	// fail: provideFoo and wire.Value both provide Foo.
+	panic(wire.Build(provideFoo, wire.Value(Foo("foo"))))
+}
+
+func injectDuplicateInterface() Bar {
+	// fail: provideBar and wire.Bind both provide Bar.
+	panic(wire.Build(provideBar, wire.Bind(new(Bar), strings.NewReader("hello"))))
+}
```
vangent: Anything else I should add? I considered adding a struct provider, do you think that's necessary?

wire/internal/wire/analyze.go
zombiezen: Oh. Yes, the test compares against the whole error string (not line-based), but you can't specify a match over multiple lines. One possible route could be to add something like a trailing backslash to indicate line continuations. A TODO seems fine.

wire/internal/wire/parse.go
```diff
@@ -38,6 +38,37 @@ type providerSetSrc struct {
...
+	}
+	var retval []string
+	if p.Provider != nil {
+		kind := "provider"
+		if p.Provider.IsStruct {
+			kind = "struct provider"
+		}
+		retval = append(retval, fmt.Sprintf("%s %s(%s)", kind, quoted(p.Provider.Name), fset.Position(p.Provider.Pos)))
+	} else if p.Binding != nil {
+		retval = append(retval, fmt.Sprintf("wire.Bind (%s)", fset.Position(p.Binding.Pos)))
```
zombiezen: This could include the types involved.

wire/internal/wire/parse.go
```diff
@@ -38,6 +38,37 @@ type providerSetSrc struct {
...
+	} else if p.Value != nil {
+		retval = append(retval, fmt.Sprintf("wire.Value (%s)", fset.Position(p.Value.Pos)))
+	} else if p.Import != nil {
+		if parent := p.Import.srcMap.At(typ); parent != nil {
+			retval = append(retval, parent.(*providerSetSrc).trace(fset, typ)...)
+		}
+		retval = append(retval, fmt.Sprintf("provider set %s(%s)", quoted(p.Import.VarName), fset.Position(p.Import.Pos)))
```
zombiezen: `VarName` might be empty.

wire/internal/wire/testdata/MultipleBindings/want/wire_errs.txt
zombiezen: I've tried to avoid adding line numbers just because it makes it harder to update the tests, but I don't have a suggestion as to how to make this better without changing the test infrastructure to permit regexes or some such. (Might not be the worst idea.)

wire/internal/wire/testdata/MultipleBindings/foo/wire.go
zombiezen: I think this covers sufficiently. If you want to add it, go for it, but I don't see it as required.

wire/internal/wire/parse.go
zombiezen: I'd avoid providing a canonical string function for the exported API for now. This function is fine as it stands.

wire/internal/wire/analyze.go
vangent: I filed an issue to track this.

wire/internal/wire/testdata/MultipleBindings/want/wire_errs.txt
vangent: Filed an issue for supporting regexes.

wire/internal/wire/parse.go
vangent: Yes, quoted() handles that; it will say something like "provider set (/.../foo:25)" if p.Import.VarName is empty.

wire/internal/wire/parse.go
vangent: I don't think that's necessary; the line number is there so it should be easy to find, and the wire.Bind type that's being provided should be the same as the one that's the duplicate anyway.

PR #501, opened 2018-10-03
samples: add a sample running "hello world" via server.Server on GAE

### LINKED ISSUE #210: server: get hello world working on GAE Standard
While `*Server.ListenAndServe` works fine for generic compute environments, it does not work in cases where an alternate listening mechanism is needed, like on Google App Engine Standard, which requires calling `appengine.Main` in place of `http.ListenAndServe`.

To fix this, `*server.Server` should be expanded to include a hook that takes in the wrapped `http.Handler` that `*server.Server` creates as well as the listening address, and then runs the server.  This would also be useful for #57, because in GCP contexts this would be the correct place to attach startup hooks like Cloud Profiler.

vangent:
Fixes #210.


samples/appengine/helloworld.go
```diff
@@ -0,0 +1,32 @@
+package main
```
ijt: Add legalese comment at the top?

vangent: Oops, yes -- thanks. Done.

Issue #470, opened 2018-09-19
all: implement escape hatches


vangent:
Design proposal has circulated, and will be added to design.md in a PR soon.

vangent:
Fixed via #507.

PR #507, opened 2018-10-09
blob: add support for escape hatches via As

vangent:
I still need to update the documentation with an overview, but I'd rather do that after getting through the review.

Feedback on naming is welcome. I ended up using func types in a couple of places instead of defining a new interface because it felt simpler.


vangent:
Issue #470.

blob/blob.go
```diff
@@ -75,7 +83,19 @@ type Attributes struct {
 	// ModTime is the time the blob object was last modified.
 	ModTime time.Time
 	// Size is the size of the object in bytes.
-	Size int64
+	Size   int64
+	asFunc func(interface{}) bool
```
zombiezen: nit: blank line before `asFunc`

blob/blob.go
```diff
@@ -61,6 +61,14 @@ func (r *Reader) Size() int64 {
 	return r.r.Attributes().Size
 }
 
+// As is an escape hatch for accessing provider-specific types.
```
zombiezen: While I think it's fine to link to the design in a "footnote" paragraph, the documentation comment should explain how to use the method without needing to look at the link.

(I believe this is what you're saying in your PR comment, but I want to call it out explicitly in case I misunderstood.)

blob/blob.go
```diff
@@ -326,6 +356,14 @@ type WriterOptions struct {
...
 	// Duplicate case-insensitive keys (e.g., "foo" and "FOO") are an error.
 	Metadata map[string]string
+
+	// Callback is a callback that will be called at most once, before any data
+	// is written. asFunc can be used as an escape hatch for accessing
+	// provider-specific types.
+	// See the provider documentation for which type(s) are supported, and
+	// https://github.com/google/go-cloud/blob/master/internal/docs/design.md#escape-hatches
+	// for more details on escape hatches.
+	Callback func(asFunc func(interface{}) bool) error
```
zombiezen: Naming: I would call this `OnStart` or some such.

(Same advice about linking to the design doc.)

blob/driver/driver.go
```diff
@@ -38,13 +38,20 @@ type Error interface {
 	BlobError() ErrorKind
 }
 
+// NoAs can be used as an As function that always returns false.
+var NoAs = func(interface{}) bool { return false }
```
zombiezen: I'm not sure how useful it is to keep versus inlining, but if you're going to do it, just create a function instead of a variable.

blob/driver/driver.go
```diff
@@ -38,13 +38,20 @@ type Error interface {
...
 	io.ReadCloser
 
 	// Attributes returns a subset of attributes about the blob.
-	// Use Bucket.Attributes to get the full set.
 	Attributes() ReaderAttributes
+
+	// As allows providers to expose provider-specific types.
+	// See https://github.com/google/go-cloud/blob/master/internal/docs/design.md#escape-hatches
+	// for more details.
+	As(interface{}) bool
```
zombiezen: Does this need to be a required method? If it was introduced in a new interface and dynamically tested for, then it would not be a breaking change.

blob/driver/driver.go
```diff
@@ -89,11 +102,21 @@ type Attributes struct {
...
+	AsFunc func(interface{}) bool
 }
 
 // Bucket provides read, write and delete operations on objects within it on the
 // blob service.
 type Bucket interface {
+	// As allows providers to expose provider-specific types.
+	// See https://github.com/google/go-cloud/blob/master/internal/docs/design.md#escape-hatches
+	// for more details.
+	As(i interface{}) bool
```
zombiezen: Same here: the required-ness seems suspect to me.

blob/drivertest/drivertest.go
```diff
@@ -42,11 +43,67 @@ type Harness interface {
...
+// 3. Fetches the blob's attributes and calls AttributeCheck.
+// 4. Creates a Reader for the blob, and calls ReaderCheck.
+// For example, an AsTest might set a provider-specific field to a custom
+// value in Write, and then verify the custom value was returned in
+// AttributesCheck and/or ReaderCheck.
+type AsTest interface {
+	// Name should return a descriptive name for the test.
+	Name() string
+	// BucketCheck will be called to allow verification of Bucket.As.
+	BucketCheck(t *testing.T, b *blob.Bucket)
```
zombiezen: I see that these methods can build up multiple failures. `Fatal` relies on the panic crossing boundaries up on the same goroutine. Instead, consider sending down an `interface{ Errorf(string, ...interface{}) }` for more harness control as the conformance tests evolve.

blob/blob.go
```diff
@@ -326,6 +356,14 @@ type WriterOptions struct {
 	// Keys may not be empty, and are lowercased before being written.
 	// Duplicate case-insensitive keys (e.g., "foo" and "FOO") are an error.
 	Metadata map[string]string
+
+	// Callback is a callback that will be called at most once, before any data
+	// is written. asFunc can be used as an escape hatch for accessing
```
zombiezen: Will it be called on a zero-length blob? Now that I'm thinking about it, it probably should always be called, but the guarantee is that it will be called before `Close`, but not necessarily before the first user-facing `Write`.

blob/blob.go

blob/blob.go
vangent: So, given that there are 4 escape hatches (with more coming after List is added), and that's just for blob, I don't want to repeat the "here's how to use As" documentation every time; the full documentation should include examples etc. My plan was to have all of that at the linked doc, and only have this short explanation here. Do you think I should repeat all of the documentation each time? Include a bit more in what's here?

blob/blob.go
vangent: How about "BeforeWrite"?

blob/driver/driver.go
vangent: Yea, it was only used once. Removed.

blob/driver/driver.go
vangent: It could be done that way, but given that we don't need to be compatible at this point, I think it's cleaner to just add it.

blob/blob.go
vangent: I updated the docstring a bit to say it will always be called exactly once, unless NewWriter or the equivalent returns an error.

I think it has to be called before the first actual write (not the first time the user calls Write); also clarified.

blob/blob.go
vangent: I updated them all to point to `Bucket.As`, and added more documentation there.

I also added an example test for blob.

blob/drivertest/drivertest.go
vangent: Maybe these should not take a testing.T, but instead return error and let the conformance test report it?

vangent: Done. PTAL.

vangent:
> remove references to the term "escape hatch" in the doc comments

Done.

Issue #416, opened 2018-09-08
sql/ocsql: allow configuration of ocsql instrumentation


basvanbeek:
Currently `ocsql` is included with all trace options enabled. This means that sql queries and query parameters are annotated in spans, which can be undesirable from a security concern as well as a performance one (high cardinality annotations or heavily annotated spans do not sit well with all tracing backends in high QPS systems). This is why default `ocsql` is much more conservative and has all extra trace options disabled. 

I think it would be best to have the conservative default mirrored in Go cloud as well as the ability for consumers to set the configuration of `ocsql` when using Go cloud.

Happy to provide a PR if I can be guided a bit on how Go cloud wants configuration done (would this be part of rdsmysql.Params and cloudmysql.Params?)

vangent:
Thanks for noting this @basvanbeek !

@ijt , can you take a look?

ijt:
Sure, looking.

ijt:
I'm thinking we can add a `[]TraceOption` parameter to the `Open` funcs in the mysql packages.

zombiezen:
This is not fixed. RDS still does not allow the configuration to be passed in.

https://github.com/google/go-cloud/blob/f4c33a8eb84a6b8164d6be83bcde0bfe79a49a0b/mysql/rdsmysql/rdsmysql.go#L129-L131

PR #511, opened 2018-10-09
sql/ocsql: allow configuration of ocsql instrumentation

ijt:


ijt:
Fixes #416

mysql/cloudmysql/cloudmysql.go
```diff
@@ -56,7 +56,7 @@ type Params struct {
 }
 
 // Open opens a Cloud SQL database.
-func Open(ctx context.Context, certSource proxy.CertSource, params *Params) (*sql.DB, error) {
+func Open(ctx context.Context, certSource proxy.CertSource, params *Params, traceOpts ocsql.TraceOption) (*sql.DB, error) {
```
vangent: It's unfortunate that this argument list is getting kind of long, and that this is a breaking change. Maybe add to Params instead?

pubsub/pubsub.go
```diff
@@ -0,0 +1,260 @@
+// Copyright 2018 The Go Cloud Authors
```
vangent: This file doesn't look like it belongs in this PR.

mysql/cloudmysql/cloudmysql.go
vangent: It looks like ocsql.TraceOption is a func; what you have here only allows a single one. I.e., what if I want WithRowsAffected and WithRowsClosed?

Maybe this should be a *ocsql.TraceOptions instead?

mysql/cloudmysql/cloudmysql.go
```diff
@@ -76,21 +76,26 @@ func Open(ctx context.Context, certSource proxy.CertSource, params *Params) (*sq
 		Passwd:               params.Password,
 		DBName:               params.Database,
 	}
-	return sql.OpenDB(connector(cfg.FormatDSN())), nil
+	var c connector
+	c.dsn = cfg.FormatDSN()
+	c.dbDriver = ocsql.Wrap(mysql.MySQLDriver{}, traceOpts)
```
vangent: Does this work if traceOpts is nil?

mysql/cloudmysql/cloudmysql.go
ijt: Done.

ijt: Done.

pubsub/pubsub.go

mysql/cloudmysql/cloudmysql.go
ijt: Sorry, I misunderstood how these are supposed to be used. An example is here:

https://github.com/opencensus-integrations/ocsql/blob/f67f654fe0de226036abcecd6354c4b8205b1afb/driver.go#L81

ijt: Fixed it by making it varargs like it's supposed to be.

mysql/cloudmysql/cloudmysql.go
ijt: Fixed with varargs.

PR #512, opened 2018-10-10
health: factor out code to write HTTP headers

gsquire:
This patch adds a small function to write out the HTTP headers in the health check handlers since they're more or less the same.

### DIFF
--- a/health/health.go
+++ b/health/health.go
@@ -43,16 +43,19 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
 	writeHealthy(w)
 }
 
-func writeHeaders(status string, w http.ResponseWriter) {
-	w.Header().Set("Content-Length", strconv.Itoa(len(status)))
+func writeHeaders(statusLen int, w http.ResponseWriter) {
+	w.Header().Set("Content-Length", strconv.Itoa(statusLen))
 	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
 	w.Header().Set("X-Content-Type-Options", "nosniff")
 }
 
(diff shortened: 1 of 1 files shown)


health/health.go
```diff
@@ -42,12 +43,18 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
 	writeHealthy(w)
 }
 
-func writeUnhealthy(w http.ResponseWriter) {
-	w.Header().Set("Content-Length", "9")
+func writeHeaders(status string, w http.ResponseWriter) {
+	w.Header().Set("Content-Length", strconv.Itoa(len(status)))
```
vangent: It's a bit unfortunate to recompute the length every time. How about passing it in as an argument to `writeHeaders` and adding

```
   const statusLen = [2 or 9]
```

in the caller?

health/health.go
```diff
@@ -43,16 +43,19 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
...
+	w.Header().Set("Content-Length", strconv.Itoa(statusLen))
 	w.Header().Set("Content-Type", "text/plain; charset=utf-8")
 	w.Header().Set("X-Content-Type-Options", "nosniff")
 }
 
 func writeUnhealthy(w http.ResponseWriter) {
-	const status = "unhealthy"
+	const (
+		status    = "unhealthy"
+		statusLen = 9
```
vangent: Since it's a constant, might as well make it a string (or do the ItoA here if you prefer).

gsquire: That's fair, I should have just made it a string in the beginning.

PR #530, opened 2018-10-11
blob/all: add support for SignedURL

vangent:


blob/blob.go
```diff
@@ -381,3 +410,12 @@ func IsNotExist(err error) bool {
 	}
 	return false
 }
+
+// IsNotImplemented returns true iff err indicates that the provider does not
+// support the given operation.
+func IsNotImplemented(err error) bool {
```
zombiezen: Testing for rationale: do we expect that calling code would branch based on this information? I can't think of a concrete example here.

blob/gcsblob/gcsblob_test.go
```diff
@@ -28,27 +30,55 @@ import (
...
+	// 3. Create a service account in your GCP project and update the
+	//    serviceAccountID constant to it.
+	// 4. Download a private key to a .pem file as described here:
+	//    https://godoc.org/cloud.google.com/go/storage#SignedURLOptions
+	//    and update the pathToPrivateKey constant with a path to the .pem file.
+	// TODO(issue #300): Use Terraform to provision a bucket, and get the bucket
+	//    name from the Terraform output instead (saving a copy of it for replay).
+	bucketName       = "go-cloud-blob-test-bucket"
+	serviceAccountID = "storage-viewer@go-cloud-test-216917.iam.gserviceaccount.com"
+	pathToPrivateKey = "/usr/local/google/home/rvangent/Downloads/storage-viewer.pem"
```
zombiezen: This should be passed via flag, as it is record dependent and not needed for comparison purposes.

blob/driver/driver.go
```diff
@@ -162,4 +166,15 @@ type Bucket interface {
 	// not exist, NewRangeReader must return an error whose Kind method
 	// returns NotFound.
 	Delete(ctx context.Context, key string) error
+
+	// SignedURL returns a URL that can be used to GET the blob for a
+	// limited time. opts is guaranteed to be non-nil.
```
zombiezen: s/a limited time/the duration specified in the options/

blob/gcsblob/gcsblob.go
```diff
@@ -37,23 +38,44 @@ import (
 	"google.golang.org/api/option"
 )
 
+type Options struct {
```
zombiezen: Doc comment for Options struct.

blob/fileblob/fileblob.go
```diff
@@ -258,13 +258,20 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
 	return nil
 }
 
+func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error) {
+	return "", fileError{msg: "SignedURL not supported", kind: driver.NotImplemented}
```
zombiezen: Please add a TODO (and perhaps even amend the error message) to refer to #546.

blob/blob.go
vangent: Interesting question. I mostly want this to be a very clear error message, where checking it is not driver-specific string parsing. I can imagine an app trying to sign a URL on startup to see if the current provider supports it, and disabling some functionality otherwise.

shantuo: Can we move the check forward? For example, the user may found out there isn't a signed URL enabled provider when they assemble their providerset.

vangent: That's what the [enforcing portability](https://github.com/google/go-cloud/blob/master/internal/docs/design.md#enforcing-portability) would achieve, but we decided to wait on implementing that until we had more evidence that users want it. I don't think there's a way to fail at startup/init time that doesn't involve the user telling us what features they want.

Issue #532, opened 2018-10-12
blob: do not use relative path to access conformance testdata


shantuo:
This will fail if the provider implementation resides outside the main repo, e.g. https://github.com/google/go-cloud/blob/master/blob/s3blob/s3blob_test.go#L58

vangent:
Yea, there is a [TODO](https://github.com/google/go-cloud/blob/master/blob/drivertest/drivertest.go#L272) to use go-bindata to inline them as source.

PR #533, opened 2018-10-12
blob: feat aliyunoss blob

aimuz:


aimuz:
I signed it!

aimuz:
Do you need to add aliyunoss accessKey?

vangent:
Hi @aimuz --

Sorry, we are unable to accept contributions that add implementations for other Cloud providers. It's just not feasible for us to commit to ownership for all of them. Rather, we expect such implementations to live in independent repositories. As described [here](https://github.com/google/go-cloud#contributing), we'd be happy to add a link to your repository from Go Cloud, and to let you know about breaking changes while the APIs are in alpha.

There are a couple of things that aren't quite ready (e.g., Issue #532) for this model, but we are actively working on making it possible.

Separately, see if you can make your tests work via [record/replay](https://github.com/google/go-cloud/blob/master/internal/docs/design.md#tests); this allows them to be run repeatably by anyone without any need for access keys. Only `-record` mode needs a real bucket etc. to be provisioned. The S3 and GCS implementations both use this model.

PR #536, opened 2018-10-12
github: add templates for bug reports, feature requests, and pull requests

vangent:


zombiezen:
@vangent, this does not appear to be working.

PR #544, opened 2018-10-15
all: fix opencensus aws exporter import path

felipeweb:
if a project has go-cloud as a dependency and try run ```go get -u``` it
fails because aws exporter path is wrong.
More details: https://github.com/golang/go/issues/28202


felipeweb:
@vangent i remove aws sdk 1.15.27, but tests still falling and i don't know why. it is my fault?

vangent:
Try removing this line from `go.mod`:

```
github.com/go-ini/ini v1.37.0
```

It looks like the newer version of one of the modules you updated requires an older version of `ini` and it's not playing nice with this one.

PR #549, opened 2018-10-15
all: go get -u && go mod tidy

vangent:
I had to regenerate the GCS goldens because something in the URL changed as part of the version change.


vangent:
I am not sure why Travis is trying to load the module that it's failing on.

vangent:
After much discussion, I am partway toward figuring out what went wrong here and continuing investigation. In the meantime, I have a workaround that updates the proxy with the right set of modules, so the Travis build works now.

PR #550, opened 2018-10-16
wire/test: support multi-line errors and -record mode in tests

### LINKED ISSUE #502: wire/tests: Allow wire_errs.txt to match multi-line errors, and to use regexes
Current the wire tests verify that the whole line from wire_errs.txt is found in one of the actual errors. It doesn't work well for errors that span multiple lines (e.g., the err.String() has \n's in it). In addition, some errors have things like line numbers that are fragile to include, maybe a regex match would be more flexible.

vangent:
Fixes #502.

This PR contains a few related changes:

1. `wire_errs.txt` now supports multi-line errors, by separating distinct errors with `\n\n` instead of just `\n`.
2. You can now use `-record` for errors, as long as the the `wire_errs.txt` file exists (indicating that you expect an error).
3. The error string is scrubbed of line numbers and line positions.
4. The entire set of errors is checked against `wire_errs.txt`, using `cmp.Diff`.
5. I had to sort something in the error for cycles to make it stable.


wire/internal/wire/wire_test.go
```diff
@@ -81,15 +82,22 @@ func TestWire(t *testing.T) {
...
 				if !test.wantWireError {
-					t.Fatal("Did not expect errors.")
+					t.Fatal("Did not expect errors. To -record an error, create want/wire_errs.txt.")
 				}
-				for _, s := range test.wantWireErrorStrings {
-					if !errorListContains(errs, s) {
-						t.Errorf("Errors did not contain %q", s)
+				if *setup.Record {
+					wireErrsFile := filepath.Join(testRoot, test.name, "want", "wire_errs.txt")
+					if err := ioutil.WriteFile(wireErrsFile, []byte(strings.Join(gotErrStrings, "\n\n")), 0666); err != nil {
```
zombiezen: This feels unfortunate, since unlike the expected output, the errors is an intentional minimum bound on what's expected instead of verbatim. This behavior would clobber existing files and the dev would have to go back and clean up diffs.

vangent: Actually, with this change, `wire_errs.txt` is not a minimum bound anymore. It's exactly the set of errors that are output (well, scrubbed). So, I think `-record` makes perfect sense.

Do you disagree with the change from it being an intentional minimum bound?

zombiezen: Oh I didn't realize the change. Yes, I suppose that if the line numbers are scrubbed and the paths are stable, having exact error output is fine.

PR #557, opened 2018-10-20
health: write error message in health response

palvarezcordoba:
The method CheckHealth of Checker interface returns an error, but it's message is not used.
writeUnhealthy func always writes unhealthy", now it writes the error message.

### DIFF
--- a/health/health.go
+++ b/health/health.go
@@ -35,7 +35,8 @@ func (h *Handler) Add(c Checker) {
 func (h *Handler) ServeHTTP(w http.ResponseWriter, _ *http.Request) {
 	for _, c := range h.checkers {
 		if err := c.CheckHealth(); err != nil {
-			writeUnhealthy(w)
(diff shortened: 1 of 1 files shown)


palvarezcordoba:
I signed it!

palvarezcordoba:
I signed it!

shantuo:
Thanks for the PR @palvarezcordoba ! IIUC you'd like to show the error message in the HTTP response of the health check page. The design of the health checks is to let the infrastructure makes health check requests and decide whether or not to send  requests to the server. When an error returns from one of the health checkers, it is likely that the underlying code returning the error handles the error and either logs the error or reports/alerts it in a proper way. To me it is 1) not very useful to print them in the response; 2) public to a user of the website for the error message which may include your stack trace as well. Therefore I am going to close this PR, but feel free to file a issue discussing it further!

PR #571, opened 2018-10-25
wire: use multiline instead of one single line for errors with traces

### LINKED ISSUE #570: wire: use multiline for errors with traces
There are a couple of `wire` errors that do the equivalent of stack traces. They are currently single-line, separate by characters like ", needed by" or "<- ", because until recently the wire tests couldn't support multi-line errors. They do now, so we should update them to be more readable.

vangent:
Fixes #570 .

### DIFF
--- a/wire/internal/wire/analyze.go
+++ b/wire/internal/wire/analyze.go
@@ -141,11 +141,12 @@ dfs:
 				index.Set(curr.t, errAbort)
 				continue
 			}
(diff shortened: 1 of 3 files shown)


zombiezen:
Spelling error in commit message? I would merge in, but I'm not sure what you meant to say.

vangent:
oops, yes -- trackes --> traces

PR #572, opened 2018-10-25
blob: remove ListOptions.PageSize

### LINKED ISSUE #564: blob: remove ListOptions.PageSize
Per discussion on #510, there's no good reason to expose this in the concrete type once #563 is fixed.

vangent:
Given that we don't think there's a major performance cost to retrieving extra blob attributes, and that the concrete type interface is an iterator, there's no need to expose the page size.

This PR removes this option from the concrete type.

The driver interface is still paged, so `PageSize` is still exposed on the driver's `ListOptions`. It is only used in tests. I think it's reasonable to have the driver expose a paged interface still; otherwise, each driver would need to implement an iterator over what is likely a paged/bulk interface to the backend.

Fixes #564.

### DIFF
--- a/blob/gcsblob/gcsblob.go
+++ b/blob/gcsblob/gcsblob.go
@@ -41,6 +41,8 @@ import (
 	"google.golang.org/api/option"
 )
 
+const defaultPageSize = 1000
+
 // Options sets options for constructing a *blob.Bucket backed by GCS.
 type Options struct {
 	// GoogleAccessID represents the authorizer for SignedURL.
@@ -130,9 +132,12 @@ func (b *bucket) ListPaged(ctx context.Context, opt *driver.ListOptions) (*drive
 			return nil, err
 		}
 	}
+	pageSize := opt.PageSize
+	if pageSize == 0 {
+		pageSize = defaultPageSize
+	}
 	iter := bkt.Objects(ctx, query)
-	pager := iterator.NewPager(iter, opt.PageSize, string(opt.PageToken))
-
+	pager := iterator.NewPager(iter, pageSize, string(opt.PageToken))
 	var objects []*storage.ObjectAttrs
 	nextPageToken, err := pager.NextPage(&objects)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


blob/gcsblob/gcsblob.go
```diff
@@ -130,9 +132,12 @@ func (b *bucket) ListPaged(ctx context.Context, opt *driver.ListOptions) (*drive
...
 		}
 	}
+	pageSize := opt.PageSize
+	if pageSize == 0 {
+		pageSize = defaultPageSize
+	}
 	iter := bkt.Objects(ctx, query)
-	pager := iterator.NewPager(iter, opt.PageSize, string(opt.PageToken))
-
+	pager := iterator.NewPager(iter, pageSize, string(opt.PageToken))
```
jba: This is overkill. Use the client's iterator, but set the page size with https://godoc.org/cloud.google.com/go/storage#ObjectIterator.PageInfo.

jba: (May make more sense in a separate PR.)

vangent: I took a look, and yes, I could set `iter.PageInfo.Token` and `iter.PageInfo.MaxSize`, and then use `iter.Next()` until `iter.PageInfo.Remaining == 0`. That's what you're suggesting, right? I'm not sure that's simpler than what I have here.... do you think it's simpler, or is there some other kind of overhead to using `NewPager`?

zombiezen:
I will wait for @jba's clarification on the one current review thread.

blob/gcsblob/gcsblob.go
jba: I see what's going on here. The driver API is at the right level, but you're using a high-level client (c.g.c/go/storage) that's already hiding pagination for you. As a result, `gcsblob.bucket.ListPaged` is doing a lot of extra work on each page.

I'll need to think about what the right thing to do here is. But I don't think it needs to block this PR.

vangent: OK, will leave it like this for now.

Issue #569, opened 2018-10-25
blob: add support for content-md5


myml:
### Is your feature request related to a problem? Please describe.

add support for [Content-md5](https://tools.ietf.org/html/rfc1864)

### Describe alternatives you've considered

```go
type WriterOptions struct{
    ContentMD5 string
    ...
}
```

jba:
Confirmed that the three major cloud providers support this on blob write.

vangent:
Thanks for the contribution @myml ! This will be in our next release.

PR #576, opened 2018-10-26
blob: add support for content-md5

myml:
see #569
Because blob import "github.com/google/go-cloud/blob/driver"
```go
dopt = &driver.WriterOptions{
  ContentMD5:  opt.ContentMD5,
  BufferSize:  opt.BufferSize,
  BeforeWrite: opt.BeforeWrite,
}
```
at next merge

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -484,6 +484,10 @@ type WriterOptions struct {
 	// http://mimesniff.spec.whatwg.org/
 	ContentType string
 
+	// Content-MD5 which may be used as a message integrity check (MIC)
+	// https://tools.ietf.org/html/rfc1864
+	ContentMD5 string
+
 	// Metadata are key/value strings to be associated with the blob, or nil.
 	// Keys may not be empty, and are lowercased before being written.
 	// Duplicate case-insensitive keys (e.g., "foo" and "FOO") are an error.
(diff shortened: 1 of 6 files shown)


blob/blob.go
```diff
@@ -484,6 +484,10 @@ type WriterOptions struct {
 	// http://mimesniff.spec.whatwg.org/
 	ContentType string
 
+	// Content-MD5 which may be used as a message integrity check (MIC)
```
vangent: Nit: end comment with a period, here and elsewhere.

blob/driver/driver.go
```diff
@@ -65,6 +65,9 @@ type WriterOptions struct {
 	// write in a single request, if supported. Larger objects will be split into
 	// multiple requests.
 	BufferSize int
+	// Content-MD5 which may be used as a message integrity check (MIC)
```
vangent: Hi @myml, thanks for the contribution!

`blob` and `blob/driver` are both in the same module, so you shouldn't need to split this change up into multiple Pull Requests.

Also, I'd like to see it working, including the implementation for `s3blob` and `gcsblob` (these should be easy, just pass-through to the provider) and `fileblob` (might be trickier...), and a test in `drivertest/drivertest.go`.

myml:
Hi @vangent .
I have a new request! including the implementation for s3blob, gcsblob and fileblob.
s3blob and fileblob passed the test, I can't test gcsblob because there is no google cloud account

vangent:
I cloned this PR over to #589 so that I could add the golden files for GCS and S3, please review there.

Issue #577, opened 2018-10-26
blob/test: improve comments and tests for List


vangent:
These should wait until #541 is done.

- Add a comment about guarantees for blob.Bucket.Next consistency (i.e., everything in UTF-8 lexicographical order) (this might already be there).
- Add a comment to driver.Bucket.ListPaged about consistency requirements (i.e., everything in UTF-8 lexicographical order, including across pages).
- Add tests with a Unicode filename, a filename with "/", a filename with "\".
- Add a test that verifies no "directory" is returned after the last file "in" it is deleted.
- Add a test that verifies consistency across ListPaged after insert & delete. E.g., "a b c d e f", fetch Page of size 3 = "a b c", then [insert after "b" | delete "b"], should get "d e f" in next Page.

vangent:
> I think the concrete implementation of Bucket.Next should account for this by calling ListPaged in a loop.

I agree; this should be fixed as part of #565, which is adding `io.EOF` to make "no more results" more explicit; I added a comment there.

For the conformance tests that call `driver.ListPaged` instead of `blob.List`, it's true that some of the tests are making assumptions about how pagination works. For example, there are tests that set page size to 1 and verify that the first RPC will return 1 entry, next page will return another, etc. This is technically too strict, as it would be valid to return, say, 1, 0, 0, 1 entries. I'm inclined to say that we can relax these tests if this becomes an issue; it feels low-priority to me. Feel free to file an issue and add a TODO referring to it. Note that it only matters in `--record` mode; in replay mode, we're replaying a recorded set of RPC results so they'll always be the same.

> Another problem with this test is that it assumes immediate (vs. eventual) consistency

I don't think it does. The second page should get "d e f" regardless. It's true that a provider that isn't handling this correctly, but also has only eventual consistency, could sometimes pass the test (so, the test would likely be flaky instead of always failing). I think this is fine.

For all of the `List` tests, I'm creating a set of blobs during initialization, and then expecting `List` to return them. This is potentially flaky for providers with eventual consistency. I currently "fix" this by not deleting the blobs after the test, so the potential flakiness is only for the first run. Another approach would be to create the blobs and then wait for them to show up before proceeding, but that's circular (the only way to tell if they've showed up is to call `List`, which is what we're supposedly testing), so the test would end up timing out for buggy providers. Unclear this is better. Again, this potential flakiness is only in `-record` mode.

vangent:
All of the tests described above have been added, in #588 and #578.

PR #579, opened 2018-10-26
blob/test: add tests for driver.ListPaged pagination consistency when keys are inserted/delete

vangent:
Issue #577.

Add tests for `driver.ListPaged` pagination consistency when keys are inserted/deleted.

Given "1 2 3", and page size 2, the first page should return "1 2".
The next page should return "3" even if a key is inserted between "1" and "2".
It should also return "3" even if "1" or "2" is deleted.


blob/drivertest/drivertest.go
```diff
@@ -317,6 +319,96 @@ func testList(t *testing.T, newHarness HarnessMaker) {
...
+			PageToken: page.NextPageToken,
+		})
+		if err != nil {
+			t.Fatal(err)
+		}
+		got = gotIndices(t, page.Objects)
+		want = []int{2}
+		if diff := cmp.Diff(got, want); diff != "" {
+			t.Errorf("got\n%v\nwant\n%v\ndiff\n%s", got, want, diff)
+		}
```
shantuo: In this case, do you define the behavior of calling ListPaged again on the newly returned page.NextPageToken? Should it return {0a}?

vangent: What do you mean by the newly returned NextPageToken?

- Do you mean calling `ListPaged` again with the same arguments as before? If so -- it should returned the same thing, just `2` again.
- Do you mean calling `ListPaged` with the `NextPageToken` returned by the last called to `ListPaged`? If so, it will be empty -- there are no more results.

If the concrete type called `ListPaged` again with an empty `PageToken`, it would start iteration again at the beginning, returning 0, 0a, 1, 2.... (with pagination).

vangent:
Hope I answered your question.

PR #587, opened 2018-10-29
wire: expand on package documentation

### LINKED ISSUE #514: wire: improve package documentation
The [package documentation](https://godoc.org/github.com/google/go-cloud/wire) should definitiively document the semantics of the wire functions. Currently the documentation is pretty thin and relies a lot on the introduction document. For example, the doc comment for Build says "its arguments are the same as NewSet", and although the doc comment for NewSet describes the possible kinds of parameters, it doesn't say what they mean. I believe that it should be possible to read the godoc and understand how to use wire and read code that uses it.

zombiezen:
Fixes #514

/cc @vangent


vangent:
I'll review this since @shantuo already has a bunch of PRs to review.

wire/wire.go
```diff
@@ -13,22 +13,60 @@
 // limitations under the License.
 
 // Package wire contains directives for Wire code generation.
+// For an overview of working with Wire, see the user guide at
+// https://github.com/google/go-cloud/blob/master/wire/README.md
+//
+// The directives in this package are used as input to the Wire code
+// generation tool. The entry point of Wire's analysis are injector functions:
+// function templates denoted by only containing a call to Build. Build names
```
vangent: How about instead of "Build names", "The arguments to Build define a set of providers" or "are a set of providers"?

wire/wire.go
```diff
@@ -13,22 +13,60 @@
...
+// a set of providers and the Wire code generation tool builds a directed
+// acylic graph of the providers' output types. The generated generated code
```
vangent: drop one "generated".

wire/wire.go
```diff
@@ -13,22 +13,60 @@
...
+// will fill in the function template by calling the providers from the
```
vangent: How about instead of "calling", something like "... by using the providers in the provider set to instantiate any needed types."?

Not all providers are "called", e.g. struct providers, and I think it might add clarity to be explicit about why it's calling the providers.

wire/wire.go
```diff
@@ -13,22 +13,60 @@
...
+// return an error as its last return value and a cleanup function as the
+// second return value. A cleanup function must be of type func() and is
+// guaranteed to be called before the cleanup function of any of the
+// provider's inputs. If any provider returns an error, the injector function
+// will call all the appropriate cleanup functions and return the error from
+// the injector function.
+//
+// Passing a struct value of type S to NewSet declares that both S and *S will
+// be provided by creating a new value of the appropriate type by filling in
+// each field of S using the provider of the field's type.
```
vangent: each exported field?

wire/wire.go
```diff
@@ -13,22 +13,60 @@
...
+//
+// Passing a ProviderSet to NewSet is the same as if the set's contents
+// were passed as arguments to NewSet directly.
```
vangent: What about Value or InterfaceValue?

wire/wire.go
```diff
@@ -13,22 +13,60 @@
...
-// providers to use. Its arguments are the same as NewSet. Its return
-// value is an error message that can be sent to panic.
+// Build is placed in the body of an injector function template to declare the
+// providers to use. The Wire code generation tool will fill in an
+// implementation of the function. The arguments to Build are interpreted the
+// same as NewSet: they determine the provider set presented to Wire's
+// dependency graph. Build returns an error message that can be sent to a call
+// to panic().
+//
+// The parameters of an injector function are used as inputs in the dependency
```
vangent: an -> the?

wire/wire.go
```diff
@@ -13,22 +13,60 @@
...
+// graph. Similar to provider functions passed into NewSet, the first return
```
vangent: Start a new paragraph at "Similar to"? I think this is independent from the previous sentence about parameters.

wire/wire.go
```diff
@@ -13,22 +13,60 @@
...
 
-// Build is placed in the body of an injector function to declare the
-// providers to use. Its arguments are the same as NewSet. Its return
-// value is an error message that can be sent to panic.
+// Build is placed in the body of an injector function template to declare the
+// providers to use. The Wire code generation tool will fill in an
+// implementation of the function. The arguments to Build are interpreted the
+// same as NewSet: they determine the provider set presented to Wire's
+// dependency graph. Build returns an error message that can be sent to a call
+// to panic().
```
vangent: This is fuzzy to me. Build doesn't return anything really, does it? The code here is never executed, it's more of a template for the code Wire is generating.

wire/wire.go
zombiezen: Done.

wire/wire.go
zombiezen: Which one? ;) Done.

wire/wire.go
zombiezen: Done.

wire/wire.go
zombiezen: Doesn't have to be, if S is in the same package as the injector.

wire/wire.go
zombiezen: Added a brief paragraph pointing to other functions' doc comments.

wire/wire.go
zombiezen: I mean, it does actually return a string. I feel that some bit of prose needs to justify why it returns a `string`, but I can't think of a better phrasing.

wire/wire.go
zombiezen: Done.

wire/wire.go
zombiezen: Done.

PR #588, opened 2018-10-29
blob/fileblob: use escaping to allow more flexible filenames, and use os.Walk to get all keys

vangent:
Also added a test for various "weird" blob names, which now passes for `fileblob`.

#577


blob/fileblob/fileblob.go
```diff
@@ -57,35 +59,18 @@ func OpenBucket(dir string) (*blob.Bucket, error) {
...
-	}
-	if key == "." {
-		return "", fmt.Errorf("invalid path \".\"")
-	}
-	if strings.HasPrefix(key, "../") {
-		return "", fmt.Errorf("starts with \"../\"")
-	}
-	return filepath.FromSlash(key), nil
+func escape(s string) string {
+	s = url.PathEscape(s)
```
zombiezen: My concern still here is that we need a specific property that URL escaping usually does not provide, which is that given the same input string, the escaping always produces the exact same string. If `PathEscape` truly provides this, let's document it here, but if not, we should have an explicitly stable naming algorithm.

If the latter, percent-encoding is still fine, but we should make the whitelisted characters explicit.

blob/fileblob/fileblob.go
```diff
@@ -57,35 +59,18 @@ func OpenBucket(dir string) (*blob.Bucket, error) {
...
+	// Unescape "/".
+	return strings.Replace(s, "%2F", "/", -1)
```
zombiezen: This should be replaced with the OS's file separator, which isn't always slash.

blob/fileblob/fileblob.go

Per meatspace discussion, during unescaping I now also return an error if `escape(unescape(s)) != s`, so that `blob` is internally consistent.

Also added tests.

blob/fileblob/fileblob.go
vangent: Done. It will be nice to get Windows tests for this package.

vangent:
PTAL.

blob/fileblob/fileblob.go
```diff
@@ -16,14 +16,21 @@
...
-// Exception: "/" is not escaped, so that it can be used as the real file
-// separator on the filesystem.
-// Filenames on disk that return an error for url.PathUnescape are not visible
-// using fileblob.
+// Blob keys are escaped before being used as filenames, and filenames are
+// unescaped when they are passed back as blob keys during List. The escape
+// algorithm is:
+// -- Alphanumeric characters (A-Z a-z 0-9) are not escaped.
+// -- Space (' '), dash ('-'), underscore ('_'), and period ('.') are not escaped.
+// -- The OS-specific path separator character (os.PathSeparator) is not escaped.
```
zombiezen: Not quite. Slash should always map to the OS-specific path separator character.

blob/fileblob/fileblob.go
```diff
@@ -16,12 +16,21 @@
...
-// or the sequence "/./" is not permitted. This is to ensure that blob names map
-// cleanly onto files underneath a directory.
+// Blob keys are escaped before being used as filenames, and filenames are
+// unescaped when they are passed back as blob keys during List. The escape
+// algorithm is:
+// -- Alphanumeric characters (A-Z a-z 0-9) are not escaped.
+// -- Space (' '), dash ('-'), underscore ('_'), and period ('.') are not escaped.
+// -- The OS-specific path separator character (os.PathSeparator) is not escaped.
+// -- All other characters are escaped similar to url.PathEscape:
+//    "%<hex ASCII code>", with capital letters ABCDEF in the hex code.
```
zombiezen:
```suggestion
//    "%<hex UTF-8 byte>", with capital letters ABCDEF in the hex code.
```

blob/fileblob/fileblob_test.go
```diff
@@ -84,3 +84,71 @@ func TestNewBucket(t *testing.T) {
...
+			want: "abc09ABC -_.",
+		},
+		{
+			key:  "~!@#$%^&*()+`=[]{}\\|;:'\",<>,",
+			want: "%7E%21%40%23%24%25%5E%26%2A%28%29%2B%60%3D%5B%5D%7B%7D%5C%7C%3B%3A%27%22%2C%3C%3E%2C",
+		},
+		{
+			key:  "☺☺",
+			want: "%E2%98%BA%E2%98%BA",
+		},
```
zombiezen: Should also test slashes.

blob/fileblob/fileblob.go

PR #596, opened 2018-10-31
wire: allow wire.Value to use values with no parent (i.e., struct fields)

vangent:
The check for "don't allow values of function scope" is too strict, it is rejecting use of "X" instead the struct, which has Parent() == nil.

I've added a testcase matching the README, and updated the README to match the actual generated code.


zombiezen:
Does this error on:

```go
package foo

var bar struct {
  X int
}

package bar

func inject() int {
  panic(wire.Build(foo.bar.X))
}
```

I would expect the above to error out. I'm convinced of the positive case you gave; I'm not convinced of the negative cases yet.

vangent:
I believe it does; I added that check (although with the `bar` struct in the `bar` package and the injector in `main`) to the existing UnexportedValue testcase, which had a similar test.

vangent:
Actually, adding that injector to the existing test masked the previous error, so I made a new test `UnexportedStruct`. PTAL and see if this addresses your example.

Issue #597, opened 2018-10-31
blob: add examples showing how to set up gcsblob and s3blob with various config options


vangent:
For example, GCS's `storage.NewClient` accepts `ClientOption`:
https://godoc.org/google.golang.org/api/option#ClientOption

It should be possible to easily pass those in. Same for S3:
https://docs.aws.amazon.com/sdk-for-go/api/aws/session/

vangent:
I think @zombiezen has convinced me that we shouldn't try to expose all of the options in our constructors, but rather accept the injection of the result (e.g., the HTTPClient). However we should make it as easy as we can for a user to construct it, even without using Wire, and to be able to customize various options.

This may involve adding some helpers to the `gcp` and/or `aws` packages. It should definitely involve writing some example code. For example `gcsblob/example_test.go` could have example for calling `gcsblob.OpenBucket` with everything defaulted, overridden credentials, with an API key, etc. Available options for GCS are discussed [here](https://cloud.google.com/docs/authentication/production) and [here](https://godoc.org/google.golang.org/api/option#ClientOption) (there's overlap).

Similarly for `s3blob`.

Issue #423, opened 2018-09-11
blob: creation of *blob.Bucket from URLs


sargun:
It would be nice if someone could provide a URL, and that would return a *blob.Bucket. If this project standardized on a schema like:

* Google Cloud: `gcp://mybucket/`
* S3: `s3://mybucket/`
* S3: `s3://AWS_ACCESS_KEY_ID: AWS_SECRET_ACCESS_KEY@mybucket/`
To specify region, maybe something like:
* S3: `s3://AWS_ACCESS_KEY_ID: AWS_SECRET_ACCESS_KEY@mybucket/?region=us-east-1`

vangent:
Hi @sargun , thanks for the suggestion! Do you have a specific use case for this feature?

If I understand correctly, to avoid having to compile in all provider drivers, this would require each provider driver to "register" a URL prefix and a parser, something like:

```
RegisterURL("s3", func(s string) (*blob.Bucket, error) {
  // Parse the rest of the URL
})
```

sargun:
@vangent I would be fine with the current approach where we implicitly include the drivers, and they do their registrations in init(). The primary use case is around configuration, and initialization of the drivers. Today, I have to implement configuration for each blob driver and all of the inputs independently. 

If we look at the RDBMS community, they've solved this to a degree with DB URLs: https://github.com/xo/dburl -- in that, the ORM can usually be initialized based on some URL schema as opposed to bespoke configuration inputs per backend.

vangent:
OK, thanks for the context. We'll discuss this for our next Sprint, so expect the next update in a couple of weeks.

zombiezen:
Next steps here would be to have a more precise design of what these locators would look like and how these strings would be dispatched to bucket-opening functions.

PR #598, opened 2018-10-31
blob: add blob.Open for creating arbitrary provider blobs from URLs

vangent:
The concrete type defines a registry, and provider implementations use `blob.Register` in their package `init()` function to register to handle specific protocols (e.g. "s3" or "gcs"). In `Open`, the concrete type parses the URL into a protocol, bucket name, and map of options, and uses the protocol to identify which provider to call.

This PR only adds the "file" protocol; "gcs" and "s3" will be added in subsequent PRs because I think it makes sense to do some refactoring (i.e., #597) first.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -524,7 +524,8 @@ type WriterOptions struct {
 	BeforeWrite func(asFunc func(interface{}) bool) error
 }
 
-// FromURLFn allows providers to convert a parsed URL to a *Bucket.
+// FromURLFn is for use by provider implementations.
+// It allows providers to convert a parsed URL from Open to a driver.Bucket.
 // It takes a bucket name and a map of key/value options. The map is guaranteed
 // to be non-nil, and all keys will be in lowercase.
 type FromURLFn func(context.Context, string, map[string]string) (driver.Bucket, error)
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -24,6 +24,7 @@ import (
 	"io/ioutil"
 	"mime"
 	"net/http"
+	"regexp"
 	"strings"
 	"time"
 
@@ -271,6 +272,7 @@ type Bucket struct {
 }
 
 // NewBucket creates a new Bucket for a group of objects for a blob service.
+// It is for use by provider implementations.
 func NewBucket(b driver.Bucket) *Bucket {
 	return &Bucket{b: b}
 }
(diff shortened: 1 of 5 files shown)


blob/blob.go
```diff
@@ -522,6 +524,69 @@ type WriterOptions struct {
...
+func Register(protocol string, fn FromURLFn) error {
+	if _, found := registry[protocol]; found {
+		return fmt.Errorf("a provider has already registered for protocol %q", protocol)
+	}
+	registry[protocol] = fn
+	return nil
+}
+
+// urlRegex is used to parse URLs in Open.
+var urlRegex = regexp.MustCompile("^([^:]+)://([^?]+)(.*)$")
```
jba: Why not url.Parse?

blob/blob.go
```diff
@@ -522,6 +524,69 @@ type WriterOptions struct {
 	BeforeWrite func(asFunc func(interface{}) bool) error
 }
 
+// FromURLFn is for use by provider implementations.
+// It allows providers to convert a parsed URL from Open to a driver.Bucket.
+// It takes a bucket name and a map of key/value options. The map is guaranteed
+// to be non-nil, and all keys will be in lowercase.
+type FromURLFn func(context.Context, string, map[string]string) (driver.Bucket, error)
```
jba: I think `FromURLFunc` would be more idiomatic.

jba:
Last comment on #423 says there should be a design discussion. Has that happened?

blob/blob.go
```diff
@@ -522,6 +524,69 @@ type WriterOptions struct {
...
+
+// Register is for use by provider implementations. It allows providers to
+// register an instantiation function for URLs with the given protocol. It is
+// expected to be called from the provider implementation's package init
+// function.
+//
+// fn will be called from Open, with a bucket name and options parsed from
+// the URL. All option keys will be lowercased.
+//
+// An error is returned if a provider has already registered for protocol.
```
jba: That leaves it up to the providers to check, and they may not be motivated to do so. I'd suggest panicking instead.

blob/blob.go
```diff
@@ -522,6 +524,69 @@ type WriterOptions struct {
...
+func Register(protocol string, fn FromURLFn) error {
+	if _, found := registry[protocol]; found {
```
jba: This needs a mutex.

blob/blob.go
vangent: Um, yeah. Why not? No good reason.

Done. I ended up just passing the parsed URL to the providers.

blob/blob.go

blob/blob.go
vangent: Good point, done.

blob/blob.go

PR #602, opened 2018-11-01
blob/drivertest: factor out common counting code

jba:


blob/drivertest/drivertest.go
```diff
@@ -796,6 +776,21 @@ func testListDelimiters(t *testing.T, newHarness HarnessMaker) {
 	}
 }
 
+func countItems(ctx context.Context, t *testing.T, iter *blob.ListIterator) int {
+	count := 0
+	for {
+		_, err := iter.Next(ctx)
```
vangent: Nit: this could be shortened a bit, right?

```
if _, err := iter.Next(ctx); err == io.EOF {
  break
} else if err != nil {
  t.Fatal(err)
}
count++
```

PR #605, opened 2018-11-01
blob: add URL schemes for gcsblob and s3blob

### LINKED ISSUE #423: blob: creation of *blob.Bucket from URLs

vangent:
Fixes #423.

### DIFF
--- a/blob/gcsblob/gcsblob.go
+++ b/blob/gcsblob/gcsblob.go
@@ -69,7 +69,7 @@ func init() {
 		if keyPath := q["private_key_path"]; len(keyPath) > 0 {
 			pk, err := ioutil.ReadFile(keyPath[0])
 			if err != nil {
-				return nil, fmt.Errorf("failed to read private key: %v", err)
+				return nil, fmt.Errorf("reading private key: %v", err)
 			}
 			opts.PrivateKey = pk
 		}
@@ -84,11 +84,11 @@ func init() {
 		} else {
 			jsonCreds, err := ioutil.ReadFile(credPath[0])
 			if err != nil {
-				return nil, fmt.Errorf("failed to read credentials: %v", err)
+				return nil, fmt.Errorf("reading credentials: %v", err)
 			}
 			creds, err = google.CredentialsFromJSON(ctx, jsonCreds)
 			if err != nil {
-				return nil, fmt.Errorf("failed to load credentials: %v", err)
+				return nil, fmt.Errorf("loading credentials: %v", err)
 			}
 		}
 
--- a/blob/gcsblob/gcsblob.go
+++ b/blob/gcsblob/gcsblob.go
@@ -14,6 +14,17 @@
 
 // Package gcsblob provides an implementation of blob that uses GCS.
 //
+// For blob.Open URLs, gcsblob registers for the "gs" protocol.
+// The URL's Host is used as the bucket name.
+// The following query options are supported:
+// - cred_path: Sets path to the Google credentials file. If unset, default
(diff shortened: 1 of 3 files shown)


blob/gcsblob/gcsblob.go
```diff
@@ -37,13 +49,57 @@ import (
...
+		opts := &Options{}
+
+		if accessID := q["access_id"]; len(accessID) > 0 {
+			opts.GoogleAccessID = accessID[0]
+		}
+
+		if keyPath := q["private_key_path"]; len(keyPath) > 0 {
+			pk, err := ioutil.ReadFile(keyPath[0])
+			if err != nil {
+				return nil, fmt.Errorf("failed to read private key: %v", err)
```
jba: nit: the usual Go style is `"reading private key: %v", err`.  In other words, we don't usually say "failed" or "error while...". 

If you've adopted a different convention for Go Cloud, that's fine; it's not a big deal.

jba: I should also mention that if something like the Go 2 error-handling proposal is adopted, it will be much easier to have one message for the whole function, so maybe we shouldn't try to be too specific here. Of course, we shouldn't consider the error message to be part of our API, so we can always change it later.

blob/gcsblob/gcsblob.go
```diff
@@ -37,13 +49,57 @@ import (
...
+
+		var creds *google.Credentials
+		if credPath := q["cred_path"]; len(credPath) == 0 {
+			var err error
+			creds, err = gcp.DefaultCredentials(ctx)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			jsonCreds, err := ioutil.ReadFile(credPath[0])
```
jba: This is going to come up a lot. Consider factoring it out to an internal function.

See https://code.googlesource.com/google-api-go-client/+/master/internal/creds.go.

blob/s3blob/s3blob.go
```diff
@@ -14,6 +14,12 @@
 
 // Package s3blob provides an implementation of blob using S3.
 //
+// For blob.Open URLs, s3blob registers for the "s3" protocol.
+// The URL's Host is used as the bucket name.
+// The following query options are supported:
+// - region: The AWS region for requests.
```
jba: Where does the auth go?

vangent: I updated the comment and added a link to the [AWS documentation for creating a session](https://docs.aws.amazon.com/sdk-for-go/api/aws/session/). Similar to GCP, AWS uses an environment variable to find the default creds. I expect we will end up adding more query parameters here to override more `aws.Config` fields.

vangent: I updated the comment to clarify and reference the [AWS documentation for creating a session](https://docs.aws.amazon.com/sdk-for-go/api/aws/session/). Similar to GCP, it uses an environment variable to find the default credentials. I expect we will end up adding support for more query parameters here to override other `aws.Config` fields.

vangent:
Travis got stuck, made a minor comment change to kickstart it.

PR #613, opened 2018-11-05
internal/contributebot: fix build breakage: filevar.NewVariable is now New

h8liu:
example build log:
https://gopkgs.io/builds/20181105-074139-ef381867


h8liu:
I am a bit surprised that your ci does not cover this..

vangent:
Thanks for the fix. I too am surprised that our CI didn't cover this; filed #614.

Issue #323, opened 2018-08-14
all: run Windows CI


zombiezen:
It would be good to run the tests on Windows to shake out any development issues with using this library with Windows. [Appveyor](https://www.appveyor.com/) seems to be the preferred way to do this.

peter-evans:
Travis now supports windows so I tested the build and there are a number of issues as you can see here:
https://travis-ci.org/peter-evans/go-cloud/jobs/445444398

I'm trying to tackle the `cannot find package "example.com/foo"` errors. The cause of the error is the forward slash in the package path. I attempted to fix it using `filepath.FromSlash` which converts to an os-specific separator, but then for some reason it escapes the back slash twice at some later point.

zombiezen:
@peter-evans Cool! Thank you for looking into this! We would really appreciate having the Windows build!

Those are the Wire tests, which run a semi-simulated Go build. The slash in the import path is correct: Go import paths don't change even if OS path separator is different. But in examining the tests, they do translation back and forth between slash-separated and OS-specific paths. It's quite plausible it's incorrect right now, so you would need to examine the data flow to progress.

zombiezen:
@peter-evans I believe I have fixed the path separator issue in the Wire tests in #616, but I don't have a Windows machine I can confirm with right now. Please give it another try and see if you still run into issues.

vangent:
I send #621 to try enabling the Windows build, and it looks like all tests passed! So I don't think there's anything more to do here.

zombiezen:
Sadly, it does seem like there is more work: https://travis-ci.com/google/go-cloud/jobs/156718327

Rolling back with #622

vangent:
@peter-evans feel free to take another look at this!

PR #616, opened 2018-11-05
wire/internal/wire: use on-disk GOPATH in generate tests

zombiezen:
The primary motivation is to permit a move to using `go/packages` instead of `go/loader`. `go/packages` runs exclusively by shelling out to the go tool, which precludes use of the in-memory "magic" GOPATH being used up to this point.

This has a secondary effect of removing a lot of code to support "magic" GOPATH from the test infrastructure. This is on the whole good, but necessitated a change in the error scrubbing: since the filenames are no longer fixed, error scrubbing also must remove the leading `$GOPATH/src` lines.

Another related change: since all callers of `Generate` needed to know the package path in order to write out wire_gen.go (necessitating a find-only import search) and `Generate` already has this information, `Generate` now returns this information to the caller. This should further reduce callers' coupling to Wire's load internals. It also eliminates code duplication.

This should hopefully shake out any difference in path separators for running on Windows, but I have not tested that yet.

Updates #78
Updates #323

### DIFF
--- a/wire/internal/wire/parse.go
+++ b/wire/internal/wire/parse.go
@@ -192,8 +192,10 @@ type Value struct {
 }
 
 // Load finds all the provider sets in the packages that match the given
-// patterns, as well as the provider sets' transitive dependencies. It may
-// return both errors and Info.
+// patterns, as well as the provider sets' transitive dependencies. It
+// may return both errors and Info. The patterns are defined by the
+// underlying build system. For the go tool, this is described at
+// https://golang.org/cmd/go/#hdr-Package_lists_and_patterns
 //
 // wd is the working directory and env is the set of environment
 // variables to use when loading the packages specified by patterns. If
@@ -283,7 +285,10 @@ func Load(ctx context.Context, wd string, env []string, patterns []string) (*Inf
 }
 
 // load typechecks the packages that match the given patterns, including
-// function body type checking for the packages that directly match.
+// function body type checking for the packages that directly match. The
+// patterns are defined by the underlying build system. For the go tool,
+// this is described at
+// https://golang.org/cmd/go/#hdr-Package_lists_and_patterns
 //
 // wd is the working directory and env is the set of environment
 // variables to use when loading the packages specified by patterns. If
--- a/wire/internal/wire/parse.go
+++ b/wire/internal/wire/parse.go
@@ -15,6 +15,7 @@
 package wire
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"go/ast"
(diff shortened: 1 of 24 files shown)


wire/internal/wire/parse.go
```diff
@@ -193,8 +194,14 @@ type Value struct {
 // Load finds all the provider sets in the given packages, as well as
```
vangent: Can you clarify the relationship between "given packages" and `patterns`? Same in `load`.

wire/internal/wire/wire.go
```diff
@@ -36,32 +38,71 @@ import (
 	"golang.org/x/tools/go/loader"
 )
 
+// A GeneratedFile stores the content of a call to Generate and the
```
vangent: Nit: drop the "A ".

wire/internal/wire/wire_test.go
```diff
@@ -332,31 +325,91 @@ func TestDisambiguate(t *testing.T) {
...
-				panic("BREAK1")
-			}
 			return false
 		}
 		i += sz
 	}
 	return true
 }
 
+func scrubError(gopath string, s string) string {
```
vangent: It's non-trivial to figure out what's being scrubbed. Can you add a docstring explaining, maybe with an example? Is this not possible to do with a regex?

wire/internal/wire/parse.go
zombiezen: Done.

wire/internal/wire/wire.go
zombiezen: Done.

wire/internal/wire/wire_test.go
zombiezen: It is difficult to do with regex, as the initial match must be a literal string. Also, I will cite [Rob Pike's advice](https://commandcenter.blogspot.com/2011/08/regular-expressions-in-lexing-and.html) on the subject of regexes in parsing.

I've added a doc comment.

wire/internal/wire/parse.go
```diff
@@ -191,9 +191,9 @@ type Value struct {
 	info *types.Info
 }
 
-// Load finds all the provider sets in the given packages, as well as
-// the provider sets' transitive dependencies. It may return both errors
-// and Info.
+// Load finds all the provider sets in the packages that match the given
```
vangent: Can you define "match"? Is pattern a regexp, or is it a partial match, or a full string match, or ...?

zombiezen: There is currently no agreed upon definition of this pattern: it is (will be) at the whims of what [`go/packages`](https://godoc.org/golang.org/x/tools/go/packages#pkg-overview) documents. How do you wish to proceed?

vangent: I would suggest just saying that. E.g., "... that match the given patterns according to the match rules defined by https://godoc.org/golang.org/x/tools/go/packages#pkg-overview..."

zombiezen: Done.

Issue #608, opened 2018-11-02
mysql: consider making an Options struct for all constructors


vangent:
For example, for `mysql/cloudmysql`, the varargs `TraceOptions` could go in an `Options` struct, and leave us more flexibility for adding options later in a non-breaking way.

Something like:

```
type Options struct {
  traceOpts []ocsql.TraceOption  
}

func (o *options) TraceOptions(traceOpts ...ocsql.TraceOption) {
  o.traceOpts = traceOpts
}
```

Similarly for `rdsmysql`.

vangent:
This is fixed now, right?

ijt:
As far as I can tell, yes. Funny that github didn't close it.

zombiezen:
Seems odd to me to have both a `Params` and an `Options` struct.

https://github.com/google/go-cloud/blob/74f68b6bde9e30b198417098b7c989767a5427d7/mysql/cloudmysql/cloudmysql.go#L48-L67

Can we consolidate these into the `Params` struct? If so, happy to open a new issue to track.

vangent:
@ijt argued that `Params` is all about the connection, and is different than `Options`. In particular `Params` is always required (with some non-default values) while `Options` is usually fine to leave nil or the zero value.

Maybe `Params` could be renamed? `Connection` or `ConnectParams`?

zombiezen:
`Params` should always be specified, but not all fields have to be filled in (see `Password` above). I believe it to satisfy the overarching need of having a struct that can be amended with optional fields later.

I also can't come up with a good enough model in my head of what should be in `Params` or `Options`. At the very least, the distinction should be made clear through comments. There are not any right now.

(I was considering opening a separate issue, but I realized while writing it that the issue would be echoing so much of the original issue that it is the same issue. Reopening.)

ijt:
I think this is done now. Closing it, but feel free to reopen if something is missing.

zombiezen:
There are still options. https://github.com/google/go-cloud/blob/f4c33a8eb84a6b8164d6be83bcde0bfe79a49a0b/mysql/rdsmysql/rdsmysql.go#L54-L62

zombiezen:
I'll take it since it's blocking me at the moment.

PR #618, opened 2018-11-05
mysql/cloudmysql: allow configuration of ocsql instrumentation

### LINKED ISSUE #416: sql/ocsql: allow configuration of ocsql instrumentation

ijt:
Fixes #416


ijt:
Fixes #608

ijt:
Tests are passing. PTAL.

PR #630, opened 2018-11-06
travis: always run Go tests for all modules, and wire check

### LINKED ISSUE #614: internal: CI does not cover internal/contributebot
E.g., see #613.

It would also be nice to ensure that running `go generate` for our Wire injectors actually produces the checked-in `wire_gen.go`.

vangent:
- Stop being clever about trying to avoid running various tests. It can result in breakages that aren't caught (e.g, #621).
- Make sure to run tests for all modules; running `go test ./...` at the top level skips sub-modules.
- Drop the `-short` argument, it doesn't do anything for our tests.
- Use `set -euxo pipefail` to simplify bash script and ensure failures are returned.

Fixes #614.


vangent:
> the heuristic that the check script uses for determining whether to skip tests is not conservative enough

Yes, this is true. I just don't think it's worth adding the complexity of getting a heuristic that works, to save a couple of minutes on a probably small percentage of test runs. For me at least, it doesn't really matter much if it's 4 minutes or 8 minutes; either way, I'm not sitting there waiting for it.

zombiezen:
The case I'm thinking of is something like #647, which is a doc-only change. While you claim that you don't need to sit in front of it, there is an element of that, since the process for submission is:

```
while branch is not up-to-date with master {
  press update branch button
  wait for Travis run to finish
}
click squash-and-merge
```

This loop is especially expensive, as the longer you wait, the more likely the branch is not up-to-date and thus causing the cycle to increase iterations.

An alternative would be to have Contribute Bot be in charge of merging in changes, only reporting back if the build failed.

vangent:
That is true, but it is also true for every other change we make, isn't it? I'm just not motivated to optimize for a fairly rare case; writing the check for "all files in this PR have this whitelisted set of extensions" is non-trivial (the code in there now is a blacklist so it would have to change).

Also, we can always add this optimization later.

I am uncomfortably excited by the idea of ContributeBot automatically doing merges.

internal/testing/runchecks.sh
```diff
@@ -18,16 +18,26 @@
...
+
+# Run Go tests for each module.
+for path in "." "./internal/contributebot" "./samples/appengine"; do
+  pushd ${path}
+  go test -race ./...
+  popd
+done
+
+# Wire checks.
+go mod vendor
```
zombiezen: Now that #78 is fixed, you can entirely skip the `go mod vendor` nonsense, and run `wire check ./...` right after `go test -race`. 🎉

PR #640, opened 2018-11-06
internal/proxy: update instructions for how to update the proxy bucket

vangent:
Previously we tried to iterate through modules, and muck with version files. This is fragile, especially in cases where we need dependencies from multiple branches.

The new approach is based on the one described in https://github.com/go-modules-by-example/index/tree/master/012_modvendor (suggested by bcmills@google.com). It creates a temporary directory for `$GOPROXY`, and runs `go mod download` for each of the modules in our repo. Repeat for each branch. Now, the module download cache is exactly what we need; we can synchronize it to the GCS bucket.


internal/proxy/README.md
```diff
@@ -17,16 +17,42 @@ dependency to the proxy:
...
+# This approach was suggested by:
+# https://github.com/go-modules-by-example/index/tree/master/012_modvendor
+
+# Create a temporary directory where we'll create a module download cache.
+tgp=$(mktemp -d)
+
+# Run this command in the master branch.
+# It runs "go mod download" in every module that we have in our repo,
+# filling the cache with all of the module dependencies we need.
+GOPROXY=file://$GOPATH/pkg/mod/cache/download GOPATH=$tgp ./internal/proxy/makeproxy.sh
```
zombiezen: Seems like this environment variable juggling could happen in the script. I would want this instruction to read:

```
./internal/proxy/makeproxy.sh "$tgp"
```

internal/proxy/README.md
```diff
@@ -17,16 +17,42 @@ dependency to the proxy:
...
+# Synchronize modvendor to the proxy.
 
 # -n: preview only
 # -r: recurse directories
 # -c: compare checksums not write times
-# -d: source directory; value should match what you passed to makeproxy.sh
-$ gsutil rsync -n -r -c -d proxy gs://go-cloud-modules
-...
-# If it looks good, repeat without the -n.
+gsutil rsync -n -r -c modvendor gs://go-cloud-modules
```
zombiezen: With this new procedure in place, it seems like either `-d` should stay in, or any `.../list` files should be appended to. (I think the latter is a better approach, but may require some scripting.)

internal/proxy/README.md
vangent: That would be nice, but the fact that you have to run "go mod download" from multiple branches makes it difficult... (see the next comment).

internal/proxy/README.md
zombiezen: SGTM.

internal/proxy/README.md
zombiezen: I don't see the relation here: what about multiple branches is incompatible with setting the environment variables within the script instead of outside it?

vangent: Oh, I see what you're saying; done.

internal/proxy/makeproxy.sh
```diff
@@ -16,40 +16,19 @@
...
 
 if [[ $# -ne 1 ]]; then
-  echo "usage: makeproxy.sh OUTDIR" 1>&2
+  echo "usage: makeproxy.sh <dir>" 1>&2
   exit 64
 fi
 
-export GO111MODULE=on
-GO="${GO:-go}"  # allows user customization of which Go command to use.
+GOPATH="$1"
```
zombiezen: Also set `GOPROXY`?

vangent: Leaving that out was intentional. The previous code set `GOPROXY` to the original file cache, which probably makes things faster, but also more fragile (e.g., if we now need to download some module that we haven't downloaded before). I don't care how fast this is so I'd rather do the simpler thing and let it download them.

zombiezen:
One thing that just occurred to me but I don't know how to address: if multiple people on the team do this without coordination, they would break each others' PRs. For instance, maybe all I know is I need to add dep A, and someone else wants to add dep B. Each of us following these instructions will cancel each other out.

I think the harder approach we identified earlier in this review: make this process effectively append only, and force someone to go through manually to remove things from the proxy.

vangent:
> if multiple people on the team do this without coordination, they would break each others' PRs

Yes. I think this operation is rare enough that this shouldn't happen. It will be fairly obvious when it does. If it does start to happen frequently we can figure it out then.

zombiezen:
Opened #661 to keep track of the idea.

PR #642, opened 2018-11-07
pubsub: add concrete API implementation

### LINKED ISSUE #607: pubsub: implement concrete API (pubsub.go) and driver interfaces
This is a sub-issue of #312.

ijt:
Fixes #607


pubsub/pubsub.go
```diff
@@ -0,0 +1,211 @@
...
+// the message has been confirmed as acknowledged on the server, or failure
+// occurs.
+func (m *Message) Ack(ctx context.Context) error {
+	// Send the ack ID back to the subscriber for batching.
+	m.sub.ackChan <- m.ackID
+	select {
+	case err := <-m.sub.ackErrChan:
+		return err
+	case <-ctx.Done():
+		return nil
```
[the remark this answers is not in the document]
ijt: Done in the implementation branch.

pubsub/pubsub.go
```diff
@@ -0,0 +1,211 @@
...
+	ctx context.Context
+}
+
+// Send publishes a message. It only returns after the message has been
+// sent, or failed to be sent. Send can be called from multiple goroutines
+// at once.
+func (t *Topic) Send(ctx context.Context, m *Message) error {
+	t.mcChan <- msgCtx{m, ctx}
+	// Wait for the batch including this message to be sent to the server.
+	return <-m.errChan
```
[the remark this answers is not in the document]
ijt: Done in the implementation branch.

pubsub/pubsub.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	go func() {
+		// Pull messages from t.mcChan and put them in batches. Send the current
+		// batch whenever it is large enough or enough time has elapsed since
+		// the last send.
+		for {
+			batch := make([]*driver.Message, 0, batchSize)
+			timeout := time.After(opts.SendDelay)
+		Loop:
+			for i := 0; i < opts.BatchSize; i++ {
```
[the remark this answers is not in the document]
ijt: Good point. I changed this in the implementation branch to loop until the batch size hits the max or the time runs out.

pubsub/pubsub.go
```diff
@@ -0,0 +1,211 @@
...
+							AckID:      m.msg.AckID,
+						}
+						batch = append(batch, dm)
+					}
+				case <-t.doneChan:
+					return
+				}
+			}
+			if len(batch) > 0 {
+				err := t.driver.SendBatch(ctx, batch)
```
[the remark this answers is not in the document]
ijt: It will have to come from a parameter of NewTopic. It can't be the ctx of any particular message because then the cancellation of that one would cancel sending the whole batch.

pubsub/pubsub.go
```diff
@@ -0,0 +1,211 @@
...
+	// of acknowledgements back to the server.
+	AckDelay time.Duration
+
+	// AckBatchSize is the maximum number of acks that should be sent to
+	// the server in a batch.
+	AckBatchSize int
+
+	// AckDeadline tells how long the server should wait before assuming a
+	// received message has failed to be processed.
+	AckDeadline time.Duration
```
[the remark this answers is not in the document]
ijt: Removed it.

pubsub/pubsub.go
```diff
@@ -0,0 +1,211 @@
...
+			if err != nil {
+				return nil, err
+			}
+			select {
+			case <-ctx.Done():
+				return nil, ctx.Err()
+			default:
+				if len(msgs) > 0 {
+					s.q = make([]*Message, len(msgs))
+					for _, m := range msgs {
```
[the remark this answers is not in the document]
ijt: Done in the implementation branch.

ijt:
I think this PR was approved by accident because of a mix-up.

I've added several tests and changed the implementation to use the bundler package. Thanks to @jba for the suggestion to use that; things are working much better now.

PTAL.

pubsub/pubsub.go
```diff
@@ -0,0 +1,243 @@
...
+	t.batcher.Flush()
+	return t.driver.Close()
+}
+
+// NewTopic makes a pubsub.Topic from a driver.Topic and opts to
+// tune how messages are sent. Behind the scenes, NewTopic spins up a goroutine
+// to bundle messages into batches and send them to the server.
+func NewTopic(ctx context.Context, d driver.Topic, opts TopicOptions) *Topic {
+	handler := func(item interface{}) {
+		mecs, ok := item.([]msgErrChan)
```
jba: Remove the `ok`. It will panic anyway.

pubsub/pubsub.go
```diff
@@ -0,0 +1,243 @@
...
+		if !ok {
+			panic("failed conversion to []msgErrChan in bundler handler")
+		}
+		dms := make([]*driver.Message, 0, len(mecs))
```
jba: I prefer
```
var dms []*driver.Message
```
What you have is a premature optimization.

pubsub/pubsub.go
```diff
@@ -0,0 +1,243 @@
...
+			dms = append(dms, dm)
+		}
+		err := d.SendBatch(ctx, dms)
+		for _, mec := range mecs {
+			mec.errChan <- err
+		}
+	}
+	b := bundler.NewBundler(msgErrChan{}, handler)
+	b.DelayThreshold = time.Millisecond
+	t := &Topic{
```
jba: `b.BundleCountThreshold = opts.BatchSize`

pubsub/pubsub.go
```diff
@@ -0,0 +1,243 @@
...
+			}
+			dms = append(dms, dm)
+		}
+		err := d.SendBatch(ctx, dms)
+		for _, mec := range mecs {
+			mec.errChan <- err
+		}
+	}
+	b := bundler.NewBundler(msgErrChan{}, handler)
+	b.DelayThreshold = time.Millisecond
```
jba: opts.SendDelay?

pubsub/acks_test.go
```diff
@@ -0,0 +1,113 @@
...
+		sendAcks: f,
+	}
+	sub := pubsub.NewSubscription(ctx, ds, pubsub.SubscriptionOptions{})
+
+	// Receive and ack the messages concurrently.
+	var wg sync.WaitGroup
+	recv := func() {
+		mr, err := sub.Receive(ctx)
+		if err != nil {
+			t.Fatal(err)
```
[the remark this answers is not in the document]
ijt: Done.

pubsub/pubsub.go

pubsub/pubsub.go

pubsub/pubsub.go

ijt:
@jba, I made the changes you suggested and fixed some bugs. PTAL.

pubsub/pubsub.go
```diff
@@ -0,0 +1,243 @@
...
+// the message has been confirmed as acknowledged on the server, or failure
+// occurs.
+func (m *Message) Ack(ctx context.Context) error {
+	// Send the message back to the subscriber for ack batching.
+	mec := msgErrChan{
+		msg:     m,
+		errChan: make(chan error),
+	}
+	// FIXME: Figure out the real size of the ackIDs.
+	size := 8
```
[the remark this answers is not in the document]
jba: Oh, right. I think the size here is really not going to matter. Google's are probably around 20 bytes, and that's fine. Or what you have, 8. Just make a comment that it's an estimate.

pubsub/pubsub_test.go
```diff
@@ -72,3 +121,89 @@ func TestSendReceive(t *testing.T) {
...
+	// Make subscriptions and start goroutines to receive from them.
+	var subs []*pubsub.Subscription
+	nSubs := 100
+	var wg sync.WaitGroup
+	for i := 0; i < nSubs; i++ {
+		ds := NewDriverSub()
+		dt.subs = append(dt.subs, ds)
+		s := pubsub.NewSubscription(ctx, ds, pubsub.SubscriptionOptions{})
+		subs = append(subs, s)
+		wg.Add(howManyToSend)
```
jba: 1

pubsub/pubsub_test.go
```diff
@@ -72,3 +121,89 @@ func TestSendReceive(t *testing.T) {
...
+		go func() {
+			for j := 0; j < howManyToSend; j++ {
```
jba: before this line: defer wg.Done()

pubsub/pubsub_test.go
```diff
@@ -72,3 +121,89 @@ func TestSendReceive(t *testing.T) {
...
+func TestLotsOfMessagesAndSubscriptions(t *testing.T) {
+	howManyToSend := int(1e3)
+	ctx := context.Background()
+	dt := &driverTopic{}
+
+	// Make subscriptions and start goroutines to receive from them.
+	var subs []*pubsub.Subscription
+	nSubs := 100
+	var wg sync.WaitGroup
+	for i := 0; i < nSubs; i++ {
```
jba: Each sub gets every message, so why do you need this loop?

pubsub/pubsub_test.go
```diff
@@ -72,3 +121,89 @@ func TestSendReceive(t *testing.T) {
...
+	}
+}
+
+func TestCancelReceive(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	ds := NewDriverSub()
+	s := pubsub.NewSubscription(ctx, ds, pubsub.SubscriptionOptions{})
+	go s.Receive(ctx)
+	cancel()
+	// Without cancellation, this Receive would hang.
```
jba: I think you're assuming that the goroutine above starts before the cancel() is executed. Actually, I'm not sure why that goroutine is there.

You could just do 
```
cancel()
err := s.Receive(ctx)
```
For a more complicated test where the Receive is in the middle of working when you cancel it, let's wait for the fake. We'll write it to make that kind of test not time-dependent.

pubsub/acks_test.go
```diff
@@ -56,58 +72,158 @@ func TestAckTriggersDriverSendAcksForOneMessage(t *testing.T) {
...
+		t.Errorf("got error %q, want %q", err.Error(), e)
+	}
+	if err := sub.Close(); err != nil {
+		t.Fatal(err)
+	}
+}
 
+func TestCancelAck(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	f := func(ctx context.Context, ackIDs []driver.AckID) error {
```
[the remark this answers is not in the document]
ijt: Done.

pubsub/acks_test.go
```diff
@@ -56,58 +72,158 @@ func TestAckTriggersDriverSendAcksForOneMessage(t *testing.T) {
...
+		<-ctx.Done()
+		return ctx.Err()
+	}
+	m := &driver.Message{}
+	ds := &ackingDriverSub{
+		q:        []*driver.Message{m},
+		sendAcks: f,
+	}
+	sub := pubsub.NewSubscription(ctx, ds, pubsub.SubscriptionOptions{})
```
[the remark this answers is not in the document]
ijt: Done.

pubsub/acks_test.go
```diff
@@ -56,58 +72,158 @@ func TestAckTriggersDriverSendAcksForOneMessage(t *testing.T) {
...
+	mr, err := sub.Receive(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	cancel()
+	if err := mr.Ack(ctx); err == nil {
+		t.Error("got nil, want cancellation error")
```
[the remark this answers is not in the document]
ijt: Done.

pubsub/empty_sub_test.go
```diff
@@ -33,4 +46,7 @@ func TestReceiveErrorIfEmptyBatchReturnedFromDriver(t *testing.T) {
 	if err == nil {
 		t.Error("error expected for Receive with buggy driver")
 	}
+	if err := sub.Close(); err != nil {
```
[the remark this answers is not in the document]
ijt: This seems okay the way it is. We can tidy it up when `check` becomes available.

pubsub/pubsub_test.go
ijt: Substituting 1 gives this result:

```
~/src/github.com/google/go-cloud/pubsub$ go test
panic: sync: negative WaitGroup counter

goroutine 32 [running]:
sync.(*WaitGroup).Add(0xc0000c0070, 0xffffffffffffffff)
	/usr/lib/google-golang/src/sync/waitgroup.go:74 +0x137
```

pubsub/acks_test.go
```diff
@@ -0,0 +1,229 @@
...
+	if !reflect.DeepEqual(sentAckBatches, want) {
+		t.Errorf("got %v, want %v", sentAckBatches, want)
+	}
+
+	if err := sub.Close(); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestMsgAckReturnsErrorFromSendAcks(t *testing.T) {
```
vangent: Note: it is interesting to have these tests of the concrete type, but I suspect that at least some of them will end up being conformance tests as well, which may test both concrete type + driver together. At that point they could be removed from here.

pubsub/acks_test.go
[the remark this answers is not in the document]
vangent: +1, same as `blob`.

pubsub/driver/driver.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2018 The Go Cloud Authors
```
vangent: This file should not be in this PR.

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+	"google.golang.org/api/support/bundler"
+)
+
+// Message contains data to be published.
+type Message struct {
+	// Body contains the content of the message.
+	Body []byte
+
+	// Attributes has key/value metadata for the message.
+	Attributes map[string]string
```
vangent: Metadata

pubsub/pubsub_test.go
ijt: Okay, I see this works with your comment below. Good point. Changed it.

pubsub/pubsub_test.go
ijt: Done.

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+
+	// AckID identifies the message on the server.
+	// It can be used to ack the message after it has been received.
+	ackID AckID
+
+	// sub is the Subscription this message was received from.
+	sub *Subscription
+}
+
+// AckID identifies a message for acknowledgement.
```
vangent: The concrete type doesn't need its own `AckID`, does it? I don't think it's exposed anywhere. You can just use `driver.AckID` for the private field.

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+// Close flushes pending ack sends and disconnects the Subscription.
+func (s *Subscription) Close() error {
+	s.ackBatcher.Flush()
+	return s.driver.Close()
+}
+
+// NewSubscription creates a Subscription from a driver.Subscription and opts to
+// tune sending and receiving of acks and messages. Behind the scenes,
+// NewSubscription spins up a goroutine to gather acks into batches and
+// periodically send them to the server.
```
vangent: Similar to `blob`, add a comment like `// It is for use by provider implementations.`.

(this is public in the package so that driver impls can call it, but will not be called by end users).

pubsub/pubsub.go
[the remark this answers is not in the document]
vangent: Consider making this a constant, and definitely add a comment explaining.

pubsub/pubsub_test.go
ijt: Done.

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+	// messages to the server.
+	SendDelay time.Duration
+
+	// BatchSize specifies the maximum number of messages that can go in a batch
+	// for sending.
+	BatchSize int
+}
+
+// SendDelayDefault is the value for TopicOptions.SendDelay if it is not set.
+const SendDelayDefault = time.Millisecond
```
vangent: Why Millisecond? Seems quite short; if you want to disable batching by default then it should be 0, no?

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+	sub *Subscription
+}
+
+// AckID identifies a message for acknowledgement.
+type AckID interface{}
+
+// Ack acknowledges the message, telling the server that it does not need to
+// be sent again to the associated Subscription. This method blocks until
+// the message has been confirmed as acknowledged on the server, or failure
+// occurs.
```
vangent: Add the same comment as you have on `Send` about it being able to be called from multiple goroutines at once.

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+			}
+			dms = append(dms, dm)
+		}
+		err := d.SendBatch(ctx, dms)
+		for _, mec := range mecs {
+			mec.errChan <- err
+		}
+	}
+	b := bundler.NewBundler(msgErrChan{}, handler)
+	if opts.SendDelay == 0 {
```
vangent: What if I actually want to set the SendDelay to 0?

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+// NewTopic makes a pubsub.Topic from a driver.Topic and opts to
+// tune how messages are sent. Behind the scenes, NewTopic spins up a goroutine
+// to bundle messages into batches and send them to the server.
+func NewTopic(ctx context.Context, d driver.Topic, opts TopicOptions) *Topic {
+	handler := func(item interface{}) {
+		mecs, ok := item.([]msgErrChan)
+		if !ok {
+			panic("failed conversion to []msgErrChan in bundler handler")
+		}
+		var dms []*driver.Message
```
vangent: You can initialize dms to the right size, e.g.:

```
dms := make([]*driver.Message, len(mecs))
```
and then

```
for i, mec := range mecs {
  dms[i] = ...
```

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+	// the server in a batch.
+	AckBatchSize int
+
+	// AckDeadline tells how long the server should wait before assuming a
+	// received message has failed to be processed.
+	AckDeadline time.Duration
+}
+
+// AckDelayDefault is the value for SubscriptionOptions.AckDelay if it is not set.
+const AckDelayDefault = time.Millisecond
```
vangent: Same comment here as above for SendDelay.

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+func (s *Subscription) getNextBatch(ctx context.Context) error {
+	msgs, err := s.driver.ReceiveBatch(ctx)
+	if err != nil {
+		return err
+	}
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+		if len(msgs) == 0 {
```
vangent: Maybe this should be allowed? E.g., some impls may do polling?
If not, make sure to document the requirement in the driver package.

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+	return m, nil
+}
+
+// getNextBatch gets the next batch of messages from the server.
+func (s *Subscription) getNextBatch(ctx context.Context) error {
+	msgs, err := s.driver.ReceiveBatch(ctx)
+	if err != nil {
+		return err
+	}
+	select {
```
vangent: Why check ctx again here?

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+		if err := s.getNextBatch(ctx); err != nil {
+			return nil, err
+		}
+	}
+	m := s.q[0]
+	s.q = s.q[1:]
+	return m, nil
+}
+
+// getNextBatch gets the next batch of messages from the server.
```
vangent: ... and saves it in s.q.

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+
+// NewSubscription creates a Subscription from a driver.Subscription and opts to
+// tune sending and receiving of acks and messages. Behind the scenes,
+// NewSubscription spins up a goroutine to gather acks into batches and
+// periodically send them to the server.
+func NewSubscription(ctx context.Context, d driver.Subscription, opts SubscriptionOptions) *Subscription {
+	handler := func(item interface{}) {
+		mecs := item.([]msgErrChan)
+		var ids []driver.AckID
+		for _, mec := range mecs {
```
vangent: same comment as above

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+			id := m.ackID
+			ids = append(ids, id)
+		}
+		err := d.SendAcks(ctx, ids)
+		for _, mec := range mecs {
+			mec.errChan <- err
+		}
+	}
+	ab := bundler.NewBundler(msgErrChan{}, handler)
+	if opts.AckDelay == 0 {
```
vangent: same comment as above about wanting to actually set to 0.

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+// Subscription receives published messages.
+type Subscription struct {
+	driver driver.Subscription
+
+	// ackBatcher makes batches of acks and sends them to the server.
+	ackBatcher *bundler.Bundler
+
+	// q is the local queue of messages downloaded from the server.
```
vangent: Move comment for q down to q, and add one for sem.

pubsub/acks_test.go
```diff
@@ -48,17 +48,16 @@ func (s *ackingDriverSub) Close() error {
...
-		sentAcks = ackIDs
-		return nil
-	}
 	id := rand.Int()
 	m := &driver.Message{AckID: id}
 	ds := &ackingDriverSub{
-		q:        []*driver.Message{m},
-		sendAcks: f,
+		q: []*driver.Message{m},
+		sendAcks: func(ctx context.Context, ackIDs []driver.AckID) error {
```
jba: nit: s/ctx/_/ (and below)

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+			mec.errChan <- err
+		}
+	}
+	ab := bundler.NewBundler(msgErrChan{}, handler)
+	if opts.AckDelay == 0 {
+		ab.DelayThreshold = AckDelayDefault
+	} else {
+		ab.DelayThreshold = opts.AckDelay
+	}
+	if opts.AckBatchSize != 0 {
```
vangent: What size is this referring to? The number of acks, or the (estimated) byte size of all of the acks? The comment on the Options structs seems to imply the former, but the fact that you've got code computing the size of each (ack, message) implies the latter.

pubsub/acks_test.go
```diff
@@ -0,0 +1,229 @@
...
+	wg.Add(2)
+	go recv()
+	go recv()
+	wg.Wait()
+
+	want := [][]driver.AckID{
+		{ids[0]},
+		{ids[1]},
+	}
+	if !reflect.DeepEqual(sentAckBatches, want) {
```
vangent: Use cmp -- see `blob` conformance tests for examples.

pubsub/acks_test.go
ijt: I think these are worthwhile tests even if there are conformance tests that are similar, but if they turn out to be redundant or burdensome then by all means we should delete them.

pubsub/acks_test.go
ijt: Done.

pubsub/driver/driver.go
ijt: Done.

pubsub/pubsub.go

pubsub/pubsub_test.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	return nil
+}
+
+func (t *driverTopic) Close() error {
+	return nil
+}
+
+type driverSub struct {
+	sem chan struct{}
```
zombiezen: I think you could replace these two with a buffered channel.

pubsub/pubsub_test.go
```diff
@@ -0,0 +1,201 @@
...
+		case <-s.sem:
+			ms := s.grabQueue()
+			if len(ms) != 0 {
+				return ms, nil
+			}
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+		}
+		time.Sleep(time.Microsecond)
```
zombiezen: Don't use sleep in tests. Receive on a channel instead.

pubsub/pubsub.go

pubsub/pubsub_test.go
```diff
@@ -0,0 +1,201 @@
...
+	dt.subs = append(dt.subs, ds)
+	s := pubsub.NewSubscription(ctx, ds, nil)
+	ng := 10
+	for i := 0; i < 10; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for j := 0; j < howManyToSend/ng; j++ {
+				if _, err := s.Receive(ctx); err != nil {
+					panic(err)
```
zombiezen: This will abort the whole test. `t.Error`?

pubsub/pubsub_test.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	if err := topic.Close(); err != nil {
+		t.Fatal(err)
+	}
+	if err := sub.Close(); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestLotsOfMessagesAndSubscriptions(t *testing.T) {
```
zombiezen: What property is this testing? I see it sending a bunch of things, but not verifying anything.

pubsub/pubsub.go
ijt: The reason is because every `Topic.Send` call blocks until the whole batch is sent. If this is happening within for example an HTTP handler, we don't want the `Send` call to add a bunch of additional latency as the default behavior.

pubsub/pubsub.go
ijt: I did a similar thing earlier with 
```go
dms := make([]*driver.Message, 0, len(mecs))
...
for i, mec := range mecs {
  ...
  dms = append(dms, dm)
}
```
and Jonathan said it was a premature optimization and asked me to put it the way it is now.

I'm making this change, but would like to leave this code alone after this.

pubsub/pubsub.go
ijt: @zombiezen, this has changed in response to @jba's comments. Please have a look.

pubsub/pubsub.go
ijt: Same response. :-)

pubsub/pubsub.go
ijt: @jba was saying it should not be, since different drivers may have different ways to efficiently block until a message is available. We can revisit this later if we find that some drivers need to be able to return an empty message slice.

pubsub/pubsub.go
jba: I guess Robert is preferring consistency across Go Cloud packages, and I am suggesting consistency across providers. I honestly don't know which is better.

pubsub/pubsub.go
jba: If the driver wants to do polling, that's its problem. The concrete impl should never have to poll, though.

pubsub/pubsub.go
zombiezen: I don't have enough context here (apologies, I am failing really hard at following this review through the PR UI), but IIUC @vangent's suggestion is to do something like `ids[i] = mecs[i].msg.ackID`. I don't have a strong opinion here. Most of the way around the inefficiencies are to set the capacity correctly anyway.

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+				ackID:      m.AckID,
+				sub:        s,
+			}
+		}
+	}
+	return nil
+}
+
+// Close flushes pending ack sends and disconnects the Subscription.
+func (s *Subscription) Close() error {
```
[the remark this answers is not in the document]
jba: I also recall coming to that decision, and renaming this to, maybe, Shutdown. For non-graceful shutdown, you could cancel all the contexts passed to Ack.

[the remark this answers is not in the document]
jba: Oops, forget about that last sentence. We ultimately rejected that idea.

pubsub/pubsub_test.go
jba: (Thinking ahead to when I steal this for the fake:) But what is the right buffer size? We should support publishing an arbitrary amount of messages before any are received, as the backend would.

pubsub/pubsub_test.go
jba: I don't even think this is necessary. Couldn't we just remove the default in the select?

vangent:
Something that came up in my 1:1 with Issac: I propose that instead of baking the `Bundler` into the concrete type, we should capture what we need from a bundler in an interface, and the concrete type should accept one as part of its construction.

Each provider implementation could then decide which bundler(s) it wants to support, and how to expose any bundler options to the user.

Feel free to add a TODO/file an issue to do this or discuss it further.

pubsub/pubsub.go
jba: I can't find anything in the readability guidelines about this, so I'll waive my objection if Robert insists. I'll just reiterate the flaws:
- More code on the page, so harder to read.
- Potential bug: writing a different slice in `make([]T, len(s))` and `for i, x := range s` (which also means extra effort on the reader's part to check for this).
- Overallocation (and complex handling of the loop index) when only some of the results are used. E.g. compare:
```
a := make([]T, len(b))
for i, x := range b {
   a[i] = x
}
```
with
```
a := make([]T, len(b))
for i := 0; i < len(b); {
   x := b[i]
   if filter(x) {
      a[i] = x
      i++
  }
}
```
Whereas if append is used, the loops look almost identical.

That's why I call this a premature optimization: it's faster, sure, but it makes the code worse overall, so it should only be done if performance is shown to matter.

pubsub/pubsub.go
vangent: I don't feel too strongly about this; I saw that it was renamed `Metadata` in the PR with just the driver and thought that this needed to be consistent.

For blob, many providers have blobs with lots of attributes (Size, ContentType, etc.), and one of the attributes is a set of arbitrary key/value pairs, which they called Metadata.

If pubsub is different, and Messages don't have any attributes other than the key/value pairs, then I'm fine with just calling the key/value pairs attributes.

pubsub/pubsub.go
jba: I found some guidance from Ian Lance Taylor, which I'll quote:

> I'm OK with [the loop that uses `a[i] = x`] when it really is that
simple.  As soon as there are several statements, or a conditional, in
the loop body, I think that using append is preferable stylistically,
as it will clearly continue to work if the code changes such that
there are cases where the value is not added to the slice.

So I'm fine with using the index in this simple case, but we should be wary in more complex loops.

ijt: Okay, I changed all of these loops to use `append`.

pubsub/pubsub.go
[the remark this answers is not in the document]
ijt: Okay. I don't mind revisiting this later if we find that some change is needed.

pubsub/acks_test.go
```diff
@@ -0,0 +1,222 @@
...
+			return ctx.Err()
+		},
+	}
+	sub := pubsub.NewSubscription(ctx, ds, nil)
+	mr, err := sub.Receive(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	cancel()
+	if err := mr.Ack(ctx); err != context.Canceled {
```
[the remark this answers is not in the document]
ijt: Done.

pubsub/pubsub.go
```diff
@@ -0,0 +1,272 @@
...
+
+	// BatchSize specifies the maximum number of messages that can go in a batch
+	// for sending.
+	BatchSize int
+}
+
+// SendDelayDefault is the value for TopicOptions.SendDelay if it is not set.
+const SendDelayDefault = time.Millisecond
+
+var DefaultTopicOptions = TopicOptions{
```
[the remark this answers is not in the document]
ijt: That's what I had before but @jba wanted to have nil denote the default.
Anyway, it will be moot if we do https://github.com/google/go-cloud/issues/665.

PR #643, opened 2018-11-07
pubsub: add driver interfaces

ijt:
This makes progress toward fixing #607 .


pubsub/driver/driver.go
```diff
@@ -0,0 +1,46 @@
...
+type AckID interface{}
+
+type Message struct {
+	// Body contains the content of the message.
+	Body []byte
+
+	// Attributes has key/value metadata for the message.
+	Attributes map[string]string
+
+	// AckID identifies the message on the server. It can be used to ack
```
vangent: Since this is the driver interface, I suggest focusing the documentation on someone who is going to implement it. There is only one user (the concrete type) so (unusually) the user is less important.

So for example, on this one, something like:

// AckID should be set to something identifying the message on the server. It may be passed to SendAcks() to ack the message.

pubsub/driver/driver.go
```diff
@@ -0,0 +1,46 @@
...
+type Subscription interface {
+	// ReceiveBatch returns a batch of messages that have queued up for the
+	// subscription on the server.
+	ReceiveBatch(ctx context.Context) ([]*Message, error)
+
+	// SendAcks acknowledges the messages with the given ackIDs on the
+	// server so that they will not be received again for this
+	// subscription. This method returns only after all the ackIDs are
+	// sent.
+	SendAcks(ctx context.Context, ackIDs []interface{}) error
```
vangent: `[]AckID` ?

pubsub/driver/driver.go
```diff
@@ -0,0 +1,46 @@
...
+	AckID AckID
+
+	// ErrChan relays back an error or nil as the result of sending the
+	// batch that includes this message.
+	ErrChan chan error
+}
+
+// Topic publishes messages.
+type Topic interface {
+	// SendBatch publishes all the messages in ms.
```
vangent: Does it also write to ErrChan for each of the messages?

pubsub/driver/driver.go
```diff
@@ -0,0 +1,46 @@
+package driver
```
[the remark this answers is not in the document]
ijt: Done.

pubsub/driver/driver.go
```diff
@@ -0,0 +1,46 @@
...
+import "context"
+
+type AckID interface{}
+
+type Message struct {
+	// Body contains the content of the message.
+	Body []byte
+
+	// Attributes has key/value metadata for the message.
+	Attributes map[string]string
```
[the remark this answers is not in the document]
jba: Let's go with the more common industry term, across both cloud providers and standard products/APIs like ActiveMQ, AMPQ, Kafka, etc.

(I don't know what that term is, just saying.)

pubsub/driver/driver.go
```diff
@@ -0,0 +1,46 @@
+package driver
+
+import "context"
+
+type AckID interface{}
```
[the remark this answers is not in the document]
ijt: Done.

pubsub/driver/driver.go
ijt: It doesn't. I'm pretty sure it was a mistake having the ErrChan field on driver.Message. Most likely the errChans can be fields of little structs like this that get sent to the message batcher:

```go
type msgErrChan struct {
    msg *Message
    errChan chan error
}
```
That should keep the Message structs cleaner.

pubsub/driver/driver.go
```diff
@@ -0,0 +1,46 @@
...
+	SendBatch(ctx context.Context, ms []*Message) error
+
+	// Close disconnects the Topic.
+	Close() error
+}
+
+// Subscription receives published messages.
+type Subscription interface {
+	// ReceiveBatch returns a batch of messages that have queued up for the
+	// subscription on the server.
```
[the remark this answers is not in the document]
ijt: Done.

pubsub/driver/driver.go
```diff
@@ -0,0 +1,46 @@
...
+	// Attributes has key/value metadata for the message.
+	Attributes map[string]string
+
+	// AckID identifies the message on the server. It can be used to ack
+	// the message after it has been received.
+	AckID AckID
+
+	// ErrChan relays back an error or nil as the result of sending the
+	// batch that includes this message.
+	ErrChan chan error
```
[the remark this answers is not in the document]
ijt: I removed it. We'll use little structs to associate err chans with messages.

pubsub/driver/driver.go
```diff
@@ -0,0 +1,46 @@
...
+// Subscription receives published messages.
+type Subscription interface {
+	// ReceiveBatch returns a batch of messages that have queued up for the
+	// subscription on the server.
+	ReceiveBatch(ctx context.Context) ([]*Message, error)
+
+	// SendAcks acknowledges the messages with the given ackIDs on the
+	// server so that they will not be received again for this
+	// subscription. This method returns only after all the ackIDs are
+	// sent.
```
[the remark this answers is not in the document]
ijt: Say it about which things? Maybe just paste in what you'd like to see.

ijt:
Changes are up. PTAL.

pubsub/driver/driver.go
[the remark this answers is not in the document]
vangent: I think @jba is suggesting that the same ("This methos returns only after...") is true for `SendBatch`, so copy that sentence there. You already have it for `ReceiveBatch` too.

pubsub/driver/driver.go
```diff
@@ -0,0 +1,63 @@
...
+	// AckID should be set to something identifying the message on the
+	// server. It may be passed to Subscription.SendAcks() to acknowledge
+	// the message.
+	AckID AckID
+}
+
+// Topic publishes messages.
+type Topic interface {
+	// SendBatch publishes all the messages in ms.
+	SendBatch(ctx context.Context, ms []*Message) error
```
vangent: For `SendBatch` and `SendAcks`, it is probably worth saying something about what the required error semantics are. For example, if a non-nil error is returned, is the driver implementation required to guaranteed that none of the message were sent? Or is it OK to have sent some message but not others? What if a driver can't guarantee that? (i.e., the underlying implementation sends one message at a time, and each can fail individually).

pubsub/driver/driver.go
```diff
@@ -0,0 +1,63 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package driver defines a set of interfaces that the pubsub package uses to
+// interact with the underlying pubsub services.
+package driver
+
+import "context"
+
```
vangent: OK to do in a separate PR if you prefer, but you should add `As` functions for `Topic`, `Subscription`, and `Message`. See `blob.Driver` for examples.

`Message` is used for both reading and writing, which makes it interesting for `As`. An `As` function ```As(i interface{}) bool``` makes sense for reading, but for writing you want to give access to the underlying type of whatever message is going to be sent to the server. In `blob`, we do this via `WriterOptions.BeforeWrite`. So you could:

1. Add a `BeforeWrite`-like function to `Message`, that's only used when writing
2. Split `Message` into separate types for reading (with `As`) and writing (with `BeforeWrite`)
3. Add an `Options` struct with `BeforeWrite` to `SendBatch`.

pubsub/driver/driver.go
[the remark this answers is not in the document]
jba: Yes, I meant that sentence. I guess GH codereview doesn't do highlights :(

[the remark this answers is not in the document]
ijt: Done.

pubsub/driver/driver.go
ijt: Added issue #657 .

pubsub/driver/driver.go
ijt: We can have a convention that this sort of driver should be paired with a bundler with a batch size of 1. The `OpenTopic` and `OpenSubscription` funcs for the implementation can take care of setting that up.

ijt: Done.

ijt:
Changes are up. PTAL.

pubsub/driver/driver.go
```diff
@@ -0,0 +1,78 @@
...
+	// server. It may be passed to Subscription.SendAcks() to acknowledge
+	// the message.
+	AckID AckID
+}
+
+// Topic publishes messages.
+type Topic interface {
+	// SendBatch publishes all the messages in ms. This method should
+	// return only after all the messages are sent, an error occurs, or the
+	// context is cancelled.
```
jba: s/cancelled/done/

PR #644, opened 2018-11-07
blob/fileblob: ignore URL's Host and drop leading / on Windows

vangent:
This appears to be more correct than including the Host as part of the string we pass `OpenBucket`, based on my reading of https://en.wikipedia.org/wiki/File_URI_scheme. Dealing with file-based URLs in Windows is pretty hokey, but it's clear you have to drop the leading "/".

`blob/example_test.go` passes on Windows after these changes.


blob/example_test.go
```diff
@@ -335,7 +337,15 @@ func ExampleOpen() {
 	}
 	fmt.Println("Got expected error opening a nonexistent path")
 
-	if _, err := blob.Open(ctx, "file://"+dir); err != nil {
+	// Ensure the path has a leading slash; fileblob ignores the URL's
+	// Host field, so URLs should always start with "file:///". On
+	// Windows, the leading "/" will be stripped, so "file:///c:/foo"
+	// will refer to c:/foo.
+	urlpath := url.PathEscape(filepath.ToSlash(dir))
```
shantuo: Should we handle converting separators to '/' in package fileblob instead of letting user to it? Is "c:\foo\bar" invalid?

vangent: No; according to [this](https://blogs.msdn.microsoft.com/ie/2006/12/06/file-uris-in-windows/), even on Windows, URLs referring to files should use "/" as a path separator. So doing this on the user side is correct.

That said, it might work either way; i.e., if they pass `file:///c:\foo\bar.txt`, it might work. "\" is definitely illegal in the host part, which is why the test was failing before, but it might be legal in the path part, and we are just passing the path through to `OpenBucket`. I don't want to document it as working though., and the example should do the correct thing.

vangent:
#323

PR #645, opened 2018-11-07
blob/fileblob: write to a temporary file and rename in Close

vangent:
The previous code would `Open` the actual file and write to it. IMHO this is bad, since it probably truncates the existing file. It also means that we `Remove` the file in `Close` without ever having closed it, which fails on Windows (seems legit).

Now, we write everything to a temp file and `Rename` to commit the change during `Close`.

I filed #649 to add a more explicit test for the behavior of aborting a write.


blob/fileblob/fileblob.go
```diff
@@ -492,22 +493,20 @@ func (w writer) Write(p []byte) (n int, err error) {
 			return 0, fmt.Errorf("updating md5 hash: %v", err)
 		}
 	}
-	return w.w.Write(p)
+	return w.f.Write(p)
```
shantuo: Should it remove the temp file if there is an error during write?

blob/fileblob/fileblob.go
```diff
@@ -492,22 +493,20 @@ func (w writer) Write(p []byte) (n int, err error) {
...
 		return err
 	}
-	if err := setAttrs(w.path, w.attrs); err != nil {
-		return fmt.Errorf("write blob attributes: %v", err)
-	}
-	err := w.w.Close()
-	if err != nil {
+	// If the write was cancelled, delete the temp file.
+	if err := w.ctx.Err(); err != nil {
+		_ = os.Remove(w.f.Name())
```
shantuo: Can we `defer` this? Whenever we exit close we should all remove the temp file. On success, the `os.Remove` would return the "not exist" error which can be ignored.

vangent: Good idea; done. I also changed it to write the attributes file first, so if that fails we can abort without having clobbered the original file (if any).

PR #646, opened 2018-11-07
wire: in tests, don't copy wire_gen.go into the temporary GOPATH

zombiezen:
This does not change the results of the test, but should not be happening. I discovered this as a red herring while debugging #623.


zombiezen:
Actually, reassigning over to @shantuo as he wanted to ramp up on Wire stuff.

ijt:
I'm not sure I follow how the code change implements the description.

zombiezen:
It happens in both record and replay mode, sadly.

PR #648, opened 2018-11-07
blob/tests: fixes for blob conformance tests found by Windows build

vangent:
-- On Windows, having the `Reader` open means that you can't overwrite the file; thus calling `r.Close` explicitly instead of `defer r.Close()`.
-- On Windows, you can't (easily) create a file with a trailing "."; it is silently dropped. I moved the `.` in the test string so it's not right before the `\`.
-- Fixed a random typo in an error string.


vangent:
#323

PR #651, opened 2018-11-08
runtimevar/filevar: close file before trying to rename it

vangent:
This worked fine on Linux but fails on Windows. Closing the file before renaming it is more correct regardless.


vangent:
Ping @ijt ?

PR #652, opened 2018-11-08
wire/tests: when reading golden files, convert all line separators to \n

vangent:
On Windows, the tests were all failing because of the "\r\n" line breaks. This change reads the files in line by line and re-writes them using "\n".

#323

### DIFF
--- a/wire/internal/wire/wire_test.go
+++ b/wire/internal/wire/wire_test.go
@@ -15,7 +15,6 @@
 package wire
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"fmt"
@@ -454,13 +453,13 @@ type testCase struct {
 //
 func loadTestCase(root string, wireGoSrc []byte) (*testCase, error) {
 	name := filepath.Base(root)
-	pkg, err := readFile(filepath.Join(root, "pkg"))
+	pkg, err := ioutil.ReadFile(filepath.Join(root, "pkg"))
 	if err != nil {
 		return nil, fmt.Errorf("load test case %s: %v", name, err)
 	}
 	var wantProgramOutput []byte
 	var wantWireOutput []byte
-	wireErrb, err := readFile(filepath.Join(root, "want", "wire_errs.txt"))
+	wireErrb, err := ioutil.ReadFile(filepath.Join(root, "want", "wire_errs.txt"))
 	wantWireError := err == nil
 	var wantWireErrorStrings []string
 	if wantWireError {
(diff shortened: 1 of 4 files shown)


vangent:
https://screenshot.googleplex.com/6sVZuXEBWb7
  shows the tests now passing on Windows.

wire/internal/wire/wire_test.go
```diff
@@ -494,7 +498,7 @@ func loadTestCase(root string, wireGoSrc []byte) (*testCase, error) {
 		if !info.Mode().IsRegular() || filepath.Ext(src) != ".go" {
 			return nil
 		}
-		data, err := ioutil.ReadFile(src)
+		data, err := readFile(src)
```
vangent: Not sure if this one needs to be converted...?

wire/internal/wire/wire_test.go
```diff
@@ -41,7 +42,7 @@ func TestWire(t *testing.T) {
 	}
 	// The marker function package source is needed to have the test cases
 	// type check. loadTestCase places this file at the well-known import path.
-	wireGo, err := ioutil.ReadFile(filepath.Join("..", "..", "wire.go"))
+	wireGo, err := readFile(filepath.Join("..", "..", "wire.go"))
```
vangent: Not sure if this one needs to be converted....?

wire/internal/wire/wire_test.go
```diff
@@ -542,3 +546,23 @@ func (test *testCase) materialize(gopath string) error {
 	}
 	return nil
 }
+
+// readFile reads data from path. It uses io.Scanner to read the lines of
+// the file, ensuring that they are separated only by "\n".
+// Otherwise, on Windows, file lines that end with "\r\n" cause spurious
+// diffs.
+func readFile(path string) ([]byte, error) {
```
zombiezen: `readTextFile` seems a more apt name for this function, as it is interpreting the bytes rather than just verbatim copying.

wire/internal/wire/wire_test.go
```diff
@@ -542,3 +546,23 @@ func (test *testCase) materialize(gopath string) error {
 	}
 	return nil
 }
+
+// readFile reads data from path. It uses io.Scanner to read the lines of
```
zombiezen: Use of `bufio.Scanner` is more complex than using `bytes.Replace`. I would recommend the latter.

wire/internal/wire/wire_test.go
zombiezen: (Same advice here as the one above.)

wire/internal/wire/wire_test.go
zombiezen: If the Go source file can't be loaded as-is by the build, then that indicates a legitimate failure. As such, I think the correct behavior to be to preserve the bytes.

wire/internal/wire/wire_test.go
```diff
@@ -453,25 +454,28 @@ type testCase struct {
...
 	var wantProgramOutput []byte
 	var wantWireOutput []byte
-	wireErrb, err := ioutil.ReadFile(filepath.Join(root, "want", "wire_errs.txt"))
+	wireErrb, err := readFile(filepath.Join(root, "want", "wire_errs.txt"))
 	wantWireError := err == nil
 	var wantWireErrorStrings []string
 	if wantWireError {
 		wantWireErrorStrings = strings.Split(string(wireErrb), "\n\n")
+		for i, s := range wantWireErrorStrings {
+			wantWireErrorStrings[i] = strings.TrimSpace(s)
```
zombiezen: I don't understand this addition. Shouldn't all the error strings already match?

PR #653, opened 2018-11-09
travis: use goveralls for code coverage

vangent:
See https://coveralls.io/github/google/go-cloud.


internal/testing/runchecks.sh
```diff
@@ -31,7 +31,8 @@ module="github.com/google/go-cloud"
 # Run Go tests for each module.
 for path in "." "./internal/contributebot" "./samples/appengine"; do
   pushd ${path}
-  go test -race ./...
+  go mod download
+  goveralls -race -service=travis-ci
```
zombiezen: I don't understand these two lines.

1. `go mod download` should happen as needed. Why does it need to be called explicitly?
2. Does `goveralls` run in this way also invoke the tests? When I've used it before, I always had it operate on a coverage file produced by a run of `go test`.

(Studying this logic critically made me realize a different issue, sent #667.)

vangent: > `go mod download` should happen as needed.
goveralls seemed to be flaky without it. I don't have a good explanation for why. It would fail with "missing package" errors. Let me try to remove it and we'll see what happens.

> Does `goveralls` run in this way also invoke the tests?
Yes: https://github.com/mattn/goveralls#continuous-integration.

vangent:
Yea, you can see the test failure running goveralls. It works fine if I run "go mod download" first. Thoughts?

vangent:
> Try running go test explicitly with the -coverprofile flag,
Trying that....

PR #660, opened 2018-11-09
wire: add FAQ for duplicate providers

### LINKED ISSUE #492: wire: document why wire doesn't allow duplicate identical providers in a provider set
`wire.NewSet()` is the perfect terminology. By that definition, this (albeit contrived example) should ideally work:

```golang
var Set = wire.NewSet(ProvideFoo, ProvideFoo)
```

Currently, this fails with a multiple bindings error. While you certainly want to prohibit the application from defining bindings in conflicting ways, automatically deduplicating _identical_ bindings would simplify application development with Wire.

A more interesting example is when you have provider sets at multiple levels of the stack with inter-dependencies:

```
                 ----------------
                /| feature1.Set |\
               / ---------------- \   ----------------
              /                    ---| library1.Set |
------------ /   ---------------- /   ----------------
| main.Set | ----| feature2.Set |/
------------ \   ----------------
              \
               \ ----------------     ----------------
                \| feature3.Set |-----| library2.Set |
                 ----------------     ----------------
```

Without deduplication, `main.Set` has to list all of the Sets in the entire application:

```golang
package main
var Set = wire.Set(
    feature1.Set,
    feature2.Set,
    feature3.Set,
    library1.Set,
    library2.Set)
```

This "top level aggregation" pattern requires knowledge of the entire tree of Sets at the top level, and can be a maintenance hazard as Sets are removed / reconfigured (i.e. dependency rot). This is a common problem with other DI frameworks like Guice.

*With* deduplication, each Set can be self-contained by including the Sets of their dependencies, even shared ones. Then the top level need only include Sets for its high level features:

```golang
package feature1
var Set = wire.Set(..., library1.Set)

package feature2
var Set = wire.Set(..., library1.Set)

package feature3
var Set = wire.Set(..., library2.Set)

package main
var MainSet = wire.Set(
    feature1.Set,
    feature2.Set,
    feature3.Set)
```

This provides better encapsulation, giving each package the sovereignty to add and remove downstream dependencies without requiring modifications to their upstream dependents.

Long story short: deduplication of identical bindings can allow applications to declare (possibly shared) provider set dependencies, improving encapsulation and making application "wiring" closer to dependency-aware build systems like Bazel.

I hope this provides some good context for this feature request. Thanks!

zombiezen:
Fixes #492


wire/README.md
```diff
@@ -600,6 +600,25 @@ func inject() *Bar {
 }
 ```
 
+### Why does Wire forbid including the same provider multiple times?
+
```
vangent: I think we should have some guidance for the higher level question described in the issue. Not "why does Wire make this illegal", but "why you shouldn't be doing this".

The scenario is something like:

```
app uses libA
app uses libB
libA uses libC
libB uses libC
```

and the problem comes up if both libA and libB have a ProviderSet that includes providers for things from libC. The argument is that either

1. libC is an implementation detail inside libA/libB. In this scenario, libA/libB should construct libC things internally and not expose it to the application layer. No need to include libC things in the ProviderSet.
2. libC things need some kind of configuration for libA/libB to use it. In this scenario, the app needs to know about libC things, and should construct them itself. For example, it might want to pass the same libC thing to both libA and libB. Or not. Again, libA/libB should not include libC things in their ProviderSet -- the application needs to know about libC so it should include libC's providers or ProviderSet explicitly.

Maybe a question along the lines of "What providers should a library include in its ProviderSet?" could address this?

zombiezen: I see. SGTM, but it sounds more like it belongs in the best practices section above. I've added more guidance there. PTAL and let me know whether I should keep this section or remove it.

PR #664, opened 2018-11-09
travis: remove comment referring to checkpr.sh, which no longer exists

vangent:
We probably don't need to get the newer bash/grep anymore, since the new `runchecks.sh` is simpler, but I don't see the harm in leaving it to save ourselves some pain in the future.


zombiezen:
You disabled maintainer edits, so I can't update and merge for you.

Issue #254, opened 2018-07-27
postgresql: add PostgreSQL support


zombiezen:
We have the dialers for MySQL, but both AWS and GCP also have hosted PostgreSQL offerings. Go Cloud should have dialers for these things.

It does raise an interesting question of whether they should be added to the `awscloud.Services` and `gcpcloud.Services` sets, since they would conflict with the MySQL providers.  How this would be resolved is TBD.

zombiezen:
I have a PR out for review that would add the RDS support. Once that is submitted, the work that remains is:

- Address the inclusion of the MySQL providers in the gcpcloud and awscloud packages.
- Add OpenCensus tracing to these packages, similarly to how they were added to the MySQL packages. I am waiting for #608's resolution to proceed.

zombiezen:
This will land in v0.7.

PR #670, opened 2018-11-12
postgres/cloudpostgres: new package

zombiezen:
Extracted common functionality from Cloud MySQL into gcp/cloudsql package.

I haven't added OpenCensus wrapping yet, as I'm waiting to understand the correct pattern on #608.

Updates #254


postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/GoogleCloudPlatform/cloudsql-proxy/proxy/proxy"
+	"github.com/lib/pq"
+)
+
+// Params specifies how to connect to a Cloud SQL database.
+type Params struct {
+	ProjectID string
+	Region    string
+	Instance  string
```
ijt: It might be worth linking to the documentation for instances: https://cloud.google.com/sql/docs/postgres/create-instance.

zombiezen: I couldn't find a good enough link target that would describe the parameters here, so I added a link to the general docs in the package documentation. I think the example helps fill in the gap.

PR #674, opened 2018-11-12
all: only run coverage for the root module

### LINKED ISSUE #625: all: introduce code coverage
It would be nice to get code coverage reports periodically.

Coveralls.io/Goveralls

vangent:
Fixes #625.

The previous code had a bug; doing `exec <cmd> && <cmd2>` never runs `cmd2`. We also only want to do code coverage on the root module.


vangent:
https://coveralls.io/github/google/go-cloud looks much better now.

PR #685, opened 2018-11-13
internal/pubsub: update design to use package-level constructors instead of a Client type 

### LINKED ISSUE #682: pubsub: update design to use package-level constructors instead of a Client type
[A `Client` type](https://github.com/google/go-cloud/blob/599092f950ec7224d9ba6dddb48f7183da6d3156/internal/docs/pubsub/design.md#driver-implementers-perspective) is not really necessary. Each provider package will have a unique signature for constructing either a Topic or Subscription because they are necessarily provider-specific (drawing upon the guidance in [Developers and Operators](https://github.com/google/go-cloud/blob/master/internal/docs/design.md#developers-and-operators)). I would advocate for promoting the `Client` methods to package functions. A `*pubsub.Topic` or `*pubsub.Subscription` is much more like a `*sql.DB`, much like a `*blob.Bucket` is like a `*sql.DB`. For example, in my mind, the exported API for GCP Pub/Sub would look like this:

```go
package gcppubsub

type TopicOptions struct { /* ... */ }
func OpenTopic(ctx context.Context, conn *grpc.ClientConn, projectID, topic string, opts *TopicOptions) (*pubsub.Topic, error) {
  // ...
}

type SubscriptionOptions struct { /* ... */ }
func OpenSubscription(ctx context.Context, conn *grpc.ClientConn, projectID, topic string, opts *SubscriptionOptions) (*pubsub.Topic, error) {
  // ...
}
```

ijt:
Fixes #682


internal/docs/pubsub/design.md
```diff
@@ -136,11 +126,7 @@ func main() {
 
 func receive() error {
 	ctx := context.Background()
-	client, err := acmepubsub.NewClient(ctx, "unicornvideohub")
-	if err != nil {
-		return err
-	}
-	s, err := client.OpenSubscription(ctx, "signup-minder", nil)
+	s, err := acmepubsub.OpenSubscription(ctx, "unicornvideohub", "signup-minder", nil)
```
jba: It's unlikely that there wouldn't be some sort of client or connection that is shared for all topics and subscriptions.

ijt: It appears we have a lack of consensus about this. I'll set up a meeting to resolve it.

zombiezen: I believe there to be consensus; I agree with @jba's statement. What is in this PR does not currently match what I proposed in #682. This PR proposes:

```go
func OpenTopic(ctx context.Context, projectID, topicName string) (*pubsub.Topic, error)
```

whereas I proposed something like:

```go
func OpenTopic(ctx context.Context, conn *grpc.ClientConn, projectID, topic string, opts *TopicOptions) (*pubsub.Topic, error)
```

Importantly, notice that the latter takes in a `*grpc.ClientConn`. Not doing so would be problematic.

ijt: Ah ok. Updating...

ijt: I would note though that for GCP, the Client already takes the projectID as a parameter, so that's no longer needed for the Topic constructor.

PR #692, opened 2018-11-13
blob: add examples for gcsblob, s3blob

### LINKED ISSUE #597: blob: add examples showing how to set up gcsblob and s3blob with various config options

vangent:
Fixes #597.

### DIFF
--- a/blob/gcsblob/example_test.go
+++ b/blob/gcsblob/example_test.go
@@ -1,0 +1,48 @@
+// Copyright 2018 The Go Cloud Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


blob/gcsblob/example_test.go
```diff
@@ -0,0 +1,48 @@
...
+	// above.
+	client, err := gcp.NewHTTPClient(gcp.DefaultTransport(), gcp.CredentialsTokenSource(creds))
+	if err != nil {
+		return
+	}
+
+	// Create a *blob.Bucket.
+	_, _ = gcsblob.OpenBucket(ctx, "my-bucket", client, nil)
+
+	// Output:
```
clausti: was there intended to be more below this?

vangent: No. In "example" tests, you put the expected output under a `// Output` comment like this, and the test "passes" if the actual output matches the expected.

https://blog.golang.org/examples

This test doesn't output anything, so there's nothing under the comment.

Depending on the environment, `OpenBucket` here may or may not work, so it's hard to output anything interesting.

PR #694, opened 2018-11-14
pubsub: no longer block on Ack() calls

### LINKED ISSUE #678: pubsub: no longer block on Ack() calls
As per @rsc's comment, it doesn't make sense to do different things conditionally on whether an Ack call succeeds, so we may as well remove the `error` return and no longer block in these calls.

ijt:
Fixes #678

### DIFF
--- a/internal/pubsub/acks_test.go
+++ b/internal/pubsub/acks_test.go
@@ -15,8 +15,6 @@ package pubsub_test
 
 import (
 	"context"
-	"errors"
-	"fmt"
 	"math/rand"
 	"reflect"
 	"sync"
@@ -50,10 +48,12 @@ func TestAckTriggersDriverSendAcksForOneMessage(t *testing.T) {
 	var sentAcks []driver.AckID
 	id := rand.Int()
 	m := &driver.Message{AckID: id}
+	ackChan := make(chan struct{})
 	ds := &ackingDriverSub{
 		q: []*driver.Message{m},
 		sendAcks: func(_ context.Context, ackIDs []driver.AckID) error {
 			sentAcks = ackIDs
+			ackChan <- struct{}{}
 			return nil
 		},
 	}
@@ -63,9 +63,8 @@ func TestAckTriggersDriverSendAcksForOneMessage(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if err := m2.Ack(ctx); err != nil {
-		t.Fatal(err)
-	}
+	m2.Ack()
+	<-ackChan
 	if len(sentAcks) != 1 {
 		t.Fatalf("len(sentAcks) = %d, want exactly 1", len(sentAcks))
 	}
@@ -76,6 +75,7 @@ func TestAckTriggersDriverSendAcksForOneMessage(t *testing.T) {
 
 func TestMultipleAcksCanGoIntoASingleBatch(t *testing.T) {
 	ctx := context.Background()
+	var wg sync.WaitGroup
 	sentAcks := make(map[driver.AckID]int)
 	ids := []int{1, 2}
 	ds := &ackingDriverSub{
@@ -83,6 +83,7 @@ func TestMultipleAcksCanGoIntoASingleBatch(t *testing.T) {
 		sendAcks: func(_ context.Context, ackIDs []driver.AckID) error {
 			for _, id := range ackIDs {
 				sentAcks[id]++
+				wg.Done()
 			}
 			return nil
 		},
@@ -93,20 +94,15 @@ func TestMultipleAcksCanGoIntoASingleBatch(t *testing.T) {
 	defer sub.Close()
 
 	// Receive and ack the messages concurrently.
-	var wg sync.WaitGroup
 	for i := 0; i < 2; i++ {
 		wg.Add(1)
 		go func() {
-			defer wg.Done()
 			mr, err := sub.Receive(ctx)
 			if err != nil {
 				t.Error(err)
 				return
 			}
-			if err := mr.Ack(ctx); err != nil {
-				t.Error(err)
-				return
-			}
+			mr.Ack()
 		}()
 	}
 	wg.Wait()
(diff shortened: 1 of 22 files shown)


internal/pubsub/pubsub.go
```diff
@@ -42,23 +42,15 @@ type Message struct {
 // be sent again to the associated Subscription. This method blocks until
```
vangent: Comment is stale.

internal/pubsub/pubsub.go
```diff
@@ -42,23 +42,15 @@ type Message struct {
...
-	}
-	return <-mec.errChan
-}
-
-type msgErrChan struct {
-	msg     *Message
-	errChan chan error
+	go func() {
+		if err := m.sub.ackBatcher.Add(m, size); err != nil {
+			panic(err)
```
vangent: `panic` doesn't seem like a good idea in a library... What are the reasons for `Add` to return an `error`? Maybe just

`go m.sub.ackBatcher.Add(m, size)`

?

internal/pubsub/pubsub.go
```diff
@@ -244,19 +241,18 @@ func (s *Subscription) Close() error {
...
-		for _, mec := range mecs {
-			m := mec.msg
+		for _, m := range ms {
 			id := m.ackID
 			ids = append(ids, id)
 		}
 		err := d.SendAcks(ctx, ids)
-		for _, mec := range mecs {
-			mec.errChan <- err
+		if err != nil {
```
vangent: Definitely do not want to panic here. 

Maybe `driver.SendAcks` should not return an error anymore, if we're not going to use it for anything....? Or maybe the concrete type should retry...?

internal/pubsub/pubsub.go
ijt: The reasons are `ErrOversizedItem` or `ErrOverflow`.

internal/pubsub/pubsub.go
```diff
@@ -247,9 +243,12 @@ func NewSubscription(ctx context.Context, d driver.Subscription, opts *Subscript
...
 		}
-		err := d.SendAcks(ctx, ids)
-		if err != nil {
-			panic(err)
+		for {
+			err := d.SendAcks(ctx, ids)
+			if err == nil {
+				break
+			}
+			time.Sleep(time.Second)
```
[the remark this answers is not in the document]

internal/pubsub/pubsub.go
ijt: Freshened.

internal/pubsub/acks_test.go
```diff
@@ -162,39 +158,16 @@ func TestTooManyAcksForASingleBatchGoIntoMultipleBatches(t *testing.T) {
...
-
-func TestCancelAck(t *testing.T) {
-	ctx, cancel := context.WithCancel(context.Background())
 	m := &driver.Message{}
 	ds := &ackingDriverSub{
 		q: []*driver.Message{m},
 		sendAcks: func(_ context.Context, ackIDs []driver.AckID) error {
-			<-ctx.Done()
-			return ctx.Err()
+			c := make(chan struct{})
```
jba: You can write this as
```
select {}
```

internal/pubsub/pubsub.go
```diff
@@ -202,17 +188,14 @@ func (s *Subscription) Close() error {
...
-			m := mec.msg
+		for _, m := range ms {
 			id := m.ackID
 			ids = append(ids, id)
 		}
-		err := d.SendAcks(ctx, ids)
-		for _, mec := range mecs {
-			mec.errChan <- err
-		}
+		// TODO(#695): Do something sensible if SendAcks returns an error.
```
jba: As Russ suggested, we should probably return an error from Receive. (In a separate PR.)

internal/pubsub/acks_test.go
ijt: Done.

internal/pubsub/pubsub.go
ijt: I'm not so sure about that. Anyway, let's move this discussion to #695.

internal/pubsub/acks_test.go
```diff
@@ -75,13 +74,15 @@ func TestAckTriggersDriverSendAcksForOneMessage(t *testing.T) {
...
 func TestMultipleAcksCanGoIntoASingleBatch(t *testing.T) {
 	ctx := context.Background()
+	var wg sync.WaitGroup
 	sentAcks := make(map[driver.AckID]int)
 	ids := []int{1, 2}
 	ds := &ackingDriverSub{
 		q: []*driver.Message{{AckID: ids[0]}, {AckID: ids[1]}},
 		sendAcks: func(_ context.Context, ackIDs []driver.AckID) error {
 			for _, id := range ackIDs {
 				sentAcks[id]++
```
jba: We probably shouldn't assume that sendAcks is called single-threaded. That will depend on details of the bundler/batcher. So I'd add a mutex.

ijt: Done.

internal/pubsub/pubsub.go
```diff
@@ -39,9 +39,7 @@ type Message struct {
 }
 
 // Ack acknowledges the message, telling the server that it does not need to
-// be sent again to the associated Subscription. This method blocks until
-// the message has been confirmed as acknowledged on the server, or failure
-// occurs.
+// be sent again to the associated Subscription.
```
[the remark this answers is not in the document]
vangent: I left a comment about this on the issue, let's continue there.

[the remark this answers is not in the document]

ijt:
Ping, @vangent, can you approve?

vangent:
> Ping, @vangent, can you approve?

Not yet; we are discussing on the issue.

internal/pubsub/pubsub.go
```diff
@@ -202,19 +194,16 @@ func (s *Subscription) Close() error {
...
-		mecs := item.([]msgErrChan)
+		ms := item.([]*Message)
 		var ids []driver.AckID
-		for _, mec := range mecs {
-			m := mec.msg
+		for _, m := range ms {
 			id := m.ackID
 			ids = append(ids, id)
 		}
-		err := d.SendAcks(ctx, ids)
```
vangent: It looks like this is no longer calling `SendAcks` at all...?

ijt: Huh? There is a call to SendAcks still.

vangent: Ah sorry, it is down below, I missed it.

PR #698, opened 2018-11-14
internal/batcher: general adaptive batching

jba:


zombiezen:
Still reviewing this; just a moment.

ijt:
I'd like to have a benchmark for bursts of activity, but I'm going to file an issue for that once this is committed.

internal/batcher/batcher_test.go
```diff
@@ -0,0 +1,108 @@
...
+		if len(items) > maxBatch {
+			maxBatch = len(items)
+		}
+		mu.Unlock()
+		defer func() { mu.Lock(); outstanding--; mu.Unlock() }()
+		time.Sleep(10 * time.Millisecond)
+		return nil
+	})
+	var wg sync.WaitGroup
+	nItems := 1000
```
zombiezen:
```suggestion
	const nItems = 1000
```

internal/batcher/batcher_test.go
```diff
@@ -0,0 +1,108 @@
...
+			time.Sleep(time.Millisecond)
+			if err := b.Add(ctx, i); err != nil {
+				log.Fatal(err)
+			}
+		}()
+	}
+	wg.Wait()
+	// Check that we saturated the batcher.
+	if max != maxOutstandingHandlers {
+		t.Errorf("got %d, want %d", max, maxOutstandingHandlers)
```
zombiezen:
```suggestion
		t.Errorf("max concurrent handlers = %d, want %d", max, maxOutstandingHandlers)
```

internal/batcher/batcher_test.go
```diff
@@ -0,0 +1,108 @@
...
+	var wg sync.WaitGroup
+	nItems := 1000
+	for i := 0; i < nItems; i++ {
+		i := i
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			time.Sleep(time.Millisecond)
+			if err := b.Add(ctx, i); err != nil {
+				log.Fatal(err)
```
zombiezen: Prefer `t.Error` here, since `log.Fatal` will stop all the tests, not just this one.

```suggestion
				t.Errorf("b.Add(ctx, %d) error: %v", err)
```

internal/batcher/batcher_test.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil
+	})
+	var wg sync.WaitGroup
+	nItems := 1000
+	for i := 0; i < nItems; i++ {
+		i := i
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			time.Sleep(time.Millisecond)
```
zombiezen: I have a vague recollection from talking with you before that while this sleep is not strictly necessary, it does increase the likelihood of saturation (because this is a probabilistic test). For future readers, can you add a comment that explains this? (Same with the other `time.Sleep` above.)

internal/batcher/batcher.go
```diff
@@ -0,0 +1,113 @@
...
+		default:
+		}
+		// Wait until either we get a result or we can do some work.
+		select {
+		case err := <-c:
+			return err
+		case <-b.sem:
+			b.callHandler()
+			b.sem <- struct{}{}
+		}
```
zombiezen: I believe this to be missing a case for receiving `<-ctx.Done()`.

internal/batcher/batcher.go
```diff
@@ -0,0 +1,113 @@
...
+
+// Add adds an item to the batcher. It blocks until the handler has
+// processed the item and reports the error that the handler returned.
+func (b *Batcher) Add(ctx context.Context, item interface{}) error {
+	// Create a channel to receive the error from the handler. Give it a buffer so
+	// that if a goroutine is in the middle of calling a handler when another
+	// finishes, the latter won't block sending to the channel.
+	c := make(chan error, 1)
+	b.mu.Lock()
+	b.pending = append(b.pending, waiter{item, c})
```
zombiezen: I am concerned that this can grow unbounded, especially since there's no explicit deadline set on the batch sends. Implementing cancel should mitigate this, but at the very least, let's add a TODO.

internal/batcher/batcher_test.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		want := []int{i}
+		if !cmp.Equal(got, want) {
+			t.Errorf("got %+v, want %+v", got, want)
+		}
+	}
+}
+
+func TestBatcherSaturation(t *testing.T) {
+	// Verify that under high load, the maximum number of handlers
```
zombiezen: nit: Remove the comma.

internal/batcher/batcher_test.go
```diff
@@ -0,0 +1,108 @@
...
+	if maxBatch <= 1 {
+		t.Errorf("got max batch size of %d, expected > 1", maxBatch)
+	}
+	// Check that handlers saw every item exactly once.
+	want := map[int]int{}
+	for i := 0; i < nItems; i++ {
+		want[i] = 1
+	}
+	if diff := cmp.Diff(count, want); diff != "" {
+		t.Error(diff)
```
zombiezen: It would be helpful to the person running the test to have context on what diff they are seeing.

```suggestion
		t.Errorf("items:\n%s", diff)
```

internal/batcher/batcher_test.go
```diff
@@ -0,0 +1,108 @@
...
+		mu.Unlock()
+		defer func() { mu.Lock(); outstanding--; mu.Unlock() }()
+		time.Sleep(10 * time.Millisecond)
+		return nil
+	})
+	var wg sync.WaitGroup
+	nItems := 1000
+	for i := 0; i < nItems; i++ {
+		i := i
+		wg.Add(1)
```
zombiezen: nit: You can run `wg.Add(nItems)` before the loop.

jba:
Thanks for the review. Holding off on this because I realized on the way in today that it can spin if maxOutstandingHandlers > 1. Will work on it.

internal/batcher/batcher.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+type waiter struct {
+	item interface{}
+	errc chan error
+}
+
+// New creates a new Batcher.
+//
+// itemExample is a value of the type that will be batched. For example, if you
```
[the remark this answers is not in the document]
jba: That would require the user to depend on, and at least sort of understand, the reflect package. The nice thing about this API is that callers don't ever need reflect. I think the doc comment about `itemExample` is pretty clear.

The bundler uses the same convention and it seems to be fine.

internal/batcher/batcher.go
```diff
@@ -0,0 +1,113 @@
...
+	"context"
+	"reflect"
+	"sync"
+)
+
+// A Batcher batches items.
+type Batcher struct {
+	handler       func(interface{}) error
+	itemSliceZero reflect.Value // nil (zero value) for slice of items
+	sem           chan struct{} // semaphore for outstanding handlers
```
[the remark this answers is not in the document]
jba: Obsolete; removed the semaphore.

internal/batcher/batcher_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"log"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestBatcherSequential(t *testing.T) {
+	// Verify sequential calls to a batcher.
```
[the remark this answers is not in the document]
jba: Done.

internal/batcher/batcher_test.go
jba: Done.

internal/batcher/batcher.go
jba: It can, but not in an interesting way (I claim).

Since `Add` is synchronous, in order to have N items in `pending` you need N goroutines. So you're already using memory proportional to the number of items. 

Put another way, you should be doing flow control with whatever is starting all those goroutines; the size of `pending` is just a side effect.

internal/batcher/batcher_test.go
jba: Done.

internal/batcher/batcher_test.go
jba: Done.

internal/batcher/batcher_test.go
jba: That is error-prone; what if I put a conditional in the loop so that I don't start a goroutine sometimes? Unless performance is an issue, I like putting the `wg.Add` just before the `go`.

jba:
I completely rewrote the implementation. It's actually even simpler now.

internal/batcher/batcher_test.go
jba: Done.

internal/batcher/batcher.go
jba: Fixed in the rewrite.

ijt:
LGTM.

Issue #1, opened 2018-05-11
runtimevar/paramstore: Watch returns same value multiple times


zombiezen:
## Reproduction case

```go
package main

import (
    "context"
    "log"
    "time"

    "github.com/aws/aws-sdk-go/aws/session"
    "github.com/google/go-cloud/runtimevar"
    "github.com/google/go-cloud/runtimevar/paramstore"
)

func main() {
    ctx := context.Background()
    sess := session.Must(session.NewSession())
    client := paramstore.NewClient(ctx, sess)
    v, err := client.NewVariable(ctx, "/foo/bar", runtimevar.StringDecoder, &paramstore.WatchOptions{
        WaitTime: 5 * time.Second,
    })
    for {
        snapshot, err := v.Watch(ctx)
        if err != nil {
            log.Println("Watch error:", err)
        }
        log.Println("Value is now:", snapshot.Value)
    }
}
```

Run with:

```shell
$ aws ssm put-parameter --name="/foo/bar" --type=String --value="ohai from AWS"
$ go run foo.go
```

## Expected output

```
Value is now: ohai from AWS
```

## Actual output

```
Value is now: ohai from AWS
Value is now: ohai from AWS
Value is now: ohai from AWS
```

PR #701, opened 2018-11-14
wire tool example

kunalbhatia18:

See


ijt:
Hi @kunalbhatia18 , thank you for filing this contribution. I think it may make more sense in a separate github repository though since our users would tend to be much more familiar with Go than with Node.js. I'm closing this out but feel free to open an issue if you would like to discuss and make a case for why something like this would work well within Go Cloud.

Issue #669, opened 2018-11-12
wire: test flakes on Travis with "failed to cache compiled Go files"


zombiezen:
### Describe the bug

The Wire tests fail intermittently on Travis with an error like:

```
        wire_test.go:96: go [list -e -json -compiled -test=false -export=false -deps=true -tags=wireinject -- example.com/foo]: exit status 1: go: finding github.com/google/go-cloud v0.2.0
            go build runtime/internal/atomic: failed to cache compiled Go files
```

### To Reproduce

This is an intermittent failure, but I've only seen it on Travis. Uncertain what the triggers are.

## Expected behavior

Wire tests should pass.

/cc @vangent, who I believe hit this failure before, but I can't find evidence of it.

vangent:
https://travis-ci.com/google/go-cloud/jobs/157991519

shows this.

zombiezen:
Yeah, worth noting that the test it fails on is not fixed, and the package that fails is not fixed. Only the error message.

Perhaps there's contention running `go list` a lot concurrently?

zombiezen:
@alandonovan @ianthehat, any ideas on what's going on here? Seems like the conditions in here are triggering regularly:

https://github.com/golang/go/blob/0a72e8eefbdaa62bf782618d50851e4740408cba/src/cmd/go/internal/work/exec.go#L599-L604

Are we holding something wrong or is this an issue with `go/packages`?

ianthehat:
Not really sure, but I know that there are known issues with running go list in module mode concurrently, that do tend to cause errors like this, is that a possibility?
This means you cant use go/packages.Load concurrently right now, from either a single process or multiple processes

ianthehat:
It's the go command (you end up with a corrupted module cache), and I know it's what Bryan has been focusing on for a while now, I think he is supposed to be done with making it work, so hopefully it will all be in 1.12, but you should check with him to be sure.
I don't know what the tracking issue is though, sorry.

zombiezen:
@bcmills, is there a Go bug tracking the corrupted module cache? I am slightly puzzled why this would occur, as each test runs in a separate, isolated `GOPATH`.

bcmills:
The known source of module-cache corruption is https://github.com/golang/go/issues/26794.

The `go: finding` message relates to the module cache, but the `failed to cache compiled Go files` message is the build cache.

zombiezen:
Thanks everyone!

[...]

PR #704, opened 2018-11-14
wire: serialize test execution

zombiezen:
We are seeing errors of the form "failed to cache compiled Go files" when running the tests on Travis. Hopefully this should reduce or eliminate the flakiness.

Updates #669

### DIFF
--- a/wire/internal/wire/wire_test.go
+++ b/wire/internal/wire/wire_test.go
@@ -70,8 +70,9 @@ func TestWire(t *testing.T) {
 	for _, test := range tests {
 		test := test
 		t.Run(test.name, func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


wire/internal/wire/wire_test.go
```diff
@@ -70,7 +70,8 @@ func TestWire(t *testing.T) {
 	for _, test := range tests {
 		test := test
 		t.Run(test.name, func(t *testing.T) {
-			t.Parallel()
+			// TODO(light): These tests should be parallelizable, but seems to flake.
```
ijt: s/but seems to flake/but they have tended to flake when run that way/.

zombiezen: Done.

Issue #638, opened 2018-11-06
pubsub: add example_test


ijt:
A few examples are already in the pubsub design doc. It would be useful to extract them and put them in an example test, so they'll be included in godoc results.

PR #705, opened 2018-11-14
internal/pubsub: add GCP PubSub driver

ijt:
Fixes #705


ijt:
The code practically wrote itself based on compiler errors. I'm thinking of adding tests in a separate PR once we have the conformance test suite, but I'm open to suggestions on that.

ijt:
The build flaked again: "failed to cache compiled Go files". Rerunning.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,130 @@
...
+
+// Package gcppubsub provides an implementation of pubsub that uses GCP
+// PubSub.
+package gcppubsub
+
+import (
+	"context"
+	"fmt"
+
+	rawgcppubsub "cloud.google.com/go/pubsub/apiv1"
```
jba: How about a shorter import alias? Like `raw`.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/google/go-cloud/internal/pubsub"
+	"github.com/google/go-cloud/internal/pubsub/driver"
+	pubsubpb "google.golang.org/genproto/googleapis/pubsub/v1"
```
jba: `pb`, since there is no other.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,130 @@
...
+	client *rawgcppubsub.PublisherClient
+}
+
+// Close closes the connection to the API service. The user should invoke this
+// when the client is no longer required.
+func (t *topic) Close() error {
+	return t.client.Close()
+}
+
+// SendBatch publishes all the messages in dms. This method returns only after
```
jba: Following @vangent's suggestion, we should write all these as `Foo implements driver.Foo`.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,130 @@
...
+// all the messages are sent, an error occurs, or the context is cancelled.
+func (t *topic) SendBatch(ctx context.Context, dms []*driver.Message) error {
+	var ms []*pubsubpb.PubsubMessage
+	for _, dm := range dms {
+		m := &pubsubpb.PubsubMessage{
+			Data:       dm.Body,
+			Attributes: dm.Metadata,
```
jba: I thought we agreed to s/Metadata/Attributes/?

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,130 @@
...
+
+// ReceiveBatch returns a batch of messages that have queued up for the
+// subscription on the server. If no messages are available yet, it blocks
+// until there is at least one, or the context is done.
+func (s *subscription) ReceiveBatch(ctx context.Context) ([]*driver.Message, error) {
+	req := &pubsubpb.PullRequest{
+		Subscription:      s.path,
+		ReturnImmediately: false,
+	}
+	resp, err := s.client.Pull(ctx, req)
```
jba: You need a loop around this. Pull can still return empty-handed, even if `ReturnImmediately` is false.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,130 @@
...
+
+// SendAcks acknowledges the messages with the given ackIDs on the server so
+// that they will not be received again for this subscription if the server
+// gets the acks before their deadlines. This method returns only after all
+// the ackIDs are sent, an error occurs, or the context is cancelled.
+func (s *subscription) SendAcks(ctx context.Context, ids []driver.AckID) error {
+	var ids2 []string
+	for _, id := range ids {
+		id2, ok := id.(string)
+		if !ok {
```
jba: Meh, I'd just let it panic.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,130 @@
...
+// Close closes the connection to the API service. The user should invoke this
+// when the client is no longer required.
+func (s *subscription) Close() error {
+	return s.client.Close()
+}
+
+// OpenTopic opens the topic on GCP PubSub for the given projectID and
+// topicName. If the topic does not exist then failure will occur when messages
+// are sent to it.
+func OpenTopic(ctx context.Context, client *rawgcppubsub.PublisherClient, projectID, topicName string) *pubsub.Topic {
```
jba: How do you feel about putting this just after `type topic`? That mimics the way godoc shows things (type, then constructor functions, then other methods) and I've gotten used to seeing code laid out that way.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,130 @@
...
+	path := fmt.Sprintf("projects/%s/topics/%s", projectID, topicName)
+	dt := &topic{path, client}
+	t := pubsub.NewTopic(ctx, dt)
```
jba: I'm confused by this. Shouldn't the driver constructors return driver types, that the user can then plug into `pubsub.NewTopic`? Is this how the other packages work?

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Done.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: I think I buy @vangent's argument that it matters more to keep consistency across the Go Cloud subpackages (in this case pubsub and blob) than it does to make particular Go Cloud subpackages consistent with some subset of providers. If we do that then if devs learn one part of Go Cloud then the conventions they learned there will be likely to hold in other parts of Go Cloud.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Okay. What error should I check for?

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Done.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Makes sense. Done.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: This is the way `blob` does it, probably because it would otherwise be too inconvenient and potentially confusing for users.

```go
dt := gcppubsub.OpenTopic(ctx, client, projectID, topicName)
t := pubsub.NewTopic(ctx, dt)
```
vs.
```go
t := gcppubsub.OpenTopic(ctx, client, projectID, topicName)
```

ijt: https://github.com/google/go-cloud/blob/master/blob/blob.go#L572

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// ReceiveBatch implements driver.Subscription.ReceiveBatch.
+func (s *subscription) ReceiveBatch(ctx context.Context) ([]*driver.Message, error) {
+	req := &pb.PullRequest{
+		Subscription:      s.path,
+		ReturnImmediately: false,
+	}
+	for {
+		resp, err := s.client.Pull(ctx, req)
```
zombiezen: Unfortunately, this error will never get seen by the end user, but it could be helpful for debugging. We could add a hook to the constructor to the effect of:

```go
// A type that implements ErrorLogger reports errors that occur while receiving a batch.
type ErrorLogger interface {
  LogError(context.Context, error)
}
```

(Probably with some filtering for transient errors or such.)

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,127 @@
...
+		if err != nil {
+			// Retry.
+			time.Sleep(time.Second)
+			continue
+		}
+		var ms []*driver.Message
```
zombiezen: This seems like it should be after the loop body. If you negate the condition above, then the structure becomes:

```go
for {
  resp, err := s.client.Pull(ctx, req)
  if err == nil {
    break
  }
  // handle error and sleep...
}
// success...

**Comment by zombiezen** [comment]
This change doesn't have tests, but I take it the conformance tests aren't ready yet?

File: internal/pubsub/gcppubsub/gcppubsub.go [hunk 11]
```diff
@@ -0,0 +1,127 @@
...
+// Close implements driver.Topic.Close.
+func (t *topic) Close() error {
+	return t.client.Close()
+}
+
+// SendBatch implements driver.Topic.SendBatch.
+func (t *topic) SendBatch(ctx context.Context, dms []*driver.Message) error {
+	var ms []*pb.PubsubMessage
+	for _, dm := range dms {
+		m := &pb.PubsubMessage{
```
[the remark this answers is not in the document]
ijt: Done.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Filed https://github.com/google/go-cloud/issues/713.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Done.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,133 @@
...
+
+type topic struct {
+	path   string
+	client *raw.PublisherClient
+}
+
+// OpenTopic opens the topic on GCP PubSub for the given projectID and
+// topicName. If the topic does not exist then failure will occur when messages
+// are sent to it.
+func OpenTopic(ctx context.Context, client *raw.PublisherClient, projectID, topicName string) *pubsub.Topic {
```
vangent: Note: you will probably want an `openTopic` function that returns the driver interface, and `OpenTopic` that calls that and then `pubsub.NewTopic`. This is so that the test harness can call `openTopic` to get a driver interface implementation.

vangent: Is there no way to get the projectID from `raw.PublisherClient`? Presumably the projectID must be known to create the `PublisherClient` so it kind of sucks to have to provide it again.

vangent: To play nice with `wire`, I don't think this function should take two string arguments; that means it can't be used as a `wire` provider function because wire won't be able to figure out which string goes where. If projectID is necessary, use `gcp.ProjectID` for it.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+type subscription struct {
+	client *raw.SubscriberClient
+	path   string
+}
+
+// OpenSubscription opens the subscription on GCP PubSub for the given
+// projectID and subscriptionName. If the subscription does not exist then
+// failure will occur when an attempt is made to receive messages from it.
```
vangent: Comments like this should be on the driver interface, as we don't want the semantics to change across driver implementations.

For example, if this is the semantics, then there should be a test that verifies that calling `OpenSubscription` for a subscription that doesn't exist returns, but that trying to receive from it returns an error.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -0,0 +1,133 @@
...
+		ReturnImmediately: false,
+	}
+	var resp *pb.PullResponse
+	for {
+		var err error
+		resp, err = s.client.Pull(ctx, req)
+		if err == nil {
+			break
+		}
+		// Retry:
```
vangent: I think I agree with you that this should be done in the concrete type.

Make sure the driver interface says something about this.

internal/pubsub/gcppubsub/gcppubsub.go
vangent: The `Open` function should take an `Options` struct, even if it's empty for now. This allows us to add options later without it being a breaking change.

vangent: function -> functions

ijt: Done.

ijt: I don't see any way to do that. The constructor for `PublisherClient` doesn't take a projectID.

ijt: Done.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: I don't understand the part about the comments. There is no `OpenSubscription` method within the driver.

The semantics described here are not very nice. It would be best if it would fail fast if the subscription doesn't exist, instead of waiting until Receive is called. I think I should change this if possible.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Okay, I removed this logic and added a comment to the driver interface.

Filed #722 to add the retry logic to the concrete API.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Done.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Changed the code to check for the existence of topics and subscriptions at construction time.

ijt:
@vangent, issue https://github.com/google/go-cloud/issues/638 is already filed for adding an example_test for pubsub. The example can be more realistic once @jba adds the ability to add subscriptions to named topics.

ijt:
@vangent and @jba, changes are up, PTAL.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Update: From our offline chat, the plan is to move the retry logic into the concrete API. See https://github.com/google/go-cloud/issues/722.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -20,35 +20,67 @@ import (
...
-// when the client is no longer required.
+// TopicOptions will contain configuration for topics.
+type TopicOptions struct{}
+
+// OpenTopic opens the topic on GCP PubSub for the given projectID and
+// topicName. If the topic does not exist then an error will be returned.
+func OpenTopic(ctx context.Context, client *raw.PublisherClient, proj gcp.ProjectID, topicName string, opts *TopicOptions) (*pubsub.Topic, error) {
+	dt, err := openTopic(ctx, client, proj, topicName)
+	if err != nil {
+		return nil, fmt.Errorf("opening topic: %v", err)
```
jba: We should prefix error messages with the package, e.g. "gcppubsub: opening topic: ..."

Could be another PR.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -57,19 +89,52 @@ func (t *topic) SendBatch(ctx context.Context, dms []*driver.Message) error {
...
+// ReceiveBatch implements driver.Subscription.ReceiveBatch.
 func (s *subscription) ReceiveBatch(ctx context.Context) ([]*driver.Message, error) {
-	req := &pubsubpb.PullRequest{
+	req := &pb.PullRequest{
 		Subscription:      s.path,
 		ReturnImmediately: false,
 	}
 	resp, err := s.client.Pull(ctx, req)
+	if err != nil {
+		return nil, fmt.Errorf("making RPC to receive next batch of messages: %v", err)
```
jba: If we wrap errors like this, it will be impossible to tell if they're retryable.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -80,51 +145,28 @@ func (s *subscription) ReceiveBatch(ctx context.Context) ([]*driver.Message, err
...
-// the ackIDs are sent, an error occurs, or the context is cancelled.
+// SendAcks implements driver.Subscription.SendAcks.
 func (s *subscription) SendAcks(ctx context.Context, ids []driver.AckID) error {
 	var ids2 []string
 	for _, id := range ids {
-		id2, ok := id.(string)
-		if !ok {
-			return fmt.Errorf("gcppubsub driver bug: cast from driver.AckID to string failed on %v", id)
-		}
+		id2 := id.(string)
```
jba: Now you can just inline the type assertion into the following line.

internal/pubsub/driver/driver.go
```diff
@@ -74,9 +74,15 @@ type Topic interface {
...
-	// for the subscription on the server. If no messages are available
-	// yet, it must block until there is at least one, or the context is
-	// done.
+	// for the subscription on the server.
+	//
+	// If there is a transient failure, this method should not retry but
+	// should return a nil slice and an error. The concrete API will take
+	// care of retry logic.
+	//
+	// If the service returns no messages for some other reason, this
```
jba: OK, so the concrete implementation will loop? But then it's important that this method waits for a while before it returns no messages. To be specific, a GCP implementation that set `ReturnImmediately` to true or false would both satisfy this requirement, but the first would cause busy-waiting.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -20,35 +20,67 @@ import (
 	"context"
 	"fmt"
 
-	rawgcppubsub "cloud.google.com/go/pubsub/apiv1"
+	raw2 "cloud.google.com/go/pubsub"
```
jba: Why do you want to use this client? It's too high-level.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Done.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Done.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Done.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: It has methods to check the existence of topics and subscriptions, but maybe this is a case of a little redundancy being better than a little dependency. Removed the import and added some redundant code.

ijt:
@jba, changes are up, PTAL.

PR #712, opened 2018-11-15
pubsub: add conformance tests

### LINKED ISSUE #639: pubsub: add conformance tests

ijt:
Fixes #639 .


internal/pubsub/drivertest/drivertest.go
```diff
@@ -0,0 +1,192 @@
...
+	"testing"
+
+	"github.com/google/go-cloud/internal/pubsub"
+)
+
+// Harness descibes the functionality test harnesses must provide to run
+// conformance tests.
+type Harness interface {
+	// MakeTopicDriver creates a Topic and associated Subscription to test.
+	MakePair() (*pubsub.Topic, *pubsub.Subscription, error)
```
vangent: This should return the driver interface type, not the concrete type. Many tests will use the result and just create the concrete type from it, but some tests may want to test the driver interface explicitly.

internal/pubsub/drivertest/drivertest.go
```diff
@@ -0,0 +1,192 @@
...
+
+	// Close closes resources used by the harness.
```
vangent: Does this close any topics/subscriptions created via `MakePair`? Or is the caller supposed to that independently?

internal/pubsub/drivertest/drivertest.go
```diff
@@ -0,0 +1,192 @@
...
+		t.Fatal(err)
+	}
+
+	// Receive from the subscription.
+	m2, err := sub.Receive(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Check that the received message matches the sent one.
```
vangent: Use `cmp.Diff` on the message objects? Or at least on `.Metadata`.

internal/pubsub/drivertest/drivertest.go
vangent: Do you expect any tests that create more than one pair? For example, a test that creates two subscribers for the same topic.

It might be better to split these (`MakeTopic`, `MakeSubscription`), and provide a way for the test to describe which topic they want. For example, we might require that every implementation provision two topics each with two subscriptions, or whatever, and then these functions would need to describe which one it wants.

vangent: (this could be a TODO or issue for a separate PR)

internal/pubsub/drivertest/drivertest.go
```diff
@@ -0,0 +1,192 @@
...
+	"math/rand"
+	"testing"
+
+	"github.com/google/go-cloud/internal/pubsub"
+)
+
+// Harness descibes the functionality test harnesses must provide to run
+// conformance tests.
+type Harness interface {
+	// MakeTopicDriver creates a Topic and associated Subscription to test.
```
vangent: Pass a `context.Context` to `Make` functions, don't save the `ctx` from `HarnessMaker`.

internal/pubsub/mempubsub/mem_conformance_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"testing"
+	"time"
+
+	"github.com/google/go-cloud/internal/pubsub"
+	"github.com/google/go-cloud/internal/pubsub/drivertest"
+	"github.com/google/go-cloud/internal/pubsub/mempubsub"
+)
+
+type harness struct {
+	ctx context.Context
```
vangent: See comment above; don't save contexts.

internal/pubsub/drivertest/drivertest.go
ijt: Done.

internal/pubsub/drivertest/drivertest.go
ijt: The caller has to do that. I'm actually not sure how useful it is for there to be this Close method on the Harness. I just copied the interface from the blob Harness. Maybe I should get rid of it.

vangent: I think you will need it; see `blob` for why. Close is where you'll close the single `HTTPClient` you make for the test, for record/replay.

I'd just suggest clarifying the comment for the `Make` functions to indicate that the caller should call `Close` on them.

ijt: I removed Harness.Close and added a cleanup closure to makePair.

internal/pubsub/drivertest/drivertest.go
ijt: Filed https://github.com/google/go-cloud/issues/717.

internal/pubsub/drivertest/drivertest.go
ijt: Done.

internal/pubsub/drivertest/drivertest.go
ijt: Done.

internal/pubsub/mempubsub/mem_conformance_test.go
ijt: Done.

ijt:
@vangent, I removed the redundant tests.

internal/pubsub/drivertest/drivertest.go
ijt: Added Harness.Close back in.

internal/pubsub/drivertest/drivertest.go
```diff
@@ -0,0 +1,257 @@
...
+	defer cleanup()
+
+	want := []*pubsub.Message{
+		{Body: []byte("a")},
+		{Body: []byte("b")},
+		{Body: []byte("c")},
+	}
+
+	for _, m := range want {
+		if err := top.Send(ctx, m); err != nil {
```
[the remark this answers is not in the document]
ijt: How about we wait for `check` to become available and then convert everything to it?

internal/pubsub/drivertest/drivertest.go
```diff
@@ -0,0 +1,257 @@
...
+			t.Fatal(err)
+		}
+	}
+
+	var got []*pubsub.Message
+	for i := 0; i < 3; i++ {
+		m2, err := sub.Receive(ctx)
```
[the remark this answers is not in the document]
ijt: Unconvinced.

internal/pubsub/drivertest/drivertest.go
```diff
@@ -0,0 +1,257 @@
...
+		got = append(got, m2)
+	}
+
+	less := func(x, y *pubsub.Message) bool { return bytes.Compare(x.Body, y.Body) < 0 }
+	if diff := cmp.Diff(got, want, cmpopts.SortSlices(less), cmpopts.IgnoreUnexported(pubsub.Message{})); diff != "" {
+		t.Error(diff)
+	}
+}
+
+func testSendReceiveWithMetadata(t *testing.T, newHarness HarnessMaker) {
```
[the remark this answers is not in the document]
ijt: Done.

internal/pubsub/drivertest/drivertest.go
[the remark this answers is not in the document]
jba: That's probably a year out. If you're not comfortable with a "must" helper, that's fine.

internal/pubsub/pubsub.go
```diff
@@ -30,12 +30,13 @@ type Message struct {
 	// Metadata has key/value metadata for the message.
 	Metadata map[string]string
 
-	// AckID identifies the message on the server.
```
jba: Why is this stuff in the conformance test PR?

ijt: I agree it should not be here. The original motivation was that when a conformance test was failing, the diff output was too spammy because of the Subscription pointer in Message. Anyway, I'm going to split that change out into a separate PR for clarity.

ijt:
@vangent, changes are up, PTAL.

ijt:
The build failed on macOS with "The command "go install ./wire/cmd/wire" failed and exited with 1 during ."

ijt:
The build failed on macOS, with this log tail:
 
> ?   	github.com/google/go-cloud/wire	[no test files]
> ?   	github.com/google/go-cloud/wire/cmd/wire	[no test files]
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

PR #715, opened 2018-11-15
wire: give wire.Bind access to the arguments to the injector function

vangent:
Before this change, `wire.Bind` did not have access to the arguments to the injector function.
-- Most conflicts (where the same type is provided from multiple sources) were identified in `buildProviderMap`, but conflicts with injector arguments were special-cased in `solve`.
-- You could not bind injector arguments to an interface type; see the `BindInjectorArg` testcase, in which wire returns an error when you try.

After this change, injector arguments are tracked in `ProviderSet`.
-- All conflicts are now identified in `buildProviderMap`.
-- The `BindInjectorArg` testcase now works and produces the expected output.


wire/internal/wire/testdata/InvalidInjector/foo/wire.go
```diff
@@ -22,7 +22,7 @@ import (
 
 func injectFoo() Foo {
 	// This non-call statement makes this an invalid injector.
-	 _ = 42
```
vangent: Note to reviewers: sorry, this is a random change from running `gofmt`.

wire/internal/wire/parse.go
```diff
@@ -89,10 +93,11 @@ type ProviderSet struct {
...
 
-	Providers []*Provider
-	Bindings  []*IfaceBinding
-	Values    []*Value
-	Imports   []*ProviderSet
+	Providers    []*Provider
+	Bindings     []*IfaceBinding
+	Values       []*Value
+	Imports      []*ProviderSet
+	InjectorArgs *InjectorArgs
```
zombiezen: nit: Add a comment that this is only filled in for `wire.Build`.

wire/internal/wire/analyze.go
```diff
@@ -149,6 +126,19 @@ dfs:
 			ec.add(errors.New(sb.String()))
 			index.Set(curr.t, errAbort)
 			continue
+		case pv.IsArg():
+			src := set.srcMap.At(curr.t).(*providerSetSrc)
```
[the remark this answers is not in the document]
zombiezen: SGTM, but does also sound like a separate change.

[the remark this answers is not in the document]
vangent: Filed https://github.com/google/go-cloud/issues/735.

wire/internal/wire/parse.go

PR #716, opened 2018-11-15
mempubsub: OpenTopic/Subscription return concrete types

### LINKED ISSUE #714: mempubsub: OpenTopic and OpenSubscription should return concrete types
See https://github.com/google/go-cloud/blob/ba6cdfb00c3d227ebaa0243978d0b9860b9f60e7/blob/fileblob/fileblob.go#L98 for a precedent.

We might decide to change how we're doing this, but if we do then it should be done consistently across all the Go Cloud packages.

jba:
Also introduce a `Broker` type to manage named sets of topics and subscribers.

Fixes #714.


internal/pubsub/mempubsub/mem.go
```diff
@@ -37,8 +52,16 @@ type topic struct {
 
 // OpenTopic establishes a new topic.
 // Open subscribers for the topic before publishing.
-func OpenTopic() driver.Topic {
-	return &topic{}
+func (b *Broker) OpenTopic(ctx context.Context, name string) *pubsub.Topic {
```
vangent: Since we decided to not have a `Client` in the "real" impls, and instead have top-level `OpenXXX` functions, WDYT about doing that here as well? I.e., something like:

```
func OpenTopic(b *Broker, name string) *pubsub.Topic
```

internal/pubsub/mempubsub/mem.go
```diff
@@ -22,13 +22,28 @@ package mempubsub
...
 	"sync"
 	"time"
 
+	"github.com/google/go-cloud/internal/pubsub"
 	"github.com/google/go-cloud/internal/pubsub/driver"
 )
 
+type Broker struct {
+	mu     sync.Mutex
+	topics map[string]*topic
```
vangent: To make this more like a real implementation, WDYT about requiring the list of topic names / subscriptions to be provided at broker construction time? (rather than created lazily)?

For example, we might want a conformance test that shows that you can call `OpenTopic` on a topic that doesn't exist, and you get an error when reading.

So, something like:

```NewBroker(topics []string)``` and initialize the map inside it.

File: internal/pubsub/mempubsub/mem.go [hunk 3]
```diff
@@ -86,20 +109,28 @@ type subscription struct {
...
 // OpenSubscription creates a new subscription for the given topic.
-// Unacknowledged messages will become available for redelivery after ackDeadline.
-func OpenSubscription(t driver.Topic, ackDeadline time.Duration) driver.Subscription {
-	tt := t.(*topic)
+func (b *Broker) OpenSubscription(ctx context.Context, topicName string, ackDeadline time.Duration) *pubsub.Subscription {
+	b.mu.Lock()
+	t := b.topics[topicName]
+	b.mu.Unlock()
+	if t == nil {
+		panic(fmt.Sprintf("no topic named %s", topicName))
```
vangent: Don't think this should panic. What would a "real" implementation do? IIUC, it would not know at this time, the `Open` would work fine but then you'd get an error when reading.

internal/pubsub/mempubsub/mem.go
```diff
@@ -37,8 +52,16 @@ type topic struct {
...
 // OpenTopic establishes a new topic.
 // Open subscribers for the topic before publishing.
-func OpenTopic() driver.Topic {
-	return &topic{}
+func (b *Broker) OpenTopic(ctx context.Context, name string) *pubsub.Topic {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	t := newTopic(name)
+	b.topics[name] = t
+	return pubsub.NewTopic(ctx, t)
```
vangent: Side note: Does `pubsub.NewTopic` really need a `ctx`? For blob it doesn't.

ijt: Yes, it needs a ctx because it spins up a bundler whose handler uses the ctx to make RPCs. If we inject the bundler as you suggested then NewTopic will no longer need the ctx arg.

jba: See #733.

internal/pubsub/mempubsub/mem.go
jba: Done.

internal/pubsub/mempubsub/mem.go
jba: Wouldn't that make it less like a real implementation? I don't know of any service that makes you list all the topics you will use when you connect to it.

But I agree with the idea because it makes this implementation more useful for testing, which IMO is its main purpose.

vangent:
Please assign back to reviewers when it's ready for review again.

internal/pubsub/mempubsub/mem.go
jba: Done.

internal/pubsub/example_test.go
```diff
@@ -28,11 +28,10 @@ import (
 func Example_sendReceive() {
 	// Open a topic and corresponding subscription.
 	ctx := context.Background()
-	dt := mempubsub.OpenTopic()
-	ds := mempubsub.OpenSubscription(dt, time.Second)
-	t := pubsub.NewTopic(dt)
+	b := mempubsub.NewBroker([]string{"t"})
```
vangent: Nit: for an example, I suggest using a more descriptive name, like "MyTopic" or something like that.

internal/pubsub/mempubsub/mem.go
```diff
@@ -37,8 +58,8 @@ type topic struct {
 
 // OpenTopic establishes a new topic.
 // Open subscribers for the topic before publishing.
-func OpenTopic() driver.Topic {
-	return &topic{}
+func OpenTopic(b *Broker, name string) *pubsub.Topic {
+	return pubsub.NewTopic(b.topic(name))
```
vangent: So this will call `pubsub.NewTopic(nil)` if you provide a nonexistent topic name, right? How/when will this fail? (Same question for `OpenSubscription` below).

To be most similar to "real" implementations, my guess would be that `OpenTopic|Subscription` should succeed, but that reading/writing from the result always fail. Or, are we asking driver implementations to do an RPC during `Open`? Either way, the driver interface should document the expected behavior (and the conformance test should test it).

jba: I opened #759 to discuss.

internal/pubsub/example_test.go
vangent: Push?

jba: Waiting for resolution on the other comment before I push. I guess I should not have commented 'Done' just because I made the change locally.

vangent: GO ahead and push; we can update the code later to match #759 if needed.

jba: pushed

Issue #723, opened 2018-11-16
wire: support function arguments as providers


vangent:
This doesn't work:

```
func injectBar(fn func() *Foo) *Bar {
	panic(wire.Build(fn, NewBar))
}
```

It currently panics. I have an easy fix for the panic, but it then results in:

```
var fn func() *example.com/foo.Foo is not a provider or a provider set
```

`fn` should be usable as a provider for `Foo`/`*Foo`.

zombiezen:
I don't know if I understand the rule that's being introduced here. Is your idea to support using any unnamed function type that is introduced into the dependency graph as a provider? Or just specifically passing arguments to `wire.Build`? I'm not sure I see the benefit, even with reviewing #690.

vangent:
A provider is just a function no? From the `README`, "a function that can produce a value". It seems like `wire.Build` should therefore accept various forms of such functions to be used in the provider graph. Why is `fn` in the example above not usable? I'm not sure exactly what "unnamed" means -- its name in this context is `fn`. What's the rationale for not allowing it?

In terms of the benefit, I tried this out as a workaround for #44. `wire.Bind`  does not allow methods to be used as providers, and this would be a way to get around that (by passing the method to the injector function and using it that way).

When I tried it it panicked, which is clearly not good. The first PR I sent fixes the panic and saves the error `wire` produces.

vangent:
I'm hoping to fix the panic for this sprint; adding real support for this is a separate discussion.

vangent:
#724 fixed the panic and recorded the current error. Deferring till next sprint for more work.

PR #724, opened 2018-11-16
wire: add a test for using a function argument as a provider

vangent:
This adds a test for using a function argument as a provider in an injector.

Before the PR, this test would panic:

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
```

at the updated line. After the PR, wire still fails, but the test captures the error and passes. Fixing this to actually work is left as a future PR.

Updates #723.


zombiezen:
I don't think I understand #723 fully. Let's discuss there.

vangent:
I responded on #723.

PR #728, opened 2018-11-16
pubsub: add example_test.go

### LINKED ISSUE #638: pubsub: add example_test

ijt:
Fixes #638


internal/pubsub/example_test.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+func ExampleSendReceive() {
+	// Open a topic and corresponding subscription.
+	ctx := context.Background()
+	dt := mempubsub.OpenTopic()
+	ds := mempubsub.OpenSubscription(dt, time.Second)
+	t := pubsub.NewTopic(ctx, dt)
+	s := pubsub.NewSubscription(ctx, ds)
+
```
vangent: Need to `Close` both of these....

internal/pubsub/example_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"context"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/google/go-cloud/internal/pubsub"
+	"github.com/google/go-cloud/internal/pubsub/mempubsub"
+)
+
+func ExampleSendReceive() {
```
vangent: IIUC, this should just be called `Example`. Or `Example_sendReceive`.

It would be good to add an example with a more realistic `Receive` loop too, can be separate PR if you prefer.

internal/pubsub/example_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// Receive a message from the subscription.
+	m, err := s.Receive(ctx)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Print out the received message.
+	fmt.Printf("%s\n", m.Body)
+
```
jba: Ack the message.

jba:
Let's be really careful about stepping on each other's toes. I had written an almost identical example that Robert asked me to remove from my PR and put at top level. That was on my to-do list.

ijt:
@jba, sorry, I didn't mean to step on your toes!

I think the easiest way to prevent toe-stepping will be for us to put all the things we're working on into the issues list. Github is pretty good about surfacing duplicates, so if we're consistent about tracking things with it then I think we'll avoid doing too much redundant work.

internal/pubsub/example_test.go
ijt: Changed the test name.

Filed #738.

internal/pubsub/example_test.go
ijt: Done.

ijt:
Changes are up, PTAL.

internal/pubsub/example_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	// The messages may be received in a different order than they were
+	// sent.
+	sort.Slice(ms2, func(i, j int) bool { return string(ms2[i].Body) < string(ms2[j].Body) })
+
+	// Print out and acknowledge the received messages.
+	for _, m2 := range ms2 {
+		fmt.Printf("%s\n", m2.Body)
+		m2.Ack()
```
vangent: How about moving `Ack` to the first loop?

Then you could save a `[]string]' instead of a `[]*pubsub.Message`, and use `sort.Strings` and `fmt.Println`.

ijt: Done.

internal/pubsub/example_test.go
```diff
@@ -0,0 +1,105 @@
...
+		if err != nil {
+			log.Fatal(err)
+		}
+		ms2 = append(ms2, string(m2.Body))
+		m2.Ack()
+	}
+
+	// The messages may be received in a different order than they were
+	// sent.
+	sort.Slice(ms2, func(i, j int) bool { return ms2[i] < ms2[j] })
```
vangent: You can use `sort.Strings(ms2)` now.

internal/pubsub/example_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// Print out and acknowledge the received messages.
+	for _, m2 := range ms2 {
+		fmt.Printf("%s\n", m2)
```
vangent: Nit: `fmt.Println(m2)`.

internal/pubsub/example_test.go
ijt: Just saw that and pushed the change.

internal/pubsub/example_test.go
ijt: Done.

Issue #505, opened 2018-10-09
wire: create injectors in _test files


zombiezen:
Currently, Wire only considers the non-test files for injectors. To enable creation of test-only injectors, it would be nice to have Wire evaluate test files as well, creating a separate `wire_gen_test.go` file for the generated code evaluated for the tests.

PR #729, opened 2018-11-16
wire: support multiple packages in Generate

### LINKED ISSUE #726: wire: generate should support multiple packages
Right now it fails if more than one package is provided. This is kind of OK because most users do `go generate`, which will call Wire with a single package at a time, but it would be nice to call Wire explicitly with a pattern.

E.g., at the root of go-cloud:

```
$ wire gen ./...
load: got 49 packages
wire: generate failed
```

--> That should work.

vangent:
Fixes #726.


wire/internal/wire/wire.go
```diff
@@ -63,34 +64,37 @@ func (gen GeneratedFile) Commit() error {
...
-	}
-	copyNonInjectorDecls(g, injectorFiles, pkgs[0].TypesInfo)
-	goSrc := g.frame()
-	fmtSrc, err := format.Source(goSrc)
-	if err != nil {
-		// This is likely a bug from a poorly generated source file.
-		// Return an error and the unformatted source.
-		return GeneratedFile{Path: outFname, Content: goSrc}, []error{err}
+	generated := make([]GeneratedFile, 0, len(pkgs))
+	ec := new(errorCollector)
```
vangent: Alternatively, we could always return exactly one `GeneratedFile` per package, remove the `errors` return value from this function, and add an `Err` field to `GeneratedFile`. That would make it easier to identify which package each error was for. Agree?

zombiezen: I mostly agree, but I think we would need to shift to returning package-centric things. I don't think we can entire drop the separate error return, because errors in unrelated packages or the load process could cause a failure that's not logically related to the other packages.

wire/internal/wire/wire_test.go
```diff
@@ -88,7 +88,14 @@ func TestWire(t *testing.T) {
 				t.Fatal(err)
 			}
 			wd := filepath.Join(gopath, "src", "example.com")
-			gen, errs := Generate(ctx, wd, append(os.Environ(), "GOPATH="+gopath), test.pkg)
+			gens, errs := Generate(ctx, wd, append(os.Environ(), "GOPATH="+gopath), []string{test.pkg})
+			if len(gens) > 1 {
```
zombiezen: It could be interesting to permit this (for testing multiple-package generation or #505), but we should file an issue separately.

wire/internal/wire/wire.go
vangent: OK, how about changing `Generate` to return a slice of `GenerateResult` plus an error? `GenerateResult` is just `GeneratedFile` renamed, plus a slice of errors.

wire/internal/wire/wire_test.go
vangent: Ack.

wire/internal/wire/wire.go
zombiezen: SGTM

vangent: PTAL.

wire/internal/wire/wire.go
```diff
@@ -38,59 +38,74 @@ import (
...
-	outDir, err := detectOutputDir(pkgs[0].GoFiles)
-	if err != nil {
-		return GeneratedFile{}, []error{fmt.Errorf("load: %v", err)}
-	}
-	outFname := filepath.Join(outDir, "wire_gen.go")
-	g := newGen(pkgs[0])
-	injectorFiles, errs := generateInjectors(g, pkgs[0])
+//
+// Generate may return one or more errors if it failed to load the packages.
+func Generate(ctx context.Context, wd string, env []string, patterns []string) ([]*GenerateResult, []error) {
```
zombiezen: nit: return `[]GenerateResult`, not `[]*GenerateResult`. No need for indirection.

```suggestion
func Generate(ctx context.Context, wd string, env []string, patterns []string) ([]GenerateResult, []error) {
```

PR #736, opened 2018-11-16
travis: run coverage with explicit -coverpkg

vangent:


vangent:
Update: it turns out `-coverpkg=./...` does the same thing, although it doesn't filter out the sample and test packages. I've updated the PR to use it. The top-level coverage number will go down because of the new packages being included, but I think that's fine.

PR #737, opened 2018-11-16
postgres/cloudpostgres: use structured parameters instead of connection string

zombiezen:
Brings consistency with the MySQL dialers and avoids pushing the user to escape usernames and passwords properly.

### DIFF
--- a/postgres/cloudpostgres/cloudpostgres.go
+++ b/postgres/cloudpostgres/cloudpostgres.go
@@ -21,6 +21,7 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"errors"
+	"fmt"
 	"net"
 	"net/url"
 	"time"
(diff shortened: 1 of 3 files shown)


postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -30,21 +31,59 @@ import (
...
+	Values url.Values
 }
 
 // Open opens a Cloud SQL database.
 func Open(ctx context.Context, certSource proxy.CertSource, params *Params) (*sql.DB, error) {
+	vals := make(url.Values)
+	for k, v := range params.Values {
+		// Only permit parameters that do not conflict with other behavior.
+		if k != "user" && k != "password" && k != "dbname" && k != "host" && k != "port" && k != "sslmode" && k != "sslcert" && k != "sslkey" && k != "sslrootcert" {
+			vals[k] = v
```
ijt: Shouldn't an error be returned if k is one of these?

zombiezen: Good idea. Done.

postgres/cloudpostgres/cloudpostgres_test.go
```diff
@@ -71,3 +75,62 @@ func TestOpen(t *testing.T) {
 		t.Error("Close:", err)
 	}
 }
+
+func TestOpenBadValue(t *testing.T) {
+	// This test will be skipped unless the project is set up with Terraform.
+
+	tfOut, err := terraform.ReadOutput(".")
+	if err != nil {
+		t.Skipf("Could not obtain harness info: %v", err)
```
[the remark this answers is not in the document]
zombiezen: My intent was to give it otherwise valid values, but only altering the parameters given to the function. I understand that in the current implementation of the function, it would abort earlier in the process, but I'm worried about more the case of the system under test not checking for these conditions, where it would produce an error for unrelated reasons.

PR #743, opened 2018-11-16
mysql: remove Options struct, move the TraceOpts field into Params

ijt:
This implements @zombiezen 's suggestion for #608.


zombiezen:
LGTM, I suspect that this would break guestbook? I remembered there being other related changes.

ijt:
I updated guestbook. PTAL.

PR #747, opened 2018-11-17
blob: bucket.List does not need a ctx, nor can it fail

### LINKED ISSUE #746: blob: List doesn't need a ctx, nor does it need to return an error
It just creates a `ListIterator` struct; everything interesting happens in that struct's `Next` method.

vangent:
Fixes #746.


vangent:
> have you considered and rejected the idea of holding the 
> ctx in the iterator so users don't have to provide it to Next?

Yes, because I think we want them to be able to provide a different timeout for each `Next` call.

PR #748, opened 2018-11-17
blob/all: use driver interface functions for identifying error types, and wrap all errors

vangent:
This is a big PR. I'm happy to split it up if we can figure out a way to do so, but I wanted to get some feedback on what I'm doing.

-- blob is hardly wrapping any errors, which is a bug. It now wraps all errors returned from the driver, and checks that it doesn't double-wrap them. I've also added tests (using a driver impl that always returns errors) to verify the wrapping.

-- I got rid of the `blob.Driver.Error` type, and added `IsNotExist` and `IsNotImplemented` to the driver interface instead. Now drivers can just return raw errors. The concrete type wraps them before giving them back to the user, and then unwraps them if the user calls `blob.IsNotExist`, in order to pass the raw error back to the driver.

-- The trickiest thing here, which I'm not sure if I've handled correctly, is the errors returned from `driver.Reader` (`Read` and `Close`) and `driver.Writer` (`Write` and `Close`). I've mostly handled them the same as the others, wrapping them, but the `io` package also inserts errors like `io.EOF` sometimes. I'm checking for `io.EOF` in my `wrapError` function, but it feels fragile. Suggestions welcome. Maybe some reflection to identify the package, and don't wrap things from `io`?


blob/blob.go
```diff
@@ -588,19 +601,43 @@ func Open(ctx context.Context, urlstr string) (*Bucket, error) {
 	return NewBucket(drv), nil
 }
 
+// wrappedError is used to wrap all errors returned by drivers so that users
+// are not given access to provider-specific errors.
+type wrappedError struct {
```
jba: This seems quite reasonable. I don't think your EOF check is fragile; that is the only pseudo-error in this part of the standard library, so it makes sense to special-case it.

How do you feel about adding an `Unwrap` method? Or maybe an `As` method? That will provide an escape hatch for provider-aware error testing. See https://go.googlesource.com/proposal/+/master/design/go2draft-error-inspection.md.

blob/blob.go
```diff
@@ -588,19 +601,43 @@ func Open(ctx context.Context, urlstr string) (*Bucket, error) {
...
+}
+
+func wrapError(b driver.Bucket, err error) error {
+	if err == nil {
+		return nil
+	}
+	if err == io.EOF {
+		return err
+	}
+	if _, ok := err.(*wrappedError); ok {
```
jba: May I suggest, instead of trying hard to avoid this, just cut yourself some slack and return `err`. I think that will make the implementation more robust overall.

vangent: Yea. Done. I updated the tests to check for double-wrapping instead.

blob/blob.go
vangent: Good idea! I filed #750.

PR #751, opened 2018-11-17
internal/retry: a package for retrying calls

### LINKED ISSUE #730: all: support RPC retry
We should establish a common pattern for retrying RPCs. It can't be done completely in the concrete implementations because different providers will have different notions of what is retryable.

@vangent suggested:

- Drivers should provide an `IsRetryable` function or method that takes an error and returns a bool and, if true, a duration. (The duration is to support providers whose RPCs return sleep hints.)

- There should be common code in `internal` that implements exponential backoff.

Alternative designs:

- _Have drivers be responsible for retry._ This puts too much burden on driver implementers.

- _Wrap errors._ Return errors that are annotated or wrapped to indicate whether they are retryable. E.g. if an error has an `IsRetryable` method that returns true, it is retryable; otherwise assume it isn't. This also makes it too hard and error-prone to write a driver, since the wrapping has to be done at every error return.

jba:
Fixes #730.

### DIFF
--- a/internal/retry/retry.go
+++ b/internal/retry/retry.go
@@ -1,0 +1,78 @@
+// Copyright 2018 The Go Cloud Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package retry provides retry logic.
+package retry
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	gax "github.com/googleapis/gax-go"
+)
+
+// Call calls the supplied function f repeatedly, using the isRetryable function and
+// the provided backoff parameters to control the repetition.
+//
+// When f returns nil, Call immediately returns nil.
+//
+// When f returns an error for which isRetryable returns false, Call immediately
+// returns that error.
+//
+// When f returns an error for which isRetryable returns true, Call sleeps for the
+// duration provided by isRetryable, or the provided backoff value if that duration
+// is zero, and invokes f again.
+//
+// When the provided context is done, Retry returns with an error that
+// includes both ctx.Error() and the last error returned by f, if any.
+func Call(ctx context.Context, bo gax.Backoff, isRetryable func(error) (bool, time.Duration), f func() error) error {
+	return call(ctx, bo, isRetryable, f, gax.Sleep)
+}
+
+// Split out for testing.
+func call(ctx context.Context, bo gax.Backoff, isRetryable func(error) (bool, time.Duration), f func() error,
+	sleep func(context.Context, time.Duration) error) error {
+	// Do nothing if context is done on entry.
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	for {
+		err := f()
+		if err == nil {
+			return nil
+		}
+		retry, dur := isRetryable(err)
+		if !retry {
+			return err
+		}
+		if dur == 0 {
+			dur = bo.Pause()
+		}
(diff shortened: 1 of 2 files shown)


internal/retry/retry.go
```diff
@@ -0,0 +1,78 @@
...
+// When f returns nil, Call immediately returns nil.
+//
+// When f returns an error for which isRetryable returns false, Call immediately
+// returns that error.
+//
+// When f returns an error for which isRetryable returns true, Call sleeps for the
+// duration provided by isRetryable, or the provided backoff value if that duration
+// is zero, and invokes f again.
+//
+// When the provided context is done, Retry returns with an error that
```
vangent: I would suggest "an error" -> "a ContextError" to make it more clear that it's a special error type; when I first read this I was confused by what it meant for an error to include 2 other errors.

internal/retry/retry.go
```diff
@@ -0,0 +1,78 @@
...
+func Call(ctx context.Context, bo gax.Backoff, isRetryable func(error) (bool, time.Duration), f func() error) error {
+	return call(ctx, bo, isRetryable, f, gax.Sleep)
+}
+
+// Split out for testing.
+func call(ctx context.Context, bo gax.Backoff, isRetryable func(error) (bool, time.Duration), f func() error,
+	sleep func(context.Context, time.Duration) error) error {
+	// Do nothing if context is done on entry.
+	if err := ctx.Err(); err != nil {
+		return err
```
vangent: Why not `ContextError` here?

internal/retry/retry.go
```diff
@@ -0,0 +1,78 @@
...
+// When f returns an error for which isRetryable returns false, Call immediately
+// returns that error.
+//
+// When f returns an error for which isRetryable returns true, Call sleeps for the
+// duration provided by isRetryable, or the provided backoff value if that duration
+// is zero, and invokes f again.
+//
+// When the provided context is done, Retry returns with an error that
+// includes both ctx.Error() and the last error returned by f, if any.
+func Call(ctx context.Context, bo gax.Backoff, isRetryable func(error) (bool, time.Duration), f func() error) error {
```
vangent: Are there multiple options for gax.Backoff? (i.e., is it necessary to expose this, or could it be created internally?)

internal/retry/retry.go
```diff
@@ -0,0 +1,78 @@
...
+			return &ContextError{CtxErr: cerr, FuncErr: err}
+		}
+	}
+}
+
+// A ContextError contains both a context error (either context.Canceled or
+// context.DeadlineExceeded), and the last error from the function being retried.
+type ContextError struct {
+	CtxErr  error // The error obtained from ctx.Err()
+	FuncErr error // The error obtained from the function being retried
```
vangent: Maybe note that it could be `nil`.

internal/retry/retry_test.go
```diff
@@ -0,0 +1,159 @@
...
+var (
+	errRetry   = errors.New("retry")
+	errNoRetry = errors.New("no retry")
+)
+
+func retryable(err error) (bool, time.Duration) {
+	if err == errRetry {
+		return true, retryDur
+	}
+	return false, 0
```
vangent: What about a test for retryable returning a non-zero `Duration`?

internal/retry/retry.go
jba: There are. I think it makes sense to expose it to our concrete implementations, since we may choose to use different values. Since this is an internal package, we can change that decision without breaking anyone.

internal/retry/retry.go
jba: There is no error from the function at this point, and so no need to return an additional error.

internal/retry/retry.go
jba: It cannot be in my design.

internal/retry/retry_test.go
jba: Those tests exist below. Look for `wantRetryableDur`.

internal/retry/retry.go
jba: Maybe you're right to make to make things more uniform, and always return a `ContextError`. I'll make that change in the new PR.

internal/retry/retry.go
jba: Done in new PR.

PR #757, opened 2018-11-19
internal/pubsub: support retry

jba:
- Add IsRetryable to driver package and all implementations.

- Add retry logic to concrete implementation.

- Test retry of SendBatch and ReceiveBatch. Retry of SendAcks is left
  for a future PR.


internal/pubsub/driver/driver.go
```diff
@@ -70,6 +73,12 @@ type Topic interface {
 	// then the call to Close should proceed and the call to SendBatch
 	// should fail immediately after Close returns.
 	Close() error
+
+	// IsRetryable should report whether err can be retried.
+	// err will always be a non-nil error returned from SendBatch.
+	// If its second return value is non-zero, it is the suggested amount
+	// of time to sleep before retrying.
+	IsRetryable(err error) (bool, time.Duration)
```
vangent: Suggest expanding to note that returning 0 is fine, a backoff will be used.

Do we even need to expose this? (i.e., why not always use backoff?).

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -175,18 +178,39 @@ func (s *subscription) SendAcks(ctx context.Context, ids []driver.AckID) error {
...
+}
+
+// Copied from cloud.google.com/go/pubsub.
+func isRetryable(err error) bool {
+	s, ok := status.FromError(err)
+	if !ok {
+		return true
+	}
+	switch s.Code() {
+	case codes.DeadlineExceeded, codes.Internal, codes.ResourceExhausted:
```
vangent: How long is the list of codes? Since subscriptions/topics fail hard for isRetryable=false, I'm wondering if a blacklist instead of a whitelist is more appropriate. E.g., if they add a new error code, we should default to retrying.

internal/pubsub/driver/driver.go
jba: Some services might provide a suggested delay. E.g. see https://code.googlesource.com/gocloud/+/master/spanner/retry.go#130. But maybe this is rare enough that we can just ignore it? It is worth noting that the [gRPC retry proposal](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#retry-policy-capabilities) doesn't allow for custom delays.

internal/pubsub/gcppubsub/gcppubsub.go
jba: The default should not be to retry. Failing fast is safer, because RPCs are not idempotent in general. That is why everyone gives whitelists for retryable codes. E.g. see the gRPC design I linked above.

vangent: Interesting. Could/should there be a different default for `SendAck` + `ReceiveBatch` (which **are** idempotent, or roughly so) than for `SendBatch`, which is not?

Failing fast doesn't seem safer to me for an application that is expected to exit/crash when our API returns an error, and retry invisibly for retryable errors.

jba: Ugh. You reminded me that the GCP client does retries (and in fact uses different codes for different RPCs), so there's nothing to do. I removed this function.

But to continue the discussion: by failing fast being safer, I'm thinking of the case where you have a non-idempotent operation, like incrementing something. If there is an error code that leaves it ambiguous whether the operation actually happened, it would be better not to retry, rather than to do something wrong (double-count).

internal/pubsub/driver/driver.go
jba: WDYT? Happy to remove it.

internal/pubsub/pubsub.go
```diff
@@ -206,8 +217,12 @@ func NewSubscription(d driver.Subscription) *Subscription {
 			id := m.ackID
 			ids = append(ids, id)
 		}
+		callCtx := context.TODO()
+		err := retry.Call(callCtx, gax.Backoff{}, d.IsRetryable, func() error {
+			return d.SendAcks(callCtx, ids)
+		})
 		// TODO(#695): Do something sensible if SendAcks returns an error.
```
ijt: Remove the TODO?

internal/pubsub/pubsub.go
```diff
@@ -206,8 +217,12 @@ func NewSubscription(d driver.Subscription) *Subscription {
 			id := m.ackID
 			ids = append(ids, id)
 		}
+		callCtx := context.TODO()
```
ijt: Should there be an issue to go with this context.TODO? Do you really mean context.Background?

internal/pubsub/driver/driver.go
ijt: Let's keep it simple for now.

internal/pubsub/pubsub.go
jba: We still have to make a failed SendAcks affect Receive and Close.

internal/pubsub/driver/driver.go
jba: Done.

internal/pubsub/gcppubsub/gcppubsub.go
jba: > what else could they do?

Alert a human. Automatically retrying non-idempotent operations introduces silent corruption into the system.

internal/pubsub/pubsub.go
jba: Done.

internal/pubsub/gcppubsub/gcppubsub.go
vangent: Isn't `ReceiveBatch` idempotent?

ijt: I believe it is not. For example if there is one message in the subscription, and you call `ReceiveBatch` then it will pull that message down and it will be marked on the server as being processed. Now say another message shows up on the server and you call `ReceiveBatch` again. Then the new message gets marked as being processed. So that's not idempotent. It's time-dependent.

Issue #750, opened 2018-11-17
blob: add 'As' for returned errors


vangent:
Per @jba on #748 : that will provide an escape hatch for provider-aware error testing. See go.googlesource.com/proposal/+/master/design/go2draft-error-inspection.md.

PR #763, opened 2018-11-19
blob/all: consistently return raw errors from drivers instead of creating new ones

vangent:
Updates #750.

This is needed to make sure `As` works correctly.

### DIFF
--- a/blob/fileblob/fileblob.go
+++ b/blob/fileblob/fileblob.go
@@ -86,10 +86,10 @@ func openBucket(dir string, _ *Options) (driver.Bucket, error) {
 	dir = filepath.Clean(dir)
 	info, err := os.Stat(dir)
 	if err != nil {
-		return nil, fmt.Errorf("open file bucket: %v", err)
+		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


blob/fileblob/fileblob.go
```diff
@@ -86,10 +86,10 @@ func openBucket(dir string, _ *Options) (driver.Bucket, error) {
 	dir = filepath.Clean(dir)
 	info, err := os.Stat(dir)
 	if err != nil {
-		return nil, fmt.Errorf("open file bucket: %v", err)
+		return nil, err
 	}
 	if !info.IsDir() {
-		return nil, fmt.Errorf("open file bucket: %s is not a directory", dir)
+		return nil, fmt.Errorf("%s is not a directory", dir)
```
jba: I don't see how removing "open file bucket" helps your purpose.

vangent: It doesn't really, but I was trying to make the error messages more consistent. They used to all have this kind of prefix, or at least mostly, but now that we're usually returning raw errors, they don't.

PR #767, opened 2018-11-19
doc: Update design doc section on errors

### LINKED ISSUE #749: doc: add a section to the design doc about errors
-- Wrap errors so that users can't peek into the provider-specific error details. Don't double-wrap.
-- All errors should be prefixed with the package name (e.g., `blob: fail`); we can do this as part of the wrapping.
-- For cases where we want to distinguish types of errors (e.g., `blob` has `NotExists`), the driver interface should have a function to do the check (e.g., `IsNotExists(err error) bool`. The concrete type should pass such functions the raw error that the driver generated (not a wrapped error).

vangent:
It looks like a accidentally checked in work-in-progress on this as part of another PR (#763, the first delete block); sorry about that.

Fixes #749.


internal/docs/design.md
```diff
@@ -109,40 +109,50 @@ canonical example is `gcpkms` and `awskms`.
...
 -   The callee is expected to return `error`s with messages that include
     information about the particular call, as opposed to the caller adding this
     information. This aligns with common Go practice.
 
+### Drivers
+
+Driver implementations should:
+
+-   Return the raw errors from the underlying provider, and not wrap them in
+    `fmt.Errorf` calls, so that they can be exposed to end users via `As`.
```
jba: `ErrorAs`?

internal/docs/design.md
```diff
@@ -109,40 +109,50 @@ canonical example is `gcpkms` and `awskms`.
...
+    a driver interface function to do the check. The concrete type must pass
+    such functions the raw error that the driver returned (not a wrapped error).
+
+-   Expose an `ErrorAs` function to allow users to access provider-specific
+    error types.
+
+-   Handle transient network errors. Retry logic is best handled as low in the
+    stack as possible to avoid [cascading failure][]. APIs should try to surface
+    "permanent" errors (e.g. malformed request, bad permissions) where
+    appropriate so that application logic does not attempt to retry idempotent
```
jba: s/idempotent/non-idempotent/

PR #768, opened 2018-11-20
internal/batcher: add Shutdown

jba:
Shutdown can be called instead of bundler.Flush in Close methods.


internal/batcher/batcher_test.go
```diff
@@ -112,31 +112,28 @@ func TestSaturation(t *testing.T) {
 func TestShutdown(t *testing.T) {
 	ctx := context.Background()
 	var nAdds int64 // atomic
-	b := New(int(0), 10, func(x interface{}) error {
+	c := make(chan int, 10)
+	b := New(int(0), cap(c), func(x interface{}) error {
```
ijt: Is it possible to change the package to `batcher_test` so this would be `batcher.New`?

internal/batcher/batcher_test.go
```diff
@@ -112,31 +112,28 @@ func TestSaturation(t *testing.T) {
...
+		for range x.([]int) {
+			c <- 0
+		}
 		atomic.AddInt64(&nAdds, int64(len(x.([]int))))
-		// Make sure some Adds are active on Shutdown.
-		time.Sleep(100 * time.Millisecond)
+		time.Sleep(time.Second) // we want Adds to be active on Shutdown
```
ijt: How do we know one second is enough time?

internal/batcher/batcher_test.go
ijt: Sorry, I only saw that this PR had been merged after I sent the comments.

jba: I did this in #791.

internal/batcher/batcher_test.go
jba: I can't be absolutely sure, but that's more than enough time to execute a couple of lines of code.

I couldn't figure out how to make this test time-independent.

PR #773, opened 2018-11-20
pubsub: replace Message.sub with a closure called ack

### LINKED ISSUE #734: pubsub: Message.sub field causes spammy debug output
This field is only being used to add acks to the subscription's batcher, so we could replace it with a closure that does that.

Alternatively, we could add a String() method to Message for nicer output. This doesn't seem quite as flexible though, since it can't do different things for %v vs %+v.

ijt:
Fixes #734


internal/pubsub/pubsub.go
```diff
@@ -193,23 +188,28 @@ func (s *Subscription) Close() error {
...
 // It is for use by provider implementations.
 func NewSubscription(d driver.Subscription) *Subscription {
 	handler := func(item interface{}) {
-		ms := item.([]*Message)
+		boxes := item.([]ackIDBox)
 		var ids []driver.AckID
-		for _, m := range ms {
-			id := m.ackID
+		for _, box := range boxes {
+			id := box.ackID
```
jba: Just inline `box.ackID` below.

internal/pubsub/pubsub.go
```diff
@@ -193,23 +188,28 @@ func (s *Subscription) Close() error {
 	return s.driver.Close()
 }
 
+// ackIDBox makes it possible to use a driver.AckID with bundler.
+type ackIDBox struct {
```
jba: Ah, I think I see why you need this. Unfortunate.

ijt: Yep.

internal/pubsub/pubsub.go
ijt: Wish I could.

jba: I'm sorry. You're saying that what you have compiles, but
```
ids = append(ids, box.ackID)
```
does not? Can you tell me what error you get from that line?

ijt: There's a closure that refers to id. If it refers to the loop variable then it will get the wrong value sometimes. It will still compile.

jba: Ah, right, thanks.

PR #777, opened 2018-11-20
blob: replace NewRangedReader with NewReader + ReaderOptions

### LINKED ISSUE #765: blob: should NewReader take Options and support chunking?
Currently `blob` exposes:

-- `NewRangeReader`: takes offset, len and calls driver.
-- `NewReader`: calls `NewRangeReader` with offset=0, len=-1 (all).

To avoid reading the entire blob into memory, the caller would have to implement their own chunking.

Perhaps we should just have `NewReader`, taking `ReaderOptions`, where `ReaderOptions` has:
-- offset
-- length
-- chunk size (defaulting to something large but reasonable)

Then the concrete type would do chunking for files larger than the chunk size.

vangent:
This simplifies the concrete type, and also adds `ReaderOptions` where we can add more things later.

Fixes #765.


blob/blob.go
```diff
@@ -505,6 +503,15 @@ type SignedURLOptions struct {
 	Expiry time.Duration
 }
 
+// ReaderOptions controls Reader behaviors.
+type ReaderOptions struct {
+	// ReadLength sets the total number of bytes to read. 0 is interpreted as
+	// no limit. Must be >= 0.
+	ReadLength int64
```
jba: nit: why not just Length? It's not like anyone is going to be confused by what that means, since it's in a ReaderOptions struct and you document it. Also, you don't say ReadOffset.

blob/driver/driver.go
```diff
@@ -22,6 +22,15 @@ import (
 	"time"
 )
 
+// ReaderOptions controls Reader behaviors.
+type ReaderOptions struct {
```
jba: OK, so drivers import the concrete package so they can call the constructor, like blob.NewBucket. Why couldn't they also get ReaderOptions from the concrete package, instead of having to duplicate it?

vangent: The concrete type imports the driver package, so the driver package can't import the concrete type without being circular. It's true that driver implementations import both, but the `ReaderOptions` need to be part of the interface.

In the past we've considered using type aliases instead of duplication, but there are enough cases where the options vary somewhat, and sometimes the documentation differs as well ("thou shalt" for drivers, "it will" for concrete type), so we've settled on this.

blob/blob.go

blob/driver/driver.go
```diff
@@ -213,12 +222,12 @@ type Bucket interface {
 	// opts is guaranteed to be non-nil.
 	ListPaged(ctx context.Context, opts *ListOptions) (*ListPage, error)
 
-	// NewRangeReader returns a Reader that reads part of an object, reading at
-	// most length bytes starting at the given offset. If length is negative, it
+	// NewReader returns a Reader that reads an object, reading at
+	// most opts.Length bytes starting at opts.Offset. If opts.Length is 0, it
```
shantuo: Should we add that implementation can expect both offset and length be >= 0?

vangent: Done (in `driver.ReaderOptions`).

zombiezen:
Updates (or fixes?) #765

vangent:
> Updates (or fixes?) #765

Yes, fixes. Updated description.

vangent:
We've decided to not do this and instead add `ReaderOptions` to the existing 2 functions; see #765 for discussion. I'm closing this PR and will open a new one.

PR #788, opened 2018-11-21
make Example_sendReceiveMultipleMessages slightly more realistic

### LINKED ISSUE #738: pubsub: add example of more realistic receive loop to example_test.go

ijt:
Fixes #738

### DIFF
--- a/internal/pubsub/example_test.go
+++ b/internal/pubsub/example_test.go
@@ -57,7 +57,8 @@ func Example_sendReceive() {
 
 func Example_sendReceiveMultipleMessages() {
 	// Open a topic and corresponding subscription.
-	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
 	b := mempubsub.NewBroker([]string{"myTopic"})
 	t := mempubsub.OpenTopic(b, "myTopic")
 	defer t.Close()
@@ -78,8 +79,11 @@ func Example_sendReceiveMultipleMessages() {
 
 	// Receive messages from the subscription.
 	ms2 := []string{}
-	for i := 0; i < len(ms); i++ {
+	for {
 		m2, err := s.Receive(ctx)
+		if err == context.DeadlineExceeded {
+			break
+		}
 		if err != nil {
 			log.Fatal(err)
 		}


internal/pubsub/example_test.go
```diff
@@ -78,8 +79,11 @@ func Example_sendReceiveMultipleMessages() {
 
 	// Receive messages from the subscription.
 	ms2 := []string{}
```
jba: Can we use
```
var ms2 []string
```
instead? It's a bit more idiomatic.

internal/pubsub/example_test.go
```diff
@@ -78,8 +79,11 @@ func Example_sendReceiveMultipleMessages() {
...
-	for i := 0; i < len(ms); i++ {
+	for {
 		m2, err := s.Receive(ctx)
+		if err == context.DeadlineExceeded {
+			break
+		}
```
jba: I can't comment on line 91 because it's not part of the diff, but to be more like a real example, the ack should happen after you process the message, which in this case is after you print it.

jba: Sorry, my mistake. You're right, it's fine as is.

internal/pubsub/example_test.go
ijt: Done.

jba:
LGTM.

internal/pubsub/example_test.go
```diff
@@ -57,7 +57,8 @@ func Example_sendReceive() {
 
 func Example_sendReceiveMultipleMessages() {
 	// Open a topic and corresponding subscription.
-	ctx := context.Background()
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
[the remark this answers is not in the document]
ijt: Done.

ijt:
@vangent, I don't think an example test is the best place for a realistic example of message processing. Probably much clearer would be to make a pair of example apps, one that sends and another that receives. Otherwise we have to distort things to work in the testing environment.

For example, if you try to signify quitting with a special message then one or both of these things will happen:
1. The test will become flaky because the order is non-deterministic. Maybe mempubsub currently happens to have deterministic ordering, but other providers do not.
2. People reading the example will get the mistaken idea that they can rely on the ordering of messages.

ijt:
I think this pull request and the associated issue are dead ends. I'd like to close them both and make nicer examples with `func main`s.

ijt:
@vangent, sounds good.

PR #790, opened 2018-11-21
internal/pubsub: remove Close methods from driver

### LINKED ISSUE #753: pubsub: remove driver Close methods
It's unclear what the driver package's `Topic.Close` and `Subscription.Close` methods should do. 

We suggest that Topics and Subscriptions be created via `OpenXXX` methods which take some sort of closeable, connection-oriented value like a client or a gRPC `ClientConn`. Since the caller establishes the connection, they should be responsible for closing it. Also, they may want to reuse the connection for multiple Topics and Subscriptions. If `Topic.Close` closed the connection, that would invalidate all the other opened Topics. The alternative would be to reference-count the connections so only unused ones were closed. That would add a lot of complexity to the driver, and we want to keep them as simple as we can.

If the `Close` methods don't close the connection, it's not clear what is left for them to do. I suppose some driver implementation might have a client-side per-topic or per-subscription resource that they want to clean up (perhaps background goroutines), so we can leave the `Close` methods there and document their purpose. Or we could just remove them.

If a driver implemented a `Client` type with `OpenTopic`, `OpenSubscription` and `Close` methods, it probably wouldn't need `Topic.Close` or `Subscription.Close`. See #679.

jba:
See the issue for rationale.

Fixes #753.

### DIFF
--- a/internal/pubsub/pubsub.go
+++ b/internal/pubsub/pubsub.go
@@ -89,10 +89,10 @@ func (t *Topic) Send(ctx context.Context, m *Message) error {
 // Close flushes pending message sends and disconnects the Topic.
 // It only returns after all pending messages have been sent.
 func (t *Topic) Close() error {
-	t.batcher.Flush()
 	t.mu.Lock()
 	t.err = errors.New("pubsub: Topic closed")
 	t.mu.Unlock()
+	t.batcher.Flush()
 	return nil
 }
 
@@ -206,10 +206,10 @@ func (s *Subscription) getNextBatch(ctx context.Context) error {
 
 // Close flushes pending ack sends and disconnects the Subscription.
 func (s *Subscription) Close() error {
-	s.ackBatcher.Flush()
 	s.mu.Lock()
 	s.err = errors.New("pubsub: Subscription closed")
 	s.mu.Unlock()
+	s.ackBatcher.Flush()
 	return nil
 }
 
--- a/internal/pubsub/pubsub.go
+++ b/internal/pubsub/pubsub.go
@@ -48,6 +48,8 @@ func (m *Message) Ack() {
 type Topic struct {
 	driver  driver.Topic
 	batcher *bundler.Bundler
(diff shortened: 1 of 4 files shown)


internal/pubsub/pubsub.go
```diff
@@ -84,7 +90,10 @@ func (t *Topic) Send(ctx context.Context, m *Message) error {
 // It only returns after all pending messages have been sent.
 func (t *Topic) Close() error {
 	t.batcher.Flush()
-	return t.driver.Close()
+	t.mu.Lock()
```
vangent: Shouldn't you do this before the `Flush`? otherwise you have a race where you `Flush` but then new `Send` calls come in before you lock+mark closed.

internal/pubsub/pubsub.go
```diff
@@ -134,7 +143,8 @@ type Subscription struct {
 	mu sync.Mutex
 
 	// q is the local queue of messages downloaded from the server.
-	q []*Message
+	q   []*Message
```
vangent: gofmt needed?

internal/pubsub/pubsub.go
```diff
@@ -194,7 +207,10 @@ func (s *Subscription) getNextBatch(ctx context.Context) error {
 // Close flushes pending ack sends and disconnects the Subscription.
 func (s *Subscription) Close() error {
 	s.ackBatcher.Flush()
-	return s.driver.Close()
+	s.mu.Lock()
```
vangent: Same as above.

jba: Done.

internal/pubsub/pubsub.go
jba: The file is correctly formatted. The type lines up with the type on the line below.

internal/pubsub/pubsub.go
jba: Done.

Issue #759, opened 2018-11-19
pubsub: determine when to fail on non-existent resources


jba:
If the `OpenXXX` methods are asked to open non-existent resources, should they fail immediately, or should failure happen only when the resource is used?

Checking during open requires an RPC and a chunk of code, so it makes it harder to implement a driver, which we want to avoid. While users will know sooner that they made a mistake, they are already used to seeing errors only on first use, since at least on Amazon and GCP, there is no "open" step&mdash;you just provide a string and start using the resource.

So for simplicity, I'd suggest failing on first use. Or more precisely, not having drivers do anything special in any of their methods: just pass through to the provider, or in the case of Open, just create client-side data structures.

vangent:
+1 for fail on first use. `Open` should be lightweight.

jba:
Decision: `OpenXXX` methods don't fail. Add `Topic.Exists`.

ijt:
Okay, we determined it. Closing this out.

PR #792, opened 2018-11-21
mempubsub: fail in driver methods on nonexistent topic

jba:
All driver methods check for a non-existent topic and return an error.
(Non-existent subscription doesn't make sense for this package.)

Addresses #759 for mempubsub.

gcppubsub is TBD.

### DIFF
--- a/internal/pubsub/mempubsub/mem_test.go
+++ b/internal/pubsub/mempubsub/mem_test.go
@@ -19,6 +19,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-cloud/internal/pubsub"
 	"github.com/google/go-cloud/internal/pubsub/driver"
 )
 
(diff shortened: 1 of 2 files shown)


internal/pubsub/mempubsub/mem_test.go
```diff
@@ -75,3 +76,18 @@ func TestReceive(t *testing.T) {
 		t.Fatalf("got %d, want %d", got, want)
 	}
 }
+
```
vangent: Fine for now, but we should stop adding mempubsub-specific tests and add them to `drivertest.go` instead. Someone implementing a new driver should ~not have to write tests for it.

jba:
Test failure:
```
FAIL: TestConformance (0.13s)
    --- FAIL: TestConformance/TestInvalidJSON (0.01s)
        drivertest.go:265: got <nil> wanted invalid-json error
FAIL
coverage: 4.1% of statements in ./...
FAIL	github.com/google/go-cloud/runtimevar/etcdvar	1.098s
```

PR #800, opened 2018-11-22
batcher: move b.wg.Done below shutdown check

### LINKED ISSUE #799: batcher: panic in Shutdown
```
$ cd internal/pubsub
$ for i in $(seq 100); do go test -count=1 --run Example_sendReceiveMultipleMessages; done
```

> panic: sync: WaitGroup is reused before previous Wait has returned [recovered]
>         panic: sync: WaitGroup is reused before previous Wait has returned.
> github.com/google/go-cloud/internal/batcher.(*Batcher).Shutdown(0xc000120230)
>         /usr/local/google/home/rvangent/go/src/github.com/google/go-cloud/internal/batcher/batcher.go:127 +0x5f
> github.com/google/go-cloud/internal/pubsub.(*Subscription).Close(0xc0000da4b0, 0x3b9aca00, 0xc0000da4b0)
>         /usr/local/google/home/rvangent/go/src/github.com/google/go-cloud/internal/pubsub/pubsub.go:191 +0xac
> github.com/google/go-cloud/internal/pubsub_test.Example_sendReceiveMultipleMessages()

ijt:
Fixes #799

### DIFF
--- a/internal/batcher/batcher.go
+++ b/internal/batcher/batcher.go
@@ -65,11 +65,11 @@ func New(itemExample interface{}, maxHandlers int, handler func(interface{}) err
 // If Shutdown has been called, Add immediately returns an error.
 func (b *Batcher) Add(ctx context.Context, item interface{}) error {
 	b.wg.Add(1)
+	defer b.wg.Done()
 	b.mu.Lock()
 	if b.shutdown {
 		return errors.New("batcher: shut down")
 	}
-	defer b.wg.Done()
 	// Create a channel to receive the error from the handler.
 	c := make(chan error, 1)
 	// Add the item to the pending list.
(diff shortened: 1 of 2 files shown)


internal/batcher/batcher.go
```diff
@@ -65,11 +65,11 @@ func New(itemExample interface{}, maxHandlers int, handler func(interface{}) err
 // If Shutdown has been called, Add immediately returns an error.
 func (b *Batcher) Add(ctx context.Context, item interface{}) error {
 	b.wg.Add(1)
```
vangent: I think the `Add` needs to be protected by the mutex as well. `sync.WaitGroup` is not threadsafe.

vangent: ... apparently I'm wrong about that.

ijt:
The changes I just uploaded appear to fix the issue, but I'm going to open a new pull request since they're completely different from the original idea of this one.

PR #802, opened 2018-11-22
pubsub: integrate internal/batcher

### LINKED ISSUE #689: pubsub: dynamically adjust batch sizes in Topic and Subscription bundler logic
A simple thing to try first would be to start with a batch size of 1 and then adjust the size to be the number of messages that piled up while the previous call to SendBatch was running.

### LINKED ISSUE #799: batcher: panic in Shutdown

ijt:
Fixes #689
Fixes #799 by using AddNoWait

### DIFF
--- a/internal/pubsub/acks_test.go
+++ b/internal/pubsub/acks_test.go
@@ -197,3 +197,72 @@ func TestAckDoesNotBlock(t *testing.T) {
 	// failure.
 	mr.Ack()
 }
+
+func TestDoubleAckCausesPanic(t *testing.T) {
+	ctx := context.Background()
+	m := &driver.Message{}
+	ds := &ackingDriverSub{
+		q: []*driver.Message{m},
+		sendAcks: func(_ context.Context, ackIDs []driver.AckID) error {
+			return nil
+		},
+	}
+	sub := pubsub.NewSubscription(ds)
+	defer sub.Close()
+	mr, err := sub.Receive(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	mr.Ack()
+	defer func() {
+		if r := recover(); r != nil {
+			// ok, panic was expected.
+			return
+		}
+		t.Errorf("second ack failed to panic")
+	}()
+	mr.Ack()
+}
+
+// For best results, run this test with -race.
+func TestConcurrentDoubleAckCausesPanic(t *testing.T) {
+	ctx := context.Background()
+	m := &driver.Message{}
+	ds := &ackingDriverSub{
+		q: []*driver.Message{m},
+		sendAcks: func(_ context.Context, ackIDs []driver.AckID) error {
+			return nil
+		},
+	}
+	sub := pubsub.NewSubscription(ds)
+	defer sub.Close()
+	mr, err := sub.Receive(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Spin up some goroutines to ack the message.
+	var mu sync.Mutex
+	panics := 0
+	var wg sync.WaitGroup
+	for i := 0; i < 2; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			defer func() {
+				if r := recover(); r != nil {
+					mu.Lock()
+					defer mu.Unlock()
+					panics++
+				}
+			}()
+			mr.Ack()
+		}()
+	}
+	wg.Wait()
+
+	// Check that one of the goroutines panicked.
+	if panics != 1 {
+		t.Errorf("panics = %d, want %d", panics, 1)
+	}
+}
--- a/internal/pubsub/acks_test.go
+++ b/internal/pubsub/acks_test.go
@@ -18,6 +18,7 @@ import (
 	"math/rand"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/google/go-cloud/internal/pubsub"
 	"github.com/google/go-cloud/internal/pubsub/driver"
@@ -177,10 +178,11 @@ func TestTooManyAcksForASingleBatchGoIntoMultipleBatches(t *testing.T) {
 func TestAckDoesNotBlock(t *testing.T) {
 	ctx := context.Background()
 	m := &driver.Message{}
+	delay := time.Second
 	ds := &ackingDriverSub{
 		q: []*driver.Message{m},
 		sendAcks: func(_ context.Context, ackIDs []driver.AckID) error {
-			select {}
+			time.Sleep(delay)
 			return nil
 		},
 	}
@@ -190,5 +192,10 @@ func TestAckDoesNotBlock(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	t0 := time.Now()
 	mr.Ack()
+	dt := time.Now().Sub(t0)
+	if dt >= delay {
+		t.Errorf("ack took %v, want less than %v", dt, delay)
+	}
 }
--- a/internal/pubsub/acks_test.go
+++ b/internal/pubsub/acks_test.go
@@ -18,7 +18,6 @@ import (
 	"math/rand"
 	"sync"
 	"testing"
-	"time"
 
 	"github.com/google/go-cloud/internal/pubsub"
 	"github.com/google/go-cloud/internal/pubsub/driver"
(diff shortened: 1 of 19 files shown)


internal/pubsub/pubsub.go
```diff
@@ -197,7 +179,11 @@ func (s *Subscription) getNextBatch(ctx context.Context) error {
 			Body:     m.Body,
 			Metadata: m.Metadata,
 			ack: func() {
-				s.ackBatcher.Add(ackIDBox{id}, size)
+				s.wg.Add(1)
```
vangent: @jba I find this pretty subtle; it's making the `ackBatcher.Add` non-blocking, but has to add the `WaitGroup` in order to synchronize the created goroutine with `Shutdown` later. WDYT about providing a non-blocking `Add` function in `driver.Batcher` instead (as `bundler.Bundler` did)? Something like `func AddNoWait(ctx context.Context, item interface{})`. IIUC, it would just mean running the final `select` in `Add` in a goroutine.

vangent:
Build failed, looks like it's hanging....

internal/pubsub/pubsub.go
jba: OK.

jba: #807

internal/pubsub/pubsub.go
```diff
@@ -233,14 +220,12 @@ func NewSubscription(d driver.Subscription) *Subscription {
...
+		return retry.Call(callCtx, gax.Backoff{}, d.IsRetryable, func() error {
 			return d.SendAcks(callCtx, ids)
 		})
-		// TODO(#695): Do something sensible if SendAcks returns an error.
-		_ = err
 	}
-	ab := bundler.NewBundler(ackIDBox{}, handler)
-	ab.DelayThreshold = time.Millisecond
+	maxHandlers := 1
+	ab := batcher.New(ackIDBox{}, maxHandlers, handler)
```
jba: ackIDBox is no longer necessary. The first arg should be `reflect.TypeOf([]driver.AckID{}).Elem()`

internal/pubsub/pubsub.go
```diff
@@ -197,7 +179,11 @@ func (s *Subscription) getNextBatch(ctx context.Context) error {
 			Body:     m.Body,
 			Metadata: m.Metadata,
 			ack: func() {
-				s.ackBatcher.Add(ackIDBox{id}, size)
+				s.wg.Add(1)
+				go func() {
+					defer s.wg.Done()
+					s.ackBatcher.Add(ctx, ackIDBox{id})
```
jba: Change to AddNoWait after #807. You'll still need to deal with the errors somehow.

internal/pubsub/pubsub.go
ijt: I'm not sure I understand how we can recover the functionality here using the new AddNoWait function from #807.

With what I have here, it's possible to wait for all these Adds to finish before shutdown. What's the solution using AddNoWait?

internal/pubsub/pubsub.go

internal/pubsub/pubsub.go
ijt: I tried integrating this AddNoWait and got some breakage that I don't know how to fix. 
TestAckDoesNotBlock hung. I fixed it by using a time.Sleep call but I don't like it.
TestRetryTopic broke: "    pubsub_test.go:236: calls: got 0, want 3"

I was close to having something that worked, and I'd like to get it checked in before attempting this approach that I don't understand as well.

internal/pubsub/pubsub.go
ijt: Done. I don't know what to do with the errors. With blocking acks it was straightforward. Now I have no idea.

internal/pubsub/pubsub.go
ijt: Filed https://github.com/google/go-cloud/issues/814

internal/pubsub/acks_test.go
```diff
@@ -177,10 +178,11 @@ func TestTooManyAcksForASingleBatchGoIntoMultipleBatches(t *testing.T) {
 func TestAckDoesNotBlock(t *testing.T) {
 	ctx := context.Background()
 	m := &driver.Message{}
+	delay := time.Second
 	ds := &ackingDriverSub{
 		q: []*driver.Message{m},
 		sendAcks: func(_ context.Context, ackIDs []driver.AckID) error {
-			select {}
+			time.Sleep(delay)
```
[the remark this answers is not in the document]
ijt: I would but it hangs the way you suggested. That's because the Subscription constructor makes its own context, as suggested earlier by @jba.

In order to fix this, I'll have to change the Subscription constructor to take a context argument again, and then update all its uses to pass in a context. I think it's worth doing but beyond the scope of this CL. Filed an issue: https://github.com/google/go-cloud/issues/815.

internal/pubsub/pubsub.go
```diff
@@ -138,13 +117,16 @@ type Subscription struct {
...
-	ackBatcher *bundler.Bundler
+	ackBatcher driver.Batcher
 
 	mu sync.Mutex
 
 	// q is the local queue of messages downloaded from the server.
 	q   []*Message
 	err error
+
+	// wg manages acking goroutines.
```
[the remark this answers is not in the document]

internal/pubsub/pubsub.go
```diff
@@ -209,7 +192,8 @@ func (s *Subscription) Close() error {
 	s.mu.Lock()
 	s.err = errors.New("pubsub: Subscription closed")
 	s.mu.Unlock()
-	s.ackBatcher.Flush()
+	s.wg.Wait()
```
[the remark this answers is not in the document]

internal/pubsub/pubsub.go
```diff
@@ -197,7 +179,8 @@ func (s *Subscription) getNextBatch(ctx context.Context) error {
 			Body:     m.Body,
 			Metadata: m.Metadata,
 			ack: func() {
-				s.ackBatcher.Add(ackIDBox{id}, size)
+				// TODO(ijt): Do something with the returned chan error.
```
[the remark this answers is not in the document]

ijt:
PTAL

internal/pubsub/acks_test.go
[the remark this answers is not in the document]
vangent: I added some comments to that issue.

Couldn't you do this with a WaitGroup instead of a sleep?

internal/pubsub/pubsub.go
```diff
@@ -117,15 +99,12 @@ func NewTopic(d driver.Topic) *Topic {
...
+		return retry.Call(callCtx, gax.Backoff{}, d.IsRetryable, func() error {
 			return d.SendBatch(callCtx, dms)
 		})
-		for _, mec := range mecs {
-			mec.errChan <- err
-		}
 	}
-	b := bundler.NewBundler(msgErrChan{}, handler)
-	b.DelayThreshold = time.Millisecond
+	maxHandlers := 1
```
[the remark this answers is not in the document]
jba: We can leave it at 1 if desired. I just wrote a more general package, in case we ever want to add concurrency for performance.

I do suggest we require drivers to be threadsafe, so we can change it to 2 or higher without races. We should probably have a test using `-race` that sets it to something greater than 1.

internal/pubsub/pubsub.go
```diff
@@ -179,11 +179,8 @@ func (s *Subscription) getNextBatch(ctx context.Context) error {
 			Body:     m.Body,
 			Metadata: m.Metadata,
 			ack: func() {
-				s.wg.Add(1)
-				go func() {
-					defer s.wg.Done()
-					s.ackBatcher.Add(ctx, ackIDBox{id})
-				}()
+				// TODO(ijt): Do something with the returned chan error.
```
jba: Based on our conversation yesterday, I think you can ignore the error. As long as you do the right thing after the call to SendAcks.

internal/pubsub/pubsub.go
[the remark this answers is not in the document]
ijt: Filed #820

internal/pubsub/pubsub.go

What is "the right thing after the call to SendAcks"?

ijt:
@vangent, PTAL.

ijt:
Pardon me, @vangent, I guess you're waiting for me to get rid of the sleep hack. I'll see what I can do.

ijt:
Okay, the sleep is gone. PTAL.

internal/pubsub/acks_test.go
```diff
@@ -175,17 +175,18 @@ func TestTooManyAcksForASingleBatchGoIntoMultipleBatches(t *testing.T) {
...
 	ds := &ackingDriverSub{
 		q: []*driver.Message{m},
 		sendAcks: func(_ context.Context, ackIDs []driver.AckID) error {
-			select {}
+			<-ctx.Done()
 			return nil
 		},
 	}
 	sub := pubsub.NewSubscription(ds)
 	defer sub.Close()
```
vangent: Could you add a comment here explaining how the test works? It's not obvious.

IIUC, `sendAcks` will block forever if `cancel` isn't called, so the test is verifying that it can call `Ack` and not block on that call.

ijt: Yes, that's how it works. Added a comment.

PR #804, opened 2018-11-26
runtimevar/test: consistently check variable type and use %T to log it

vangent:


vangent:
> on line 42 there's a typo (s/iff/if)

That's on purpose actually, "iff" standing for "if and only if".

PR #805, opened 2018-11-26
runtimevar/test: use driver.WatchVariable directly in some tests

### LINKED ISSUE #797: runtimevar: tests can be flaky when using the concrete type
The polling implementations (paramstore and runtimeconfigurator) both use record/replay. If you use a short polling interval, then they can be flaky because the number of polls they do is indeterminate. If you use a longer polling interval then the tests are slow.

Other tests can be flaky because of other timing issues with the polling being different in replay mode.

The concrete type behavior is well tested in concrete type tests, this is probably a case where we should have a couple of basic end-to-end tests in drivertest, but then do other tests using the driver interface instead of the concrete type.

vangent:
Fixes #797.

This simplifies the tests by not generally mucking around with the `WaitDuration`.

### DIFF
--- a/runtimevar/constantvar/constantvar_test.go
+++ b/runtimevar/constantvar/constantvar_test.go
@@ -34,7 +34,7 @@ func newHarness(t *testing.T) (drivertest.Harness, error) {
 	return &harness{vars: map[string][]byte{}}, nil
 }
 
-func (h *harness) MakeWatcher(ctx context.Context, name string, decoder *runtimevar.Decoder, wait time.Duration) (driver.Watcher, error) {
+func (h *harness) MakeWatcher(ctx context.Context, name string, decoder *runtimevar.Decoder) (driver.Watcher, error) {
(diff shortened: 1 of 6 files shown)


runtimevar/constantvar/constantvar_test.go
```diff
@@ -34,7 +34,7 @@ func newHarness(t *testing.T) (drivertest.Harness, error) {
 	return &harness{vars: map[string][]byte{}}, nil
 }
 
-func (h *harness) MakeWatcher(ctx context.Context, name string, decoder *runtimevar.Decoder, wait time.Duration) (driver.Watcher, error) {
```
shantuo: Just curious, was wait not used at all before?

vangent: Yes, it was used before; the conformance tests passed different values (`shortWait` or `defaultWait`) depending on the test. Now that they use the driver interface explicitly, that's no longer needed.

PR #807, opened 2018-11-26
internal/batcher: add AddNoWait

jba:
Calling AddNoWait doesn't block.


internal/batcher/batcher.go
```diff
@@ -64,36 +64,49 @@ func New(itemType reflect.Type, maxHandlers int, handler func(interface{}) error
...
+	// Wait until either our result is ready or the context is done.
+	select {
+	case err := <-c:
+		return err
+	case <-ctx.Done():
+		return ctx.Err()
 	}
+}
+
+// Add adds an item to the batcher and returns immediately. When the handler is
```
vangent: Add -> AddNoWait

PR #808, opened 2018-11-26
pubsub/test: add Receive examples with worker pools

vangent:
One with a traditional worker pool where it spins up N goroutines that all loop forever, and one with an inverted worker pool that uses a semaphore to limit the concurrent goroutines.

### DIFF
--- a/internal/pubsub/example_test.go
+++ b/internal/pubsub/example_test.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"log"
 	"sort"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/google/go-cloud/internal/pubsub"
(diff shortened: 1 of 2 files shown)


internal/pubsub/example_test.go
```diff
@@ -101,3 +103,147 @@ func Example_sendReceiveMultipleMessages() {
...
+			log.Fatal(err)
+		}
+	}
+
+	// In order to make our test exit, we keep track of how many messages were
+	// processed and cancel the receiveCtx when we've processed them all.
+	var nProcessed int32
+	receiveCtx, cancel := context.WithCancel(ctx)
+	go func() {
+		for {
```
jba: This burns CPU.

You could use a waitgroup instead:
- Add(100) before publishing.
- Call Done when finished processing.
- Call Wait instead of this loop.

internal/pubsub/example_test.go
```diff
@@ -101,3 +103,147 @@ func Example_sendReceiveMultipleMessages() {
...
+		}
+	}()
+
+	// Process messages using an inverted worker pool, as described here:
+	// https://www.youtube.com/watch?v=5zXAHh5tJqQ&t=26m58s
+	// It uses a buffered channel, sem, as a semaphore to manage the maximum
+	// number of workers.
+	const poolSize = 10
+	sem := make(chan struct{}, poolSize)
+	for {
```
jba: Why not just
```
for atomic.LoadInt32(&nProcessed) < nMessages {
```

internal/pubsub/example_test.go
```diff
@@ -101,3 +103,147 @@ func Example_sendReceiveMultipleMessages() {
...
+			// up the slot for another goroutine.
+			<-sem
+		}()
+	}
+
+	// Wait for all workers to finish.
+	for n := poolSize; n > 0; n-- {
+		sem <- struct{}{}
+	}
+	fmt.Printf("Read %d messages", nProcessed)
```
jba: Close the sub to make sure all acks are sent.

internal/pubsub/example_test.go
vangent: I think this is more realistic; normal users will likely have an infinite loop with a `break` rather than an explicit condition. Also, I find that misleading, as you might think it would work without the cancellation of the ctx, but it doesn't -- you can enter the loop when `nMessages` is small, call `Receive`, and block forever as `nMessages` is incremented.

internal/pubsub/example_test.go
vangent: There is a `defer s.Close()` up above already. Are you suggesting to remove that and add an explicit `s.Close()` down here?

jba: No, that's fine.

internal/pubsub/example_test.go
```diff
@@ -101,3 +103,149 @@ func Example_sendReceiveMultipleMessages() {
...
+		}
+	}
+
+	// In order to make our test exit, we keep track of how many messages were
+	// processed and cancel the receiveCtx when we've processed them all.
+	var nProcessed int32
+	receiveCtx, cancel := context.WithCancel(ctx)
+	go func() {
+		wg.Wait()
+		if atomic.LoadInt32(&nProcessed) == nMessages {
```
jba: The test is unnecessary now.

vangent: Ah, sorry -- in fact, `nProcessed` isn't needed at all anymore.

internal/pubsub/example_test.go
```diff
@@ -101,3 +102,141 @@ func Example_sendReceiveMultipleMessages() {
...
+			// Record that we've processed this message, and Ack it.
+			msg.Ack()
+			wg.Done()
+			// Read a token from the semaphore before exiting this goroutine, freeing
+			// up the slot for another goroutine.
+			<-sem
+		}()
+	}
+
+	// Wait for all workers to finish.
```
ijt: I don't believe this code is waiting for workers to finish. I think it's allowing them to start.

internal/pubsub/example_test.go
```diff
@@ -101,3 +102,141 @@ func Example_sendReceiveMultipleMessages() {
...
+		// Write a token to the semaphore; if there are already poolSize workers
+		// active, this will block until one of them completes.
+		sem <- struct{}{}
+		// Process the message. For many applications, this can be expensive, so
+		// we do it in a goroutine, allowing this loop to continue and Receive more
+		// messages.
+		go func() {
+			// Record that we've processed this message, and Ack it.
+			msg.Ack()
+			wg.Done()
```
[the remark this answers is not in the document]
vangent: I updated the comments up above.

internal/pubsub/example_test.go
vangent: I don't think so. If no workers ran at all, then this would be able to write `poolSize` tokens to `sem` immediately. This is supposed to block while any workers are running.

vangent:
PTAL.

PR #812, opened 2018-11-26
runtimevar/test: Add a test that updates the var but keeps the same error

### LINKED ISSUE #798: runtimevar: add a test for when there's an update, but the error is the same

vangent:
I also fixed a couple of bugs/flakes that I found as part of adding this test.

Fixes #798.


runtimevar/drivertest/drivertest.go
```diff
@@ -144,14 +156,19 @@ func testString(t *testing.T, newHarness HarnessMaker) {
...
-	tCtx, cancel := context.WithTimeout(ctx, 10*time.Millisecond)
+	tCtx, cancel := context.WithTimeout(ctx, 100*time.Millisecond)
 	defer cancel()
 	got, err = v.Watch(tCtx)
 	if err == nil {
 		t.Errorf("got %v want error", got)
 	}
-	if tCtx.Err() == nil {
-		t.Errorf("got err %v; want Watch to have blocked until context was Done", err)
+	// tCtx should be cancelled. However, tests using record/replay mode can
```
vangent: This was a potential flake before, that became more obvious when I added another copy of it down below, so I fixed them both.

runtimevar/etcdvar/etcdvar.go
```diff
@@ -146,8 +146,9 @@ func (w *watcher) watch(ctx context.Context, cli *clientv3.Client, name string,
 				val, err := decoder.Decode(kv.Value)
 				if err != nil {
 					cur = w.updateState(&state{err: err}, cur)
+				} else {
+					cur = w.updateState(&state{val: val, updateTime: time.Now(), version: kv.Version}, cur)
```
vangent: Not having this in an `else` was a bug before that this test caught.

runtimevar/drivertest/drivertest.go
```diff
@@ -392,3 +409,92 @@ func testDelete(t *testing.T, newHarness HarnessMaker) {
...
+	defer cancel()
+	state, _ = drv.WatchVariable(tCtx, state)
+	if state == nil {
+		// OK: nil indicates no change.
+	} else {
+		// WatchVariable should have blocked until tCtx was cancelled, and we
+		// should have gotten that error back.
+		got, err := state.Value()
+		if err == nil {
+			t.Fatalf("got %v want error", got)
```
shantuo: s/error/nil

runtimevar/etcdvar/etcdvar.go
```diff
@@ -97,7 +97,7 @@ func (w *watcher) WatchVariable(ctx context.Context, _ driver.State) (driver.Sta
 // It always return s.
 func (w *watcher) updateState(s, prev *state) *state {
 	if s.err != nil && prev != nil && prev.err != nil {
-		if s.err == prev.err {
+		if s.err == prev.err || s.err.Error() == prev.err.Error() {
```
shantuo: Does `s.err.Error() == prev.err.Error()` include `s.err == prev.err`?

runtimevar/drivertest/drivertest.go
vangent: No, this is correct -- the test is expecting that err should not be nil, and it was nil. I clarified it a bit.

runtimevar/etcdvar/etcdvar.go
vangent: Yes, but it's less efficient.

vangent:
PTAL.

PR #813, opened 2018-11-26
runtimevar/filevar: always add files to the notifier after errors

vangent:
The current code turns out to be broken, as I discovered with Printfs while adding #812. During that test, the `filevar` implementation blocks as it's supposed to, but it does it because it doesn't even notice the `Rename`, not because it notices it and then notices that the errors are the same. It is difficult to test because it's hard to inject different errors.

Updates #812.


vangent:
> Don't want to block this PR, but is this a bug with fsnotify?

Maybe? It may also be OS-specific; for some OS's, the "Add" follows the file descriptor as part of a rename, and others it's sticky to the filename. There are lots of open issues for `fsnotify` about different behavior on different OSs.

Issue #675, opened 2018-11-12
blob/tests: get code coverage up to 90% for concrete type, 85% for providers


vangent:
[Coveralls](https://coveralls.io/github/google/go-cloud)
  shows code coverage for `blob` at around 73%.

 73.85 blob/
 82.5 fileblob/
 71.0 gcsblob/
 81.91 s3blob/
 53.2 blob.go

Let's see if we can get them all up to 90%.

vangent:
The coverage is pessimistic, see https://github.com/google/go-cloud/issues/741.

vangent:
New snapshot after fixing #741:

84.75 blob/
  85.0 fileblob/
  76.19 gcsblob/
  83.28 s3blob/
  94.0 blob.go

vangent:
Latest snapshot:

86.33 fileblob/
72.93 gcsblob/
82.99 s3blob/
95.39 blob.go

vangent:
Latest snapshot:

86.33 fileblob
87.61 gcsblob/
87.63 s3blob/
95.45 blob.go

Done!

PR #819, opened 2018-11-27
blob/gcsblob: add tests for OpenURL

vangent:
The `gcsblob.go` is just moving the function from being defined inline to being defined at the top level. It's more obvious if you "hide whitespace" in the diff.

Updates #675.


vangent:
PTAL.

blob/gcsblob/gcsblob.go
```diff
@@ -59,46 +59,48 @@ import (
...
+	opts := &Options{}
 
-		if keyPath := q["private_key_path"]; len(keyPath) > 0 {
-			pk, err := ioutil.ReadFile(keyPath[0])
-			if err != nil {
-				return nil, err
-			}
-			opts.PrivateKey = pk
-		}
+	if accessID := q["access_id"]; len(accessID) > 0 {
```
shantuo: nit: `if accessID, ok := q["access_id"]; ok {`

blob/gcsblob/gcsblob.go
```diff
@@ -59,46 +59,48 @@ import (
...
-		} else {
-			jsonCreds, err := ioutil.ReadFile(credPath[0])
-			if err != nil {
-				return nil, err
-			}
-			creds, err = google.CredentialsFromJSON(ctx, jsonCreds)
-			if err != nil {
-				return nil, err
-			}
+	if keyPath := q["private_key_path"]; len(keyPath) > 0 {
```
shantuo: same

blob/gcsblob/gcsblob.go
```diff
@@ -59,46 +59,48 @@ import (
...
+		pk, err := ioutil.ReadFile(keyPath[0])
+		if err != nil {
+			return nil, err
 		}
+		opts.PrivateKey = pk
+	}
 
-		client, err := gcp.NewHTTPClient(gcp.DefaultTransport(), gcp.CredentialsTokenSource(creds))
+	var creds *google.Credentials
+	if credPath := q["cred_path"]; len(credPath) == 0 {
```
shantuo: same

blob/gcsblob/gcsblob.go
vangent: Hmm, I think checking that `len > 0` is more correct here; technically the map could have an `access_id` key with a 0-length slice value, and `ok` would be `true` but the access of `[0]` on the next line would panic. It's possible that `net.URL` would never produce a 0-length slice value, but I think its better to check the thing we actually care about.

PR #822, opened 2018-11-27
mysql/rdsmysql: add OpenCensus configuration

### LINKED ISSUE #416: sql/ocsql: allow configuration of ocsql instrumentation

### LINKED ISSUE #608: mysql: consider making an Options struct for all constructors

zombiezen:
Remove Options struct. Regenerated Guestbook Wire injectors.

Fixes #416
Fixes #608


zombiezen:
@ijt, when you get the chance, can you please review?

ijt:
The Params are supposed to be parameters for connecting to a db. Adding traceopts to this struct changes what it's about.

I don't feel strongly enough about this to block it though. Just my two cents.

zombiezen:
> The Params are supposed to be parameters for connecting to a db. Adding traceopts to this struct changes what it's about.

This struct was added in #618. Since then, the consensus on #608 was that having the two structs was confusing. In #743, you performed the same operation this PR does but did not address this package. I am continuing the change for consistency. If you think this is the wrong decision, then we should discuss on #608.

> I don't feel strongly enough about this to block it though. Just my two cents.

I am blocked on having a firm resolution here, because I don't want to propagate further inconsistency as I introduce tracing into the PostgreSQL packages (#254).

ijt:
Okay, let's keep it consistent. Sorry for the noise.

PR #830, opened 2018-11-28
pubsub: return after t.Error call in goroutine to avoid nil dereference

### LINKED ISSUE #829: pubsub: TestConcurrentReceivesGetAllTheMessages fails intermittently
Found on Travis:

```
ok  	github.com/google/go-cloud/internal/batcher	3.904s	coverage: 1.0% of statements in ./...
[signal 0xc0000005 code=0x0 addr=0x0 pc=0xa6039f]
goroutine 1400 [running]:
github.com/google/go-cloud/internal/pubsub_test.TestConcurrentReceivesGetAllTheMessages.func1(0xc0003ac5a0, 0xc000317770, 0x1779180, 0xc00053e5c0, 0xc000309000, 0xc0003ac5b0, 0xc0003af770)
	C:/Users/travis/gopath/src/github.com/google/go-cloud/internal/pubsub/pubsub_test.go:157 +0x8f
created by github.com/google/go-cloud/internal/pubsub_test.TestConcurrentReceivesGetAllTheMessages
	C:/Users/travis/gopath/src/github.com/google/go-cloud/internal/pubsub/pubsub_test.go:146 +0x448
FAIL	github.com/google/go-cloud/internal/pubsub	2.199s
```

ijt:
Fixes #829

### DIFF
--- a/internal/pubsub/pubsub_test.go
+++ b/internal/pubsub/pubsub_test.go
@@ -152,6 +152,7 @@ func TestConcurrentReceivesGetAllTheMessages(t *testing.T) {
 						return
 					}
 					t.Error(err)
+					return
 				}
 				mu.Lock()
 				receivedMsgs[string(m.Body)]++


internal/pubsub/pubsub_test.go
```diff
@@ -152,6 +152,7 @@ func TestConcurrentReceivesGetAllTheMessages(t *testing.T) {
 						return
 					}
 					t.Error(err)
```
vangent: Or `t.Fatal` instead?

vangent:
Also, the test will still fail, right?

internal/pubsub/pubsub_test.go
ijt: `t.Fatal` calls `t.FailNow`, and that must not be called from goroutines other than the one running the test. https://golang.org/pkg/testing/#B.FailNow

ijt:
The test will still fail if it hits this error, yes.

internal/pubsub/pubsub_test.go
vangent: Hmm, I did not know that! Learn something every day. Well, some days.

Issue #828, opened 2018-11-28
wire: use new Wire repository


zombiezen:
Wire is now a separate repository and Go module: https://github.com/google/wire

We should remove `wire/internal/` and `wire/cmd/`, and leave `README.md` and `wire.go` as pointers over to the new repository.

vangent:
May also need to update the Travis tests for Go Cloud that use `wire` to get it from the right place.

PR #831, opened 2018-11-28
wire: remove code generator tool

zombiezen:
The marker function package remains where it is, but now aliases to the new Wire package. I will update the import paths in this repository in a separate PR so as to shake out any unforeseen issues that would be encountered by end-users as well.

I removed the content of the README and left behind a pointer to the new repository at https://github.com/google/wire. I fixed any dangling documentation links I could find.

Updates #828

/cc @vangent @clausti


README.md
```diff
@@ -24,8 +24,8 @@ and being able to run that code on any cloud you want, avoiding all the ceremony
 of cloud-specific authorization, tracing, SDKs and all the other code required
 to make an application portable across cloud platforms.
 
-Another key part of the project is a code generator called
-[Wire](https://github.com/google/go-cloud/blob/master/wire/README.md). It
+The project works with a code generator called
```
vangent: Maybe "works well with"?
I don't think that we want to imply that it requires Wire.

doc.go
```diff
@@ -26,8 +26,8 @@ gcsblob.OpenBucket, s3blob.OpenBucket, or any other provider. Then,
 the blob.Bucket can be used throughout your application without worrying
 about the underlying implementation.
 
-This project also provides a code generator called Wire
-(https://github.com/google/go-cloud/blob/master/wire/README.md). It creates
+This project works with a code generator called Wire
```
vangent: Same.

samples/wire/README.md
```diff
@@ -405,12 +405,12 @@ There is support for [binding interfaces][interfaces], [binding
 values][values], as well as support for [cleanup functions][cleanup]. See the
```
vangent: Should this whole sample be moved to the `wire` repo? It seems to be all about wire.

Fine to do in a separate PR.

vangent:
Tests failed. I suspect you need to update the proxy with the packages for `wire`.

Are we sharing the same proxy? If so, we'll need to update the instructions to always build wire as well as go-cloud. It'll probably be easier/better to have separate proxy buckets.

samples/wire/README.md
shantuo: Agree to separate the sample, but we'll want some pointers in this repo to Wire as well.

.travis.yml
```diff
@@ -23,7 +23,7 @@ install:
   # converting \n to \r\n.
   - "git config --global core.autocrlf input"
   - "git checkout -- ."
-  - "go install ./wire/cmd/wire"
+  - "go install github.com/google/wire/cmd/wire"
```
shantuo: Are we generating code in the tests though?

zombiezen: We're running `wire check` to ensure that there are no errors in the provider sets we give. See here: https://github.com/google/go-cloud/blob/87e273223b452e987ff654f016887ab80a6f9252/internal/testing/runchecks.sh#L38

README.md
zombiezen: Done.

doc.go
zombiezen: Done.

samples/wire/README.md
zombiezen: Agreed. Opened #836 to track removing the sample from this repository.

zombiezen:
Yup, I forgot to update the proxy before. I have updated it now.

zombiezen:
@shantuo, I'm blocked on your review.

shantuo:
Discussed offline, agreed on removing from the current version.

PR #832, opened 2018-11-29
runtimevar: wrap driver errors

### LINKED ISSUE #795: runtimevar: ensure runtimevar consistently wraps driver errors

vangent:
Fixes #795.

### DIFF
--- a/runtimevar/runtimevar.go
+++ b/runtimevar/runtimevar.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"encoding/gob"
 	"encoding/json"
-	"fmt"
 	"reflect"
 	"time"
 
@@ -91,8 +90,7 @@ func (c *Variable) Watch(ctx context.Context) (Snapshot, error) {
 		c.prev = cur
 		v, err := cur.Value()
 		if err != nil {
-			// Mask underlying errors.
-			return Snapshot{}, fmt.Errorf("Variable.Watch: %v", err)
+			return Snapshot{}, wrapError(c.watcher, err)
 		}
 		return Snapshot{Value: v, UpdateTime: cur.UpdateTime()}, nil
 	}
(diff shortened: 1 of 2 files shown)


runtimevar/runtimevar.go
```diff
@@ -91,16 +90,33 @@ func (c *Variable) Watch(ctx context.Context) (Snapshot, error) {
 		c.prev = cur
 		v, err := cur.Value()
 		if err != nil {
-			// Mask underlying errors.
-			return Snapshot{}, fmt.Errorf("Variable.Watch: %v", err)
+			return Snapshot{}, wrapError(err)
```
shantuo: The UI doesn't let me comment on L77, but should L77 also wraps the error?

runtimevar/runtimevar_test.go
```diff
@@ -193,6 +192,45 @@ func TestVariable(t *testing.T) {
 	}
 }
 
+var errFake = errors.New("fake")
+
+// fakeError implements driver.Watcher.
+// WatchVariable always returns a state with errFake, and Close
+// always returns errFake.
+type fakeErrorer struct {
```
shantuo: fakeErrorer sounds weird, especially it doesn't have a function called `fakeError`. How about just `fakeWatcher` or `errorWatcher`?

runtimevar/runtimevar.go
vangent: No, that isn't necessary, because that's guaranteed to be a context-canceled error, not an error from a provider.

runtimevar/runtimevar_test.go
vangent: Yeah, you are right. Changed to `erroringWatcher`. I also updated the similar fakes in `blob_test`, to be consistent.

vangent:
PTAL.

Issue #820, opened 2018-11-27
pubsub: add tests that set maxHandlers >= 2 and run them with -race in Travis


ijt:
We'll need one for sending to a topic and another for acking messages.

There will have to be a way to set maxHandlers in Topics and Subscriptions to make this work.

vangent:
Please also update the docstring for `SendBatch` to make it clear that driver implementations need to support multiple concurrent calls to it (and similar for other driver APIs that don't have this).

vangent:
@jba is adding a test for RabbitMQ that tests this; I think it should be moved to be a conformance test as part of this issue.

jba:
The tricky part is: we don't want to expose `maxHandlers` as a knob to users, but the conformance test is in a different package so if we give it the knob, we expose it to everyone. 

In the past I've named exported symbols `Internal...` and documented that they shouldn't be used. (e.g. https://github.com/GoogleCloudPlatform/google-cloud-go/blob/master/iam/iam.go#L102). When I've done that before there was no reason for a user to call those functions anyway; in the current case, we'd clearly be exposing a knob that someone might want to turn.

An alternative is to just set `maxHandlers` to some small number greater than 1.

vangent:
It seems to me we should do some benchmark testing to find a "good" value for `maxHandlers`, and use that all the time (including tests). I'm not sure what the value of exposing the knob to tests but not externally is.

Note that an alternative way to manage this in tests is to inject a different bundler, so you can more precisely control what it does (even setting `maxHandlers` to a value > 1 will not guarantee anything in terms of concurrency of calls to the driver).

ijt:
@vangent, I don't know how we could find a value of `maxHandlers` that would work well for all pubsub providers. The best setting could vary depending on the bandwidth of the connection as well.

vangent:
How do you plan to reconcile that with "we don't want to expose `maxHandlers` as a knob to users"? Or do you think it will be modified dynamically in some way?

Regardless, I still think that you are going to end up wanting to be able to inject a fake bundler in tests, so that you can test what you want to test without the "real" bundler getting in the way.

ijt:
@vangent, yes, I was thinking it could be adjusted dynamically.

Ideally I guess we would want to start sending the next batch immediately after the current batch has been sent, before the response comes back for the current batch's RPC. I doubt whether gRPC makes this possible, but it's probably worthwhile for us to have a clear picture in mind of what the optimal strategy would do.

[...]

PR #833, opened 2018-11-29
internal/pubsub/rabbitpubsub: RabbitMQ implementation: publish

jba:
Implementation of driver.Topic, including SendBatch.

### DIFF
--- a/internal/pubsub/rabbitpubsub/examples_test.go
+++ b/internal/pubsub/rabbitpubsub/examples_test.go
@@ -1,0 +1,73 @@
+// Copyright 2018 The Go Cloud Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rabbitpubsub_test
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/google/go-cloud/internal/pubsub"
+	"github.com/google/go-cloud/internal/pubsub/rabbitpubsub"
+	"github.com/streadway/amqp"
+)
+
+const rabbitURL = "amqp://guest:guest@localhost:5672/"
+
+func Example_publish() {
+	// Connect to RabbitMQ.
+	conn, err := amqp.Dial(rabbitURL)
+	if err != nil {
+		log.Printf("cannot connect to RabbitMQ: %v\n", err)
+		return
+	}
+	defer conn.Close()
+
+	// Declare a fanout exchange to be used as a Pub/Sub topic.
+	ch, err := conn.Channel()
+	if err != nil {
+		log.Fatalf("creating channel: %v", err)
+	}
+	const exchangeName = "my-topic"
+	err = ch.ExchangeDeclare(
+		exchangeName,
+		"fanout", // kind
+		false,    // durable
+		false,    // delete when unused
+		false,    // internal
+		false,    // no-wait
+		nil)      // args
+	if err != nil {
+		log.Fatalf("declaring exchange: %v", err)
+	}
+	ch.Close()
+
+	// Open a topic, using the exchange name.
+	topic, err := rabbitpubsub.OpenTopic(conn, exchangeName)
+	if err != nil {
+		log.Fatalf("opening topic: %v", err)
+	}
+	ctx := context.Background()
+	err = topic.Send(ctx, &pubsub.Message{Body: []byte("hello")})
+	if err != nil {
+		log.Fatalf("sending: %v", err)
+	}
+	if err := topic.Close(); err != nil {
+		log.Fatalf("closing topic: %v", err)
+	}
+	fmt.Println("message sent")
+
+	// Output: message sent
+}
--- a/internal/pubsub/rabbitpubsub/examples_test.go
+++ b/internal/pubsub/rabbitpubsub/examples_test.go
@@ -68,6 +68,4 @@ func Example_publish() {
 		log.Fatalf("closing topic: %v", err)
 	}
 	fmt.Println("message sent")
-
-	// Output: message sent
 }
--- a/internal/pubsub/rabbitpubsub/rabbit.go
+++ b/internal/pubsub/rabbitpubsub/rabbit.go
@@ -35,7 +35,7 @@ const (
 	mandatory = true
 
 	// If there are no waiting consumers, enqueue the message instead of dropping it
-	immediate = true
+	notImmediate = false
 )
 
 // OpenTopic returns a *pubsub.Topic corresponding to the named exchange. The
@@ -46,7 +46,7 @@ const (
 //
 // OpenTopic uses the supplied amqp.Connection for all communication. It is the caller's
 // responsibility to establish this connection before calling OpenTopic, and to close
-// it when it is no longer in use.
+// it when Close has been called on all topics opened with it.
 //
 // The documentation of the amqp package recommends using separate connections for
 // publishing and subscribing.
@@ -111,7 +111,7 @@ func (t *topic) SendBatch(ctx context.Context, ms []*driver.Message) error {
 	}
 
 	// Read from the channel established with NotifyPublish.
-	// Do so concurrently or we will be deadlock with the Publish RPC.
+	// Do so concurrently or we will deadlock with the Publish RPC.
 	// (The amqp package docs recommend setting the capacity of the channel
 	// to the number of messages to be published, but we can't do that because
 	// we want to reuse the channel for all calls to SendBatch.)
@@ -152,7 +152,7 @@ func (t *topic) SendBatch(ctx context.Context, ms []*driver.Message) error {
 
 	for _, m := range ms {
 		pub := toPublishing(m)
-		if err := t.ch.Publish(t.exchange, routingKey, mandatory, !immediate, pub); err != nil {
+		if err := t.ch.Publish(t.exchange, routingKey, mandatory, notImmediate, pub); err != nil {
 			t.ch = nil // AMQP channel is broken after error
 			return err
 		}
--- a/internal/pubsub/rabbitpubsub/rabbit.go
+++ b/internal/pubsub/rabbitpubsub/rabbit.go
@@ -1,0 +1,179 @@
+package rabbitpubsub
+
+import (
+	"context"
+	"errors"
+	"sync"
+
+	"github.com/google/go-cloud/internal/pubsub"
+	"github.com/google/go-cloud/internal/pubsub/driver"
+	"github.com/streadway/amqp"
+)
+
+type topic struct {
+	exchange string // the AMQP exchange
+	conn     *amqp.Connection
+
+	mu   sync.Mutex
+	ch   *amqp.Channel            // AMQP channel used for all communication.
+	pubc <-chan amqp.Confirmation // Go channel for server acks of publishes
+}
+
+// Values for the amqp client.
+// See https://www.rabbitmq.com/amqp-0-9-1-reference.html.
+const (
+	// Many methods of the amqp client take a "no-wait" parameter, which
+	// if true causes the client to return without waiting for a server
+	// response. We always want to wait.
+	wait = false
+
+	// Always use the empty routing key. This driver expects to be used with topic
+	// exchanges, which disregard the routing key.
+	routingKey = ""
+
+	// If the message can't be enqueued, return it to the sender rather than silently dropping it.
+	mandatory = true
+
+	// If there are no waiting consumers, enqueue the message instead of dropping it
+	immediate = true
+)
+
+// OpenTopic returns a *pubsub.Topic corresponding to the named exchange. The
+// exchange must have been previously created (for instance, by using
+// amqp.Channel.ExchangeDeclare). For the model of Go Cloud Pub/Sub to make sense,
+// the exchange should be a fanout exchange, although nothing in this package
+// enforces that.
+//
+// OpenTopic uses the supplied amqp.Connection for all communication. It is the caller's
+// responsibility to establish this connection before calling OpenTopic, and to close
+// it when it is no longer in use.
+//
+// The documentation of the amqp package recommends using separate connections for
+// publishing and subscribing.
+func OpenTopic(conn *amqp.Connection, name string) (*pubsub.Topic, error) {
+	// TODO(jba): support context.Context
+	t, err := newTopic(conn, name)
+	if err != nil {
+		return nil, err
+	}
+	return pubsub.NewTopic(t), nil
+
+}
+
+func newTopic(conn *amqp.Connection, name string) (*topic, error) {
+	t := &topic{
+		exchange: name,
+		conn:     conn,
+	}
+	if err := t.establishChannel(); err != nil {
+		return nil, err
+	}
+	return t, nil
+}
+
+// establishChannel creates an AMQP channel if necessary. According to the amqp
+// package docs, once an error is returned from the channel, it must be discarded and
+// a new one created.
+//
+// Must be called with the t.mu held.
+func (t *topic) establishChannel() error {
+	// TODO(jba): support context.Context
+	if t.ch != nil {
+		// We already have a channel; nothing to do.
+		return nil
+	}
+	// Create a new channel.
+	ch, err := t.conn.Channel()
+	if err != nil {
+		return err
+	}
+	// Put the channel into a mode where confirmations are delivered for each publish.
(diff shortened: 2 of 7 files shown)


internal/pubsub/rabbitpubsub/rabbit_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+func mustDialRabbit(t *testing.T) *amqp.Connection {
+	conn, err := amqp.Dial(rabbitURL)
+	if err != nil {
+		t.Skipf("skipping because the RabbitMQ server is not up (dial error: %v)", err)
+	}
+	return conn
+}
+
+func TestPublishConcurrently(t *testing.T) {
```
vangent: Can you run the conformance tests instead? We shouldn't be writing provider-specific tests except for very provider-specific things that aren't exposed via the concrete type (this should be rare).

internal/pubsub/rabbitpubsub/rabbit_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/google/go-cloud/internal/pubsub/driver"
+	"github.com/streadway/amqp"
+)
+
+const rabbitURL = "amqp://guest:guest@localhost:5672/"
+
+func mustDialRabbit(t *testing.T) *amqp.Connection {
+	conn, err := amqp.Dial(rabbitURL)
+	if err != nil {
+		t.Skipf("skipping because the RabbitMQ server is not up (dial error: %v)", err)
```
vangent: Have you figured out the record/replay strategy yet?

jba: It isn't feasible to do record/replay because the protocol is a custom one directly over TCP. 

I'm going with what Ross did on mysql: setup instructions to a human, along with skipping tests if the setup wasn't done.

internal/pubsub/rabbitpubsub/rabbit_test.go
jba: I added a call to the conformance tests, but can't run them yet because I don't have subscriptions implemented.

I removed one of the two tests, but kept the second for reasons I documented in the file.

internal/pubsub/rabbitpubsub/examples_test.go
```diff
@@ -0,0 +1,73 @@
...
+		nil)      // args
+	if err != nil {
+		log.Fatalf("declaring exchange: %v", err)
+	}
+	ch.Close()
+
+	// Open a topic, using the exchange name.
+	topic, err := rabbitpubsub.OpenTopic(conn, exchangeName)
+	if err != nil {
+		log.Fatalf("opening topic: %v", err)
```
vangent: Why `log.Fatalf` here and `log.Printf` + `return` above?

internal/pubsub/rabbitpubsub/examples_test.go
```diff
@@ -0,0 +1,73 @@
...
+		"fanout", // kind
+		false,    // durable
+		false,    // delete when unused
+		false,    // internal
+		false,    // no-wait
+		nil)      // args
+	if err != nil {
+		log.Fatalf("declaring exchange: %v", err)
+	}
+	ch.Close()
```
vangent: Consider adding a comment here? It's not obvious to me why it's ok for `ch` to be closed before using the exchange, especially since you'll next hand `conn` off to `OpenTopic`.

internal/pubsub/rabbitpubsub/examples_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// Open a topic, using the exchange name.
+	topic, err := rabbitpubsub.OpenTopic(conn, exchangeName)
+	if err != nil {
+		log.Fatalf("opening topic: %v", err)
+	}
+	ctx := context.Background()
+	err = topic.Send(ctx, &pubsub.Message{Body: []byte("hello")})
```
vangent: `defer topic.Close()` ?

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -0,0 +1,179 @@
...
+// The documentation of the amqp package recommends using separate connections for
+// publishing and subscribing.
+func OpenTopic(conn *amqp.Connection, name string) (*pubsub.Topic, error) {
+	// TODO(jba): support context.Context
+	t, err := newTopic(conn, name)
+	if err != nil {
+		return nil, err
+	}
+	return pubsub.NewTopic(t), nil
+
```
vangent: nit: extra blank line here?

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -0,0 +1,179 @@
...
+// OpenTopic returns a *pubsub.Topic corresponding to the named exchange. The
+// exchange must have been previously created (for instance, by using
+// amqp.Channel.ExchangeDeclare). For the model of Go Cloud Pub/Sub to make sense,
+// the exchange should be a fanout exchange, although nothing in this package
+// enforces that.
+//
+// OpenTopic uses the supplied amqp.Connection for all communication. It is the caller's
+// responsibility to establish this connection before calling OpenTopic, and to close
+// it when it is no longer in use.
```
vangent: Consider being more explicit here about the lifetime of `conn`. E.g., "... and to close it after all pubsub Topics using it have been Closed."

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil, err
+	}
+	return t, nil
+}
+
+// establishChannel creates an AMQP channel if necessary. According to the amqp
+// package docs, once an error is returned from the channel, it must be discarded and
+// a new one created.
+//
+// Must be called with the t.mu held.
```
vangent: Nit: The call to this function in `newTopic` does not hold `mu`. Probably fine, but contradicts this comment.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -0,0 +1,179 @@
...
+	t.ch = ch
+	// Get a Go channel which will hold acks from the server. The server
+	// will send an ack for each published message to confirm that it was received.
+	t.pubc = ch.NotifyPublish(make(chan amqp.Confirmation)) // NotifyPublish returns its arg
+	return nil
+}
+
+// SendBatch implements driver.SendBatch.
+func (t *topic) SendBatch(ctx context.Context, ms []*driver.Message) error {
+	// It is simplest to allow only one SendBatch at a time. Allowing concurrent
```
vangent: Outside this PR, but to me this is an indication that perhaps we should update `driver` to explicitly state that `SendBatch` will not be called concurrently, making driver implementations easier. At least we should be very explicit in the driver docstrings about whether this can happen and what the expectations are.

IIUC, we would enforce this by always setting the `max???` arg for the bundler to 1 (which it currently is), or getting rid of it.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -0,0 +1,179 @@
...
+	// confirmations and returns. We can go that route if performance warrants it.
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	if err := t.establishChannel(); err != nil {
+		return err
+	}
+
+	// Read from the channel established with NotifyPublish.
+	// Do so concurrently or we will be deadlock with the Publish RPC.
```
vangent: nit: "will be deadlocked" or "will deadlock".

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -0,0 +1,179 @@
...
+		if !ok {
+			errc <- errors.New("rabbitpubsub: ack failed on publish")
+		} else {
+			errc <- nil
+		}
+	}()
+
+	for _, m := range ms {
+		pub := toPublishing(m)
+		if err := t.ch.Publish(t.exchange, routingKey, mandatory, !immediate, pub); err != nil {
```
vangent: Why `!immediate`? That is confusing (to define a constant but only use it with `!`).

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -0,0 +1,179 @@
...
+					t.ch = nil // re-create the channel on next use
+					return
+				}
+				if !conf.Ack {
+					ok = false
+				}
+			}
+		}
+		if !ok {
+			errc <- errors.New("rabbitpubsub: ack failed on publish")
```
vangent: What happens if some sends were acked and others were not? Does `driver` say anything about this?

internal/pubsub/rabbitpubsub/rabbit_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (h *harness) Close() {
+	h.conn.Close()
+}
+
+// This test is important for the RabbitMQ driver because the underlying client is
+// poorly designed with respect to concurrency, so we must make sure to exercise the
+// driver with concurrent calls.
+//
+// We can't make this a conformance test at this time because there is no way
```
vangent: If there is no way to get the concrete type to exercise the driver with concurrent calls, then why is it important to test it?

internal/pubsub/rabbitpubsub/rabbit_test.go
vangent: I guess this is fine for now, but I feel like we need to do better here. E.g., maybe we need to make our own fake.

https://github.com/NeowayLabs/wabbit ?

vangent:
Also, tests are failing; you likely need to [update the proxy](https://github.com/google/go-cloud/tree/master/internal/proxy) with the new dependencies.

internal/pubsub/rabbitpubsub/rabbit_test.go
jba: Cool. Added issue #837.

internal/pubsub/rabbitpubsub/examples_test.go
jba: If I used log.Fatal above, then this test couldn't pass unless a RabbitMQ instance was running. (Equivalent of doing t.Skip in the "real" tests.)

internal/pubsub/rabbitpubsub/examples_test.go
jba: Done.

internal/pubsub/rabbitpubsub/examples_test.go
jba: If I deferred the close, I couldn't (a) check for errors or (b) legitimately print that the message was sent (since it might be buffered).

internal/pubsub/rabbitpubsub/rabbit.go
jba: done

internal/pubsub/rabbitpubsub/rabbit.go
jba: Fixed by grabbing mu. (Probably unnecessary, but simpler.)

internal/pubsub/rabbitpubsub/rabbit.go
jba: done.

internal/pubsub/rabbitpubsub/rabbit.go
jba: That would close off an important optimization. In cloud.google.com/go/pubsub, we make [many](https://github.com/GoogleCloudPlatform/google-cloud-go/blob/master/pubsub/topic.go#L439) concurrent publish RPCs to maximize throughput.

internal/pubsub/rabbitpubsub/rabbit.go
jba: I think it should be an error if even one message isn't sent successfully, which is what I do here.

I think we should work on the doc for driver.SendBatch. Currently it reads
```
// Only one RPC should be made to send the messages, and the returned                                                                                                                                                                     
// error should be based on the result of that RPC. Implementations                                                                                                                                                                       
// that send only one message at a time should return a non-nil error                                                                                                                                                                     
// if len(ms) != 1.  
```
But I don't think the concrete impl does anything to make sure `len(ms) ==  1`.

jba: I think that should say that ideally only one RPC is necessary, but more may be, and len(ms) may end up being > 1.

internal/pubsub/rabbitpubsub/rabbit_test.go
jba: Because we want drivers to support concurrent calls, even though we choose not to enable them at this time. See #820.

jba:
PTAL.

internal/pubsub/rabbitpubsub/rabbit.go
jba: @ijt

internal/pubsub/rabbitpubsub/rabbit.go

internal/pubsub/rabbitpubsub/rabbit.go
ijt: I think this is a design mistake: "Implementations                                                                                                                                                           that send only one message at a time should return a non-nil error                                                                                                                                                                     
if len(ms) != 1."

A better approach would be to have a method like `driver.Topic.MaxBatchSize()` that says how many messages can go in a batch.

internal/pubsub/rabbitpubsub/examples_test.go
vangent: Oh never mind, for some reason I missed the `Close` and thought you had left it out.

internal/pubsub/rabbitpubsub/rabbit_test.go
vangent: OK. I will add a comment to that #820 that this test should be moved to be a conformance test as part of it.

internal/pubsub/rabbitpubsub/rabbit.go
vangent: I filed #841 for this discussion.

vangent:
Tests are still failing....

```
2018/11/29 20:30:23 cannot connect to RabbitMQ: dial tcp 127.0.0.1:5672: connect: connection refused
--- FAIL: Example_publish (0.00s)
got:
want:
message sent
FAIL
```

jba:
The example cannot be a test, because the server may not be running.

Removed the `// Output` line.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	return amqp.Publishing{
+		Headers: h,
+		Body:    m.Body,
+	}
+}
+
+// IsRetryable implements driver.Topic.IsRetryable.
+func (*topic) IsRetryable(error) bool {
+	// TODO(jba): figure out what errors can be retried.
```
ijt: Filed https://github.com/google/go-cloud/issues/843

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -0,0 +1,182 @@
...
+	wait = false
+
+	// Always use the empty routing key. This driver expects to be used with topic
+	// exchanges, which disregard the routing key.
+	routingKey = ""
+
+	// If the message can't be enqueued, return it to the sender rather than silently dropping it.
+	mandatory = true
+
+	// If there are no waiting consumers, enqueue the message instead of dropping it
```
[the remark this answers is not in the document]
jba: Done (in forthcoming PR).

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -0,0 +1,182 @@
...
+	pubc <-chan amqp.Confirmation // Go channel for server acks of publishes
+}
+
+// Values for the amqp client.
+// See https://www.rabbitmq.com/amqp-0-9-1-reference.html.
+const (
+	// Many methods of the amqp client take a "no-wait" parameter, which
+	// if true causes the client to return without waiting for a server
+	// response. We always want to wait.
+	wait = false
```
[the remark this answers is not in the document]
jba: Then I'd have the problem Robert pointed out with immediate: I'd have
```
const noWait = true
```
and the calls would all say `!noWait`.

internal/pubsub/rabbitpubsub/rabbit.go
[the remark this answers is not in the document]

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -0,0 +1,182 @@
...
+
+	// Read from the channel established with NotifyPublish.
+	// Do so concurrently or we will deadlock with the Publish RPC.
+	// (The amqp package docs recommend setting the capacity of the channel
+	// to the number of messages to be published, but we can't do that because
+	// we want to reuse the channel for all calls to SendBatch.)
+	errc := make(chan error, 1)
+	go func() {
+		// Consume all the acknowledgments for the messages we are publishing.
+		// Since this method holds the lock, we expect exactly as many acks as messages.
```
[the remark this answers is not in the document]
jba: The user can pass in a context that has a timeout.

ijt:
@jba, calling it `wait` is inaccurate. Could we call it `immediate` and have a helper function like `func shouldWait() bool { return !immediate }`?

PR #834, opened 2018-11-29
internal/pubsub/drivertest: minor improvements

jba:
- Make the test table-driven, allowing harness creation/close to happen in one place.

- Remove the HarnessMaker type.

- Remove the testing.T arg from the harness maker (it can just return an error).

- Change empty slices to nil slices (more idiomatic).

### DIFF
--- a/internal/pubsub/drivertest/drivertest.go
+++ b/internal/pubsub/drivertest/drivertest.go
@@ -48,7 +48,7 @@ type Harness interface {
 //
 // newHarness is called exactly once per test to construct a Harness; its Close method
 // will be called when the test is complete.
-func RunConformanceTests(t *testing.T, newHarness func(context.Context) (Harness, error)) {
+func RunConformanceTests(t *testing.T, newHarness func(context.Context, *testing.T) (Harness, error)) {
 	ctx := context.Background()
 	for _, test := range []struct {
 		name string
@@ -59,7 +59,7 @@ func RunConformanceTests(t *testing.T, newHarness func(context.Context) (Harness
 		{"TestErrorOnReceiveFromClosedSubscription", testErrorOnReceiveFromClosedSubscription},
 		{"TestCancelSendReceive", testCancelSendReceive},
 	} {
-		h, err := newHarness(ctx)
+		h, err := newHarness(ctx, t)
 		if err != nil {
 			t.Fatal(err)
 		}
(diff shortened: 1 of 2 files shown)


vangent:
Please take a look at `blob` or `runtimevar` conformance tests; this code follows the same pattern. Consistency is good, so if you see something to be improved, it should be changed for all 3.

However, there are good reasons for most of the things you've changed. Until `pubsub` has a working driver implementation using record/replay, it may not be obvious.

> * Make the test table-driven, allowing harness creation/close to happen in one place.

This doesn't work well for tests with record/replay; the harness is recreated each time to allow for a separate record/replay file per test.

> * Remove the HarnessMaker type.

Same.

> * Remove the testing.T arg from the harness maker (it can just return an error).

The record/replay utilities we have need a `testing.T` (e.g., so they can fail the test during replay if no matching request is found).

> * Change empty slices to nil slices (more idiomatic).

This seems fine, separate PR?

jba:
> However, there are good reasons for most of the things you've changed. Until `pubsub` has a working driver implementation using record/replay, it may not be obvious.
> 
> > * Make the test table-driven, allowing harness creation/close to happen in one place.
> 
> This doesn't work well for tests with record/replay; the harness is recreated each time to allow for a separate record/replay file per test.
> 
> > * Remove the HarnessMaker type.
> 
> Same.

My commit message was misleading. Nothing has changed about the dynamic behavior. I just refactored the code to remove duplication.

I put back the testing.T arg.

internal/pubsub/drivertest/drivertest.go
```diff
@@ -44,42 +44,40 @@ type Harness interface {
...
-	})
+//
+// newHarness is called exactly once per test to construct a Harness; its Close method
+// will be called when the test is complete.
+func RunConformanceTests(t *testing.T, newHarness func(context.Context, *testing.T) (Harness, error)) {
+	ctx := context.Background()
+	for _, test := range []struct {
+		name string
+		f    func(context.Context, *testing.T, Harness)
+	}{
```
vangent: So I tried to follow this pattern before, but ended up getting rid of it in favor of the individual calls to `t.Run` because some of the `testXXX` functions ended up wanting to do multiple tests in a TDD way. For example:
https://github.com/google/go-cloud/blob/master/blob/drivertest/drivertest.go#L218

Also, some tests ended up needing slightly different arguments, e.g.:
https://github.com/google/go-cloud/blob/master/blob/drivertest/drivertest.go#L189

I suspect you will end up wanting the same (or at least, I don't think the pubsub drivertest coverage is good enough yet to be sure that you won't), so I still recommend we leave it alone at least for now.

jba: OK, closing.

PR #835, opened 2018-11-29
blob/all: add ReaderOptions to NewReader funcs

### LINKED ISSUE #765: blob: should NewReader take Options and support chunking?

vangent:
Fixes #765.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Imagine writing this to read from blob storage (like Google Cloud Storage or
 S3):
 
 ```go
-blobReader, err := bucket.NewReader(context.Background(), "my-blob")
+blobReader, err := bucket.NewReader(context.Background(), "my-blob", &blob.ReaderOptions{})
 ```
 
 and being able to run that code on any cloud you want, avoiding all the ceremony
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Imagine writing this to read from blob storage (like Google Cloud Storage or
 S3):
 
 ```go
-blobReader, err := bucket.NewReader(context.Background(), "my-blob", &blob.ReaderOptions{})
+blobReader, err := bucket.NewReader(context.Background(), "my-blob", nil)
 ```
 
 and being able to run that code on any cloud you want, avoiding all the ceremony
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -308,7 +308,7 @@ func (b *Bucket) As(i interface{}) bool {
 	return b.b.As(i)
 }
 
-// ReadAll is a shortcut for creating a Reader via NewReader with default
+// ReadAll is a shortcut for creating a Reader via NewReader with nil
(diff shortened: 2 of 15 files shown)


blob/blob.go
```diff
@@ -308,9 +308,10 @@ func (b *Bucket) As(i interface{}) bool {
 	return b.b.As(i)
 }
 
-// ReadAll is a shortcut for creating a Reader via NewReader and reading the entire blob.
+// ReadAll is a shortcut for creating a Reader via NewReader with default
+// ReaderOptions, and reading the entire blob.
```
vangent: Could also add `ReaderOptions` arg here and pass it through, WDYT?

README.md
```diff
@@ -17,7 +17,7 @@ Imagine writing this to read from blob storage (like Google Cloud Storage or
 S3):
 
 ```go
-blobReader, err := bucket.NewReader(context.Background(), "my-blob")
+blobReader, err := bucket.NewReader(context.Background(), "my-blob", &blob.ReaderOptions{})
```
zombiezen: `nil`?
```suggestion
blobReader, err := bucket.NewReader(context.Background(), "my-blob", nil)
```

blob/blob.go
zombiezen: I don't feel like we should until we know what `ReaderOptions` will contain.

blob/blob.go
```diff
@@ -366,8 +367,8 @@ func (b *Bucket) Attributes(ctx context.Context, key string) (Attributes, error)
 // is not found by the given key, which can be checked by calling IsNotExist.
```
zombiezen: Docs (here and below) should explicitly state that `nil` options are treated the same as passing the zero value.

blob/blob.go
```diff
@@ -505,6 +510,9 @@ type SignedURLOptions struct {
 	Expiry time.Duration
 }
 
+// ReaderOptions controls Reader behaviors.
```
zombiezen: Consider leaving a comment that it is empty for future compatibility.

blob/driver/driver.go
```diff
@@ -22,6 +22,9 @@ import (
 	"time"
 )
 
+// ReaderOptions controls Reader behaviors.
```
zombiezen: Same advice here: consider leaving a comment that it is empty for future compatibility.

PR #844, opened 2018-11-30
blob/tests: add more tests for OpenBucket, OpenURL

vangent:
This improves test coverage for `gcsblob` and `s3blob`.

Updates #675.


blob/gcsblob/gcsblob.go
```diff
@@ -122,9 +122,12 @@ type Options struct {
 }
 
 // openBucket returns a GCS Bucket that communicates using the given HTTP client.
-func openBucket(ctx context.Context, bucketName string, client *gcp.HTTPClient, opts *Options) (driver.Bucket, error) {
+func openBucket(ctx context.Context, bucketName string, client *gcp.HTTPClient, opts *Options) (*bucket, error) {
 	if client == nil {
-		return nil, errors.New("OpenBucket requires an HTTP client")
+		return nil, errors.New("gcsblob.OpenBucket: client is required")
+	}
+	if bucketName == "" {
```
shantuo: We used to have a simple check for bucket and object names, we then decided to remove it and rely on the backend service to check for the name. Is there any reason to add an empty name check now?

blob/s3blob/s3blob_test.go
```diff
@@ -145,3 +147,100 @@ func (verifyContentLanguage) ListObjectCheck(o *blob.ListObject) error {
...
+		url      string
+		wantName string
+		wantErr  bool
+	}{
+		{
+			url:      "s3://mybucket?region=foo",
+			wantName: "mybucket",
+		},
+		{
+			url:      "gs://mybucket2?region=bar",
```
shantuo: Can S3 open a bucket under `gs://`?

vangent: No, that was a typo. Fixed.

PR #847, opened 2018-11-30
internal/pubsub/rabbitpubsub: Subscription

jba:
Implement driver.Subscription: the ReceiveBatch and SendAcks methods.


internal/pubsub/rabbitpubsub/doc.go
```diff
@@ -20,5 +20,6 @@
 // A Pub/Sub topic is an AMQP exchange. The exchange kind should be "fanout" to match
 // the Pub/Sub model, although publishing will work with any kind of exchange.
 //
-// TODO(jba): describe the subscription side.
+// A Pub/Sub subscription is an AMQP queue. The queue should be bound to exchange
```
ijt: s/exchange/the exchange/

internal/pubsub/rabbitpubsub/examples_test.go
```diff
@@ -26,7 +27,7 @@ import (
 
 const rabbitURL = "amqp://guest:guest@localhost:5672/"
 
-func Example_publish() {
+func Example() {
 	// Connect to RabbitMQ.
 	conn, err := amqp.Dial(rabbitURL)
 	if err != nil {
```
vangent: Please verify that this test passes when you do have a local RabbitMQ running.

Consider adding some comments here explaining that the example code aborts early if rabbitURL isn't available.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -180,3 +183,166 @@ func (*topic) IsRetryable(error) bool {
 	// TODO(jba): figure out what errors can be retried.
 	return false
 }
+
+// OpenSubscription returns a *pubsub.Subscription corresponding to the named queue. The
+// queue must have been previously created (for instance, by using
+// amqp.Channel.QueueDeclare) and bound to an exchange.
+//
+// OpenSubscription uses the supplied amqp.Connection for all communication. It is
+// the caller's responsibility to establish this connection before calling OpenTopic,
```
vangent: OpenTopic -> OpenSubscription

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -180,3 +183,166 @@ func (*topic) IsRetryable(error) bool {
...
+			if !ok { // channel closed
+				s.ch = nil // re-establish the channel
+				if len(ms) > 0 {
+					return ms, nil
+				}
+				// TODO(jba): get more information for a better message
+				return nil, errors.New("delivery channel closed")
+			}
+			ms = append(ms, toMessage(d))
+		default:
```
vangent: Why have the `default`? IIUC, you could leave out the `default` and `loop`, and the `select` would block on `delc`; isn't that what you want? To avoid blocking forever you could move your  `<-time.After()` case up here and get rid of the lower `select`.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -180,3 +183,166 @@ func (*topic) IsRetryable(error) bool {
...
+	return &driver.Message{
+		Body:     d.Body,
+		AckID:    d.DeliveryTag,
+		Metadata: md,
+	}
+}
+
+// SendAcks implements driver.Subscription.SendAcks.
+func (s *subscription) SendAcks(ctx context.Context, ackIDs []driver.AckID) error {
+	s.mu.Lock()
```
vangent: It seems unfortunate that `SendAcks` is blocked by `ReceiveBatch` due to the mutex. Maybe they should be on different channels? OK to file an issue to follow up on this separately.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -180,3 +183,166 @@ func (*topic) IsRetryable(error) bool {
...
+	if err != nil {
+		return nil, err
+	}
+	return pubsub.NewSubscription(s), nil
+}
+
+type subscription struct {
+	conn     *amqp.Connection
+	queue    string // the AMQP queue name
+	consumer string // the client-generated name for this particular subscriber
```
vangent: Do you know what this is used for inside RabbitMQ? Maybe we should expose it in `OpenSubscription`...?

internal/pubsub/rabbitpubsub/doc.go
jba: Done.

internal/pubsub/rabbitpubsub/rabbit.go

internal/pubsub/rabbitpubsub/examples_test.go
jba: OK, I did something that makes it work as a test in both cases.

internal/pubsub/rabbitpubsub/rabbit.go
jba: Yes, that's how I had it originally, and that's why I added the TODO at the top of the function (which I forgot to remove). 

In the old way, if a message comes in every 40ms and `maxMessages` is 100, ReceiveBatch will stay around for 4s, collecting 100 messages. With the current code, it will return one message immediately. I think that's better.

The story is probably not over, though. The channel returned by Consume (`delc`) is unbuffered, which may mean that we'll never get more than one message even if others are waiting somewhere. I've changed my TODO to reflect that.

internal/pubsub/rabbitpubsub/rabbit.go
jba: #853.

internal/pubsub/rabbitpubsub/rabbit.go
jba: It's a client-side thing and only important for canceling a Consume. I forgot to add that cancel logic, but now I have.

internal/pubsub/rabbitpubsub/rabbit.go
vangent: WDYT about creating a `time.After` channel outside the loop instead? Something like:

```
maxTime := time.After(50*time.Millisecond)
done := false
for !done && len(ms) < maxMessages {
  select {
    case <-ctx.Done(): ...
    case d, ok := <-s.delc: ...
    case <-maxTime: ... done = true; break
  }
}
return ms, nil
```

That way the total time of the call is bounded, but we'll be more likely to collect multiple messages that are delivered close together in time.

jba:
PTAL.

internal/pubsub/rabbitpubsub/rabbit.go
vangent: WDYT?

jba: Messages might arrive every 26ms, but ReceiveBatch would return one message every 50ms, halving possible throughput.

vangent: The first call to `ReceiveBatch` would take 50ms and return 1 message, but the 2nd call would take 50ms and return 2 messages, no?

jba: Ah, right. So maybe that's okay.

Made the change. PTAL.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -180,3 +183,164 @@ func (*topic) IsRetryable(error) bool {
...
+			return nil, ctx.Err()
+
+		case d, ok := <-s.delc:
+			if !ok { // channel closed
+				s.ch = nil // re-establish the channel next time
+				if len(ms) > 0 {
+					return ms, nil
+				}
+				// TODO(jba): get more information for a better message
+				return nil, errors.New("delivery channel closed")
```
vangent: This is retryable, right?

jba: I need to understand better why it would be closed. That's one of my next tasks.

PR #848, opened 2018-11-30
internal/pubsub/rabbitpubsub: Subscription

jba:
Implement driver.Subscription: the ReceiveBatch and SendAcks methods.

* internal/pubsub/rabbitpubsub: get returned messages

RabbitMQ will "return" undeliverable messages. The client we're using
exposes these returns as a channel. Receive from this channel so we
can provide a good error when a message us undeliverable.


vangent:
This is a dupe of #847, right?

PR #849, opened 2018-11-30
pubsub: clarify and expand docstring+semantics for SendBatch

### LINKED ISSUE #841: pubsub: clarify and expand docstring+semantics for SendBatch
- What should happen is some message sends succeed and others fail?
- "Implementations that send only one message at a time should return a non-nil error                                                                                                                                                                     if len(ms) != 1." seems like a mistake. @ijt suggested "A better approach would be to have a method like driver.Topic.MaxBatchSize() that says how many messages can go in a batch."

ijt:
Fixes #841


internal/pubsub/driver/driver.go
```diff
@@ -61,10 +61,8 @@ type Topic interface {
 	// Only the Body and (optionally) Metadata fields of the Messages in ms
 	// should be set by the caller of SendBatch.
```
vangent: "should be set" -> "will be set"? This documentation is really for the driver implementor, not the caller.

internal/pubsub/driver/driver.go
```diff
@@ -61,10 +61,8 @@ type Topic interface {
...
 	//
-	// Only one RPC should be made to send the messages, and the returned
-	// error should be based on the result of that RPC. Implementations
-	// that send only one message at a time should return a non-nil error
-	// if len(ms) != 1.
+	// If any message in the batch fails to send, SendBatch should return an
+	// error.
 	//
 	// If there is a transient failure, this method should not retry but should
 	// return an error. The concrete API will take care of retry logic.
```
vangent: "return an error" -> "return an error for which IsRetryable returns true"?

internal/pubsub/driver/driver.go
```diff
@@ -61,10 +61,8 @@ type Topic interface {
...
 	// should be set by the caller of SendBatch.
 	//
-	// Only one RPC should be made to send the messages, and the returned
-	// error should be based on the result of that RPC. Implementations
-	// that send only one message at a time should return a non-nil error
-	// if len(ms) != 1.
+	// If any message in the batch fails to send, SendBatch should return an
+	// error.
 	//
 	// If there is a transient failure, this method should not retry but should
```
vangent: Instead of "SendBatch should be safe for concurrent access from multiple goroutines", how about "SendBatch may be called concurrently from multiple goroutines"?

vangent:
Can you also improve `ReceiveBatch` (in this PR or separately)?

internal/pubsub/driver/driver.go
jba: +1

internal/pubsub/driver/driver.go
jba: That's fine with me.

internal/pubsub/driver/driver.go
ijt: Done.

internal/pubsub/driver/driver.go
ijt: Done.

internal/pubsub/driver/driver.go
ijt: Done.

Issue #827, opened 2018-11-06
all: travis should verify that `go generate` output matches checked-in `wire_gen.go`


vangent:
Travis should verify that running `go generate` for the Wire injectors in our repo actually produces the checked-in `wire_gen.go`.

PR #850, opened 2018-11-30
samples/guestbook: go generate

isopov:
Sample sources and wire_gen.go seems to be unsynced. `go generate` produces diff. This PR is this diff.


isopov:
I signed it!

vangent:
Thanks for the PR! @zombiezen can you review, these look like changes due to things you've been doing.

BTW, we have #827 to run `wire diff` as part of our Travis tests, which will prevent this from happening again.

Issue #546, opened 2018-10-15
blob/fileblob: support signed URLs


zombiezen:
This is a feature we'd like to develop, but don't want to block #530's submission on it. It's important that our local implementations continue to provide the functionality for testing features.

`fileblob` should take in a URL or an interface to a signed URL creator, as well as providing some implementation (in-process or out-of-process HTTP handler) that serves signed blob URLs. We might want to make this optional, as some applications may very well be fine without having signed URLs for their blobs, but it should be Easy™️ to do.

/cc @clausti @vangent

vangent:
Next step on this is a design proposal for how to do it.

vangent:
This is fixed!

PR #852, opened 2018-11-30
blob/fileblob: add URLSigner to Options, set errorURLSigner by default

clausti:
Step 1 in adding a `FakeURLSigner` for local use and also allowing users to supply their own local implementation. 

updates https://github.com/google/go-cloud/issues/546


blob/fileblob/fileblob.go
```diff
@@ -74,15 +74,26 @@ func init() {
...
+	URLSigner URLSigner
+}
 
 type bucket struct {
-	dir string
+	dir       string
+	urlSigner URLSigner
+}
+
+func (b *bucket) path(key string) (string, error) {
```
zombiezen: nit: Let's move this method below the constructor functions so it's grouped with the other methods.

blob/fileblob/fileblob.go
```diff
@@ -565,6 +582,20 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
-	// TODO(Issue #546): Implemented SignedURL for fileblob.
+	path, err := b.path(key)
+	if err != nil {
+		return "", err
+	}
+
+	return b.urlSigner.SignedURL(ctx, key, path, opts)
+}
+
+type URLSigner interface {
```
zombiezen: Since this is introducing new exported API, it should have a doc comment explaining what the interface is. Usually for one-method interfaces like this, the method documentation goes in the interface doc comment, like in [`io.Reader`](https://godoc.org/io#Reader).

blob/fileblob/fileblob.go
```diff
@@ -565,6 +582,20 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+	SignedURL(ctx context.Context, key string, path string, opts *driver.SignedURLOptions) (string, error)
```
zombiezen: It would be good to give this method a different name, since it is a different signature than the bucket driver's `SignedURL` method now. There has been confusion in Go Cloud in the past with methods that have identical names but ever-so-slightly different functionality.

Perhaps `SignedFileURL`? If you're on board with that, I don't think the interface name has to change, as the full qualified identifier (`fileblob.URLSigner`) already contains the word "file".

blob/fileblob/fileblob.go
```diff
@@ -565,6 +582,20 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+}
+
+type errorURLSigner struct{}
```
zombiezen: And while this is unexported, it would be good to include a brief description of this type. Perhaps:

```suggestion
// errorURLSigner is a stub implementation of URLSigner used when no URLSigner is passed to OpenBucket.
type errorURLSigner struct{}
```

blob/fileblob/fileblob.go
vangent: Nit: This refactor (pulling out `path`) looks good, but it's unrelated to the rest of this PR, can you pull it out into a separate PR?

blob/fileblob/fileblob.go
```diff
@@ -74,15 +74,26 @@ func init() {
 }
 
 // Options sets options for constructing a *blob.Bucket backed by fileblob.
-type Options struct{}
+type Options struct {
+	URLSigner URLSigner
```
vangent: Please add docstring for URLSigner. In particular, IIUC it is only needed if you want to use `SignedURL`, right?

zombiezen:
When squashing this change, don't forget to mention in the commit message that this updates #546.

blob/fileblob/fileblob.go
```diff
@@ -91,7 +102,13 @@ func openBucket(dir string, _ *Options) (driver.Bucket, error) {
 	if !info.IsDir() {
 		return nil, fmt.Errorf("%s is not a directory", dir)
 	}
-	return &bucket{dir}, nil
+	b := &bucket{dir: dir}
```
vangent: I don't think it's going to be possible to support `SignedURL` if the user hasn't provided `opts.URLSigner`. You could conceivably provide a default for the signing part, but if they haven't provided a hook for HTTP requests, the URLs won't work.

How about this:

```
if opts == nil {
  opts = &Options{}
}
return &bucket{dir: dir, opts: opts}
```

(and update `bucket` to have `opts *Options` instead of the `URLSigner` explicitly).

blob/fileblob/fileblob.go
```diff
@@ -565,6 +582,20 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
 }
 
 func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error) {
-	// TODO(Issue #546): Implemented SignedURL for fileblob.
+	path, err := b.path(key)
+	if err != nil {
+		return "", err
+	}
+
+	return b.urlSigner.SignedURL(ctx, key, path, opts)
```
vangent: Why does SignedURL need `path`? Isn't `key` sufficient?

blob/fileblob/fileblob.go
vangent: It's not clear to me what we're asking the provider of this interface to do here. It looks like "everything". Can't we share more of the implementation? For example, we're going to want to look at `opts.Expiry` and include that in the URL somewhere, and then when we're serving the URL we'll want to verify that the expiration time is in the future. There's no reason to ask the user to implement that.

So, what are the things that we can reasonably expect a user to want to control? At least:

- The base URL / URL prefix for URLs.
- The signing mechanism (e.g., given a URL, produce a signature).
- The hook for handling incoming HTTP requests.

We can provide a default for the signing mechanism but not the other two. So maybe something like:

```
type Options struct {
  SignedURLPrefix string  // required to use SignedURL
  SignedURLHandler func(...) // required to use SignedURL
  SignedURLSigner func(u *url.URL) ([]byte, error) // defaults to MD5 hash or something
}
```

zombiezen: I think the approach here is [SOLID](https://en.wikipedia.org/wiki/SOLID) (pun intended) in that it allows different URL construction schemes (both lexical or perhaps involving RPCs) and makes the fileblob bucket only responsible for serving content. A bucket is then composed of an interchangeable signer that handles the details of signing, which could in fact have the knobs you suggest.

blob/fileblob/fileblob.go
clausti: Yeah, it's only necessary if you want to use SignedURL.

blob/fileblob/fileblob.go
clausti: PTAL

blob/fileblob/fileblob.go
clausti: Can confirm, was not planning on trying to support `SignedURL` if no `opts.URLSigner` was supplied. 

The plan was to leave `errorURLSigner` as the default, which would have no impact on users who don't intend or attempt to use signed urls, though I could give `errorURLSigner` a more informative message to the effect of 'you need to supply a URLSigner if you want to use this', as right now it just returns an error with the message "not implemented"?

blob/fileblob/fileblob.go
clausti: I chatted about this with @zombiezen, as I had the same question,  and his comment (please correct me if I'm paraphrasing this inaccurately) was that it will need the actual path info to serve the file, but since the dir information to assemble the full path is only available directly on the bucket, one cannot derive the full path from the key OR the key from the full path.

clausti:
Going to close this for the time being to write a design doc so that we can have clearer consensus without making one giant PR. 🙂

PR #855, opened 2018-11-30
internal/pubsub/rabbitpubsub: get returned messages

jba:
RabbitMQ will "return" undeliverable messages. The client we're using
exposes these returns as a channel. Receive from this channel so we
can provide a good error when a message is undeliverable.


internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -166,6 +139,54 @@ func (t *topic) SendBatch(ctx context.Context, ms []*driver.Message) error {
...
+	// Since SendBatch (the only caller of this method) holds the lock, we expect
+	// exactly as many acks as messages.
+	var err error
+	nAcks := 0
+	for nAcks < nMessages {
+		select {
+		case <-ctx.Done():
+			// Channel will be in a weird state (not all publish acks consumed, perhaps)
+			// so re-create it next time.
+			t.ch = nil
```
vangent: Does `t.ch` need an explicit close?

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -166,6 +139,54 @@ func (t *topic) SendBatch(ctx context.Context, ms []*driver.Message) error {
...
+			if !ok {
+				// Channel closed. Handled in the pubc case below. But set
+				// the channel to nil to prevent it from being selected again.
+				t.retc = nil
+			} else if err == nil {
+				// The message was returned from the server because it is unroutable.
+				// This will be the error we return, but continue so we drain all
+				// items from pubc. We don't need to re-establish the channel on this
+				// error.
+				err = fmt.Errorf("rabbitpubsub: message returned from %s: %s (code %d)",
```
vangent: Note: eventually we probably will want to expose `ret` for `ErrorAs`. Is `ret` an error? If so, consider just returning it instead of creating a new error. If not, we may need to wrap it in an error.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -166,6 +139,54 @@ func (t *topic) SendBatch(ctx context.Context, ms []*driver.Message) error {
...
+					ret.Exchange, ret.ReplyText, ret.ReplyCode)
+			}
+
+		case conf, ok := <-t.pubc:
+			if !ok {
+				// t.pubc was closed
```
vangent: Nit: trailing period.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -166,6 +139,54 @@ func (t *topic) SendBatch(ctx context.Context, ms []*driver.Message) error {
...
+				// Channel closed. Handled in the pubc case below. But set
+				// the channel to nil to prevent it from being selected again.
+				t.retc = nil
+			} else if err == nil {
+				// The message was returned from the server because it is unroutable.
+				// This will be the error we return, but continue so we drain all
+				// items from pubc. We don't need to re-establish the channel on this
+				// error.
+				err = fmt.Errorf("rabbitpubsub: message returned from %s: %s (code %d)",
+					ret.Exchange, ret.ReplyText, ret.ReplyCode)
```
vangent: Don't you need to increment `nAcks` here?

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -166,6 +139,54 @@ func (t *topic) SendBatch(ctx context.Context, ms []*driver.Message) error {
...
+				t.ch = nil // re-create the channel on next use
+				return err
+			}
+			nAcks++
+			if !conf.Ack && err == nil {
+				err = errors.New("rabbitpubsub: ack failed on publish")
+			}
+		}
+	}
+	return err
```
vangent: It might be more valuable to have the error include a count of successes/failures in the batch (e.g., "4 of 4 messages failed to be published, first error: %v"). WDYT?

vangent: You could even have a wrapped error type that you expose via `ErrorAs` that includes the full slice of errors.

internal/pubsub/rabbitpubsub/rabbit.go
jba: Good call, probably. 

Done.

internal/pubsub/rabbitpubsub/rabbit.go
jba: It's not an error.

Wait for the next PR before discussing error-handling. It is all about errors.

internal/pubsub/rabbitpubsub/rabbit.go
jba: No. The contract is that I will get exactly one item on pubc for each message I send, whether there's a problem or not, but I'll only get something on retc for returned messages.

internal/pubsub/rabbitpubsub/rabbit.go
jba: #889

internal/pubsub/rabbitpubsub/rabbit.go
vangent: I don't see the `close`, did you push?
Can you check the file for other places where `t.ch` may need to be closed? I think there's at least one more.

internal/pubsub/rabbitpubsub/rabbit.go
vangent: Interesting. Can you add a comment explaining this? It's not obvious; I assumed you'd either get a `pubc` message OR a `retc` message, and then outer loop of `nAcks < nMessages` looks like a bug. Maybe at the top of the loop?

jba:
> Can you check the file for other places where t.ch may need to be closed? I think there's at least one more.

I don't think there is. The other failures come from the server, so we can assume the channel is borked (and closed).

internal/pubsub/rabbitpubsub/rabbit.go
jba: Rewrote the comment just below the func line.

jba:
Waiting for review/LGTM from @ijt.

ijt:
Looking.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -169,6 +142,58 @@ func (t *topic) SendBatch(ctx context.Context, ms []*driver.Message) error {
 	return <-errc
 }
 
+// Read from the channels established with NotifyPublish and NotifyReturn.
+// Must be called with t.mu held.
+func (t *topic) receiveFromPublishChannels(ctx context.Context, nMessages int) error {
+	// Consume all the acknowledgments for the messages we are publishing, and also get returned messages.
```
ijt: nit: Reflow comment to 80 columns here and elsewhere.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -119,44 +121,15 @@ func (t *topic) SendBatch(ctx context.Context, ms []*driver.Message) error {
 		return err
 	}
 
-	// Read from the channel established with NotifyPublish.
-	// Do so concurrently or we will deadlock with the Publish RPC.
-	// (The amqp package docs recommend setting the capacity of the channel
+	// Receive from Go channels concurrently or we will deadlock with the Publish RPC.
+	// (The amqp package docs recommend setting the capacity of the Go channel
 	// to the number of messages to be published, but we can't do that because
 	// we want to reuse the channel for all calls to SendBatch.)
```
ijt: Why do we want to reuse the channel for all calls to SendBatch?

jba:
PTAL.

PR #858, opened 2018-11-30
internal/pubsub: remove ackIDBox

### LINKED ISSUE #814: pubsub: remove ackIDBox type
@jba says we can do without it.

jba:
Fixes #814.


internal/pubsub/acks_test.go
```diff
@@ -176,7 +176,7 @@ func TestTooManyAcksForASingleBatchGoIntoMultipleBatches(t *testing.T) {
 
 func TestAckDoesNotBlock(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
-	m := &driver.Message{}
+	m := &driver.Message{AckID: 0}
```
ijt: What does this have to do with removing ackIDBox?

jba: The reflect package chokes on nil interface values. I don't fully understand why, but AckIDs will never be nil in practice so it shouldn't matter except for tests.

ijt: Probably worth a comment.

jba: Done.

PR #860, opened 2018-11-30
pubsub/test: fix flaky TestConcurrentReceivesGetAllTheMessages test

vangent:
It was flaky before because sometimes a worker goroutine would not exit until after the test was done and the `defer Close` happened, so they'd get a `subscription closed` error, and `isCanceled` would return false. Now we consider all errors after we've received all messages as "expected".

I also noticed that the test was busy-looping while waiting for the workers to receive messages; I updated the test to use a WaitGroup for waiting for all messages to be processed instead of busy-waiting on `len(receivedMsgs)`.

Updates #829.


internal/pubsub/pubsub_test.go
```diff
@@ -129,76 +130,88 @@ func TestSendReceive(t *testing.T) {
...
-			t.Errorf("got %d for %q, want %d", v2, k, v)
+	got := 0
+	for i := 0; i < howManyToSend; i++ {
+		key := fmt.Sprintf("message #%d", i)
+		if val, ok := receivedMsgs[key]; !ok {
+			t.Errorf("never received message %q", key)
+		} else {
+			v, err := strconv.Atoi(val)
+			if err != nil {
+				t.Error(err)
```
jba: t.Fatal, since something is deeply wrong.

internal/pubsub/pubsub_test.go
```diff
@@ -129,76 +130,88 @@ func TestSendReceive(t *testing.T) {
...
+			}
+			got += v
 		}
 	}
+	if got != wantSum {
+		t.Errorf("got sum %d want %d", got, wantSum)
```
jba: nit: comma after first %d

PR #863, opened 2018-12-01
all: run three travis builds in parallel: win, mac, linux

### LINKED ISSUE #862: all: Travis takes 12m to run, want 5m
Even after splitting wire out into its own github repository, it's taking an excessive amount of time for our tests to run.

ijt:
Fixes #862


ijt:
This isn't needed. The builds are already parallel. Windows builds are holding up the show.

PR #868, opened 2018-12-03
pubsub/tests: simplify TestConcurrentReceivesGetAllTheMessages

### LINKED ISSUE #829: pubsub: TestConcurrentReceivesGetAllTheMessages fails intermittently

vangent:
Instead of keeping a map with counter (string->int) for each unique message key, keep a set (string->bool). The counter would never be anything other than 0 or 1; the test would hang if it failed to deliver them all.

Also, just verify the expected keys are in the set at the end instead of creating a second set containing them.

Fixes #829.


internal/pubsub/pubsub_test.go
```diff
@@ -185,17 +183,10 @@ func TestConcurrentReceivesGetAllTheMessages(t *testing.T) {
...
-	if sum != howManyToSend {
-		t.Errorf("received %d messages, want %d", sum, howManyToSend)
-	}
-	for k, v := range sentMsgs {
-		v2 := receivedMsgs[k]
-		if v2 != v {
-			t.Errorf("got %d for %q, want %d", v2, k, v)
+	for i := 0; i < howManyToSend; i++ {
+		key := fmt.Sprintf("message #%d", i)
+		if _, found := receivedMsgs[key]; !found {
```
jba: In this case, I think this is equivalent to
```
if !receivedMsgs[key] {
```

vangent: Yup; done.

PR #873, opened 2018-12-03
travis: only generate coverage data for the linux build

vangent:
Now we'll only run `-race` and coverage -> Coveralls for the `linux` build.

Updates #862.

### DIFF
--- a/internal/testing/runchecks.sh
+++ b/internal/testing/runchecks.sh
@@ -28,12 +28,17 @@ fi
 
 result=0
 
-# Run Go tests for the root, including coverage.
-go test -race -coverpkg=./... -coverprofile=coverage.out ./... || result=1
-if [ -f coverage.out ]; then
-  # Filter out test and sample packages.
-  grep -v test coverage.out | grep -v samples > coverage2.out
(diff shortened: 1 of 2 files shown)


internal/testing/runchecks.sh
```diff
@@ -28,12 +28,17 @@ fi
...
+# because they are slow, and Coveralls will only save the last one anyway.
+if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+  go test -race -coverpkg=./... -coverprofile=coverage.out ./... || result=1
+  if [ -f coverage.out ]; then
+    # Filter out test and sample packages.
+    grep -v test coverage.out | grep -v samples > coverage2.out
+    goveralls -coverprofile=coverage2.out -service=travis-ci
+  fi
+else
+  go test ./... || result=1
```
zombiezen: Optional: Running the race detector seems useful to run generally for CI. It doesn't get run frequently on dev workstations (being non-default).

```suggestion
  go test -race ./... || result=1
```

vangent: OK we can try that. There may be races that only happen on Windows for example. Done.

Issue #781, opened 2018-11-20
secrets: add a driver package


shantuo:
- [x] Decryptor
- [ ] Encryptor

PR #875, opened 2018-12-03
internal/secrets/driver: Add Crypter interface to group Encrypt() and Decrypt()

clausti:
Updates #781

### DIFF
--- a/internal/secrets/driver/driver.go
+++ b/internal/secrets/driver/driver.go
@@ -31,3 +31,9 @@ type Encrypter interface {
 	// Encrypt encrypts the plaintext and returns the cipher message.
 	Encrypt(ctx context.Context, plaintext []byte) ([]byte, error)
 }
+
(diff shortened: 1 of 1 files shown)


zombiezen:
How do you envision this interface being used? I suspect it to be useful for conformance tests, but it would be helpful to have context.

vangent:
Sorry, I took a look at that PR and it's still not obvious to me why `Crypter` is needed. Users should be able to use `Encrypter` and `Decrypter` independently, so why are the conformance tests requiring providers to expose a single unioned type?

clausti:
@vangent I think our rationale was that you kind of can't test decryption without having first encrypted something, but I have also closed the associated PR.

vangent:
> you kind of can't test decryption without having first encrypted something

Yes, that is true. It's hard to imagine writing a conformance test for one of the two interfaces in isolatino. But I don't think that that requirement for testing implies that the driver interface needs to have a unioned interface; the test can just require that the harness provide one of each.

PR #876, opened 2018-12-03
docs: document design decisions regarding Options structs in APIs

vangent:


internal/docs/design.md
```diff
@@ -107,6 +107,36 @@ symbols stable over time.
...
+    constructor. Inside a driver interface or in a concrete type like `blob`,
+    use more descriptive names like `ReaderOptions` or `WriterOptions`.
+-   When similar `Options` are part of a driver interface and also part of the
+    concrete type (e.g., `blob.WriterOptions`), duplicate the struct instead of
+    aliasing or embedding it, and copy the struct fields explicitly where
+    needed. This allows the godoc for each type to be tailored to the
+    appropriate audience (concrete type: users, driver interface: provider
+    implementors), and also allows the structs to diverge over time if
+    appropriate.
+-   Arguments that are optional, including ones for which a default value is
```
zombiezen: I recommend wording this more strongly. Anything that is required should not be in an options struct. Adding a new field to the options struct must have a reasonable default, else it is a breaking change.

internal/docs/design.md
```diff
@@ -107,6 +107,36 @@ symbols stable over time.
...
+    common, should generally go into the `Options` struct. Avoid adding
+    `Options` fields for which the default value doesn't make sense.
+-   Regarding empty `Options` structs: we considered only adding them the first
```
zombiezen: This prose feels like it should be part of the first paragraph when it is introducing the idea. The rest of the bullets work nicely to explain the guidelines efficiently.

internal/docs/design.md
```diff
@@ -107,6 +107,36 @@ symbols stable over time.
...
+    possible that we'll add arguments, add `Options`.
+-   Name the `Options` struct appropriately. `Options` is usually fine for
+    provider constructors since the package generally only exposes a
+    constructor. Inside a driver interface or in a concrete type like `blob`,
+    use more descriptive names like `ReaderOptions` or `WriterOptions`.
+-   When similar `Options` are part of a driver interface and also part of the
+    concrete type (e.g., `blob.WriterOptions`), duplicate the struct instead of
+    aliasing or embedding it, and copy the struct fields explicitly where
+    needed. This allows the godoc for each type to be tailored to the
+    appropriate audience (concrete type: users, driver interface: provider
```
zombiezen: It took me a while to parse this aside. I suggest:

```suggestion
    appropriate audience (e.g. end-users for the concrete type, provider implementors for the driver interface)
```

internal/docs/design.md
ijt: we considered only adding them the first -> we considered only adding them _when_ the first

internal/docs/design.md

internal/docs/design.md
vangent: It's a pretty long aside, and I think distracts from the main points. I pulled it out of the bullet list and left it at the end.

vangent:
PTAL.

internal/docs/design.md
```diff
@@ -107,6 +107,36 @@ symbols stable over time.
...
+-   When similar `Options` are part of a driver interface and also part of the
+    concrete type (e.g., `blob.WriterOptions`), duplicate the struct instead of
+    aliasing or embedding it, and copy the struct fields explicitly where
+    needed. This allows the godoc for each type to be tailored to the
+    appropriate audience (e.g. end-users for the concrete type, provider implementors for the driver interface)
+    implementors), and also allows the structs to diverge over time if
+    appropriate.
+-   Required arguments must not be in an `Options` struct.
+-   All fields of the `Options` struct must have reasonable defaults.
+
```
jba: Add:
- The argument should be of type `*Options`, so `nil` can be passed in the default case.

vangent: This is already submitted, but done in #884.

Issue #657, opened 2018-11-09
pubsub/driver: add As functions for Topic, Subscription, and Message


ijt:
This change was suggested by @vangent on pull #643:

> See blob.Driver for examples.
> 
> Message is used for both reading and writing, which makes it interesting for As. An As function As(i interface{}) bool makes sense for reading, but for writing you want to give access to the underlying type of whatever message is going to be sent to the server. In blob, we do this via WriterOptions.BeforeWrite. So you could:
> 
> Add a BeforeWrite-like function to Message, that's only used when writing
> Split Message into separate types for reading (with As) and writing (with BeforeWrite)
> Add an Options struct with BeforeWrite to SendBatch.

ijt:
... and I suppose it would make sense to have the same method on `Subscription` to handle the conversion for received messages.

vangent:
`Message.As` is better than `Subscription.MessageAs` for received messages.


For sent messages, I don't think your suggestion works for the actual lifecycle of the message objects. It definitely did not for `blob`. The lifecycle looks something like:

1. User creates a `pubsub.Message` and sets some fields.
2. User calls concrete type's `Send` and passes the message <blocks here>.
3. Concrete type converts `pubsub.Message` to `driver.Message` and adds the message to a batch.
4. ... at some later point, the batcher decides to send the batch and calls `driver.SendBatch`.
5. The driver converts the `driver.Message` to the provider-specific type and sends it.
6. User's `Send` returns.

Note that the provider-specific type is not created until 5). It definitely does not exist before 2) when the user could call something (they are blocked after 2) ). It is also desirable that the provider-specific type the user gets access to via `As` is filled in as much as possible (e.g., the `Body` field, etc.) so that they can override things if they want to. That doesn't happen until 5).

The `blob` solution is for the user to provide a callback function (in `Options`) that accepts the `As` function the user can use:

```
BeforeWrite func(asFunc func(interface{}) bool) error
```

jba:
Looks like we'll need `TopicOptions.BeforeSend`.

I'm not really sure on why we need `asFunc`. Couldn't BeforeSend take an `interface{}` which holds the actual message type? The user of BeforeSend must already know that type to call `asFunc`.

jba:
BTW, I think the same criticism applies to blob's BeforeWrite.

ijt:
[...]

PR #877, opened 2018-12-03
pubsub: add As methods to Topic and Subscription

ijt:
Work towards #657.

Support for `Message`s will go in a separate PR.


internal/pubsub/gcppubsub/as_test.go
```diff
@@ -0,0 +1,49 @@
...
+		t.Errorf("cast failed for %T", &c)
+	}
+}
+
+func TestSubscriptionAsCanFail(t *testing.T) {
+	proj := ""
+	subName := "some-sub"
+	sub := gcppubsub.OpenSubscription(context.Background(), client, proj, subName, nil)
+	var c raw.SubscriptionClient
+	if top.As(&c) {
```
jba: s/top/sub/

internal/pubsub/gcppubsub/as_test.go
```diff
@@ -0,0 +1,49 @@
...
+		t.Errorf("cast succeeded for %T, want failure", &c)
+	}
+}
+
+func TestSubscriptionAsCanSucceed(t *testing.T) {
+	proj := ""
+	subName := "some-sub"
+	top := gcppubsub.OpenSubscription(context.Background(), client, proj, subName, nil)
```
jba: sub

internal/pubsub/gcppubsub/as_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"context"
+	"testing"
+
+	raw "cloud.google.com/go/pubsub/apiv1"
+	"github.com/google/go-cloud/internal/pubsub/gcppubsub"
+)
+
+func TestTopicAsCanFail(t *testing.T) {
```
jba: Combine this test with the next, since the setup is the same.

internal/pubsub/drivertest/drivertest.go
```diff
@@ -34,10 +34,18 @@ type Harness interface {
 	// MakeTopic makes a driver.Topic for testing.
 	MakeTopic(ctx context.Context) (driver.Topic, error)
 
+	// MakeNonexistentTopic makes a driver.Topic referencing a topic that
```
jba: Ignoring (from another PR?)

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -104,6 +97,16 @@ func (t *topic) IsRetryable(error) bool {
 	return false
 }
 
+// As implements driver.Topic.As.
+func (t *topic) As(i interface{}) bool {
+	t2, ok := i.(*topic)
```
jba: I thought *raw.PublisherClient?

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -188,3 +184,13 @@ func (s *subscription) IsRetryable(error) bool {
 	// The client handles retries.
 	return false
 }
+
+// As implements driver.Subscription.As.
+func (s *subscription) As(i interface{}) bool {
+	s2, ok := i.(*subscription)
```
jba: ditto

internal/pubsub/gcppubsub/as_test.go
ijt: Done.

internal/pubsub/gcppubsub/as_test.go
ijt: Done.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Done.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Done.

ijt:
PTAL

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -74,6 +74,17 @@ func (t *topic) IsRetryable(error) bool {
 	return false
 }
 
+// As implements driver.Topic.As.
```
vangent: Please document the types supported for each `As` function in the package godoc; see [gcsblob](https://github.com/google/go-cloud/blob/master/blob/gcsblob/gcsblob.go#L28) for an example.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -184,6 +184,17 @@ func (*topic) IsRetryable(error) bool {
 	return false
 }
 
+// As implements driver.Topic.As.
```
vangent: Same: document the types supported by the various `As` functions in the package godoc.

internal/pubsub/gcppubsub/gcppubsub.go
ijt: Why wouldn't the documentation for what `As` does go right above `As`?

vangent: Because the struct you are documenting here is not exported, and therefore it won't show up in godoc at all.

ijt: For the rabbit test, I'm not sure how to check that the connection pointer has the right value anymore, without making `pubsub.Topic.driver` public.

vangent: Look at the generated godoc for the package (e.g., https://godoc.org/github.com/google/go-cloud/blob/s3blob for s3blob). If you put the documentation on `As`, it won't show up in godoc, because `As` is a function on an unexported struct. The only exported functions in driver packages are generally the constructors.

So, if you want users to see it, you need to put the documentation on the constructors or in the package docs.

internal/pubsub/drivertest/drivertest.go
```diff
@@ -55,8 +56,41 @@ type Harness interface {
 // It is called exactly once per test; Harness.Close() will be called when the test is complete.
 type HarnessMaker func(ctx context.Context, t *testing.T) (Harness, error)
 
+// AsTest represents a test of As functionality.
+// The conformance test:
+// 1. Calls TopicCheck.
+// 2. Calls SubscriptionCheck.
+type AsTest interface {
```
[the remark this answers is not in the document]
vangent: @ijt is copying the model I used for `blob`. It allows both the `nil` test to be built-in as a conformance test, but also for providers to add more specific ones.

internal/pubsub/gcppubsub/gcppubsub_test.go
```diff
@@ -89,5 +90,36 @@ func (h *harness) Close() {
 }
 
 func TestConformance(t *testing.T) {
-	drivertest.RunConformanceTests(t, newHarness)
+	asTests := []drivertest.AsTest{gcpAsTest{}}
+	drivertest.RunConformanceTests(t, newHarness, asTests)
```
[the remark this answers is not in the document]
vangent: The `As` part of the test is driver-specific, but there can be a fair amount of set-up required to construct the types needed for `As` to be called on. So far it's just a `Topic` and a `Subscription` so it's not bad, but soon there will be tests for `As` on a `Message` before it is sent, `As` on a received `Message`, `As` on an `error`.... I think there's value in doing that once in the conformance tests and letting the driver test focus on just the `As` part of it. See the `blob` [drivertest](https://github.com/google/go-cloud/blob/master/blob/drivertest/drivertest.go#L1553) for an example.

Your code snippet would work for some of the kinds of tests, but not all; specifically, the "before write" hooks. Again, see `blob`'s drivertest for examples.

internal/pubsub/drivertest/drivertest.go
```diff
@@ -113,9 +158,7 @@ func testNonExistentSubscriptionSucceedsOnOpenButFailsOnSend(t *testing.T, newHa
 
 	ds, err := h.MakeNonexistentSubscription(ctx)
 	if err != nil {
-		// Failure shouldn't happen for non-existent subscriptions until messages are
-		// received from them.
-		t.Fatalf("creating a local subscription that doesn't exist on the server: %v", err)
+		t.Skipf("failed to make non-existent subscription: %v", err)
```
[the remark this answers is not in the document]
vangent: Ping on this comment, I think it's the last remaining open comment.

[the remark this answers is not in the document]
vangent: I don't think that can be true, as you're not changing anything about `MakeNonexistentSubscription` in this PR, and the test is passing right now....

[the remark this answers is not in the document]
vangent: Ah, I suspect `rabbitpubsub` is skipping the test if a RabbitMQ server isn't running locally, which is the case on Travis. The test likely fails if you do have one running locally. I'd rather fix this on the rabbit side than here....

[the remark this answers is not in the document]
ijt: I'm open to suggestions but I don't understand what you're asking for.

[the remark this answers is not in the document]
vangent: Revert this change (put back the comments and the `t.Fatalf`, and remove the `t.Skip`). It will pass on Travis. It is only failing for you locally because you (presumably) have a RabbitMQ server running locally; HEAD will also fail in the same way.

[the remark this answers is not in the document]
ijt: Filed #891.

Issue #811, opened 2018-11-26
secrets: add a in-memory implementation


shantuo:
Use crypto package.

PR #878, opened 2018-12-03
internal/secrets: add localsecrets implementation and conformance tests

clausti:
updates #811


clausti:
Following discussions about local fileblob signedURL implementations, I am closing this  because I think it either needs to be renamed or refactored to take a secret.

PR #880, opened 2018-12-03
tests: run "wire diff" in addition to "wire check"

### LINKED ISSUE #827: all: travis should verify that `go generate` output matches checked-in `wire_gen.go`

vangent:
This ensures that the checked-in `wire_gen.go` files in our repository match what a fresh run of `wire` would generate.

Fixes #827.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,7 +38,7 @@ require (
 	github.com/google/btree v0.0.0-20180813153112-4030bb1f1f0c // indirect
 	github.com/google/go-cmp v0.2.0
 	github.com/google/martian v2.1.0+incompatible // indirect
-	github.com/google/wire v0.1.0
+	github.com/google/wire v0.2.0
 	github.com/googleapis/gax-go v2.0.0+incompatible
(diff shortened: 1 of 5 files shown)


zombiezen:
I would argue that both should be run: `wire check` ensures that any exported provider sets pass analysis, while `wire diff` ensures that the generated code has not drifted.

vangent:
> I would argue that both should be run: `wire check` ensures that any exported provider sets pass analysis, while `wire diff` ensures that the generated code has not drifted.

IIUC, `wire diff` checks the "provider sets pass analysis" too, no? Are there scenarios where `wire.Load` will fail but `wire.Generate` succeeds?

zombiezen:
The roots of `wire.Generate` are any injectors discovered in the named packages. The roots of `wire.Load` are any provider sets found in top-level variable declarations in the named packages. If a provider set is not used in an injector (i.e. a library provider set), then `wire.Generate` will not perform analysis.

vangent:
Ah, I see. OK, restoring `wire check`.

PR #882, opened 2018-12-04
pubsub: change Subscription.Close() to Shutdown(ctx)

### LINKED ISSUE #766: pubsub: Subscription.Shutdown(ctx)
At one point we considered a `Shutdown` method for `Subscription` that would replace `Close`. It would take a context so users could choose between allowing the outstanding `SendAcks` to complete, and stopping them (via cancel or timeout).

In the current implementation there is no way to stop a `SendAcks` call (since it takes `context.Background`). That's probably fine, but `Shutdown` is an alternative that gives the user more control.

ijt:
Fixes #766


internal/pubsub/acks_test.go
```diff
@@ -266,3 +263,26 @@ func TestConcurrentDoubleAckCausesPanic(t *testing.T) {
...
+		q: []*driver.Message{m},
+		sendAcks: func(ctx context.Context, ackIDs []driver.AckID) error {
+			<-ctx.Done()
+			return ctx.Err()
+		},
+	}
+	sub := pubsub.NewSubscription(ds)
+	defer func() {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Millisecond)
+		sub.Shutdown(ctx)
```
jba: This should cancel the sendAcks call.

internal/pubsub/acks_test.go
```diff
@@ -266,3 +263,26 @@ func TestConcurrentDoubleAckCausesPanic(t *testing.T) {
 		t.Errorf("panics = %d, want %d", panics, 1)
 	}
 }
+
+func TestSubShutdownCanBeCanceledEvenWithHangingSendAcks(t *testing.T) {
```
jba: This test should verify that canceling the Shutdown context results in:
- outstanding SendAcks calls (of which there must be one) returning quickly
- Shutdown itself returning quickly

So there should be a `cancel` call somewhere, with Shutdown running in a goroutine and then closing or sending to a channel, and lastly something like
```
select {
    case <-shutdownDone:
    case <-time.After(5*time.Second):
           t.Fatal(...)
}
```
Yes, it's time sensitive, but the idea is to avoid the entire test running for 10 minutes and then timing out.

internal/pubsub/pubsub.go
```diff
@@ -197,12 +200,21 @@ func (s *Subscription) getNextBatch(ctx context.Context) error {
 }
 
 // Close flushes pending ack sends and disconnects the Subscription.
-func (s *Subscription) Close() error {
+func (s *Subscription) Shutdown(ctx context.Context) error {
+	c := make(chan struct{})
```
jba: I think this is all correct, but I find it hard to think about because the order in which things happen is complicated (first the function returns, then the channel gets closed, then the goroutine exits). I think the following is equivalent:
```
s.mu.Lock()
s.err = ...
s.mu.Unlock()
c := make(chan struct{})
go func() { defer close(c); s.ackBatcher.Shutdown() }
select {
case <-ctx.Done();
    s.cancel()
    return ctx.Err()
case <-c:
   return nil
}
```
except that it's easier to think about: "shut down the batcher, then wait for either that, or ctx.Done."

jba: One more thing, you're always supposed to call `cancel`, so I think the second select case should also say `s.cancel`. Or maybe:
```
...
select {
case <-ctx.Done():
case <-c:
}
s.cancel()
return ctx.Err()
```

internal/pubsub/acks_test.go
ijt: It does. `Subscription.Shutdown` has a call to `s.cancel` that runs when the passed in `ctx` is done. That gets triggered when the 1 msec timeout happens. `s.cancel` is the `cancel` func for the context passed into all the calls to `SendAcks`.

internal/pubsub/acks_test.go
ijt: Done.

internal/pubsub/pubsub.go

PR #885, opened 2018-12-04
blob/test: add/extend tests to improve concrete type coverage

vangent:


blob/blob_test.go
```diff
@@ -144,15 +146,21 @@ func (b *erroringBucket) SignedURL(ctx context.Context, key string, opts *driver
 // wrapped exactly once by the concrete type.
 func TestErrorsAreWrapped(t *testing.T) {
 	ctx := context.Background()
+	buf := bytes.Repeat([]byte{65}, sniffLen)
```
shantuo: nit: s/[]byte{65}/[]byte{'A'}

vangent: Yea. Done.

PR #888, opened 2018-12-04
all: only run Windows Travis builds on cron, not during pre-commit checks

### LINKED ISSUE #862: all: Travis takes 12m to run, want 5m

ijt:
Fixes #862.


ijt:
The Travis check took 5m to run. https://github.com/google/go-cloud/pull/888/checks?check_run_id=37705476

ijt:
Reassigning to @jba since @vangent is TAing a Go class today.

internal/testing/runchecks.sh
```diff
@@ -28,6 +28,11 @@ fi
 
 result=0
 
+if [[ "$TRAVIS_OS_NAME" == "windows" && "$TRAVIS_EVENT_TYPE" != "cron" ]]; then
```
vangent: Could you make this `== "pullrequest"` or whatever to make it opt-out instead of opt-in?

ijt: Done.

PR #892, opened 2018-12-05
server/sdserver: add cleanup func for Stackdriver exporter

### LINKED ISSUE #867: server/sdserver: NewExporter should return a cleanup function for Exporter.Flush()
### Describe the bug

Seems like with the current design of providing a StackDriver exporter, it's the clients responsibility to inject an exporter, and then remember to call `contrib.go.opencensus.io/exporter/stackdriver.Exporter.Flush()`. While `wire` supports cleanup, I thought it maybe good idea to augment `NewExporter` to also provide a cleanup function.

Not sure how this will affect existing users though; but, maybe, we can added a new provider that supports this.

## Expected behavior

When one inject a `stackdriver.Exporter`, she should expect the cleanup function should be returned as well so remind of her to call to flush.

### Version

v0.6.0

### Additional context

I'm happy to put together a PR if this is going along the right direction. :)

garukun:
Adds a cleanup function fto the Stackdriver exporter, so that client
doesn't have to be aware of exporter.Flush() call.

This change also aligns to xrayserver.NewExporter's return signature.

Samples are updated and regenerated as well.

Fixes #867. 

NOTE: I can't seem to find specific tests needed to validate `sdserver`, (or `xrayserver`). Not sure if I had missed something obvious.

### DIFF
--- a/server/sdserver/server.go
+++ b/server/sdserver/server.go
@@ -44,6 +44,9 @@ var Set = wire.NewSet(
 )
 
 // NewExporter returns a new OpenCensus Stackdriver exporter.
+//
+// The second return value is a Wire cleanup function that calls Flush
+// on the exporter.
 func NewExporter(id gcp.ProjectID, ts gcp.TokenSource, mr monitoredresource.Interface) (*stackdriver.Exporter, func(), error) {
 	tokOpt := option.WithTokenSource(oauth2.TokenSource(ts))
 	exp, err := stackdriver.NewExporter(stackdriver.Options{
--- a/server/sdserver/server.go
+++ b/server/sdserver/server.go
@@ -44,14 +44,23 @@ var Set = wire.NewSet(
 )
 
 // NewExporter returns a new OpenCensus Stackdriver exporter.
-func NewExporter(id gcp.ProjectID, ts gcp.TokenSource, mr monitoredresource.Interface) (*stackdriver.Exporter, error) {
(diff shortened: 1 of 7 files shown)


server/sdserver/server.go
```diff
@@ -44,14 +44,23 @@ var Set = wire.NewSet(
 )
 
 // NewExporter returns a new OpenCensus Stackdriver exporter.
```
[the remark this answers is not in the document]
garukun: Done.

garukun:
Thanks for reviewing :)

server/sdserver/server.go
```diff
@@ -44,14 +44,26 @@ var Set = wire.NewSet(
...
 	})
+	if err != nil {
+		return nil, nil, err
+	}
+
+	cleanup := func() {
+		exp.Flush()
+	}
+
+	return exp, cleanup, err
```
shantuo: nit: it seems you can just `return exp, exp.Flush(), err`

garukun: You meant
```go
return exp, func() { exp.Flush() }, err
```
?

shantuo: My bad, either like you said or just `return exp, exp.Flush, err`

garukun: Thanks, I ended up with `func() { exp.Flush() }` since its style would be more consistent with `xrayserver`

Issue #891, opened 2018-12-05
pubsub: rabbitpubsub conformance test fails if rabbit server is running locally


ijt:
### Describe the bug

`go test` fails in `rabbitpubsub` if a local instance of RabbitMQ is running.

### To Reproduce

```sh
cd internal/pubsub/rabbitpubsub
go test
--- FAIL: TestConformance (0.13s)
    --- FAIL: TestConformance/TestNonExistentSubscriptionSucceedsOnOpenButFailsOnSend (0.00s)
        drivertest.go:118: creating a local subscription that doesn't exist on the server: unimplemented
FAIL
exit status 1
FAIL    github.com/google/go-cloud/internal/pubsub/rabbitpubsub 0.361s
```

## Expected behavior

`PASS`

### Version

b2ef1a6b9c08bd6cc4c1e1e83a6a089ca77cd712

PR #893, opened 2018-12-05
internal/pubsub/rabbitpubsub: succeed on open of non-existent resource

jba:
OpenTopic and OpenSubscription now succeed even if the resource doesn't exist.
They no longer need to return an error.

Also, implement the test harness MakeNonexitentSubscription method.

### DIFF
--- a/internal/pubsub/rabbitpubsub/rabbit.go
+++ b/internal/pubsub/rabbitpubsub/rabbit.go
@@ -45,10 +45,10 @@ const (
 )
 
 // OpenTopic returns a *pubsub.Topic corresponding to the named exchange. The
-// exchange must have been previously created (for instance, by using
-// amqp.Channel.ExchangeDeclare). For the model of Go Cloud Pub/Sub to make sense,
-// the exchange should be a fanout exchange, although nothing in this package
-// enforces that.
+// exchange should already exist (for instance, by using
+// amqp.Channel.ExchangeDeclare), although this won't be checked until the first call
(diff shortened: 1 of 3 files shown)


ijt:
I'm guessing this fixes #891.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -228,18 +211,12 @@ type subscription struct {
...
-	s := &subscription{
+func newSubscription(conn *amqp.Connection, name string) *subscription {
+	return &subscription{
 		conn:     conn,
 		queue:    name,
 		consumer: fmt.Sprintf("c%d", atomic.AddInt64(&nextConsumer, 1)),
 	}
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	if err := s.establishChannel(); err != nil {
```
ijt: Doesn't this call have to be done somewhere else now? I don't see it.

jba: It's in ReceiveBatch and SendAcks.

jba:
I'll add the Fixes line before I submit.

jba:
Awaiting @ijt LGTM.

jba:
@ijt PTAL

internal/pubsub/rabbitpubsub/rabbit.go
ijt: Ah, right.

PR #897, opened 2018-12-05
pubsub: add Topic.MessageAs and Subscription.MessageAs

### LINKED ISSUE #657: pubsub/driver: add As functions for Topic, Subscription, and Message

ijt:
Fixes #657

### DIFF
--- a/internal/pubsub/driver/driver.go
+++ b/internal/pubsub/driver/driver.go
@@ -84,6 +84,13 @@ type Topic interface {
 	// https://github.com/google/go-cloud/blob/master/internal/docs/design.md#as
 	// for more background.
 	As(i interface{}) bool
+
+	// As allows providers to expose provider-specific types.
+	//
+	// See
+	// https://github.com/google/go-cloud/blob/master/internal/docs/design.md#as
+	// for more background.
+	MessageAs(m *Message, i interface{}) bool
 }
 
 // Subscription receives published messages.
@@ -130,4 +137,11 @@ type Subscription interface {
 	// https://github.com/google/go-cloud/blob/master/internal/docs/design.md#as
 	// for more background.
 	As(i interface{}) bool
+
+	// As allows providers to expose provider-specific types.
+	//
+	// See
+	// https://github.com/google/go-cloud/blob/master/internal/docs/design.md#as
+	// for more background.
+	MessageAs(m *Message, i interface{}) bool
 }
--- a/internal/pubsub/gcppubsub/gcppubsub.go
+++ b/internal/pubsub/gcppubsub/gcppubsub.go
@@ -94,7 +94,7 @@ func (t *topic) MessageAs(m *driver.Message, i interface{}) bool {
 	if !ok {
 		return false
 	}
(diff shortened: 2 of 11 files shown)


internal/pubsub/driver/driver.go
```diff
@@ -84,6 +84,13 @@ type Topic interface {
 	// https://github.com/google/go-cloud/blob/master/internal/docs/design.md#as
 	// for more background.
 	As(i interface{}) bool
+
+	// As allows providers to expose provider-specific types.
+	//
+	// See
+	// https://github.com/google/go-cloud/blob/master/internal/docs/design.md#as
+	// for more background.
+	MessageAs(m *Message, i interface{}) bool
```
jba: So I don't think this is right, as discussed on #657.

internal/pubsub/gcppubsub/gcppubsub.go
```diff
@@ -160,3 +171,15 @@ func (s *subscription) As(i interface{}) bool {
 	*c = s.client
 	return true
 }
+
+// MessageAs implements driver.Subscription.MessageAs.
+func (s *subscription) MessageAs(m *driver.Message, i interface{}) bool {
```
jba: I think converting back and forth is ultimately going to be lossy and unsatisfying.

If we don't make `driver.Message` an interface, then at least it should have a field
```
Raw interface{}
```
that drivers can populate with the raw value.

internal/pubsub/pubsub.go
```diff
@@ -237,6 +249,14 @@ func (s *Subscription) As(i interface{}) bool {
 	return s.driver.As(i)
 }
 
+// MessageAs converts a message to an implementation-specifc message
+// type, returning true or false depending on whether the conversion
+// succeeded.
+func (s *Subscription) MessageAs(m *Message, i interface{}) bool {
+	dm := toDriverMessage(m)
```
jba: Hmm, at this point we've already lost information... not sure how to do this. I guess `pubsub.Message` should hold a `driver.Message` which should hold the raw driver message type.

internal/pubsub/gcppubsub/gcppubsub.go
vangent: That is not sufficient, as one of the benefits of `As` is that the provider can support multiple raw types. It needs to embed a field here that's an `asFunc`; see `blob.Attributes` for an example.

ijt:
I think this PR is a dead-end. Closing it.

PR #900, opened 2018-12-06
docs: add CNAME file

jba:
The CNAME file is part of configuring GitHub Pages to serve from a custom domain.

The [GitHub Help page](https://help.github.com/articles/adding-or-removing-a-custom-domain-for-your-github-pages-site/)
describes how to do this via the UI, but that doesn't work: GitHub complains about committing
to a protected branch. So instead we create CNAME file manually. It's described (vaguely) at
https://help.github.com/articles/troubleshooting-custom-domains.

### DIFF
--- a/docs/CNAME
+++ b/docs/CNAME
@@ -1,0 +1,1 @@
+gocloud.dev


zombiezen:
Apologies if I missed an update to plan-of-record here. Is the plan to still have the Travis build generate the site? If that's the case, then I think the CNAME file needs to be part of the `gh-pages` branch that gets created, not in the docs source directory.

jba:
Ross, you're saying that using master/docs to host the page is incompatible with doing it from Travis? I hadn't realized that.

jba:
I've just been focused on verifying that the new domain serves the right pages, I haven't gotten to the automation yet.

zombiezen:
Ah gotcha. It could be done, but it's IMO easier to have Travis write to a separate branch.

jba:
Abandoning. We'll use gh-pages instead.

PR #904, opened 2018-12-06
internal/pubsub/rabbitpubsub: use errors from channel-closed notifications

jba:
Get notified when an AMQP channel closes, so we can produce better
error messages.


ijt:
Can you add something to the description comparing the old error messages to the new ones?

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -184,6 +199,27 @@ func (t *topic) receiveFromPublishChannels(ctx context.Context, nMessages int) e
 	return err
 }
 
+// Return the error from a Go channel monitoring the closing of an AMQP channel.
+// closec must have been registered via Channel.NotifyClose.
+// When closeErr is called, we expect closec to be closed. If it isn't, we also
+// consider that an error.
+func closeErr(closec <-chan *amqp.Error) error {
+	select {
+	case aerr := <-closec:
```
ijt: Why not just `err`?

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -299,8 +344,13 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*dr
 				if len(ms) > 0 {
 					return ms, nil
 				}
-				// TODO(jba): get more information for a better message
-				return nil, errors.New("delivery channel closed")
+				// t.closec must be closed too. See if it has an error.
```
ijt: t.closec -> s.closec?

internal/pubsub/rabbitpubsub/rabbit.go
jba: To make it a little more clear that it's not of type `error`.

internal/pubsub/rabbitpubsub/rabbit.go

jba:
> Can you add something to the description comparing the old error messages to the new ones?

How can I update just the commit message? On gerrit I use commit --amend, but I think that will lose comments on github.

internal/pubsub/rabbitpubsub/rabbit.go
ijt: Oh, I see, it only satisfies `error`. Okay.

ijt:
`git commit --amend` locally, then `git push --force-with-lease` .

More details here about how to change the commit msg on an older commit: https://help.github.com/articles/changing-a-commit-message/.

internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -184,6 +199,27 @@ func (t *topic) receiveFromPublishChannels(ctx context.Context, nMessages int) e
...
+		// returned it directly (effectively assigning it to a variable of
+		// type error), then the return value would not be a nil interface
+		// value even if aerr was a nil pointer, and that would break tests
+		// like "if err == nil ...".
+		if aerr == nil {
+			return nil
+		}
+		return aerr
+	default:
+		return errors.New("rabbitpubsub: channel not closed")
```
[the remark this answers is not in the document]
jba: No, "channel closed with no error given" is the `aerr == nil` check above. (The AMQP channel is closed, which closes the Go channel, but without an `amqp.Error`.)

This line is a bug or behavior change in the underlying client, where one of the Go channels (e.g. `pubc`) is closed but `closec` isn't. I don't know how much detail it's worth going into in the error message; the important thing is that we can find this line of code when a user reports the error. 

I'll change it to something better.

Issue #905, opened 2018-12-06
blob: figure out a design/strategy for provider-specific string restrictions


vangent:
@vsaroopchand has discovered two things while working on Azure support:

1. Azure only allows C# identifiers in Metadata keys.
2. Azure has subtle behavior regarding "\" in blob names. Something like:
  -- It is not illegal in the name.
  -- It is illegal as a delimiter.
  -- It is interpreted as a "/" when doing virtual directories.

We don't have a good story for this right now. There's some discussion [here](https://github.com/google/go-cloud/blob/master/internal/docs/design.md#enforcing-portability).

For now, I'm going to disable some conformance tests so that the Azure contribution is unblocked.

vangent:
- Consider strongly typing names instead of using `string`? E.g., `blob.Name`. This might provide hooks for validation, etc.
- Consider having provider implementations accept translation functions (e.g., Azure constructor could take a function for metadata keys that converts arbitrary keys to C# identifiers). This can only partially work since you can't recover the original name, so writing and then reading back would give a changed result.

PR #907, opened 2018-12-06
blob: use _ in metadata keys instead of -

vangent:
See #905 for context; TL;DR, Azure doesn't support - in metadata keys.

Updates #905.


vangent:
cc/ @vsaroopchand

PR #913, opened 2018-12-07
internal/pubsub/rabbitpubsub: implement IsRetryable

jba:


internal/pubsub/rabbitpubsub/rabbit.go
```diff
@@ -233,9 +233,40 @@ func toPublishing(m *driver.Message) amqp.Publishing {
...
-func (*topic) IsRetryable(error) bool {
-	// TODO(jba): figure out what errors can be retried.
-	return false
+func (*topic) IsRetryable(err error) bool {
+	return isRetryable(err)
+}
+
+func isRetryable(err error) bool {
+	aerr, ok := err.(*amqp.Error)
+	if !ok {
```
vangent: What error do we get if the connection was just closed (e.g., server restarted)?

jba: ChannelError (504).

We shouldn't retry in this case. Our connections come from outside (we never create or close them), so this would require user code to retry by re-creating a connection.

Issue #837, opened 2018-11-29
pubsub: use a fake for testing RabbitMQ


jba:
Currently we can only test against an actual RabbitMQ instance. We should also support an in-memory fake, perhaps https://github.com/NeowayLabs/wabbit.

jba:
Use the fake to test context-awareness of establishChannel methods.

jba:
`wabbit` doesn't just pass through to `streadway/amqp`, it adds some functionality and makes some choices. So instead I'm going to try rolling my own.

PR #914, opened 2018-12-07
internal/pubsub/rabbitpubsub: make establishChannels context-sensitive

jba:
Make the establishChannel methods responsive to context doneness, by
making the RPCs in a separate goroutine.


jba:
I added a test for runWithContext. For better testing we'll need some sort of fake. I added a comment to #837.

jba:
@ijt ping

internal/pubsub/rabbitpubsub/rabbit_test.go
```diff
@@ -147,6 +149,25 @@ func TestUnroutable(t *testing.T) {
...
+	e := errors.New("")
+	// f sleeps for a bit just to give the scheduler a chance to run.
+	f := func() error { time.Sleep(100 * time.Millisecond); return e }
+	got := runWithContext(context.Background(), f)
+	if want := e; got != want {
+		t.Errorf("got %v, want %v", got, want)
+	}
+
+	// runWithContext will return ctx.Err if context is done.
+	ctx, cancel := context.WithCancel(context.Background())
```
ijt: Looks like this and the chunk above could be split into two subtests.

PR #915, opened 2018-12-07
internal/imports: script for generating import path redirect files

jba:
This script fills a directory with the html files needed to redirect
import paths to gocloud.dev/...

Later, we'll configure Travis to run the script and push the generated
files to the gh-pages branch, where they will be served by GitHub Pages.


internal/imports/makeimports.sh
```diff
@@ -0,0 +1,57 @@
...
+fi
+
+
+shopt -s nullglob  # glob patterns that don't match turn into the empty string, instead of themselves
+
+function files_exist() {  # assumes nullglob
+  [[ ${1:-""} != "" ]]
+}
+
+rm -rf $OUTDIR
```
zombiezen: Optional: Seems like checking for existence and bailing is a bit safer than removing. Consider changing this.

jba: Done.

jba:
@vangent ping

jba:
nvm, I see you're not assigned.

PR #924, opened 2018-12-10
blob/all: add ListObject.MD5 and Attributes.MD5, with MD5 hash of the blob content

### LINKED ISSUE #615: blob: add a way to determine if a file differs from the existing contents
We've got it in `WriterOptions`, but it makes sense to add it as a read-only Attribute as well.

vangent:
These MD5 hashes are intended to allow `blob` users to see if content differs without necessarily loading the whole blob.

The docstring for `MD5` states that it may be `nil` if the data is not available. For the current providers, it *will* be `nil` if:

- `fileblob`: Always, unless `fileblob` wrote the blob, *and* the user provided a `ContentMD5` header for us to verify while writing; we save that MD5 hash in the attributes file. We could more aggressively fill in the MD5 field, but it would be at the cost of computing the MD5 hash when it may not be needed.

- `s3blob`: S3's `ETag` header is *usually* the quoted hexstring of the MD5. But not always. In particular, for blobs that were uploaded using multi-part upload. The code handles this gracefully and just returns `nil` if we can't parse the `ETag` header.

Fixes #615.


blob/drivertest/drivertest.go
```diff
@@ -561,62 +564,62 @@ func testListDelimiters(t *testing.T, newHarness HarnessMaker) {
...
-					Sub: []listResult{
-						listResult{Key: keyPrefix + "\\dir1\\a.txt"},
-						listResult{Key: keyPrefix + "\\dir1\\b.txt"},
-						listResult{
-							Key:   keyPrefix + "\\dir1\\subdir\\",
-							IsDir: true,
-							Sub: []listResult{
-								listResult{Key: keyPrefix + "\\dir1\\subdir\\c.txt"},
-								listResult{Key: keyPrefix + "\\dir1\\subdir\\d.txt"},
+			{
```
vangent: Note for reviewers: I must have not run `gofmt` when I commented this out before. You can set github to `No Whitespace` for diffs to hide it.

blob/drivertest/drivertest.go
```diff
@@ -1346,6 +1349,86 @@ func testMetadata(t *testing.T, newHarness HarnessMaker) {
...
+	}
+	defer h.Close()
+	drv, err := h.MakeDriver(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	b := blob.NewBucket(drv)
+
+	// Write the two blobs. Include an MD5 hash while writing in case provider
+	// implementations use that to produce the MD5 for List/Attributes.
```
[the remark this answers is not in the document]
vangent: Yes, `fileblob` does.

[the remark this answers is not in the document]
zombiezen: I don't see a reason to not have fileblob compute the MD5 automatically as it is being written. The time it takes to write to disk >> the time it takes to compute an MD5.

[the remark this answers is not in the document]
vangent: OK, changed `fileblob` to always compute and write MD5 when writing, and removed the `ContentMD5` from the drivertest.

Note that we may still not have MD5 for files in `fileblob` that we didn't write (they won't have an attributes file at all).

blob/drivertest/drivertest.go
```diff
@@ -1368,6 +1371,86 @@ func testMetadata(t *testing.T, newHarness HarnessMaker) {
...
 
+// testMD5 tests reading MD5 hashes via List and Attributes.
+func testMD5(t *testing.T, newHarness HarnessMaker) {
+	ctx := context.Background()
+
+	// Define two blobs with different content; we'll write them and then verify
+	// their returned MD5 hashes.
+	const aKey, bKey = "blob-for-md5-aaa", "blob-for-md5-bbb"
+	aContent, bContent := []byte("hello"), []byte("goodbye")
+	hash := md5.New()
```
jba: I think you can use the top-level function `md5.Sum` instead of these three lines.

blob/s3blob/s3blob.go
```diff
@@ -395,6 +398,32 @@ func (b *bucket) NewRangeReader(ctx context.Context, key string, offset, length
...
+	}
+	// Strip the expected leading and trailing quotes.
+	quoted := *etag
+	if quoted[0] != '"' || quoted[len(quoted)-1] != '"' {
+		return nil
+	}
+	unquoted := quoted[1 : len(quoted)-1]
+	// Un-hex; we return nil on error. In particular, we'll get an error here
+	// for multi-part uploaded blobs, whose ETag will look something like
+	// "aaaaaaa-2".
```
jba: So there's no way a multi-part ETag can ever look like hex? If it weren't for the `-`, your example would be. 

I'm hoping your comment can say (truthfully) "multi-part upload ETags always contain `-`, so they can never be legal hex" or something like that.

blob/drivertest/drivertest.go
vangent: Done, although it returns a `[16]byte` so I needed to add `[:]` down below in `bytes.Equal`.

blob/s3blob/s3blob.go
vangent: I believe that is true, so I made the comment stronger.

vangent:
PTAL.

blob/blob.go
```diff
@@ -88,6 +88,8 @@ type Attributes struct {
 	ModTime time.Time
 	// Size is the size of the object in bytes.
 	Size int64
+	// MD5 is an MD5 hash of the blob contents, or nil if not available.
```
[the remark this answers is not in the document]
jba: Hah, I would have suggested the opposite.

What, ever.

PR #927, opened 2018-12-11
pubsub: change Topic.Close() to Topic.Shutdown(ctx) 

### LINKED ISSUE #883: pubsub: Topic.Shutdown(ctx)
`Subscription.Close()` is being converted to `Shutdown(ctx)`. It seems inconsistent not to do the same for `Topic.Close()`.

ijt:
Fixes #883

### DIFF
--- a/internal/pubsub/pub_test.go
+++ b/internal/pubsub/pub_test.go
@@ -36,8 +36,8 @@ func TestTopicShutdownCanBeCanceledEvenWithHangingSend(t *testing.T) {
 
 	go func() {
 		m := &pubsub.Message{}
-		if err := top.Send(context.Background(), m); err != nil {
-			t.Fatal(err)
+		if err := top.Send(context.Background(), m); err == nil {
+			t.Fatal("nil err from Send, expected context cancellation error")
 		}
 	}()
 
--- a/internal/pubsub/pub_test.go
+++ b/internal/pubsub/pub_test.go
@@ -1,0 +1,72 @@
+package pubsub_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/google/go-cloud/internal/pubsub"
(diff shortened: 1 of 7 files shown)


internal/pubsub/pub_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (t *funcTopic) Close() error {
+	return nil
+}
+
+func (s *funcTopic) IsRetryable(error) bool { return false }
+
+func (s *funcTopic) As(i interface{}) bool { return false }
+
+func TestTopicShutdownCanBeCanceledEvenWithHangingSend(t *testing.T) {
```
vangent: Consider adding this test in a separate PR, it is not related to the rename.

internal/pubsub/pubsub.go
```diff
@@ -88,14 +91,23 @@ func (t *Topic) Send(ctx context.Context, m *Message) error {
...
+// Shutdown flushes pending message sends and disconnects the Topic.
 // It only returns after all pending messages have been sent.
-func (t *Topic) Close() error {
+func (t *Topic) Shutdown(ctx context.Context) error {
 	t.mu.Lock()
 	t.err = errors.New("pubsub: Topic closed")
 	t.mu.Unlock()
-	t.batcher.Shutdown()
-	return nil
+	c := make(chan struct{})
```
vangent: Same here, this isn't related to the rename is it?

internal/pubsub/pub_test.go
ijt: The rename isn't just a rename. The new name reflects new functionality, and this test demonstrates the new functionality.

internal/pubsub/pubsub.go
ijt: Same answer.

PR #928, opened 2018-12-11
internal/contributebot: obtain title patterns from new ".contributebot" file

zombiezen:
Both the regular expression and the text of the response can be changed using this new configuration file, which is plain JSON. I have added an in-memory cache so as to avoid eating into the GitHub API quota for every event.

I have tested the integration manually on my own test repository.

Updates google/wire#82


zombiezen:
You don't need to block submission on tests; the tests block me due to them being red.

It looks like a transient Travis network failure: Goveralls failed to communicate in the Linux build and the Mac build gave a cryptic failure while communicating with GCS to download dependencies.

Issue #922, opened 2018-12-10
internal/secrets: Consolidate interfaces


clausti:
Pursuant to the [discussion](https://github.com/google/go-cloud/pull/899#discussion_r239982555) on the `localsecrets` implementation, should the `secrets` package have:

 - 2 interface types: `Encrypter` and `Decrypter`
 - 3 interface types: `Encrypter`,  `Decrypter`, and `Crypter` (composed of the first two)
 - 1 interface type: `Crypter`

(This is intended as an independent decision from whether the `secrets` package should have a driver or not.)

clausti:
summarizing in-person discussion: consensus that 1 interface type is the way to go.

PR #933, opened 2018-12-11
internal/secrets/driver: make driver one interface

shantuo:
Combine Decrypter and Encrypter into one interface.


shantuo:
Fixes #922

Issue #931, opened 2018-12-11
blob: remove global registry for Open


zombiezen:
[`blob.Open`](https://godoc.org/github.com/google/go-cloud/blob#Open) currently operates from a global, mutable registry (populated with [`blob.Register`](https://godoc.org/github.com/google/go-cloud/blob#Register)). Go Cloud emphatically tries to avoid the pattern of having a global registry due to the confusion that can come in the face of conflicting keys and the difficulty of reasoning about side effects, especially when incorporating into a large application or when multiple incompatible versions are present. Avoiding global mutable state in favor of explicit construction is preferred.

## Proposed Change

The `blob` package should grow a new type (called `Opener` below, but exact name TBD) that manages the mapping to different implementations. A reasonable default package can exist in the Go Cloud repository that imports GCS, S3, and Azure.

```go
package blob

// An Opener returns buckets that map to pseudo-URLs.
type Opener struct {
  // ...
}

// NewOpener returns a new Opener that uses the given mapping of protocol schemes.
func NewOpener(schemes map[string]FromURLFunc) *Opener

// Open returns a bucket for the given URL based on its mapping.
func (o *Opener) Open(ctx context.Context, urlstr string) (*Bucket, error) {
}
```

and elsewhere:

```go
package cloudbuckets

import (
  ".../azureblob"
  ".../gcsblob"
  ".../s3blob"
)

// Opener returns a *blob.Opener that maps URLs to common cloud providers.
func Opener() *blob.Opener {
  // ...
}
```

While this adds a bit more setup code, it:

- Allows different `*blob.Opener`s to be present in different parts of the application
- Makes the connection between the `Open` call and the bucket much more explicit
- Permits decoration or alteration of the `FromURLFunc`s being sent to the opener

/cc @vangent @shantuo

jba:
Is it worth currying this? We could just have 
```
blob.OpenURL(urlstr string, schemes map[string]FromURLFunc) (*Bucket, error)
```

vangent:
On https://github.com/google/go-cloud/pull/981#discussion_r242239468, @clausti suggested that maybe the scheme(s) that a provider registers for should be exposed as public constants for the package.

zombiezen:
One additional consideration that occurred to me while hacking on this (and this might be a separate issue): do we want to support being able to open readers and writers by URL? It seems like a natural extension.

zombiezen:
Obsoleted by design decisions in #1174.

PR #935, opened 2018-12-11
internal/docs/design.md: forbid global state

zombiezen:
Updates #931


zombiezen:
@jba, a Bitcoin for your thoughts?

internal/docs/design.md
```diff
@@ -96,6 +96,14 @@ https://godoc.org/github.com/google/go-cloud/runtimevar#Variable
 https://godoc.org/github.com/google/go-cloud/blob#Bucket.NewWriter
 [`database/sql`]: https://godoc.org/database/sql
 
+## No Global State
+
+As a library, Go Cloud should not introduce global state. Global state is
```
jba: I think this advice is too strict. We already [violate it](https://github.com/google/go-cloud/blob/master/blob/blob.go#L618). 

Maybe no exported global state? Still too strict for me, actually.

jba: I guess I was thinking of globals that one sometimes writes instead of declaring flags inside a package.

Anyway, I just wanted to record my thoughts. I won't block this.

PR #937, opened 2018-12-12
docs: add a section about conformance tests to design.md

vangent:


internal/docs/design.md
```diff
@@ -322,48 +322,65 @@ on the [mailing list](https://groups.google.com/forum/#!forum/go-cloud).
...
+-   Ensure you've run `goimports` on your code to properly group import
+    statements.
 
 ## Tests
 
+### Conformance Tests
+
+Since our goal is for users to be able to use provider implementations
+interchangeably, it is critical that they behave similarly. To this end, for
+each portable API (e.g., `blob`), we provide a suite of conformance tests that
```
zombiezen: Let's keep the tone more imperative to match with the rest of this document:

> To this end, each portable API (e.g., `blob`) must provide a suite of conformance tests that provider implementations should run.

internal/docs/design.md
```diff
@@ -322,48 +322,65 @@ on the [mailing list](https://groups.google.com/forum/#!forum/go-cloud).
...
+each provider implementation should run. The conformance tests are
+comprehensive, so provider implementations shouldn't need additional tests.
```
zombiezen: I'm not as certain about this last bit of advice. For instance, it could be advisable for the provider implementation to test whether writing a blob sets the correct bits of provider-specific metadata that the conformance tests don't test for.

I think the principle you are trying to describe is that the conformance tests validate all (most?) of the driver requirements. The provider should not need to test those: they should let our conformance tests verify that.

zombiezen:
LGTM after addressing these comments.

internal/docs/design.md

internal/docs/design.md

Issue #896, opened 2018-12-05
all: review godoc for formatting and clarity


vangent:
There are a bunch of places with bad formatting (e.g., things that look like bullet points in the code are collapsed onto a single line, accidental code blocks, etc.).

vangent:
godoc -http=localhost:8080
  should work locally for me.

shantuo:
I noticed this as well on godoc.org, maybe that's a bug on gddo rendering.

vangent:
I've done a pass over runtimevar and blob and made a bunch of changes. We can make incremental changes after this, but I'm going to close this particular issue.

PR #944, opened 2018-12-13
blob/docs: more improvements for blob's godoc

vangent:
Updates #896.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -414,10 +414,8 @@ func (b *Bucket) NewReader(ctx context.Context, key string, opts *ReaderOptions)
 }
 
 // NewRangeReader returns a Reader to read content from the blob object
-// represented by key. It reads at most length bytes starting at offset. If
-// length is negative, it will read till the end of the object.
-//
-// offset must be >= 0, and length cannot be 0.
+// represented by key. It reads at most length (!= 0) bytes starting at
+// offset (>= 0). If length is negative, it will read till the end of the object.
 //
 // If the blob object does not exist, NewRangeReader returns an error for which
 // IsNotExist will return true. Attributes is a lighter-weight way
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -59,20 +59,21 @@ import (
 	"github.com/google/go-cloud/blob/driver"
 )
 
-// Reader implements io.ReadCloser to read a blob. It must be closed after
+// Reader reads bytes from a blob.
+// It implements io.ReadCloser, and must be closed after
 // reads are finished.
 type Reader struct {
 	b driver.Bucket
 	r driver.Reader
 }
 
-// Read implements io.ReadCloser to read from this reader.
+// Read implements io.Reader (https://golang.org/pkg/io/#Reader).
 func (r *Reader) Read(p []byte) (int, error) {
 	n, err := r.r.Read(p)
 	return n, wrapError(r.b, err)
 }
 
-// Close implements io.ReadCloser to close this reader.
+// Close implements io.Closer (https://golang.org/pkg/io/#Closer).
 func (r *Reader) Close() error {
 	return wrapError(r.b, r.r.Close())
 }
@@ -82,12 +83,12 @@ func (r *Reader) ContentType() string {
 	return r.r.Attributes().ContentType
 }
 
-// ModTime is the time the blob object was last modified.
+// ModTime returns the time the blob object was last modified.
 func (r *Reader) ModTime() time.Time {
 	return r.r.Attributes().ModTime
 }
 
-// Size returns the content size of the blob object.
+// Size returns the size of the blob object's content.
 func (r *Reader) Size() int64 {
 	return r.r.Attributes().Size
 }
@@ -111,7 +112,7 @@ type Attributes struct {
 	Metadata map[string]string
 	// ModTime is the time the blob object was last modified.
 	ModTime time.Time
-	// Size is the size of the object in bytes.
+	// Size is the size of the blob object's content.
 	Size int64
 
 	asFunc func(interface{}) bool
@@ -126,8 +127,8 @@ func (a *Attributes) As(i interface{}) bool {
 	return a.asFunc(i)
 }
 
-// Writer implements io.WriteCloser to write to blob. It must be closed after
-// all writes are done.
+// Writer writes bytes to a blob.
+// It implements io.WriteCloser, and must be closed after all writes are done.
 type Writer struct {
 	b driver.Bucket
 	w driver.Writer
@@ -149,11 +150,11 @@ type Writer struct {
 // sniffLen is the byte size of Writer.buf used to detect content-type.
 const sniffLen = 512
 
-// Write implements the io.Writer interface.
+// Write implements the io.Writer interface (https://golang.org/pkg/io/#Writer).
 //
-// The writes happen asynchronously, which means the returned error can be nil
-// even if the actual write fails. Use the error returned from Close to
-// check and handle errors.
+// Writes may happen asynchronously, so the returned error can be nil
+// even if the actual write eventually fails. The write is only guaranteed to
+// have succeeded if Close returns no error.
 func (w *Writer) Write(p []byte) (n int, err error) {
 	if w.w != nil {
 		n, err := w.w.Write(p)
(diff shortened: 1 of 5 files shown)


vangent:
I will try to leave my demo server up at http://keebler3.svl.corp.google.com:8080/pkg/github.com/google/go-cloud/blob/.

blob/blob.go
```diff
@@ -126,8 +127,8 @@ func (a *Attributes) As(i interface{}) bool {
 	return a.asFunc(i)
 }
 
-// Writer implements io.WriteCloser to write to blob. It must be closed after
-// all writes are done.
+// Writer writes bytes to a blob.
+// It implements io.WriteCloser, and must be closed after all writes are done.
```
zombiezen: Link here to `io.WriteCloser`?

blob/blob.go
```diff
@@ -177,9 +178,10 @@ func (w *Writer) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
-// Close flushes any buffered data and completes the Write. It is the user's
-// responsibility to call it after finishing the write and handle the error if returned.
-// Close will return an error if the context provided to create w is canceled or times out.
+// Close completes the Write, which is not guaranteed to have succeeded until
```
zombiezen:
```suggestion
// Close closes the blob writer. The write operation is not guaranteed to have succeeded until
```

blob/blob.go
```diff
@@ -177,9 +178,10 @@ func (w *Writer) Write(p []byte) (n int, err error) {
...
+// Close returns with no error.
+// Close will return an error if the context provided to create the Writer is
```
zombiezen:
```suggestion
// Close may return an error if the context provided to create the Writer is
```

blob/blob.go
```diff
@@ -177,9 +178,10 @@ func (w *Writer) Write(p []byte) (n int, err error) {
...
+// canceled or times out.
```
zombiezen:
```suggestion
// canceled or reaches its deadline.
```

blob/blob.go
```diff
@@ -206,7 +208,7 @@ func (w *Writer) open(p []byte) (int, error) {
 	return n, wrapError(w.b, err)
 }
 
-// ListOptions sets options for listing objects.
+// ListOptions sets options for listing objects via List.
```
zombiezen:
```suggestion
// ListOptions sets options for listing objects via Bucket.List.
```

blob/blob.go
```diff
@@ -279,7 +281,7 @@ type ListObject struct {
 	Key string
 	// ModTime is the time the blob object was last modified.
 	ModTime time.Time
-	// Size is the size of the object in bytes.
+	// Size is the size of the blob object's content.
```
zombiezen:
```suggestion
	// Size is the size of the blob's content in bytes.
```

blob/blob.go
```diff
@@ -299,8 +301,10 @@ func (o *ListObject) As(i interface{}) bool {
 	return o.asFunc(i)
 }
 
-// Bucket manages the underlying blob service and provides read, write and delete
-// operations on objects within it.
+// Bucket provides an easy and portable way to interact with blob objects
```
zombiezen:
```suggestion
// Bucket provides an easy and portable way to interact with blobs
```

blob/blob.go
```diff
@@ -346,7 +359,7 @@ func (b *Bucket) ReadAll(ctx context.Context, key string) ([]byte, error) {
 	return ioutil.ReadAll(r)
 }
 
-// List returns an object that can be used to iterate over objects in a
+// List returns a ListIterator that can be used to iterate over objects in a
```
zombiezen:
```suggestion
// List returns a ListIterator that can be used to iterate over blobs in a
```

blob/blob.go
```diff
@@ -367,7 +380,10 @@ func (b *Bucket) List(opts *ListOptions) *ListIterator {
 	return &ListIterator{b: b.b, opts: dopts}
 }
 
-// Attributes reads attributes for the given key.
+// Attributes returns attributes for the blob object represented by key.
```
zombiezen:
```suggestion
// Attributes returns attributes for the blob represented by key.
```

blob/blob.go
```diff
@@ -367,7 +380,10 @@ func (b *Bucket) List(opts *ListOptions) *ListIterator {
...
+//
+// If the blob object does not exist, Attributes returns an error for which
```
zombiezen:
```suggestion
// If the blob does not exist, Attributes returns an error for which
```

blob/blob.go
```diff
@@ -392,24 +408,18 @@ func (b *Bucket) Attributes(ctx context.Context, key string) (Attributes, error)
...
+// NewReader is a shortcut for NewRangedReader with offset=0 and length=-1.
 func (b *Bucket) NewReader(ctx context.Context, key string, opts *ReaderOptions) (*Reader, error) {
 	return b.NewRangeReader(ctx, key, 0, -1, opts)
 }
 
-// NewRangeReader returns a Reader that reads part of an object, reading at
-// most length bytes starting at the given offset. If length is negative, it
-// will read till the end of the object. offset must be >= 0, and length cannot
-// be 0.
+// NewRangeReader returns a Reader to read content from the blob object
```
zombiezen:
```suggestion
// NewRangeReader returns a Reader to read content from the blob
```

blob/blob.go
```diff
@@ -392,24 +408,18 @@ func (b *Bucket) Attributes(ctx context.Context, key string) (Attributes, error)
...
+// represented by key. It reads at most length (!= 0) bytes starting at
```
zombiezen:
```suggestion
// stored at key. It reads at most length (!= 0) bytes starting at
```

blob/blob.go
```diff
@@ -392,24 +408,18 @@ func (b *Bucket) Attributes(ctx context.Context, key string) (Attributes, error)
...
+// offset (>= 0). If length is negative, it will read till the end of the object.
 //
-// NewRangeReader returns an error if the object does not exist, which can be checked
-// by calling IsNotExist. Bucket.Attributes is a lighter-weight way to check for
-// existence.
+// If the blob object does not exist, NewRangeReader returns an error for which
```
zombiezen:
```suggestion
// If the blob does not exist, NewRangeReader returns an error for which
```

blob/blob.go
```diff
@@ -445,16 +455,17 @@ func (b *Bucket) WriteAll(ctx context.Context, key string, p []byte, opts *Write
...
+// NewWriter returns a Writer that writes to the blob object represented by key.
 // A nil WriterOptions is treated the same as the zero value.
 //
-// A new object will be created unless an object with this key already exists.
-// Otherwise any previous object with the same key will be replaced. The object
-// is not guaranteed to be available until Close has been called.
+// If a blob object with this key already exists, it will be replaced.
+// The blob object being written is not guaranteed to be readable until Close
+// has been called; until then, any previous object will still be readable.
+// Even after CLose is called, newly written objects are not guaranteed to be
```
zombiezen:
```suggestion
// Even after Close is called, newly written objects are not guaranteed to be
```

blob/blob.go
```diff
@@ -82,12 +83,12 @@ func (r *Reader) ContentType() string {
...
 }
 
-// ModTime is the time the blob object was last modified.
+// ModTime returns the time the blob object was last modified.
 func (r *Reader) ModTime() time.Time {
 	return r.r.Attributes().ModTime
 }
 
-// Size returns the content size of the blob object.
+// Size returns the size of the blob object's content.
```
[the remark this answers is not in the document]

blob/blob.go
vangent: Done, here and elsewhere.

blob/blob.go
```diff
@@ -299,8 +303,10 @@ func (o *ListObject) As(i interface{}) bool {
 	return o.asFunc(i)
 }
 
-// Bucket manages the underlying blob service and provides read, write and delete
-// operations on objects within it.
+// Bucket provides an easy and portable way to interact with blobs
+// within a "bucket", including read, write, and list operations.
+// To create a Bucket, use constructors found in provider-specific
```
clausti: Do we define "provider" anywhere? Or define or label "provider-specific subpacke[s]" ? I'm not sure this term is completely obvious with respect to where I would want to go next if I were first reading it. 

I think an 'eg' and an example of calling one, or a label at the top of the provider-specific packages that says "this is the <SERVICE> provider-specific package", or a definition in DESIGN.md, or some combination thereof would make it clearer.

vangent: I tried to introduce this better here, PTAL. I think with the example most people will understand what "provider" means.

PR #953, opened 2018-12-13
pubsub: add awspubsub

### LINKED ISSUE #634: pubsub: AWS SNS + SQS driver

ijt:
Fixes #634


ijt:
I'm adding record/replay to the test, but that shouldn't change anything about the code that's already here.

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,181 @@
...
+	params := sqs.ReceiveMessageInput{QueueUrl: &s.qURL}
+	req, output := s.client.ReceiveMessageRequest(&params)
+	err := req.Send()
+	if err != nil {
+		return nil, fmt.Errorf("awspubsub: message receive failed: %v", err)
+	}
+	var ms []*driver.Message
+	for _, m := range output.Messages {
+		type Attribute struct {
+			Type  string
```
vangent: Is this just unused? What are possible values (given that the `Value` is always of type `string`)?

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,181 @@
...
+		b := string(dm.Body)
+		attrs := map[string]*sns.MessageAttributeValue{}
+		for k, v := range dm.Metadata {
+			attrs[k] = &sns.MessageAttributeValue{
+				DataType:    aws.String("String"),
+				StringValue: aws.String(v),
+			}
+		}
+		params := sns.PublishInput{
+			Message:           &b,
```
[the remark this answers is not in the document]
ijt: Done.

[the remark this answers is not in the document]
ijt: Filed https://github.com/google/go-cloud/issues/956 to add a conformance test for this.

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,181 @@
...
+
+type topic struct {
+	client *sns.SNS
+	arn    string
+}
+
+// TopicOptions will contain configuration for topics.
+type TopicOptions struct{}
+
+// OpenTopic opens the topic on AWS SNS for the given projectID and
```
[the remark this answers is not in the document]
ijt: Done.

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,181 @@
...
+type subscription struct {
+	client *sqs.SQS
+	qURL   string
+}
+
+// SubscriptionOptions will contain configuration for subscriptions.
+type SubscriptionOptions struct{}
+
+// OpenSubscription opens the queue on AWS SQS for the given
+// projectID and subscriptionName.
```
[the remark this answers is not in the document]
ijt: Done.

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,181 @@
...
+	return &subscription{client, qURL}
+}
+
+// ReceiveBatch implements driver.Subscription.ReceiveBatch.
+func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*driver.Message, error) {
+	params := sqs.ReceiveMessageInput{QueueUrl: &s.qURL}
+	req, output := s.client.ReceiveMessageRequest(&params)
+	err := req.Send()
+	if err != nil {
+		return nil, fmt.Errorf("awspubsub: message receive failed: %v", err)
```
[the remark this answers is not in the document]
ijt: Done.

internal/pubsub/awspubsub/awspubsub.go
ijt: Oh, right, it's unused. Removed it.

ijt: Possible values for Type are String, String.Array, Number, and Binary. awspubsub on Go Cloud only creates messages whose attribute values have type String for now but messages could be sent to the same topic not by Go Cloud.

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,202 @@
...
+			return err
+		}
+	}
+	return nil
+}
+
+func (t *topic) send(params *sns.PublishInput) error {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	req, _ := t.client.PublishRequest(params)
```
jba: So AWS has no way to batch requests? Sad.

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,202 @@
...
+
+// As implements driver.Topic.As.
+func (t *topic) As(i interface{}) bool {
+	c, ok := i.(**sns.SNS)
+	if !ok {
+		return false
+	}
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	*c = t.client
```
jba: If `t.client` doesn't change after construction, it doesn't need to be protected.

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,202 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/sns"
+	"github.com/aws/aws-sdk-go/service/sqs"
+	"gocloud.dev/internal/pubsub"
+	"gocloud.dev/internal/pubsub/driver"
+)
+
+type topic struct {
+	mu     sync.Mutex
```
jba: You may not need the mutex—nothing changes, right?

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,202 @@
...
+	var ms []*driver.Message
+	for _, m := range output.Messages {
+		type Attribute struct {
+			Value string
+		}
+		type MsgBody struct {
+			Message           string
+			MessageAttributes map[string]Attribute
+		}
+		var body MsgBody
```
jba:
```
var body struct { ... }
```
I think you can also inline the `Attributes` struct.

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,202 @@
...
+			return nil, err
+		}
+		attrs := map[string]string{}
+		for k, v := range body.MessageAttributes {
+			attrs[k] = v.Value
+		}
+		m := &driver.Message{
+			Body:     []byte(body.Message),
+			Metadata: attrs,
+			AckID:    *m.ReceiptHandle,
```
jba: nit: there's no need to dereference it, is there?

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		ms = append(ms, m)
+	}
+	return ms, nil
+}
+
+func (s *subscription) receiveMessages() (*sqs.ReceiveMessageOutput, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	params := sqs.ReceiveMessageInput{QueueUrl: &s.qURL}
```
jba: You don't need `params`

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,202 @@
...
+	req, output := s.client.ReceiveMessageRequest(&params)
+	err := req.Send()
```
jba: maybe 
```
if err := req.Send(); err != nil {
   return nil, err
}
return output nil
```
? Probably best not to return `output` if there's an error.

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,202 @@
...
+	return output, err
+}
+
+// SendAcks implements driver.Subscription.SendAcks.
+func (s *subscription) SendAcks(ctx context.Context, ids []driver.AckID) error {
+	for _, id := range ids {
+		rh := id.(string)
```
jba: Don't need `rh`

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,202 @@
...
+		err := s.deleteMessage(rh)
```
jba: if err := ...; err != nil ...

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,202 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *subscription) deleteMessage(receiptHandle string) error {
```
jba: After you get rid of the lock, you can just inline this into the above function.

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// TopicOptions will contain configuration for topics.
+type TopicOptions struct{}
+
+// OpenTopic opens the topic on AWS SNS for the given projectID and
+// topicName.
+func OpenTopic(ctx context.Context, client *sns.SNS, topicARN string, opts *TopicOptions) *pubsub.Topic {
+	dt := openTopic(ctx, client, topicARN)
+	return pubsub.NewTopic(dt)
```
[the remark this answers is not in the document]
ijt: Done.

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// SubscriptionOptions will contain configuration for subscriptions.
+type SubscriptionOptions struct{}
+
+// OpenSubscription opens the queue on AWS SQS for the given
+// projectID and subscriptionName.
+func OpenSubscription(ctx context.Context, client *sqs.SQS, qURL string, opts *SubscriptionOptions) *pubsub.Subscription {
+	ds := openSubscription(ctx, client, qURL)
+	return pubsub.NewSubscription(ds)
```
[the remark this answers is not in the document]
ijt:
```go
return pubsub.NewSubscription(openSubscription(ctx, client, qURL))
```
doesn't look more readable to me. Mind if I leave it?

internal/pubsub/awspubsub/awspubsub.go
ijt: I haven't found anything for batching sends, though it does appear to at least potentially receive in batches.

internal/pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,181 @@
...
+			}
+		}
+		params := sns.PublishInput{
+			Message:           &b,
+			MessageAttributes: attrs,
+			TopicArn:          &t.arn,
+		}
+		req, _ := t.client.PublishRequest(&params)
+		if err := req.Send(); err != nil {
+			return fmt.Errorf("awspubsub: failed to send message: %v", err)
```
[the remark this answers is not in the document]
ijt: Done.

internal/pubsub/awspubsub/awspubsub.go
ijt: I get a race when running the test in replace mode if I remove the mutex for the subscription client, probably because fields within the client are being mutated.

internal/pubsub/awspubsub/awspubsub.go
ijt: What if fields within it are mutated during sends? I tried to check, but I don't yet have clear evidence one way or the other by looking at the AWS Go SDK.

internal/pubsub/awspubsub/awspubsub.go
ijt: I don't know. Running with -race seems ok without this mutex, but maybe it's just because of insufficient test coverage.

ijt: Oh, never mind. Their docs say the SNS clients are safe to use concurrently. https://docs.aws.amazon.com/sdk-for-go/api/service/sns/

internal/pubsub/awspubsub/awspubsub.go
ijt: This is surprising considering that the docs say SQS clients are safe for concurrent use: https://docs.aws.amazon.com/sdk-for-go/api/service/sqs/.

internal/pubsub/awspubsub/awspubsub.go
ijt: Update: their docs say the SNS clients are safe for concurrent use: https://docs.aws.amazon.com/sdk-for-go/api/service/sns/.

internal/pubsub/awspubsub/awspubsub.go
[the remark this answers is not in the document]
jba: If you prefer.

internal/pubsub/awspubsub/awspubsub.go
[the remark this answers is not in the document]
jba: whatevs

internal/pubsub/awspubsub/awspubsub.go
jba: What's the actual data that's racy?

internal/pubsub/awspubsub/awspubsub.go
ijt: Done.

internal/pubsub/awspubsub/awspubsub.go
ijt: Apparently there's not. Done.

internal/pubsub/awspubsub/awspubsub.go
ijt: That's true, but omitting it doesn't make it more readable for me.

internal/pubsub/awspubsub/awspubsub.go
ijt: Done.

internal/pubsub/awspubsub/awspubsub.go
ijt: Done.

ijt:
Hmm, well that's strange. I ran `git rebase -i 44d80268d88af0f39540cd8827b6fbf32aa0f6f4` in order to squash some commits and now this PR appears to be hosed, with the "So there's good news and bad news" message above. So I'm starting a fresh PR to replace this one: https://github.com/google/go-cloud/pull/1084.

Issue #420, opened 2018-09-11
secrets: add new secret management provider packages


shantuo:
- [x] Google Cloud [KMS](https://cloud.google.com/kms/docs/reference/libraries)
- [ ] AWS [KMS](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/using-kms-with-go-sdk.html)
- [ ] [Vault](https://godoc.org/github.com/hashicorp/vault/api)

PR #974, opened 2018-12-14
internal/secrets: add Google Cloud KMS implementation

shantuo:
Update #420


internal/secrets/gcpkms/kms_test.go
```diff
@@ -0,0 +1,79 @@
...
+func newHarness(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
+	conn, done := setup.NewGCPgRPCConn(ctx, t, endPoint)
+	client, err := cloudkms.NewKeyManagementClient(ctx, option.WithGRPCConn(conn))
+	if err != nil {
+		return nil, err
+	}
+	return &harness{
+		client: client,
+		close: func() {
+			_ = client.Close()
```
clausti: Why does `client.Close()` get assigned to anything here? (I am assuming there's a reason it needs to be, just curious what it is.)

internal/secrets/gcpkms/example_test.go
```diff
@@ -0,0 +1,75 @@
...
+package gcpkms_test
+
+import (
+	"context"
+
+	"github.com/google/go-cloud/internal/secrets/gcpkms"
+)
+
+func ExampleCrypter_Encrypt() {
+	ctx := context.Background()
```
zombiezen: For here and the other example: it's helpful to over-explain each step in examples so that the reader knows what's going and on, and especially what they need to change.

internal/secrets/gcpkms/kms.go
```diff
@@ -0,0 +1,88 @@
...
+
+// DefaultClient returns a client to use with Cloud KMS and a clean-up function
+// to close the client after used.
+func DefaultClient(ctx context.Context, ts gcp.TokenSource) (*cloudkms.KeyManagementClient, func(), error) {
+	c, err := cloudkms.NewKeyManagementClient(ctx, option.WithTokenSource(ts))
+	return c, func() { c.Close() }, err
+}
+
+// NewCrypter returns a new Crypter to to encryption and decryption.
+func NewCrypter(ki *KeyInfo, client *cloudkms.KeyManagementClient) *Crypter {
```
zombiezen: nit: Flip the order of these parameters. Go from most general type to least general.

```suggestion
func NewCrypter(client *cloudkms.KeyManagementClient, ki *KeyInfo) *Crypter {
```

internal/secrets/gcpkms/kms.go
```diff
@@ -0,0 +1,88 @@
...
+	"google.golang.org/api/option"
+	kmspb "google.golang.org/genproto/googleapis/cloud/kms/v1"
+)
+
+// endPoint is the address to access Google Cloud KMS API.
+const endPoint = "cloudkms.googleapis.com:443"
+
+// DefaultClient returns a client to use with Cloud KMS and a clean-up function
+// to close the client after used.
+func DefaultClient(ctx context.Context, ts gcp.TokenSource) (*cloudkms.KeyManagementClient, func(), error) {
```
zombiezen: nit: `Dial` for consistency with other packages?

internal/secrets/gcpkms/kms.go
```diff
@@ -0,0 +1,88 @@
...
+func NewCrypter(ki *KeyInfo, client *cloudkms.KeyManagementClient) *Crypter {
+	return &Crypter{
+		keyInfo: ki,
+		client:  client,
+	}
+}
+
+// KeyInfo includes related information to construct a key name that is managed
+// by Cloud KMS.
+type KeyInfo struct {
```
zombiezen: This concept is called the [key resource ID](https://cloud.google.com/kms/docs/object-hierarchy#key) in the docs. So perhaps `KeyID` is more suitable here, as well as a link to that doc?

internal/secrets/gcpkms/kms.go
```diff
@@ -0,0 +1,88 @@
...
+	ProjectID, Location, KeyRing, KeyID string
+}
+
+// Crypter contains information to construct the pull path of a key.
+type Crypter struct {
+	keyInfo *KeyInfo
+	client  *cloudkms.KeyManagementClient
+}
+
+func (ki *KeyInfo) keyPath() string {
```
zombiezen: Move this to be next to the `KeyInfo` type. Also seems reasonable to make this the `String()` method for the type.

internal/secrets/gcpkms/kms.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// KeyInfo includes related information to construct a key name that is managed
+// by Cloud KMS.
+type KeyInfo struct {
+	ProjectID, Location, KeyRing, KeyID string
+}
+
+// Crypter contains information to construct the pull path of a key.
+type Crypter struct {
```
zombiezen: Since secrets is going to use the driver pattern, this type should not be exported.

internal/secrets/gcpkms/kms_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/google/go-cloud/internal/testing/setup"
+	"google.golang.org/api/option"
+
+	cloudkms "cloud.google.com/go/kms/apiv1"
+	"github.com/google/go-cloud/internal/secrets/driver"
+	"github.com/google/go-cloud/internal/secrets/drivertest"
+)
+
+const (
+	// These constants capture values that were used during the last --record.
```
zombiezen: Move the doc comment to right before the `const` keyword. Otherwise, the doc is just associated with `projectID`.

runtimevar/runtimevar.go
```diff
@@ -122,10 +122,11 @@ func (w *wrappedError) Error() string {
 // Decode is a function type for unmarshaling/decoding bytes into given object.
 type Decode func([]byte, interface{}) error
 
-// Decoder is a helper for decoding bytes into a particular Go type object.  The Variable objects
-// produced by a particular driver.Watcher should always contain the same type for Variable.Value
-// field.  A driver.Watcher can use/construct a Decoder object with an associated type (Type) and
-// decoding function (Func) for decoding retrieved bytes into Variable.Value.
+// Decoder is a helper for decoding bytes into a particular Go type object. The
+// Variable objects produced by a particular driver.Watcher should always
```
zombiezen: Unrelated change. Revert?

internal/secrets/gcpkms/example_test.go
shantuo: Added some comments, thanks!

shantuo:
PTAL

Issue #970, opened 2018-12-14
all:  documenting design pattern for constructors and make packages consistent with it


vangent:
@jba 
@zombiezen 

We seem to have several patterns for how driver implementations expose their constructors. We should document our preference and aim for consistency unless there are good reasons for variance.

- `blob` has `OpenBucket` at the top level. This seems simplest to me.
- `runtimevar/paramstore` has a `Client` struct ([here](https://github.com/google/go-cloud/blob/master/runtimevar/paramstore/paramstore.go#L42)). `Client` then exposes `NewVariable`, which is how you get an instance of the concrete type.
  - I think we can just get rid of this intervening type and have `NewVariable` take whatever is being passed to `NewClient`.
- `pubsub/mempubsub` has a `Broker` ([here](https://github.com/google/go-cloud/blob/master/internal/pubsub/mempubsub/mem.go#L39)) and then `NewTopic` takes an instance of the `Broker`.

jba:
I'm in favor of removing `mempubsub.Broker`. That was my original design.

I don't know if we can make this a hard-and-fast rule. We may want a client-like type if there are many things that are shared among the resources. But as a guideline, I'm fine with it.

vangent:
@jba -- I think this was my suggestion. At the time, I thought it was necessary in order to do the kinds of conformance tests we need (IIRC, the previous version had no  way to connect specific topics to specific subscriptions). If there's a better way that enables the tests we need but gets rid of `Broker`, let's at least try it out.

jba:
If we removed `Broker`, we would need a global map of topics by name. That would violate @zombiezen's rule about no global state. But since I think of `mempubsub` as test code, that doesn't bother me.

I started the PR. Let me know if it's worth finishing.

zombiezen:
I don't believe that any naming scheme is necessary for `mempubsub`, and thus no global state. The sketch I made to Robert was:

```go
package mempubsub

// Topic is an in-memory topic that can have in-memory subscriptions created.
type Topic struct {
  *pubsub.Topic

  // ...
}

// NewTopic creates a new in-memory topic with no subscriptions.
func NewTopic() *Topic { /* ... */ }

// NewSubscription returns a new subscription to the topic.
func (t *Topic) NewSubscription() *pubsub.Subscription { /* ... */ }
```

One could also get fancy with `As()` methods and instead make `NewSubscription` just take in a `*pubsub.Topic`:

```go
package mempubsub

// NewTopic returns a new in-memory topic.
[...]

PR #976, opened 2018-12-14
runtimevar: drop Client types in favor of direct constructors

vangent:
This is a straw man for the discussion on #970, although I think I like it and would be willing to submit it.

My proposal would be to then add a section to `design.md` saying something like:
- provider implementation packages should expose top-level constructors that return the concrete type.
- they should not create an intervening "builder" type
- they should not create a "wrapper" type

Updates #970.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -423,7 +423,7 @@ func (b *Bucket) NewReader(ctx context.Context, key string, opts *ReaderOptions)
 }
 
 // NewRangeReader returns a Reader to read content from the blob stored at key.
-// It reads at most length (!= 0) bytes starting at offset (>= 0).
+// It reads at most length bytes starting at offset (>= 0).
 // If length is negative, it will read till the end of the blob.
 //
 // If the blob does not exist, NewRangeReader returns an error for which
@@ -436,9 +436,6 @@ func (b *Bucket) NewReader(ctx context.Context, key string, opts *ReaderOptions)
 func (b *Bucket) NewRangeReader(ctx context.Context, key string, offset, length int64, opts *ReaderOptions) (*Reader, error) {
 	if offset < 0 {
 		return nil, errors.New("blob.NewRangeReader: offset must be non-negative")
-	}
-	if length == 0 {
-		return nil, errors.New("blob.NewRangeReader: length cannot be 0")
 	}
 	if opts == nil {
 		opts = &ReaderOptions{}
--- a/blob/drivertest/drivertest.go
+++ b/blob/drivertest/drivertest.go
@@ -824,17 +824,16 @@ func testRead(t *testing.T, newHarness HarnessMaker) {
 			wantErr: true,
 		},
 		{
-			name:       "length 0 read fails",
-			key:        key,
-			wantErr:    true,
-			skipCreate: true,
-		},
-		{
 			name:       "negative offset fails",
 			key:        key,
 			offset:     -1,
 			wantErr:    true,
 			skipCreate: true,
+		},
+		{
+			name: "length 0 read",
+			key:  key,
+			want: []byte{},
 		},
 		{
 			name:         "read from positive offset to end",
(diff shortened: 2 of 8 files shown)


samples/guestbook/inject_gcp.go
```diff
@@ -71,13 +72,13 @@ func gcpSQLParams(id gcp.ProjectID, flags *cliFlags) *cloudmysql.Params {
 
 // gcpMOTDVar is a Wire provider function that returns the Message of the Day
 // variable from Runtime Configurator.
-func gcpMOTDVar(ctx context.Context, client *runtimeconfigurator.Client, project gcp.ProjectID, flags *cliFlags) (*runtimevar.Variable, func(), error) {
+func gcpMOTDVar(ctx context.Context, client pb.RuntimeConfigManagerClient, project gcp.ProjectID, flags *cliFlags) (*runtimevar.Variable, func(), error) {
```
shantuo: I'm generally in favor of reducing the middle layer, or wrapper. One minor concern is that user may need to figure out want's inside package "google.golang.org/genproto/googleapis/cloud/runtimeconfig/v1beta1". Though this can be trivial and only affect users using all default options we provide. I would imagine in real applications user would  do that anyway to create their own custom clients?

vangent: Probably. And this is not any worse than before; before this PR, they needed to create one of these to create the `Client`, now they need to create one to create the `Variable`.

runtimevar/paramstore/paramstore.go
```diff
@@ -33,32 +33,21 @@ import (
...
-func NewClient(p client.ConfigProvider) *Client {
-	return &Client{sess: p}
-}
-
-// NewVariable constructs a runtimevar.Variable object with this package as the driver
-// implementation.
-func (c *Client) NewVariable(name string, decoder *runtimevar.Decoder, opts *Options) (*runtimevar.Variable, error) {
-	w, err := c.newWatcher(name, decoder, opts)
+// NewVariable constructs a runtimevar.Variable that watches AWS Parameter Store.
+func NewVariable(name string, sess client.ConfigProvider, decoder *runtimevar.Decoder, opts *Options) (*runtimevar.Variable, error) {
```
zombiezen: `sess` usually comes earlier in the argument list, since it's a general credential-ish thing.

```suggestion
func NewVariable(sess client.ConfigProvider, name string, decoder *runtimevar.Decoder, opts *Options) (*runtimevar.Variable, error) {
```

runtimevar/runtimeconfigurator/runtimeconfigurator.go
```diff
@@ -64,33 +56,23 @@ func Dial(ctx context.Context, ts gcp.TokenSource) (pb.RuntimeConfigManagerClien
...
-func NewClient(stub pb.RuntimeConfigManagerClient) *Client {
-	return &Client{client: stub}
-}
-
 // NewVariable constructs a runtimevar.Variable object with this package as the driver
 // implementation. Provide a decoder to unmarshal updated configurations into similar
 // objects during the Watch call.
-func (c *Client) NewVariable(name ResourceName, decoder *runtimevar.Decoder, opts *Options) (*runtimevar.Variable, error) {
-	w, err := c.newWatcher(name, decoder, opts)
+func NewVariable(name ResourceName, client pb.RuntimeConfigManagerClient, decoder *runtimevar.Decoder, opts *Options) (*runtimevar.Variable, error) {
```
zombiezen: Similar advice here: move the client up in the parameter list since it is more general (i.e. could be shared by other calls to `NewVariable`) than the other parameters.

```suggestion
func NewVariable(client pb.RuntimeConfigManagerClient, name ResourceName, decoder *runtimevar.Decoder, opts *Options) (*runtimevar.Variable, error) {
```

runtimevar/paramstore/paramstore.go
vangent: I'm going to open a bug to discuss, because we have the same thing in `blob`; e.g.:

```
func OpenBucket(ctx context.Context, bucketName string, sess client.ConfigProvider, opts *Options) (*blob.Bucket, error) 
```

PR #980, opened 2018-12-14
pubsub: make failure output of TestSendReceive more readable

### LINKED ISSUE #979: pubsub: output of TestSendReceive conformance test is hard to read on failure
Here is an example of a failing run of this test:
```
[ ~/src/gocloud.dev/internal/pubsub/gcppubsub ] go test -record
--- FAIL: TestConformance (1.78s)
    --- FAIL: TestConformance/TestSendReceive (1.14s)
        replay.go:196: Recording into golden file testdata/TestConformance/TestSendReceive.replay
        drivertest.go:213: {[]*pubsub.Message}[0->?]:
            	-: &pubsub.Message{Body: []uint8{0x34, 0x31}, ack: (func())(0x01436670), isAcked: true}
            	+: <non-existent>
            {[]*pubsub.Message}[?->0]:
            	-: <non-existent>
            	+: &pubsub.Message{Body: []uint8{0x30}, Metadata: map[string]string{"a": "0"}}
            {[]*pubsub.Message}[1->?]:
            	-: &pubsub.Message{Body: []uint8{0x34, 0x33}, ack: (func())(0x01436670), isAcked: true}
            	+: <non-existent>
            {[]*pubsub.Message}[?->1]:
            	-: <non-existent>
            	+: &pubsub.Message{Body: []uint8{0x31}, Metadata: map[string]string{"a": "1"}}
            {[]*pubsub.Message}[2->?]:
            	-: &pubsub.Message{Body: []uint8{0x34, 0x33}, ack: (func())(0x01436670), isAcked: true}
            	+: <non-existent>
            {[]*pubsub.Message}[?->2]:
            	-: <non-existent>
            	+: &pubsub.Message{Body: []uint8{0x32}, Metadata: map[string]string{"a": "2"}}
            
FAIL
```
One way to make the output more readable would be to make a local Message-like type with a string field for its Body and use that for the items in `got` and `want`.

ijt:
Fixes #979 

Here's a failing run of the newly added test showing how Messages printed out before this change:
```
[ ~/src/gocloud.dev/internal/pubsub ] go test
--- FAIL: TestMessageString (0.00s)
    pubsub_test.go:40: got &{Body:[] Metadata:map[] ack:<nil> mu:{state:0 sema:0} isAcked:false}, want {"" map[]}
    pubsub_test.go:40: got &{Body:[97 108 108 32 115 121 115 116 101 109 115 32 97 114 101 32 103 111] Metadata:map[] ack:<nil> mu:{state:0 sema:0} isAcked:false}, want {"all systems are go" map[]}
    pubsub_test.go:40: got &{Body:[] Metadata:map[x:1] ack:<nil> mu:{state:0 sema:0} isAcked:false}, want {"" map[x:1]}
FAIL
exit status
```

### DIFF
--- a/internal/pubsub/drivertest/drivertest.go
+++ b/internal/pubsub/drivertest/drivertest.go
@@ -17,7 +17,6 @@
 package drivertest // import "gocloud.dev/internal/pubsub/drivertest"
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"strconv"
@@ -184,7 +183,11 @@ func testSendReceive(t *testing.T, newHarness HarnessMaker) {
 	defer cleanup()
 
 	// Send to the topic.
-	var want []*pubsub.Message
+	type Message struct {
+		Body     string
+		Metadata map[string]string
+	}
+	var want []Message
 	for i := 0; i < 3; i++ {
 		m := &pubsub.Message{
 			Body:     []byte(strconv.Itoa(i)),
@@ -193,23 +196,29 @@ func testSendReceive(t *testing.T, newHarness HarnessMaker) {
 		if err := top.Send(ctx, m); err != nil {
 			t.Fatal(err)
 		}
-		want = append(want, m)
+		want = append(want, Message{
+			Body:     string(m.Body),
+			Metadata: m.Metadata,
+		})
 	}
 
 	// Receive from the subscription.
-	var got []*pubsub.Message
+	var got []Message
 	for i := 0; i < len(want); i++ {
 		m, err := sub.Receive(ctx)
 		if err != nil {
 			t.Fatal(err)
 		}
-		got = append(got, m)
+		got = append(got, Message{
+			Body:     string(m.Body),
+			Metadata: m.Metadata,
+		})
 		m.Ack()
 	}
 
 	// Check that the received messages match the sent ones.
-	less := func(x, y *pubsub.Message) bool { return bytes.Compare(x.Body, y.Body) < 0 }
-	if diff := cmp.Diff(got, want, cmpopts.SortSlices(less), cmpopts.IgnoreUnexported(pubsub.Message{})); diff != "" {
+	less := func(x, y Message) bool { return x.Body < y.Body }
+	if diff := cmp.Diff(got, want, cmpopts.SortSlices(less)); diff != "" {
 		t.Error(diff)
 	}
 }
--- a/internal/pubsub/pubsub.go
+++ b/internal/pubsub/pubsub.go
@@ -48,6 +48,10 @@ type Message struct {
 	isAcked bool
 }
 
+func (m *Message) String() string {
+	return fmt.Sprintf("{%q %v}", m.Body, m.Metadata)
+}
+
 // Ack acknowledges the message, telling the server that it does not need to be
 // sent again to the associated Subscription. It returns immediately, but the
 // actual ack is sent in the background, and is not guaranteed to succeed.


internal/pubsub/drivertest/drivertest.go
```diff
@@ -184,7 +183,11 @@ func testSendReceive(t *testing.T, newHarness HarnessMaker) {
 	defer cleanup()
 
 	// Send to the topic.
-	var want []*pubsub.Message
+	type Message struct {
```
vangent: Isn't this kind of fragile? We'll have to remember to update this if we ever add fields to `Message`.

internal/pubsub/drivertest/drivertest.go
```diff
@@ -193,23 +196,29 @@ func testSendReceive(t *testing.T, newHarness HarnessMaker) {
...
+			Metadata: m.Metadata,
+		})
 		m.Ack()
 	}
 
 	// Check that the received messages match the sent ones.
-	less := func(x, y *pubsub.Message) bool { return bytes.Compare(x.Body, y.Body) < 0 }
-	if diff := cmp.Diff(got, want, cmpopts.SortSlices(less), cmpopts.IgnoreUnexported(pubsub.Message{})); diff != "" {
+	less := func(x, y Message) bool { return x.Body < y.Body }
+	if diff := cmp.Diff(got, want, cmpopts.SortSlices(less)); diff != "" {
```
vangent: Why new `.replay` files?

internal/pubsub/drivertest/drivertest.go
```diff
@@ -193,23 +196,29 @@ func testSendReceive(t *testing.T, newHarness HarnessMaker) {
...
+			Body:     string(m.Body),
+			Metadata: m.Metadata,
+		})
 		m.Ack()
 	}
 
 	// Check that the received messages match the sent ones.
-	less := func(x, y *pubsub.Message) bool { return bytes.Compare(x.Body, y.Body) < 0 }
-	if diff := cmp.Diff(got, want, cmpopts.SortSlices(less), cmpopts.IgnoreUnexported(pubsub.Message{})); diff != "" {
+	less := func(x, y Message) bool { return x.Body < y.Body }
```
vangent: WDYT about adding a `String() string` func to `Message instead?

internal/pubsub/drivertest/drivertest.go
ijt: This type is only defined within this test. If we add fields to `Message` then this test will still work. If we add usage of those fields to this test then we will find out right away that we need to add them to this type.

internal/pubsub/drivertest/drivertest.go
ijt: I'm not sure how those got into the change. Removing them.

internal/pubsub/drivertest/drivertest.go
ijt: That works. Using that for now.

ijt: Okay, based on our chat offline, I'm going to see if I can add a `String()` method to `pubsub.Message` that gives readable output and also deals with the `Body` field possibly being non-UTF-8.

ijt:
Closing this. It probably makes more sense to start a new PR adding a `Format` method to `Message`.

internal/pubsub/drivertest/drivertest.go
ijt: Adding a String method to pubsub.Message has no effect on cmp.Diff, so that's a dead-end.

The root of the problem is all these unexported fields on Message. We can't easily get rid of them because so many things depend on them, but we can make a new type called say MessageValue that contains them and add a method to convert Message to MessageValue for testing purposes. Making a new PR for that: https://github.com/google/go-cloud/pull/1014.

PR #981, opened 2018-12-15
blob/test: add more example tests

vangent:
These are kind of lame examples, but IMHO they are still useful as concrete examples of how to use the constructors, and it's nice to be consistent about providing them.

### DIFF
--- a/blob/fileblob/example_test.go
+++ b/blob/fileblob/example_test.go
@@ -12,38 +12,49 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package s3blob_test
+package fileblob_test
 
 import (
 	"context"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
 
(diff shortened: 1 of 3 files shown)


blob/fileblob/example_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+func Example_open() {
+	// For this example, create a temporary directory.
+	dir, err := ioutil.TempDir("", "go-cloud-fileblob-example")
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer os.RemoveAll(dir)
+
+	_, err = blob.Open(context.Background(), path.Join("file://", dir))
```
clausti: this line just makes me wonder, should the URL scheme that a provider registers for be a constant of the package?

vangent: Interesting idea; I'm going to add it as a suggestion on #931.

PR #993, opened 2018-12-17
rabbitpubsub: collect all errors from SendBatch

### LINKED ISSUE #889: rabbitpubsub: improve error handling
When batch methods fail, consider returning an error that summarizes the failure (e.g. N of M messages failed), or a multi-error that includes all the individual errors.

jba:
Instead of just returning the first error we see when publishing, we collect
and return them all.

It doesn't make sense to do this for ReceiveBatch: we want to stop on the first error.

SendAcks is debatable. I don't think it's worth acking the remaining
messages if the RPC fails. We should just fail fast. (That logic doesn't apply to SendBatch, because
some publish errors can be continued from.)

Fixes #889.


ijt:
I'd like to see a test showing multiple errors.

jba:
PTAL

jba:
Ping @ijt

PR #998, opened 2018-12-17
mempubsub: remove broker

jba:
As discussed in #970, simplify mempubsub by getting rid of Broker
and named topics.


internal/pubsub/mempubsub/mem.go
```diff
@@ -91,7 +70,15 @@ func (t *topic) SendBatch(ctx context.Context, ms []*driver.Message) error {
 func (*topic) IsRetryable(error) bool { return false }
 
 // As implements driver.Topic.As.
-func (t *topic) As(i interface{}) bool { return false }
+// It supports *topic for NewSubscription (see below).
```
vangent: Hmm, I don't think this is useful to expose externally for `As`, because the `topic` type isn't exported. I suggest leaving `As` returning `false` and doing a regular type cast to `*topic` in `NewSubscription`.

internal/pubsub/mempubsub/conformance_test.go
```diff
@@ -23,38 +23,29 @@ import (
...
 func (h *harness) MakeTopic(ctx context.Context) (driver.Topic, error) {
-	dt := h.b.topic("t")
-	return dt, nil
+	return &topic{}, nil
 }
 
 func (h *harness) MakeNonexistentTopic(ctx context.Context) (driver.Topic, error) {
-	dt := h.b.topic("nonexistent-topic")
-	return dt, nil
+	return (*topic)(nil), nil
```
vangent: Might be worth commenting here that a nil `*topic` is handled correctly as a nonexistent topic, it's not obvious.

internal/pubsub/mempubsub/mem.go
jba: What would you like me to cast to `*topic`? Not `*pubsub.Topic`—that is a concrete type.

This is approximately Ross's second design in #970, the one you preferred. In Ross's words,

> One could also get fancy with As() methods and instead make NewSubscription just take in a `*pubsub.Topic`

Can you describe in detail the design you actually want?

vangent: Can you expand this a bit? Something like:

```
// It supports *topic so that NewSubscription can recover a *topic
// from the concrete type (see below). External users won't be able
// to use As because topic isn't exported.
```

jba: Done.

PR #999, opened 2018-12-18
internal/admin: codify project configuration into Terraform

zombiezen:
This allows us to version control and review changes to our repository configuration just like other changes, as well as keep both Wire and Go Cloud mostly in sync in terms of setup. Testing this already corrected multiple such discrepancies between the projects, so this seems like a good thing to have.


internal/admin/main.tf
```diff
@@ -0,0 +1,75 @@
...
+
+provider "google" {
+  version = "~> 1.15"
+}
+
+module "go_cloud_repo" {
+  source       = "./repository"
+  name         = "go-cloud"
+  description  = "A library and tools for open cloud development in Go."
+  homepage_url = "https://godoc.org/github.com/google/go-cloud"
```
jba: godoc.org/gocloud.dev

zombiezen: Done. And this is why I like having code-reviewed configurations. 🙂

zombiezen:
@jba, I need your approval before merging.

jba:
Sorry!

Issue #996, opened 2018-12-17
runtimevar: expose a non-blocking accessor on the concrete type


vangent:
Currently `runtimevar.Variable` exposes a blocking `Watch` function. Users are expected to wrap this in a goroutine that calls it in a loop.

Perhaps we could make runtimevar easier to use by doing that for them, either as a helper, a wrapper around `runtimevar.Variable`, or a new function on `runtimevar.Variable`.

vangent:
@zombiezen

vangent:
Reopening:

--- FAIL: TestVariable_Watch (0.14s)
    runtimevar_test.go:167: Watch interrupted by Close returned context canceled, want ErrClosed
FAIL

PR #1000, opened 2018-12-18
runtimevar: add a goroutine-safe Latest function that always returns the latest good value

vangent:
This is a straw man for the discussion on #996.

It adds `Latest` to `runtimevar.Variable`. Users should use `Latest` or `Watch`, but not both. It blocks on the first call until a "good" value is received (or `Close` is called), but subsequently returns the latest "good" value without blocking. It can be used concurrently (unlike `Watch`).

No changes to the driver interface were needed.

More documentation are updated are needed (e.g., to the package godoc and examples), but I'd prefer to handle those separately if needed.


runtimevar/runtimevar.go
```diff
@@ -42,32 +43,76 @@ type Snapshot struct {
...
+// goroutine to receive variable updates. Errors returned by Watch are ignored
+// in favor of returning the latest "good" value.
 //
-// If method returns an error, the returned Snapshot object is a zero value and cannot be used.
+// Applications should call Latest or Watch, not both.
 //
-// The first call to this method should return the current variable unless there are errors in
-// retrieving the value.
+// Latest is safe to be called from multiple goroutines.
+func (c *Variable) Latest() interface{} {
```
vangent: Or should this return a `Snapshot`?

runtimevar/runtimevar.go
```diff
@@ -98,6 +143,7 @@ func (c *Variable) Watch(ctx context.Context) (Snapshot, error) {
 
 // Close cleans up any resources used by the Variable object.
 func (c *Variable) Close() error {
+	c.cancel()
```
zombiezen: This also needs to synchronize with the background goroutine finishing.

runtimevar/runtimevar.go
```diff
@@ -42,32 +43,76 @@ type Snapshot struct {
...
+		ctx:     ctx,
+		cancel:  cancel,
+	}
 }
 
-// Watch blocks until there are variable changes, the Context's Done channel
-// closes or a new error occurs.
+// Latest returns the latest successfully retrieved value of the variable.
+//
+// The first call to Latest will block until the variable is successfully
```
zombiezen: If this is to block, then it should take in a context for cancellation, which implies having an error.

The more complex (but correct) thing is that what should happen here is that the first watch should start (but not block) during server startup, but block the server health checks from reporting healthy, and cause any calls to `Latest` to immediately fail without blocking (since a server can still receive traffic while in unhealthy).

runtimevar/runtimevar.go
```diff
@@ -42,32 +43,76 @@ type Snapshot struct {
...
+		for c.ctx.Err() == nil && c.lastGood == nil {
+			s, _ := c.Watch(c.ctx)
+			if s.Value != nil {
+				c.lastGood = s.Value
+			}
+		}
+		// ... and spin up a goroutine to Watch for updates.
+		go func() {
+			for c.ctx.Err() == nil {
+				s, _ := c.Watch(c.ctx)
```
zombiezen: This error should be presented to the caller somehow, which is why `Watch` has historically been presented to the end-user.

runtimevar/runtimevar.go
```diff
@@ -42,32 +43,76 @@ type Snapshot struct {
...
+	return c.lastGood
+}
+
+// Watch blocks until there are variable changes, the Context's Done channel
+// closes or a new error occurs. It returns a Snapshot containing the variable's
+// Value, or an error.
 //
-// Users should not call this method from multiple goroutines as implementations may not guarantee
-// safety in data access.  It is typical to use only one goroutine calling this method in a loop.
+// Applications should call Latest or Watch, not both.
```
zombiezen: If this is the case, then there should be a runtime signifier of this being an invalid state, or even better, an anti-affordance to prevent this from happening.

runtimevar/runtimevar.go
vangent: Updated to take a context and return an error.

runtimevar/runtimevar.go
vangent: It now panics.

runtimevar/runtimevar.go
vangent: Done using `errCh`.

runtimevar/runtimevar.go
vangent: Done.

vangent:
PTAL, let me know what you think.

I think this is better:
- It has two separate functions, `Latest` and `InitLatest`.
- You don't have to call the `InitLatest` unless you want to prewarm and/or process errors. `InitLatest` is intended to be called once during initialization; it can block via a `ctx`, or not if you pass `nil`.
- `Latest` also takes a context (or `nil` if you don't want any blocking). It returns a `Snapshot` or a specific error `ErrNoValue`. It is intended to be called early on during request processing; the rest of the request processing can deal with the `Snapshot` (or something(s) pulled out of it).
- `Close` now synchronizes the goroutine being shut down properly.
- `Watch` panics if you call it after `InitLatest` or `Latest`; applications should do one or the other.

If this looks good I'll add some more package-level documentation, and examples (possibly in separate CLs, LMKWYT).

vangent:
PTAL.

zombiezen:
Apologies: I haven't gotten to this because I've had higher priority work preempting. This still doesn't seem ideal to use, but I haven't had time to collect my thoughts and give a constructive counter-proposal yet.

Let me know if you want me to bump priority.

vangent:
Closing, will re-send as part of design.

PR #1005, opened 2018-12-18
runtimevar: add support for As functions

### LINKED ISSUE #1004: runtimevar: add support for As

vangent:
In particular, `Snapshot.As` and `ErrorAs`.

I used the same pattern as in `blob` and `pubsub`, including for the tests.

Fixes #1004.


runtimevar/runtimeconfigurator/runtimeconfigurator.go
```diff
@@ -106,19 +113,36 @@ type Options struct {
...
 func (s *state) UpdateTime() time.Time {
 	return s.updateTime
 }
 
+// As implements driver.State.As.
+func (s *state) As(i interface{}) bool {
+	if s.raw == nil {
+		return false
+	}
+	switch p := i.(type) {
```
ijt: Why is this done differently from the others?

vangent: I think because I copy/pasted from `paramstore`, which supports multiple types so used a `switch` instead of a type assertion. Fixed.

vangent:
PTAL.

PR #1014, opened 2018-12-18
pubsub: add MessageValue type and Message.Value()

### LINKED ISSUE #979: pubsub: output of TestSendReceive conformance test is hard to read on failure

ijt:
Fixes #979 

Here is a crude comparison of how the diffs look before and after this change. These are from different test failures but hopefully the idea comes across anyway:

Before:
```
[ ~/src/gocloud.dev/internal/pubsub/gcppubsub ] go test -record
--- FAIL: TestConformance (1.78s)
    --- FAIL: TestConformance/TestSendReceive (1.14s)
        replay.go:196: Recording into golden file testdata/TestConformance/TestSendReceive.replay
        drivertest.go:213: {[]*pubsub.Message}[0->?]:
            	-: &pubsub.Message{Body: []uint8{0x34, 0x31}, ack: (func())(0x01436670), isAcked: true}
            	+: <non-existent>
```

After:
```
[ ~/src/gocloud.dev/internal/pubsub/gcppubsub ] go test -record ./...
--- FAIL: TestConformance (4.05s)
    --- FAIL: TestConformance/TestSendReceive (3.45s)
        replay.go:196: Recording into golden file testdata/TestConformance/TestSendReceive.replay
        drivertest.go:212: {[]pubsub.MessageValue}:
            	-: []pubsub.MessageValue(nil)
            	+: []pubsub.MessageValue{{Body: "0", Metadata: map[string]string{"a": "0"}}, {Body: "1", Metadata: map[string]string{"a": "1"}}, {Body: "2", Metadata: map[string]string{"a": "2"}}}
```

### DIFF
--- a/internal/pubsub/pubsub.go
+++ b/internal/pubsub/pubsub.go
@@ -61,6 +61,24 @@ func (m *Message) Ack() {
 	m.isAcked = true
 }
 
+// Value returns a MessageValue based on the Message receiver.
+func (m *Message) Value() MessageValue {
+	return MessageValue{
+		Body:     string(m.Body),
+		Metadata: m.Metadata,
+	}
+}
+
+// MessageValue contains the exported fields of a Message with a UTF8 Body, making it
+// easier to read and compare in diffs and other output.
+type MessageValue struct {
+	// Body contains the content of the message.
+	Body string
+
+	// Metadata has key/value metadata for the message.
+	Metadata map[string]string
+}
+
 // Topic publishes messages to all its subscribers.
 type Topic struct {
 	driver  driver.Topic
--- a/internal/pubsub/drivertest/drivertest.go
+++ b/internal/pubsub/drivertest/drivertest.go
@@ -17,7 +17,6 @@
 package drivertest // import "gocloud.dev/internal/pubsub/drivertest"
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"strconv"
@@ -184,7 +183,7 @@ func testSendReceive(t *testing.T, newHarness HarnessMaker) {
 	defer cleanup()
 
 	// Send to the topic.
-	var want []*pubsub.Message
+	var want []pubsub.MessageValue
 	for i := 0; i < 3; i++ {
 		m := &pubsub.Message{
 			Body:     []byte(strconv.Itoa(i)),
@@ -193,23 +192,23 @@ func testSendReceive(t *testing.T, newHarness HarnessMaker) {
 		if err := top.Send(ctx, m); err != nil {
 			t.Fatal(err)
 		}
-		want = append(want, m)
+		want = append(want, m.Value())
 	}
 
 	// Receive from the subscription.
-	var got []*pubsub.Message
+	var got []pubsub.MessageValue
 	for i := 0; i < len(want); i++ {
 		m, err := sub.Receive(ctx)
 		if err != nil {
 			t.Fatal(err)
 		}
-		got = append(got, m)
+		got = append(got, m.Value())
 		m.Ack()
 	}
 
 	// Check that the received messages match the sent ones.
-	less := func(x, y *pubsub.Message) bool { return bytes.Compare(x.Body, y.Body) < 0 }
-	if diff := cmp.Diff(got, want, cmpopts.SortSlices(less), cmpopts.IgnoreUnexported(pubsub.Message{})); diff != "" {
+	less := func(x, y pubsub.MessageValue) bool { return x.Body < y.Body }
+	if diff := cmp.Diff(got, want, cmpopts.SortSlices(less)); diff != "" {
 		t.Error(diff)
 	}
 }


ijt:
See also https://www.infoq.com/presentations/Value-Values.

internal/pubsub/pubsub.go
```diff
@@ -61,6 +61,24 @@ func (m *Message) Ack() {
...
 
+// Value returns a MessageValue based on the Message receiver.
+func (m *Message) Value() MessageValue {
+	return MessageValue{
+		Body:     string(m.Body),
+		Metadata: m.Metadata,
+	}
+}
+
+// MessageValue contains the exported fields of a Message with a UTF8 Body, making it
```
vangent: This just feels like overkill to me -- a whole new exported struct type just so diffs look nicer in our tests? It's not obvious to me that it will be useful to end users at all; they may want to log Messages, but it's not obvious that MessageValue.String() will produce something useful.

Why not a best-effort `String()` function that does something like:
- Take the first N bytes of `Body`
  - If it's a string, print it, with "... (%d bytes total)" appended if len(Body) > N.
  - If it's not a string, print "body with binary data (%d bytes)".
- Append the number of metadata keys (e.g., "(%d metadata keys)") if len(Metadata) > 0, or maybe a string representation of the metadata if it's short enough.

The test could capture a slice of both `*Message` and strings from `Message.String()`, and compare the strings first (resulting in a nice diff if they are not the same), and the actual messages (hiding unexported fields) second (resulting in the existing diff if `String()` returned the same thing but the underlying `Message`s are different; this should be extremely unlikely for the short messages used in the test).

ijt: Yes, I think the improvement in readability of test output is well worth the cost of an additional exported type.

Why is it not obvious that `MessageValue.String()` will produce something useful? Its fields are the ones with the data users have provided and presumably care about.

The main reason not add a best-effort `String()` method on Message is because it fails to change the output of `cmp.Diff`, which was what motivated this change in the first place. I don't think it would give a very readable result to use `cmp.Diff` on a string that contains a bunch of message `String`s.

Also `String` is not very flexible, forcing you to choose between breaking "%v" or "%+v". It's possible to implement `Formatter` but the code for that is more complex and harder to test. I got partway through doing that and wasn't sure if I had covered all the cases. With this PR, all the cases are covered.

jba: @ijt I strongly disagree. We should not be changing the API just for tests. I might understand if it was necessary to exercise some behavior that we couldn't otherwise, but this is just cosmetic.

ijt: Why not? Testability is important, and having readable output when tests fail is also important.

vangent: I think the difference is the size of the audience. The number of end users exposed to the exported API will (hopefully) be orders of magnitude larger than the users exposed to the test code.

Personally I don't find readable output when tests fail super high on my list of priorities. When the test fails, you end up in the code debugging anyway, and if it's not obvious what the problem is you can usually easily add code to make it obvious. I'd rather have the test be simple and straightforward.

I still find it interesting to consider how to make it easy/useful for users to log/print a `Message`, but for me at least the impact on the test is secondary.

ijt: Ah, if it's a low priority then I won't spend any more time on it. Closing this.

PR #1016, opened 2018-12-19
pubsub: add gcmsg sample command

### LINKED ISSUE #978: pubsub: add cross-cloud examples
These should go into the top-level `samples` directory once `pubsub` moves out of internal.

Some ideas:
* `pub`: command line tool to publish a message to a topic, like `pub -env=gcp -t "/fully/qualified/path/to/topic" "message"`
* `sub`: command line tool to tail from an existing subscription, like `sub -env=aws -s "SQS queue URL"`

### LINKED ISSUE #1024: pubsub: use helpers from PR #1022 in pubsub/samples programs
They should shorten the code.

ijt:
Fixes #978 
Fixes #1024


internal/pubsub/samples/pub/main.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		log.Println(err)
+		return subcommands.ExitFailure
+	}
+	return subcommands.ExitSuccess
+}
+
+type delCmd struct{}
+
+func (*delCmd) Name() string     { return "del" }
```
vangent: Why the `new` and `delete` commands? I buy a simple app that shows how to use the pubsub APIs to send to a topic and subscribe/tail a topic, but this functionality isn't in our API.

ijt: Done. PTAL.

internal/pubsub/samples/pub/main.go
```diff
@@ -0,0 +1,64 @@
...
+	// Get lines from stdin and send them as messages to the topic.
+	scanner := bufio.NewScanner(os.Stdin)
+	for scanner.Scan() {
+		line := scanner.Text()
+		m := &pubsub.Message{Body: []byte(line)}
+		if err := topic.Send(ctx, m); err != nil {
+			return err
+		}
+	}
+
```
jba: Check scanner.Err() here.

internal/pubsub/samples/pubsub_test.go
```diff
@@ -0,0 +1,91 @@
...
+var subFlag = flag.String("sub", "test-subscription-1", "subscription to be used for testing pub and sub commands")
+var envsFlag = flag.String("pubsub_envs", "gcp,rabbit", "environments in which to run the test")
+
+func TestMain(m *testing.M) {
+	flag.Parse()
+	os.Exit(m.Run())
+}
+
+func TestPubAndSubCommands(t *testing.T) {
+	if !commandExists("pub") {
```
jba: Couldn't you just let runWithInput or run fail?

internal/pubsub/samples/pub/main.go
ijt: Done.

internal/pubsub/samples/pubsub_test.go
ijt: That would break the build on Travis and for anyone who hasn't compiled the `pub` and `sub` example programs.

ijt:
PTAL.

internal/pubsub/samples/doc.go
```diff
@@ -0,0 +1,2 @@
+// package samples contains sample programs using the pubsub API.
```
vangent: Please add the Copyright header.

internal/pubsub/samples/pub/gcp.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	topic := gcppubsub.OpenTopic(ctx, pubClient, proj, topicID, nil)
+	cleanup2 := func() {
+		topic.Shutdown(ctx)
+		cleanup()
+	}
+	return topic, cleanup2, nil
+}
+
+func getGCPProjectID() (gcp.ProjectID, error) {
```
vangent: FYI, there is a `gcp.DefaultProjectID` function, maybe use that instead of defining your own? It gets them from the credentials instead of the environment which is probably more robust.

Maybe openGCPPubClient / openGCPSubClient could return the `gcp.ProjectID`.

internal/pubsub/samples/pub/gcp.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	return gcp.ProjectID(p), nil
+}
+
+func openGCPPubClient(ctx context.Context) (*raw.PublisherClient, func(), error) {
+	creds, err := gcp.DefaultCredentials(ctx)
+	if err != nil {
+		return nil, nil, fmt.Errorf("getting default credentials: %v", err)
+	}
+	opts := []grpc.DialOption{
```
vangent: FYI I have PR #1022 out that adds some helpers to `gcspubsub.go` to do the `Dial` and create the `raw.(Publisher/Subscriber)Client` for you; using them would shorten this considerably.

internal/pubsub/samples/pub/gcp.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, nil, fmt.Errorf("making publisher client: %v", err)
+	}
+	cleanup := func() {
+		conn.Close()
+		pubClient.Close()
+	}
+	return pubClient, cleanup, nil
+}
+
+func gcpTopicName(proj gcp.ProjectID, topicID string) string {
```
vangent: Where is this used?

internal/pubsub/samples/pub/main.go
```diff
@@ -0,0 +1,77 @@
...
+func main() {
+	log.SetFlags(0)
+	log.SetPrefix("pub: ")
+	envFlag := flag.String("env", "", "environment [gcp|rabbit]")
+	flag.Parse()
+	if *envFlag == "" {
+		log.Fatal("missing -env flag\n")
+	}
+	if flag.NArg() != 1 {
+		fmt.Fprintf(os.Stderr, "usage: pub -env=<env> [flags] <topic-name>\n")
```
vangent: Is there a bug yet about `pubsub` supporting URLs? That would simplify this, it would be more like:

`pub gcspubsub://my-topic`

internal/pubsub/samples/pub/main.go
```diff
@@ -0,0 +1,77 @@
...
+	case "rabbit":
+		topic, cleanup, err = openRabbitTopic(topicID)
+	default:
+		return fmt.Errorf("unrecognized -env: %s", env)
+	}
+	if err != nil {
+		return err
+	}
+	defer cleanup()
+
```
vangent: Maybe a `fmt.Println` to stdout (stderr ?) here telling the user that the connection was successful and that they can enter lines to be sent? Otherwise it seems like "pub ..." will just hang.

internal/pubsub/samples/pub/main.go
```diff
@@ -0,0 +1,77 @@
...
+	// Get lines from stdin and send them as messages to the topic.
+	scanner := bufio.NewScanner(os.Stdin)
+	for scanner.Scan() {
+		line := scanner.Text()
```
vangent: How does the user get out of this loop?

internal/pubsub/samples/pub/rabbit.go
```diff
@@ -0,0 +1,34 @@
...
+
+package main
+
+import (
+	"github.com/streadway/amqp"
+	"gocloud.dev/internal/pubsub"
+	"gocloud.dev/internal/pubsub/rabbitpubsub"
+)
+
+const rabbitURL = "amqp://guest:guest@localhost:5672/"
```
vangent: Shouldn't the user be able to specify this?

internal/pubsub/samples/pub/rabbit.go
```diff
@@ -0,0 +1,34 @@
...
+
+// openRabbitTopic returns the RabbitMQ topic for the given topic ID.
+func openRabbitTopic(topicID string) (*pubsub.Topic, func(), error) {
+	conn, err := amqp.Dial(rabbitURL)
```
vangent: It is surprising that this doesn't take a context. What happens if the server isn't running?

internal/pubsub/samples/sub/main.go
```diff
@@ -0,0 +1,76 @@
+// Copyright 2018 The Go Cloud Authors
```
vangent: `sub` seems to overlap heavily with `pub`... WDYT about combining them? `gocloudpubsub pub mytopic`, `gocloudpubsub sub mytopic`....

internal/pubsub/samples/sub/gcp.go
```diff
@@ -0,0 +1,86 @@
...
+func openGCPSubscriberClient(ctx context.Context) (*raw.SubscriberClient, func(), error) {
+	creds, err := gcp.DefaultCredentials(ctx)
+	if err != nil {
+		return nil, nil, fmt.Errorf("getting default credentials: %v", err)
+	}
+	opts := []grpc.DialOption{
+		grpc.WithTransportCredentials(grpccreds.NewClientTLSFromCert(nil, "")),
+		grpc.WithPerRPCCredentials(oauth.TokenSource{TokenSource: gcp.CredentialsTokenSource(creds)}),
+	}
+	conn, err := grpc.DialContext(ctx, gcppubsub.EndPoint, opts...)
```
zombiezen: Seems like there's a fair amount of shared code between this and the `pub` command. Would joining them into a single binary make sense?

internal/pubsub/samples/pub/gcp.go
ijt: Done.

internal/pubsub/samples/pub/gcp.go
ijt: Created issue https://github.com/google/go-cloud/issues/1024 for this.

internal/pubsub/samples/pub/gcp.go
ijt: It's not. Removed it.

internal/pubsub/samples/pub/main.go
ijt: Filed https://github.com/google/go-cloud/issues/1025 for the more general question of whether to support URLs throughout Go Cloud.

ijt: There doesn't seem to be an issue for that specific to pubsub.

internal/pubsub/samples/pub/gcp.go
ijt: This doesn't seem to work well. After making this change, it outputs this error 
```
pub: rpc error: code = NotFound desc = Requested project not found or user does not have access to it (project=). Make sure to specify the unique project identifier and not the Google Cloud Console display name.
```
even after running
```
$ gcloud config set project go-cloud-test-216917
```

internal/pubsub/samples/pub/main.go
ijt: Done.

internal/pubsub/samples/pub/rabbit.go
ijt: It quickly returns with an error:
```
[ ~/gocloud.dev/internal/pubsub/samples/pub ] pub -rabbiturl=amqp://nonexistent -env rabbit foo 
pub: dial tcp: lookup nonexistent: no such host
[ ~/gocloud.dev/internal/pubsub/samples/pub ] pub -rabbiturl=amqp://localhost:8080 -env rabbit foo 
pub: dial tcp 127.0.0.1:8080: connect: connection refused
```

internal/pubsub/samples/doc.go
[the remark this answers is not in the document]
ijt: Done.

internal/pubsub/samples/sub/main.go
ijt: It's a fair point about overlapping code, but I'm not liking the verbosity and stuttering.

How about we call it `msgs`. Then we can write for example:
```
msgs sub $subscription
echo "the eagle has landed" | msgs pub $topic
```

ijt: Or maybe `msgs send $topic` and `msgs recv $subscription`.

ijt: It turns out that `msgs` is the name of an existing unix command. Other possibilities include `msg` and `gcmsg`.

ijt: Done.

internal/pubsub/samples/pub/gcp.go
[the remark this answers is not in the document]
ijt: Done.

internal/pubsub/samples/pub/gcp.go
```diff
@@ -0,0 +1,85 @@
...
+	"google.golang.org/grpc/credentials/oauth"
+)
+
+// openGCPTopic returns the GCP topic based on the topic ID.
+func openGCPTopic(ctx context.Context, topicID string) (*pubsub.Topic, func(), error) {
+	pubClient, cleanup, err := openGCPPubClient(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+	proj, err := getGCPProjectID()
```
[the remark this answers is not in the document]
ijt: This could be combined with @vangent's suggestion of using URLs, so everything including the project ID would go into a single parameter. Then we could write things like
```
$ t=gcppubsub://projects/go-cloud-test-216917/topics/test-topic
$ s=gcppubsub://projects/go-cloud-test-216917/subscriptions/test-subscription-1
$ echo foo | gcmsg pub $t
$ gcmsg sub $s
```

internal/pubsub/samples/pubsub_test.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return "", fmt.Errorf("running %v: %s", c.Args, out)
+	}
+	wg.Wait()
+	return string(out), nil
+}
+
+func commandExists(name string) bool {
+	c := exec.Command(name)
+	err := c.Start()
```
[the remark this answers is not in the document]
ijt: Done.

internal/pubsub/samples/sub/gcp.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	sub := gcppubsub.OpenSubscription(ctx, subClient, proj, subID, nil)
+	cleanup2 := func() {
+		sub.Shutdown(ctx)
+		cleanup()
+	}
+	return sub, cleanup2, nil
+}
+
+func getGCPProjectID() (gcp.ProjectID, error) {
```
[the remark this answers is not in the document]
ijt: obsolete.

internal/pubsub/samples/sub/gcp.go
ijt: Done.

internal/pubsub/samples/pub/gcp.go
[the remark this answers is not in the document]
ijt: Done.

ijt:
I changed it to the URL-based approach we (@jba, @vangent) agreed on during this morning's meeting.

PTAL

internal/pubsub/samples/gcmsg/main.go
```diff
@@ -0,0 +1,224 @@
...
+		}
+		m := rx.FindStringSubmatch(rightPart)
+		if len(m) != 3 {
+			return nil, nil, fmt.Errorf(`rabbitpubsub topic URL contents "%s" failed to match "%s"`, rightPart, rx)
+		}
+		serverURL := "amqp://" + m[1]
+		topicID := m[2]
+		return openRabbitTopic(serverURL, topicID)
+	case "":
+		return nil, nil, fmt.Errorf(`scheme missing from URL: "%s"`, topicURL)
```
jba: Replace
  "%s"
with
  %q

Then you can use ordinary "" for the string instead of backquotes.

zombiezen:
I still don't have a clear answer on [this question I raised](https://github.com/google/go-cloud/pull/1016#discussion_r243443839):

> But then these tests never run on Travis. What's the behavior this is really testing beyond what is already tested in the Pub/Sub tests?

internal/pubsub/samples/pubsub_test.go
ijt: This tests the logic in gcmsg/main.go. I feel I may not be quite understanding the question.

ijt: I changed the test to no longer rely on the `gcmsg` command being on the `$PATH` and now your point seems clearer. We could get most of the value from the test by extracting the URL parsing logic and writing unit tests for that. Then we wouldn't need any record/replay on it.

zombiezen: I was making a higher-level point: tests have maintenance cost, too. Ideally, this sample should be straightforward enough that basic manual usage and code review should be enough to have confidence that the sample works. If there is logic in the sample complex enough that tests are buying us confidence in its correctness, then that's signal that the API may be deficient. If the tests never run on Travis, then we don't receive much benefit from them and we incur maintenance cost.

(The only other places in Go Cloud that I'm aware of right now that have tests that don't run on Travis are the server and SQL dialer packages, and this is because record/replay is infeasible due to their nature. It would bring significantly more benefit if they ran regularly, but that brings other costs.)

internal/pubsub/samples/gcmsg/gcp.go
```diff
@@ -0,0 +1,109 @@
...
+	topic := gcppubsub.OpenTopic(ctx, pubClient, gcp.ProjectID(proj), topicID, nil)
+	cleanup2 := func() {
+		topic.Shutdown(ctx)
+		cleanup()
+	}
+	return topic, cleanup2, nil
+}
+
+func openGCPPubClient(ctx context.Context, creds *google.Credentials) (*raw.PublisherClient, func(), error) {
+	creds, err := gcp.DefaultCredentials(ctx)
```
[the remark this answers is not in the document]
ijt: Fixed.

internal/pubsub/samples/gcmsg/gcp.go
```diff
@@ -0,0 +1,109 @@
...
+	conn, cleanup, err := gcppubsub.Dial(ctx, ts)
+	if err != nil {
+		return nil, nil, fmt.Errorf("dialing grpc endpoint: %v", err)
+	}
+	pubClient, err := raw.NewPublisherClient(ctx, option.WithGRPCConn(conn))
+	if err != nil {
+		return nil, nil, fmt.Errorf("making publisher client: %v", err)
+	}
+	cleanup2 := func() {
+		cleanup()
```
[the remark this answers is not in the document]
ijt: Done.

internal/pubsub/samples/gcmsg/main.go
```diff
@@ -0,0 +1,224 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("parsing URL: %v", err)
+	}
+	rightPart := topicURL[len(u.Scheme+"://"):]
+	if rightPart == "" {
+		return nil, nil, errors.New("empty contents in topic URL")
+	}
+	switch u.Scheme {
+	case "gcppubsub":
+		rx, err := regexp.Compile(`^projects/([^/]+)/topics/([^/]+)$`)
```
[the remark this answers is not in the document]
ijt: Done.

internal/pubsub/samples/gcmsg/main.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// openTopic opens a topic on a supported pubsub provider, given a topic URL in a format
+// specific to gcmsg.
+func openTopic(ctx context.Context, topicURL string) (*pubsub.Topic, func(), error) {
+	u, err := url.Parse(topicURL)
+	if err != nil {
+		return nil, nil, fmt.Errorf("parsing URL: %v", err)
+	}
+	rightPart := topicURL[len(u.Scheme+"://"):]
```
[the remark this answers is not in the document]
ijt: Changed over to using a regex for this.

internal/pubsub/samples/gcmsg/main.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	switch u.Scheme {
+	case "gcppubsub":
+		rx, err := regexp.Compile(`^projects/([^/]+)/topics/([^/]+)$`)
+		if err != nil {
+			panic(fmt.Sprintf("gcppubsub topic regex failed to copile: %v", err))
+		}
+		m := rx.FindStringSubmatch(rightPart)
+		if len(m) != 3 {
+			return nil, nil, fmt.Errorf(`gcppubsub topic URL contents "%s" failed to match "%s"`, rightPart, rx)
```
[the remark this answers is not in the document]
ijt: It's not impossible. Users can provide bad input like "gcppubsub://" in which case len(m) is 0.

internal/pubsub/samples/gcmsg/main.go
ijt: %q also escapes characters within the string, but I don't want that.

internal/pubsub/samples/pubsub_test.go
ijt: I think I generally agree with your point here.

The URL parsing logic is complex enough to need tests. This logic is only provisionally meant to live in the sample. The plan is to move it into the pubsub API if users say they want that, or if we decide it makes sense for some other reason such as consistency with blob's URL functionality.

ijt:
Changes are up. PTAL.

internal/pubsub/samples/gcmsg/main.go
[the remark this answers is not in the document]
jba: I looked at the new version. I'm still confused why you're not using `net/url.Parse`. It seems to do some of this work for you, at least the initial split of the scheme.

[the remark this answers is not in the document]
ijt: The only work that needs to be done is to split on "://". Using `net/url.Parse` for this means I would then have to extract all the stuff to the right of "://" using different fields of the `URL` object since apparently "all the stuff to the right of ://" doesn't have a name. Using a regex makes it easy and clear to say what I want. Why block on this?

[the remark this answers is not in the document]
jba: Not insisting.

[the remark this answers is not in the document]
zombiezen: You aren't guaranteed that the characters will actually be `"://"`, though. Note that `url.Parse` will accept a URI like `scheme:opaque?query`, in which case this code would misbehave.

URI/URL parsing is extraordinarily subtle. I would recommend using the `net/url` package consistently, especially if this function is eventually going to be made part of the larger Pub/Sub package. Using this string-based method does not handle percent encoding, for example.

At the very least, we should have an issue open for this.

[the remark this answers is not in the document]
ijt: Created issue https://github.com/google/go-cloud/issues/1090.

PR #1018, opened 2018-12-19
runtimevar/doc: improve/add examples for driver implementations

vangent:
Updates #896.

### DIFF
--- a/runtimevar/etcdvar/example_test.go
+++ b/runtimevar/etcdvar/example_test.go
@@ -12,16 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package filevar_test
+package etcdvar_test
 
 import (
-	"context"
-	"fmt"
-	"io/ioutil"
 	"log"
 
+	"github.com/coreos/etcd/clientv3"
 	"gocloud.dev/runtimevar"
-	"gocloud.dev/runtimevar/filevar"
+	"gocloud.dev/runtimevar/etcdvar"
 )
 
 // MyConfig is a sample configuration struct.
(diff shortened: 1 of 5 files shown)


runtimevar/etcdvar/example_test.go
```diff
@@ -0,0 +1,52 @@
...
+	// For this example, the etcd variable being referenced should have a
+	// JSON string that decodes into MyConfig.
+	v, err := etcdvar.New("myconfig", client, decoder, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer v.Close()
+
+	// You can now read the current value of the variable from v.
+	// Snapshot.Value will be of type MyConfig.
```
ijt: Where is Snapshot defined?

runtimevar/paramstore/example_test.go
```diff
@@ -0,0 +1,50 @@
...
+	// For this example, the Parameter Store variable being referenced
+	// should have a JSON string that decodes into MyConfig.
+	v, err := paramstore.NewVariable(session, "myconfig", decoder, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer v.Close()
+
+	// You can now read the current value of the variable from v.
+	// Snapshot.Value will be of type MyConfig.
```
ijt: Where is Snapshot defined. Also, why not give an example of using Snapshot.Value or v rather than just mentioning them in the comment?

runtimevar/runtimeconfigurator/example_test.go
```diff
@@ -35,37 +43,25 @@ func Example() {
...
-	snapshot, err := v.Watch(ctx)
-	if err != nil {
-		log.Fatal(err)
-	}
-
-	// Value will always be of the decoder's type.
-	cfg := snapshot.Value.(*MyAppConfig)
-	log.Println(cfg.MsgOfTheDay)
+	// You can now read the current value of the variable from v.
+	// Snapshot.Value will be of type MyConfig.
```
ijt: Same question here.

runtimevar/etcdvar/example_test.go
vangent: `Snapshot` is defined in runtimevar:
https://github.com/google/go-cloud/blob/master/runtimevar/runtimevar.go#L52

I updated the comment to fully qualify it as `runtimevar.Snapshot` (here and elsewhere).

runtimevar/paramstore/example_test.go
vangent: See above regarding `Snapshot`'s definition.

Yeah, extended the example. I find these examples where they don't actually run weird.

runtimevar/runtimeconfigurator/example_test.go
vangent: Done.

runtimevar/runtimeconfigurator/example_test.go
```diff
@@ -63,5 +63,10 @@ func ExampleNew() {
 	defer v.Close()
 
 	// You can now read the current value of the variable from v.
-	// Snapshot.Value will be of type MyConfig.
+	snapshot, err := v.Watch(context.Background())
+	if err != nil {
+		log.Fatal(err)
+	}
+	// The resulting runtimevar.Snapshot.Value will be of type MyConfig.
```
ijt: You could write `var val MyConfig = runtimevar.Snapshot.Value` instead of the comment to have the test check the claim.

vangent: `Snapshot.Value` is in interface{}, so that assignment won't work without a cast.

ijt: Why not do the cast?

Why not add an `// Output` line?

vangent: > Why not do the cast?
? I do do the cast, it's on the next line. I don't think it makes it clearer to split it into two lines.

> Why not add an // Output line?
With what? These examples are for non-local providers, so they aren't going to actually work if run. All it could do is print out some static text, or an error.

PR #1022, opened 2018-12-20
all: make the User-Agent we sent to GCP include the api name

### LINKED ISSUE #144: all: add go-cloud to GCP client library User-Agent string
Post launch.

### LINKED ISSUE #941: pubsub: gcppubsub should provide a Dial option similar to runtimeconfigurator
https://github.com/google/go-cloud/blob/master/runtimevar/runtimeconfigurator/runtimeconfigurator.go#L56

It's not required to use, but plays well with `wire` and the `gcp` provider set. It also will give us a good place to insert a `WithUserAgent` option where at least a good percentage of users will use it.

`Endpoint` can then be `endPoint`.

`example_test.go` should also be updated.

vangent:
This will allow us to differentiate between requests from different APIs.

I also added a couple of helpers for `gcppubsub` that set the option.

Fixes #941.
Fixes #144.

### DIFF
--- a/blob/fileblob/attrs.go
+++ b/blob/fileblob/attrs.go
@@ -30,7 +30,7 @@ var errAttrsExt = fmt.Errorf("file extension %q is reserved", attrsExt)
 type xattrs struct {
 	ContentType string            `json:"user.content_type"`
 	Metadata    map[string]string `json:"user.metadata"`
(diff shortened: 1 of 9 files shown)


vangent:
Tests will fail due to mismatched User-Agent headers; I'll re-record before submitting.

PR #1029, opened 2018-12-21
all: add script to update proxy

ijt:
The new script `internal/proxy/add.sh` automates the previously tedious process of adding a dependency to the module proxy.


internal/proxy/add.sh
```diff
@@ -0,0 +1,63 @@
...
+function log() {
+	echo "add.sh: $@" 1>&2
+}
+
+function die() {
+	log "$@"
+	exit 1
+}
+
+if ! [[ -d .git ]]; then die 'this script must be run from the root of the repository'; fi
```
zombiezen: This doesn't have to be an error: you can use the relative position from [the script's source directory](https://stackoverflow.com/questions/59895/getting-the-source-directory-of-a-bash-script-from-within) to make any working directory be okay.

internal/proxy/add.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+# add.sh adds new dependencies to the Go Cloud module proxy. This script
+# should be run from the root of the repository.
+
+set -o pipefail
```
zombiezen: Check the other bash scripts: we are consistently setting `-e` and a few others.

```suggestion
# https://coderwall.com/p/fkfaqq/safer-bash-scripts-with-set-euxo-pipefail
set -euxo pipefail
```

internal/proxy/add.sh
```diff
@@ -0,0 +1,63 @@
...
+
+if ! [[ -d .git ]]; then die 'this script must be run from the root of the repository'; fi
+
+# This approach was suggested by:
+# https://github.com/go-modules-by-example/index/tree/master/012_modvendor
+
+log "creating temp dir where we'll create a module download cache"
+tgp="$(mktemp -d)"
+if [[ $? != 0 ]]; then die 'failed'; fi
+
```
zombiezen: Once you've created the temporary directory, the script should use `trap [...] EXIT` to ensure that the directory is cleaned up. (Same with the temporary log file below.)

internal/proxy/add.sh
```diff
@@ -0,0 +1,63 @@
...
+temp=$(mktemp)
+if [[ $? != 0 ]]; then die 'failed to create temp file for verbose command output'; fi
+
+# Copy current module cache into temporary directory as basis.
+# Periodically, someone on the project should go through this process without
```
zombiezen: There should be a command-line option to skip this initial `rsync` to permit the workflow that this documentation was stating. This comment should live as part of user documentation.

internal/proxy/add.sh
```diff
@@ -0,0 +1,63 @@
...
+# running this step to prune unused dependencies.
+log 'making temp download dir'
+if ! mkdir -p "$tgp/pkg/mod/cache/download"; then die 'failed'; fi
+log "downloading current module cache, logging to $temp"
+if ! gsutil -m rsync -r gs://go-cloud-modules "$tgp/pkg/mod/cache/download" > $temp 2>&1; then die 'failed'; fi
```
zombiezen: Here and elsewhere: I think I'm okay with having the logs be directed away on the happy path, but it is much easier for an end-user to see the logs on failure rather than having to open a file that isn't being cleaned up.

internal/proxy/add.sh
```diff
@@ -0,0 +1,63 @@
...
+
+log "filling cache with all module dependencies from current branch, logging to $temp"
+if ! ./internal/proxy/makeproxy.sh "$tgp" > $temp 2>&1; then die 'failed while running makeproxy.sh'; fi
+
+log 'moving the temporary cache to modvendor'
+if ! rm -rf modvendor; then die 'failed to remove modvendor dir'; fi
```
zombiezen: Copying the cache to a local directory is helpful for human introspection, but not necessary AFAIK for scripted operation. I believe you can do the `rsync` without ill effects.

internal/proxy/add.sh
zombiezen: Here and elsewhere: make sure to quote `$temp`.

internal/proxy/add.sh
```diff
@@ -0,0 +1,63 @@
...
+
+log 'previewing synchronization of modvendor to the proxy'
+# -n: preview only
+# -r: recurse directories
+# -c: compare checksums not write times
+# -d: delete remote files that aren't present locally
+if ! gsutil rsync -n -r -c -d modvendor gs://go-cloud-modules; then die 'gsutil rsync failed'; fi
+# If the set of packages being added and removed looks good,
+# repeat without the -n.
+echo 'if this looks good, hit ENTER to continue, otherwise hit ctrl-C to quit'
```
zombiezen: An interrupt signal would leave the temporary files behind. Let's prompt for an explicit "yes" and all other inputs `exit 1` after cleanup.

internal/proxy/README.md
```diff
@@ -17,48 +17,6 @@ dependency to the proxy:
 1.  Ensure that the new dependency is covered under one of the `notice`,
     `permissive`, or `unencumbered` licences under the categories defined by
     [Google Open Source](https://opensource.google.com/docs/thirdparty/licenses/).
-2.  Gather the new set of dependencies and sync them to the GCS bucket.
+2.  Gather the new set of dependencies and sync them to the GCS bucket by running
+    `internal/proxy/add.sh` from the root of the Go Cloud repository.
```
zombiezen: I recommend adding a sentence like the following to show that this is an interactive script:

> Review the files to be sent to the proxy when prompted by the script.

internal/proxy/add.sh
ijt: makeproxy.sh is called by add.sh and makeproxy.sh also assumes it's being called from the root dir of the repo. I don't think it's worth it to make these scripts runnable from anywhere. Better to just keep them simple and fail fast.

internal/proxy/add.sh
ijt: Adding e just in case (not really needed since I check for errors everywhere, but may catch something if someone forgets) and u because that can catch real bugs. I'm leaving out x because I want the script to calmly describe what it's doing in English instead of filling the screen with noise.

internal/proxy/add.sh
ijt: Done.

internal/proxy/add.sh
ijt: I added traps to clean up the temp dir and modvendor but I left the temp dir cleanup for the end so users can manually inspect it in case of failure. I agree that it's easier in a sense if that file gets automatically sent to the terminal where the script is being run, but that also means it can be harder to see the context of what the script was trying to do when it failed.

internal/proxy/README.md
ijt: Done.

internal/proxy/add.sh
ijt: Done.

ijt: It seems to want to remove a lot of things. Maybe we should talk about this.

internal/proxy/add.sh
ijt: Done.

ijt: You may be right about this, but I'm not familiar enough with gsutil rsync and the proxy to be sure it will work with this change. Adding a TODO.

internal/proxy/add.sh
```diff
@@ -0,0 +1,91 @@
...
+function log() {
+  echo "add.sh: $@" 1>&2
+}
+
+function die() {
+  log "$@"
+  exit 1
+}
+
+rsyncing=1
```
jba: You could set `rsyncing` to `true`/`false` and then just write `if $rsyncing; ...`.

But if you prefer this way, that's fine with me.

internal/proxy/add.sh
jba: Probably all the modules named "github.com/google/go-cloud/...".

internal/proxy/add.sh
zombiezen: Please open a "code health" issue to track it.

internal/proxy/add.sh
zombiezen: That seems unfortunate for `makeproxy.sh`. Sent #1031 to fix that.

My main objection here is that the presence of the `.git` directory isn't really sufficient to test the condition that you're checking for. The most robust mechanism I can think of is by using the trick I linked to, which allows you to guarantee the working directory you wish, which means that there isn't an error condition for the script's user.

internal/proxy/add.sh
zombiezen: `"$temp"`.

Unexpected results can occur if the value of the variable contains spaces. I believe [shellcheck](https://www.shellcheck.net/) (available from package managers) will complain about this.

internal/proxy/add.sh
ijt: Done.

internal/proxy/add.sh
ijt: Done. https://github.com/google/go-cloud/issues/1043

internal/proxy/add.sh
ijt: Done.

internal/proxy/add.sh
ijt: Done.

internal/proxy/add.sh
ijt: Ah, makes sense.

PR #1031, opened 2018-12-26
internal/proxy: remove makeproxy.sh assumptions about working directory

zombiezen:
Coming from a thread on #1029.

/cc @ijt


internal/proxy/makeproxy.sh
```diff
@@ -24,11 +24,18 @@ if [[ $# -ne 1 ]]; then
...
+export GOPATH="$1"
+if ! [[ "$GOPATH" =~ ^/ ]]; then
+  GOPATH="$(pwd)/$GOPATH"
+fi
+
+# Script is in internal/proxy, so repository root is two levels up.
+repo_root="$( cd "$( dirname "${BASH_SOURCE[0]}" )/../.." >/dev/null 2>&1 && pwd )"
 
 # Download modules for each of the modules in our repo.
 for path in "." "./internal/contributebot" "./samples/appengine"; do
```
ijt: These should probably be updated to "", "internal/contributebot" and "samples/appengine".

zombiezen: Done, except I kept the first argument as `"."` for clarity.

PR #1035, opened 2018-12-27
.travis.yml: generate gh-pages import redirects

jba:
Provide the necessary Travis config to create our gh-pages site,
with import-path redirects for our packages, on pushes to master.

Requires a personal GitHub access token, which I've already set up
on the Travis settings page.

The only way to test this is to submit the PR and see what happens.


zombiezen:
The deploy job hasn't run yet because the Windows job keeps not coming up. 🙁

PR #1038, opened 2018-12-27
.travis.yml: remove windows build

jba:
Temporary, so we can observe the gh-pages deploy.


jba:
That probably would have been better, but whatevs, it's temporary.

PR #1039, opened 2018-12-27
internal/contributebot: don't log in event process functions

zombiezen:
Keeps test logs cleaner.


zombiezen:
Now waiting on #1041 to fix imports.

PR #1041, opened 2018-12-27
internal/imports: generate top-level index.html and fix generated comment

zombiezen:


zombiezen:
I've run the script manually on my local box and output LGTM.

PR #1047, opened 2018-12-28
travis: run Windows only on master and ignore failures

### LINKED ISSUE #920: travis: don't launch the windows build for pull requests
Currently we launch the windows VM, and then abort it for a subset of build requests. It would be clearer (and more efficient) to not launch the windows build at all for those cases.

I think this can be done via `matrix.include` and conditionals:
https://docs.travis-ci.com/user/multi-os/
https://docs.travis-ci.com/user/conditional-builds-stages-jobs/

zombiezen:
There is a low false positive rate, but a very high false negative rate to Windows builds. This allows us to keep the information, but prevents failures from emitting build errors.

(Not sending this out for review immediately because I want to observe the Travis behavior first.)

Fixes #920

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 os:
   - linux
   - osx
-  - windows
 
 language: go
 go_import_path: gocloud.dev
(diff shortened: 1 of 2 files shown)


zombiezen:
Neat, looks like this worked at least for pull requests:

A screenshot of an in-progress Travis build with only Linux and Mac build jobs

Feel free to merge this without me, but be warned that I have no way of verifying the master configuration. Whoever merges will need to confirm that the master build works as intended: it runs Windows (but may ignore errors).

Issue #975, opened 2018-12-14
pubsub: testing awspubsub in replay mode has a data race


ijt:
The order of calls to ReceiveBatch and SendAcks is nondeterministic with internal/batcher using AddNoWait. The race goes away when the call to `s.ackBatcher.AddNoWait` is replaced with a blocking call to `s.ackBatcher.Add`.

One way to solve this problem is for `NewSubscription` to take its batcher as an argument, and for the test on AWS to pass in a batcher that blocks when adding items. There should be a separate PR to inject the batcher this way, and the PR for `awspubsub` (https://github.com/google/go-cloud/pull/953) should no longer fail on Travis once that is merged in and integrated.

jba:
This should be fixed by a better replayer (see #770). I'll make that a priority for the next sprint.

ijt:
Okay, handing this issue off to you.

vangent:
Per discussion in standup, a better replayer won't resolve this.

ijt:
Being able to pass in an ack batcher means this won't be a problem anymore: a5b87531. Closing it.

PR #1050, opened 2019-01-02
pubsub: factor out DefaultAckBatcher func

ijt:
This is a step towards fixing #975.


internal/pubsub/pubsub.go
```diff
@@ -257,18 +257,34 @@ func (s *Subscription) As(i interface{}) bool {
...
+
+// MakeAckBatchHandler makes a handler func suitable for constructing a Batcher
+// that can be used as an argument to NewSubscription. This handler wraps
+// d.SendAcks with retry logic. MakeAckBatchHandler is for use by provider
+// implementations.
+func MakeAckBatchHandler(ctx context.Context, d driver.Subscription) func(interface{}) error {
+	return func(items interface{}) error {
+		ids := items.([]driver.AckID)
+		// TODO: Consider providing a way to stop this call. See #766.
+		return retry.Call(ctx, gax.Backoff{}, d.IsRetryable, func() error {
```
vangent: Do you want the retry behavior for tests?

internal/pubsub/pubsub.go
```diff
@@ -257,18 +257,34 @@ func (s *Subscription) As(i interface{}) bool {
...
+	handler := MakeAckBatchHandler(callCtx, d)
+	ackBatcher := batcher.New(reflect.TypeOf([]driver.AckID{}).Elem(), maxHandlers, handler)
+	return NewSubscriptionWithBatcher(d, ackBatcher, cancel)
+}
+
+// NewSubscriptionWithBatcher creates a Subscription from a driver.Subscription
+// and a batcher that sends batches of acks to the provider. Typically the
+// batcher is constructed using the result of MakeAckBatchHandler.
+// NewSubscriptionWithBatcher is for use by provider implementations.
+func NewSubscriptionWithBatcher(d driver.Subscription, ab driver.Batcher, cancel func()) *Subscription {
```
vangent: Another way to do this (not sure if it's better, but consider it):

```
func NewSubscription(d driver.Subscription, makeBatcher func(ctx context.Context, handler func(interface{}) error)) {
  callCtx, cancel := context.WithCancel(context.Background())
  handler := /* handler code inline here */
  batcher := makeBatcher(callCtx, handler)
  return &Subscription{...}
}
```

and then have a `DefaultBatcher` function that provider implementations can pass to as `makeBatcher`. Tests would pass something like `makeTestBatcher`.

vangent: This avoids the awkwardness of exposing the `MakeAckBatchHandler` function, but requiring the caller to pass in a `driver.Batcher`, so they have to connect those dots.

internal/pubsub/pubsub.go
```diff
@@ -257,18 +257,34 @@ func (s *Subscription) As(i interface{}) bool {
 // It is for use by provider implementations.
 func NewSubscription(d driver.Subscription) *Subscription {
```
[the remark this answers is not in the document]
ijt: Added `DefaultAckBatcher`.

internal/pubsub/pubsub.go
ijt: It doesn't seem to be hurting anything.

internal/pubsub/pubsub.go

ijt:
PTAL

pubsub/pubsub.go
```diff
@@ -280,18 +280,30 @@ func (s *Subscription) As(i interface{}) bool {
...
+	ctx, cancel := context.WithCancel(context.Background())
 	s := &Subscription{
 		driver: d,
 		cancel: cancel,
 	}
+	s.ackBatcher = newAckBatcher(ctx, s)
+	return s
+}
+
+type BatcherMaker func(context.Context, *Subscription) driver.Batcher
```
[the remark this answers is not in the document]
ijt: `DefaultAckBatcher` also uses `s.mu` and `s.err`, which is why I changed it to take a `*Subscription`. The diff hides this by default; you have to expand something to see it.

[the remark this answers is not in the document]
vangent: Ah, sorry, I missed that.

jba:
A simpler idea would be a variable `driver.AckBatcher`, which would be set by drivertest. Then nothing in the concrete API would look different.

jba:
If you don't like global variables, then the driver interface could have an AckBatcher method.

Another idea: the drivertest package could have the global, and the pubsub package could depend on drivertest. Even I don't like that one, though.

jba:
Just looked at pubsub.go carefully. We definitely don't want two new exported symbols, BatcherMaker and DefaultAckBatcher, just to support tests.

How about a `driver.SetAckBatcher` function. If you look at what we did with `NewPageInfo` in https://godoc.org/google.golang.org/api/iterator, using a variable assigned to an unexported function, we completely hide the function signature.

ijt:
I just had a look and it seems that it's not really feasible to add a `driver.Subscription.AckBatcher` method since the ack batcher handler has to know about the `pubsub.Subscription` to set its `err` field in case of an unrecoverable error. 

I removed the `BatcherMaker` type, so now there's less API exposed just for tests. PTAL.

jba:
Why does DefaultAckBatcher need to be exported? Couldn't non-test code pass nil?

ijt:
I think that's possible. Trying it.

ijt:
Done.

ijt:
Done. PTAL.

PR #1054, opened 2019-01-03
blob/azblob: various small doc and style changes

vangent:
These are items that I noticed during the code review but didn't want to keep nitpicking Vishal on. Let me know if you notice anything else :-).

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -12,35 +12,32 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package azureblob provides a blob implementation that uses Azure Storage’s BlockBlob. Use OpenBucket
-// to construct a blob.Bucket.
+// Package azureblob provides a blob implementation that uses Azure Storage’s
+// BlockBlob. Use OpenBucket to construct a *blob.Bucket.
 //
 // Open URLs
 //
-// For blob.Open URLs, azureblob registers for the protocol "azblob".
+// For blob.Open URLs, azureblob registers for the scheme "azblob"; URLs start
+// with "azblob://".
+//
 // The URL's Host is used as the bucket name.
-//
-// Authentication Options:
-//
-// Option 1: Use the Storage Account Name and Account Key (Primary or Secondary)
-// To use this option, pass query parameter cred_path
-// Example URL: azblob://mybucket?cred_path=pathToCredentials
-// Example credentials file in JSON format:
-// 	{
-// 		"AccountName": "gocloud",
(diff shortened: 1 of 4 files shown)


blob/azureblob/azureblob.go
```diff
@@ -80,24 +77,30 @@ type Options struct {
...
 }
 
-// Azure does not handle backslashes in the blob key well. As a workaround, all backslashes are converted to forwardslashes during bucket operations.
-// This is needed to ensure the directories from Windows file system is represented correctly in Azure Storage.
-// For example:
-// Windows path C:\Users\UserName\Test.json is converted to C:/Users/UserName/Test.json before uploads
-// This retains the original directory structure in Azure Storage as C:/Users/UserName/Test.json, where forwardslash represents the virtual directory
-// For more naming rules and limitations see https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata
+// Azure does not handle backslashes in the blob key well. As a workaround, all
+// backslashes are converted to forwardslashes during bucket operations.
```
jba: nit: "forward slashes" (two words)

blob/azureblob/azureblob.go
```diff
@@ -161,48 +163,45 @@ func openURL(ctx context.Context, u *url.URL) (driver.Bucket, error) {
...
+		return openBucket(ctx, serviceURL, u.Host, opts), err
+	}
 
-	} else {
-		return nil, fmt.Errorf("azureblob: fail, missing querystring cred_path")
+	serviceURL, err := ServiceURLFromSASToken(ac.AccountName, ac.SASToken)
+	if err != nil {
+		return nil, err
 	}
+	return openBucket(ctx, serviceURL, u.Host, opts), err
```
jba: nit: s/err/nil/ (more idiomatic to be explicit)

blob/azureblob/azureblob.go
```diff
@@ -513,40 +496,28 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
...
 	}
-
 	if w.w == nil {
 		pr, pw := io.Pipe()
 		w.w = pw
 		if err := w.open(pr); err != nil {
 			return 0, err
 		}
 	}
-
```
jba: Removing this seems more significant than just a cleanup. Why isn't it needed?

blob/azureblob/azureblob.go
vangent: Done.

blob/azureblob/azureblob.go
vangent: Done.

vangent:
PTAL

PR #1057, opened 2019-01-03
blob/azureblob: add tests for OpenBucket and OpenURL

vangent:


blob/azureblob/azureblob_test.go
```diff
@@ -123,3 +126,154 @@ func TestConformance(t *testing.T) {
...
+			if !test.nilServiceURL {
+				serviceURL, _ = ServiceURLFromAccountKey("x", "y")
+			}
+
+			// Create driver impl.
+			drv, err := openBucket(ctx, serviceURL, test.containerName, nil)
+			if (err != nil) != test.wantErr {
+				t.Errorf("got err %v want error %v", err, test.wantErr)
+			}
+			if drv != nil {
```
jba: if err == nil && ...

blob/azureblob/azureblob_test.go
```diff
@@ -123,3 +126,154 @@ func TestConformance(t *testing.T) {
...
+
+func TestOpenURL(t *testing.T) {
+
+	const (
+		accountName = "my-accoun-name"
+		accountKey  = "aGVsbG8=" // must be base64 encoded string, this is "hello"
+		sasToken    = "my-sas-token"
+	)
+
+	makeCredFile := func(name string, content string) *os.File {
```
jba: name, content string

blob/azureblob/azureblob_test.go
vangent: Done.

blob/azureblob/azureblob_test.go
vangent: Done.

vangent:
PTAL

PR #1064, opened 2019-01-04
travis: don't allow windows builds to fail

### LINKED ISSUE #920: travis: don't launch the windows build for pull requests

vangent:
It seems that by setting this, the windows builds are not even running. They get started, and then are aborted to continue to the next stage. See any build [here](https://travis-ci.com/google/go-cloud/builds).

The Windows build is actually broken right now due to a Contribute use of `unix`, and we didn't notice at all; I'd rather have occasional flakiness.

Fixes #920.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,11 +56,7 @@ jobs:
         verbose: true  # temporarily, while verifying
         on:
           branch: master
-  # Windows is in early access and has been unreliable. Don't block builds on it,
-  # but provide data (for now).
-  allow_failures:
-    - os: windows
-      
+
 stages:
   - test
   - name: imports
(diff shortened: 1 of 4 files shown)


zombiezen:
@vangent I don't think this was the right fix, can we roll this PR back?

vangent:
Yes, sent a rollback.

PR #1065, opened 2019-01-04
travis: don't run wire checks or non-core module tests except on linux

vangent:
There's no real reason to keep ContributeBot running on windows, and there's no reason to do the wire checks per-OS either.

Also configure the OSs to run all in one place. Pretty sure the semantics are the same, but I think this is more clear.

Updates #920.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,3 @@
-os:
-  - linux
-  - osx
-
 language: go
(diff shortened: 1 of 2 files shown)


vangent:
This should get the Windows build to pass tests, if it actually runs them. I'm still not sure why the master/push builds for Windows don't even run the tests.

PR #1069, opened 2019-01-05
blob/tests: add a conformance test for ListObject.As for a directory

vangent:


blob/s3blob/s3blob_test.go
```diff
@@ -140,9 +140,16 @@ func (verifyContentLanguage) ReaderCheck(r *blob.Reader) error {
...
+		var commonPrefix s3.CommonPrefix
+		if !o.As(&commonPrefix) {
+			return errors.New("ListObject.As for directory returned false")
+		}
+		return nil
+	}
 	var obj s3.Object
 	if !o.As(&obj) {
-		return errors.New("ListObject.As returned false")
+		return errors.New("ListObject.As for blob returned false")
```
shantuo: nit: s/blob/object

blob/azureblob/azureblob_test.go
```diff
@@ -195,9 +195,16 @@ func (verifyContentLanguage) ReaderCheck(r *blob.Reader) error {
...
+		var prefix azblob.BlobPrefix
+		if !o.As(&prefix) {
+			return errors.New("ListObject.As for directory returned false")
+		}
+		return nil
+	}
 	var item azblob.BlobItem
 	if !o.As(&item) {
-		return errors.New("ListObject.As returned false")
+		return errors.New("ListObject.As for blob returned false")
```
shantuo: nit: s/blob/object

blob/s3blob/s3blob_test.go
vangent: Done.

blob/azureblob/azureblob_test.go
vangent: Done.

Issue #76, opened 2018-06-06
all: add Azure support


vsekhar:
For post-launch.

zombiezen:
@VPashkov You should check over at Azure/azure-sdk-for-go#2296. Folks may already be working on this.

joshgav:
@VPashkov we (Azure Go team) aren't able to work on this quite yet but would be happy to support you! Please do open issues in our repo or use https://github.com/Azure/azure-sdk-for-go/issues/2296 for questions, and would be great if you could @-mention me @jhendrixMSFT @marstr @vladbarosan in PRs, we'd love to review :) Thank you!

vsaroopchand:
Disclaimer: Although I work for Microsoft, my comments represent my own personal interest in this project.

I can help with the implementation as well. Just out of curiosity, the TerraForm Azure Provider does a great job wrapping the Azure Management SDK for Go, see config.go

https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/config.go

For Azure Storage (Blob as an example) requires distinct steps:

1)	Authenticate with ServicePrincipal
2)	Provision a Resource Group (requires a Resource Group Name or Random Generated)
3)	Provision a Storage Account (requires a Storage Account Name or Random Generated)
4)	Retrieve the Access Keys to the Storage Account
5)	Create a Container (Based on Access Levels)
6)	Create / Upload Blob

Should the implementation automatically provision Resource Name, Storage Accounts or will this be supplied by the caller (exists)?    

@VPashkov – lets connect so we can coordinate our efforts.

zombiezen:
Awesome, sounds good!

@vsaroopchand Take a look at the AWS and GCP implementations as well as [design.md](https://github.com/google/go-cloud/blob/master/internal/docs/design.md). Implementations should assume that provisioning has happened. Feel free to send questions to our [mailing list](https://groups.google.com/forum/#!forum/go-cloud).

vsaroopchand:
I have a question on Reader.go (https://github.com/google/go-cloud/blob/master/blob/blob.go); there is no operation to return the read bytes? How should the retrieved bytes be returned?

vangent:
Hi @vsaroopchand,

Reader.Read reads into the slice of bytes it is passed, p. There's an explanation of io.Reader and examples here:
https://tour.golang.org/methods/21

HTH!

BTW, PR #311 is providing a conformance test for Blob implementations which you should be able to use to test yours.

[...]

PR #1071, opened 2019-01-07
blob/azureblob: update example_test.go

vsaroopchand:
Fixes examples for #76


blob/azureblob/example_test.go
```diff
@@ -42,22 +42,33 @@ func Example() {
 	// Creates a *blob.Bucket backed by Azure Storage Account
 	b, _ := azureblob.OpenBucket(ctx, serviceURL, bucketName, &azureOpts)
 
-	// Obtain the native blob attributes
+	// Get the native blob attributes
```
vangent: Nit: end comments with a period, here and elsewhere.
https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

blob/azureblob/example_test.go
```diff
@@ -42,22 +42,33 @@ func Example() {
...
+		if as(&nativeWriteOptions) {
+			// Use the native SDK type
+		}
+		return nil
+	}
 	opts := &blob.WriterOptions{
 		ContentType: "application/json",
+		BeforeWrite: beforeWrite,
 	}
 	w, _ := b.NewWriter(ctx, key, opts)
```
vangent: Since reading/writing aren't the key points for this example,consider modifying it to use `b.WriteAll` and `b.ReadAll`.

vangent: Or at least `WriteAll`, you need the `Reader`.

blob/azureblob/example_test.go
```diff
@@ -42,22 +42,33 @@ func Example() {
 	// Creates a *blob.Bucket backed by Azure Storage Account
 	b, _ := azureblob.OpenBucket(ctx, serviceURL, bucketName, &azureOpts)
 
-	// Obtain the native blob attributes
```
vangent: Using the `As` functionality should be rare for users (hopefully). Could you split this into two examples, one which focuses on creating the bucket and doing a read (similar to `gcsblob/example_test.go`), and another one with all of the `As` stuff? Note that you can have `// Output` for the former at least to make sure it runs.

blob/azureblob/example_test.go
vsaroopchand: Noted & Fixed

blob/azureblob/example_test.go
vsaroopchand: Done.

blob/azureblob/example_test.go
vsaroopchand: Done.

vsaroopchand:
PTAL

blob/azureblob/example_test.go
```diff
@@ -17,82 +17,160 @@ package azureblob_test
...
+	"log"
 
 	"github.com/Azure/azure-storage-blob-go/azblob"
 	"gocloud.dev/blob"
 	"gocloud.dev/blob/azureblob"
 )
 
-func Example() {
-	ctx := context.Background()
+func openAzureBucket(ctx context.Context) (*blob.Bucket, error) {
```
vangent: Sorry, but can you copy/paste this into each of the examples? It's the most interesting code for a user in the examples, and I don't think godoc will show it at all if it's in a helper.

vsaroopchand: Sure, done.

blob/azureblob/example_test.go
```diff
@@ -25,74 +26,171 @@ import (
...
 	bucketName := "my-bucket"
-	key := "my-file.json"
 
-	// Initializes an azblob.ServiceURL which represents the Azure Storage Account using Shared Key authorization
-	serviceURL, _ := azureblob.ServiceURLFromAccountKey(accountName, accountKey)
+	// Initializes an azblob.ServiceURL which represents the Azure Storage Account using Shared Key authorization.
+	serviceURL, err := azureblob.ServiceURLFromAccountKey(accountName, accountKey)
+	if err != nil {
+		// This is expected since accountName and accountKey is empty.
+		log.Fatal(err)
```
vangent: The test is failing. I think you may need to `fmt.Printf` instead?

blob/azureblob/example_test.go
```diff
@@ -25,74 +26,171 @@ import (
 
 func Example() {
 	ctx := context.Background()
-
 	accountName, accountKey := "", ""
```
vangent: What happens if you provide invalid credentials instead of empty ones? It would be nice (because more of the code would actually run) if you got the error from `ReadAll` instead of from `ServiceURLFromAccountKey`, similar to the GCS example.

vsaroopchand: Using invalid account/keys will result in a fail test due to a timeout exception. The Azure download operation (for ReadAll) will fail after 30 seconds with a HTTP timeout exception. 30 seconds represents the default timeout on a blob operation.

panic: test timed out after 30s
https://docs.microsoft.com/en-us/rest/api/storageservices/setting-timeouts-for-blob-service-operations

vsaroopchand: I can fix this with a custom retry policy on the pipeline to ensure the calls return under 30 seconds. Let me know if this is OK.

opts := azblob.PipelineOptions{
		Retry : azblob.RetryOptions{	
			Policy: azblob.RetryPolicyFixed,
			TryTimeout : 2 * time.Second,		
			MaxTries : 1,
			RetryDelay : 0 * time.Second,
			MaxRetryDelay : 0 * time.Second,
		},
	}
	**p := azblob.NewPipeline(creds, opts)
	*serviceURL = serviceURL.WithPipeline(p)**

blob/azureblob/example_test.go
```diff
@@ -25,74 +29,215 @@ import (
...
+	// Example URL: azblob://my-bucket?cred_path=replace-with-path-to-credentials-file
+	// The schema "azblob" represents the provider for Azure Blob (Azure Storage Account).
+	// The host "my-bucket" represents the blob bucket (Azure Storage Account's Container).
+	// The queryString "cred_path: represents the credentials file in JSON format. See the package documentation for credentials file schema.
+
+	_, err := blob.Open(context.Background(), "azblob://my-bucket?cred_path=replace-with-path-to-credentials-file")
+	if err != nil {
+		// Windows will return error: replace-with-path-to-credentials-file: The system cannot find the file specified.
+		// macOS and Linux will return error: open replace-with-path-to-credentials-file: no such file or directory
+		fmt.Println(err)
```
vangent: Don't print the specific error, just print something like "blob.Open failed", that way it's not dependent on the OS or Go version.

blob/azureblob/example_test.go
```diff
@@ -25,74 +29,215 @@ import (
...
+	}
 
-	// Credential represents the authorizer for SignedURL
-	creds, _ := azblob.NewSharedKeyCredential(accountName, accountKey)
+	// Override the default retry policy so calls can return promptly for this test. Please review the timeout guidelines and set accordingly.
+	// See https://docs.microsoft.com/en-us/rest/api/storageservices/setting-timeouts-for-blob-service-operations for more information.
+	opts := azblob.PipelineOptions{
+		Retry: azblob.RetryOptions{
+			Policy:        azblob.RetryPolicyFixed,
+			TryTimeout:    10 * time.Second,
```
vangent: Does this mean the test takes 10 seconds to run? If so, please make it much shorter (here and below).

blob/azureblob/example_test.go
```diff
@@ -25,74 +29,215 @@ import (
...
+	}
+
+	// Create a *blob.Reader.
+	r, err := b.NewReader(ctx, key, &blob.ReaderOptions{})
+	if err != nil {
+		// This is expected due to the fake credentials we used above.
+		fmt.Println("ReadAll failed due to invalid credentials")
+		return
+	}
+
```
vangent: Can you add a comment here clarifying that nothing below here actually runs in the example?

blob/azureblob/example_test.go
vsaroopchand: TryTimeout is the HttpClient timeout; I will experiment and see what works best.

vsaroopchand: Essentially, TryTimeout needs to be anything but zero. Zero triggers the default wait time which is 30 seconds. In the latest commit I set it to 5; however, the tests run under 1 seconds.

blob/azureblob/example_test.go
vsaroopchand: Ok added.

blob/azureblob/example_test.go
vsaroopchand: Done.

Issue #687, opened 2018-11-13
internal/contributebot: merge in pull requests triggered by comment


zombiezen:
### Is your feature request related to a problem? Please describe.

From [a comment on #630](https://github.com/google/go-cloud/pull/630#issuecomment-436809324):

> [Go Cloud contributors sit in front of their PRs], since the process for submission is:
>
> ```
> while branch is not up-to-date with master {
>   press update branch button
>   wait for Travis run to finish
> }
> click squash-and-merge
> ```
>
> This loop is especially expensive, as the longer you wait, the more likely the branch is not up-to-date and thus causing the cycle to increase iterations.

### Describe the solution you'd like

Contribute Bot could be in charge of merging in changes via a comment like `/merge`, only reporting back if the build failed. The commit message could be derived from the PR title and first comment.

### Describe alternatives you've considered

We could use off-the-shelf solutions, but the amount of effort required to get them working would be roughly equivalent to just getting Contribute Bot to do it.

Solutions we found:

- [Bulldozer](https://github.com/palantir/bulldozer)
- [Refined GitHub](https://github.com/sindresorhus/refined-github), a browser extension, but @vangent has not gotten this to work.
- [mergify.io](https://mergify.io/)
- Kubernetes's [Prow](https://github.com/kubernetes/test-infra/tree/master/prow)
- [Bors-NG](https://github.com/bors-ng/bors-ng)

zombiezen:
@vangent and I brainstormed how to move this feature forward incrementally before the holidays. Basically, this can be broken down into two major technical pieces:

1. Keep requested PRs up-to-date with master &mdash; call this _auto-sync_. This requires watching for push events and label events and performing Git merges where appropriate.
2. Squash and merge PRs into master when all checks pass &mdash; call this _auto-merge_. This requires watching for check run events, verifying that no unapproved commits slip in between the time a maintainer kicks off the submit and the merge event, and then invoking the [GitHub API to merge](https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button) the pull request.

Auto-merge is not very useful without auto-sync, but auto-sync is useful even without auto-merge.

I have a working auto-sync implementation ready to go. To keep the implementation simple for the first pass, the trigger is a GitHub label named "ready to submit" instead of a comment. You can see this in action on zombiezen/go-cloud-bot-test#8, one of my test PRs: I added the "ready to submit" label, I made an upstream commit, the bot merged it into the PR, and then I merged the change in manually.
 
[...]

PR #1073, opened 2019-01-07
internal/contributebot: keep PRs with "ready to submit" in sync with upstream

zombiezen:
This is an incremental (but significant) step toward accomplishing #687. It does not submit PRs, but it manages them so that they are more likely to be in a mergeable state. Since this is taking some shortcuts, I am not updating the design doc with details of this change yet, but this proves (to me at least) the feasibility of an approach where all state is stored in GitHub. Hopefully this eases enough maintainer toil that I can do design work for the remaining elements on a more relaxed schedule.

When a pull request is labeled with "ready to submit", Contribute Bot calls syncPullRequest to update the pull request. Using a label makes validation easy because only users with write access to a repository can make this change.

On any branch push to the repository, Contribute Bot will search for any open pull requests with the "ready to submit" label that have a base branch of the updated branch. Contribute Bot will call syncPullRequest for all the matching pull requests.

If any errors occur while merging, Contribute Bot will remove the "ready to submit" label and leave a comment. This gives notification of required manual intervention.

I've changed the built container to include a copy of Git. This does bump up the container to requiring debian-slim, but I don't see a good way around that.

Updates #687


internal/contributebot/process.go
```diff
@@ -82,6 +96,16 @@ func hasLabel(iss *github.Issue, label string) bool {
 	return false
 }
 
+// prHasLabel returns true iff the issue has the given label.
```
vangent: Nit: issue -> pull request

zombiezen:
Argh, this approach hit a snag. I was assuming that the "Maintainers Can Modify" would include GitHub applications installed on a repository, but a sample PR (https://github.com/zombiezen/go-cloud-bot-test/pull/9) shows that this is not the case. 🙁

This would mean that either Contribute Bot would need to obtain user credentials (act as a user) or we somehow find a set of GitHub APIs that are equivalent to pressing the "Update branch" button.

zombiezen:
Alright, good news: I have a workaround for the auth issue. GitHub Applications can obtain [user-to-server credentials](https://developer.github.com/apps/building-github-apps/identifying-and-authorizing-users-for-github-apps/) so that the server can perform actions on behalf of a user instead of the application itself. I have manually tested that access tokens obtained in this way can be used to authorize pushes to branches in pull request forks that have the "Maintainers Can Modify" bit.

Bad news: things get more complicated. Instead of just keeping credentials for the application, Contribute Bot now needs to store credentials per maintainer and serve a 3-legged OAuth flow. When performing syncs, Contribute Bot needs to query for who requested the sync and authorize using their credentials.

zombiezen:
Closing out this PR until the above changes are made.

Issue #1075, opened 2019-01-07
travis: re-enable windows build


vangent:
It is currently disabled (or allowed to fail) because it's always failing. We think that the failure is related to a Travis bug regarding secrets:
https://travis-ci.community/t/windows-instances-hanging-before-install/250/19

Apparently Travis only fills in secrets for PUSH builds, which is why this problem can't be reproduced on Pull Requests.

vangent:
https://travis-ci.community/t/current-known-issues-please-read-this-before-posting-a-new-topic/264/8

is a better place to check for status on the secrets issue.

vangent:
```
--- FAIL: TestOpenVariableURL (0.03s)
    filevar_test.go:251: file://C:\Users\travis\AppData\Local\Temp\gcdk-filevar-example084267913\mysecret.txt?decoder=decrypt: got error open runtimevar.Variable: parse file://C:\Users\travis\AppData\Local\Temp\gcdk-filevar-example084267913\mysecret.txt?decoder=decrypt: invalid character "\\" in host name, want error false
    filevar_test.go:251: file://C:\Users\travis\AppData\Local\Temp\gcdk-filevar-example084267913\mysecret.txt?decoder=decrypt+bytes: got error open runtimevar.Variable: parse file://C:\Users\travis\AppData\Local\Temp\gcdk-filevar-example084267913\mysecret.txt?decoder=decrypt+bytes: invalid character "\\" in host name, want error false
    filevar_test.go:251: file://C:\Users\travis\AppData\Local\Temp\gcdk-filevar-example084267913\mysecret.txt?decoder=decrypt+jsonmap: got error open runtimevar.Variable: parse file://C:\Users\travis\AppData\Local\Temp\gcdk-filevar-example084267913\mysecret.txt?decoder=decrypt+jsonmap: invalid character "\\" in host name, want error false
    filevar_test.go:251: file://C:\Users\travis\AppData\Local\Temp\gcdk-filevar-example084267913\mysecret.txt?decoder=decrypt%2Bjsonmap: got error open runtimevar.Variable: parse file://C:\Users\travis\AppData\Local\Temp\gcdk-filevar-example084267913\mysecret.txt?decoder=decrypt%2Bjsonmap: invalid character "\\" in host name, want error false
FAIL
FAIL	gocloud.dev/runtimevar/filevar	0.421s
```

```
--- FAIL: TestConformance (1.37s)
    --- FAIL: TestConformance/TestListDelimiters (0.21s)
[...]

PR #1077, opened 2019-01-07
travis: disable windows builds for now instead of allowing them to fail

vangent:
Updates #1075.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,9 +35,10 @@ jobs:
   include:
     - os: linux
     - os: osx
-    # Only run Windows build job on master; it's very slow.
-    # TODO(#1075): Re-enable Windows build once it passes; currently it always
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -35,11 +35,11 @@ jobs:
   include:
     - os: linux
     - os: osx
-    # Only run Windows build job on master. This would be listed in `os:` above,
-    # but matrix expansion does not allow conditionals. See:
-    # https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#conditional-jobs
-    - os: windows
-      if: branch = master AND type = push
+    # Only run Windows build job on master; it's very slow.
```
zombiezen: I recommend removing this line: it's misleading now. Windows does not run ever.

PR #1078, opened 2019-01-07
blob/all: verify ContentMD5 in the concrete type

### LINKED ISSUE #930: blob: enforce MD5 integrity check in the concrete type
Currently, the meaning of the `WriterOptions.ContentMD5` field is underspecified:

https://github.com/google/go-cloud/blob/623057483416703a7029ee5e7827cbab28171d48/blob/blob.go#L575-L577

Between this and the doc comment for `driver.WriterOptions`, Go Cloud does not specify:

- Whether the concrete type or the driver should be performing this check. (Currently it is up to the driver; the concrete type does nothing.)
- What the behavior of a `nil` `ContentMD5` is. (Currently driver-specific, but assumed to mean "don't check".)
- What happens if the bytes sent to the `Writer` don't match the MD5. (Currently driver-specific, and it's unclear what the behavior is.)

## Proposed Change

I would like to perform the integrity check in the concrete type, so that drivers can safely ignore the field without being nonconformant. An error would be returned in `Close`, and the concrete type would cancel the Context passed to the driver's `NewTypedWriter` method so as to abort the write so it has no effect. Drivers can pass along the MD5 header to their underlying network service so as to guarantee the integrity of the bytes in transit.

Of course, this new behavior should all be reflected in the documentation and conformance tests. AFAICT the only new conformance test is to verify that writing a blob that does not match MD5 does not change an existing blob.

/cc @vangent @shantuo

vangent:
*  Verifies `ContentMD5` in the concrete type.
*  `fileblob` and `memblob` still compute the MD5 hash, but only so they can store it to later return it for `Attributes` or `List`; they don't verify it.
*  Clarified comments about how `ContentMD5` is used in the concrete type and driver.
*  Added a conformance test that checks that the write is correctly aborted if `ContentMD5` verification in the concrete type fails.

Fixes #930.


blob/driver/driver.go
```diff
@@ -63,8 +63,12 @@ type WriterOptions struct {
 	// ContentLanguage specifies the language used in the blob's content, if any.
 	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Language
 	ContentLanguage string
-	// ContentMD5 may be used as a message integrity check (MIC).
-	// https://tools.ietf.org/html/rfc1864
+	// ContentMD5 is used as a message integrity check.
+	// The concrete type checks that the MD5 hash of the bytes written matches
+	// ContentMD5.
+	// If len(ContentMD5) > 0, driver implementations may pass it to their
+	// underlying/ network service to guarantee the integrity of the bytes in
```
zombiezen:
```suggestion
	// underlying network service to guarantee the integrity of the bytes in
```

PR #1080, opened 2019-01-07
pubsub: add conformance test for two subscriptions

### LINKED ISSUE #787: pubsub: add test for multiple subscribers to a single topic in drivertest

jba:
Add a test that verifies that both subscriptions get all published messages.

Fixes #787.


pubsub/drivertest/drivertest.go
```diff
@@ -185,35 +188,91 @@ func testSendReceive(t *testing.T, newHarness HarnessMaker) {
...
+		return pubsub.NewSubscription(ds)
+	}
+	s0 := makeSub(0)
+	s1 := makeSub(1)
+	defer s0.Shutdown(ctx)
+	defer s1.Shutdown(ctx)
+
+	want := publishN(t, ctx, top, 3)
+	gotc := make(chan []*pubsub.Message, 2)
+	go func() { gotc <- receiveN(t, ctx, s0, len(want)) }()
```
ijt: I don't think it's necessary to have these two goroutines here. It's probably clearer to write it like this:
```go
ss := []*pubsub.Subscription{makeSub(0), makeSub(1)}
...
for _, s := range ss {
    got := receiveN(t, ctx, s, len(want))
    if diff := diffMessageSets(got, want); diff != "" {
        t.Error(diff)
    }
}
```
though maybe a little slower while recording.

jba: Done.

Issue #676, opened 2018-11-12
runtimevar/tests: get code coverage up to 90% for concrete type, 85% for providers


vangent:
[Coveralls](https://coveralls.io/github/google/go-cloud)
shows code coverage for `runtimevar` at around 73%.

76.89 runtimevar/
72.73 constantvar/
85.19 etcdvar/
85.32 filevar/
67.96 paramstore/
63.83 runtimeconfigurator/
88.89 runtimevar.go

Let's see if we can get them all up to 90%.

vangent:
The coverage is pessimistic, see https://github.com/google/go-cloud/issues/741.

vangent:
New snapshot after fixing #741:

76.71 runtimevar/
72.73 constantvar/
100.0 driver/
80.25 etcdvar/
82.57 filevar/
67.96 paramstore/
63.83 runtimeconfigurator/
98.15 runtimevar.go

vangent:
Latest snapshot:

100.0 constantvar/
100.0 driver/
80.49 etcdvar/
81.19 filevar/
73.08 paramstore/
69.47 runtimeconfigurator/
98.15 runtimevar.go

vangent:
Done.

PR #1083, opened 2019-01-08
runtimevar/etcdvar: add a Timeout option so that initial Watch doesn't block forever

vangent:
I ran into this when trying to get the example to actually run; it blocked forever.

Updates #676.


runtimevar/etcdvar/example_test.go
```diff
@@ -51,8 +55,14 @@ func ExampleNew() {
 	// You can now read the current value of the variable from v.
 	snapshot, err := v.Watch(context.Background())
 	if err != nil {
-		log.Fatal(err)
+		// This is expected because we can't connect to the fake server URL.
```
shantuo: There is an option to omit the error as well as the output line altogether if we want the example simple to understand and focusing on the main function flow.

vangent: We could, but I prefer having the example run because it ensures it is more realistic. Anyway, this pattern is followed across almost all of our examples now, and I'm updating this one to be consistent; if you feel strongly that they shouldn't run, please file a separate issue and we can discuss.

PR #1084, opened 2019-01-08
pubsub: add awspubsub

ijt:
Fixes #953


pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,182 @@
...
+	return &topic{client: client, arn: topicARN}
+}
+
+// SendBatch implements driver.Topic.SendBatch.
+func (t *topic) SendBatch(ctx context.Context, dms []*driver.Message) error {
+	for _, dm := range dms {
+		attrs := map[string]*sns.MessageAttributeValue{}
+		for k, v := range dm.Metadata {
+			attrs[k] = &sns.MessageAttributeValue{
+				DataType:    aws.String("String"),
```
jba: Put `aws.String("String")` in a global to avoid allocating on each call.

pubsub/awspubsub/awspubsub.go
```diff
@@ -0,0 +1,182 @@
...
+
+// ReceiveBatch implements driver.Subscription.ReceiveBatch.
+func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*driver.Message, error) {
+	output, err := s.receiveMessages()
+	if err != nil {
+		return nil, err
+	}
+	var ms []*driver.Message
+	for _, m := range output.Messages {
+		type Attribute struct {
```
jba: Since you never use this type, you can omit it and define the struct inline on line 124.

pubsub/awspubsub/awspubsub.go
ijt: Done.

pubsub/awspubsub/awspubsub.go
ijt: Done.

ijt:
I'm blocked on trying to programmatically subscribe SQS queues to SNS topics in the conformance tests. See https://stackoverflow.com/questions/54204855/how-does-one-programmatically-subscribe-an-sqs-queue-to-an-sns-topic-in-go.

jba:
Text at the end is hilarious.

ijt:
No one has answered the stack overflow question. I filed an issue against the AWS Go SDK: https://github.com/aws/aws-sdk-go/issues/2415.

ijt:
Issue https://github.com/aws/aws-sdk-go/issues/2415 doesn't seem to have been solved yet.

jba:
Is there a way to do the setup you need via a UI? Maybe do that to unblock?

ijt:
Sure. I could do that and just leave the `Create*` functions unimplemented with a TODO. If you're okay with that then I'll go ahead.

ijt:
I added something to allow provider implementations to skip conformance tests based on their name and provide a reason. This allowed skipping TestSendReceive, TestSendReceiveTwo and TestAs in replay mode on awspubsub since these were sometimes panicking and sometimes hanging in replay mode, apparently due to some issue with the recorder.

I filed https://github.com/google/go-cloud/issues/1233 about the recorder problem.

Tests are passing now. PTAL.

jba:
You have approval.

PR #1092, opened 2019-01-09
blob/azureblob: set Azure User-Agent with go-cloud prefix

vsaroopchand:
For #76 ; this change adds **go-cloud/<version>** as prefix to the Blob SDK's User-Agent 
Example: go-cloud/0.9.0 Azure-Storage/0.3 (go1.10.3; darwin)


internal/useragent/useragent.go
```diff
@@ -39,6 +39,11 @@ func GRPCDialOption(api string) grpc.DialOption {
 	return grpc.WithUserAgent(userAgentString(api))
 }
 
+// AzureUserAgentPrefix returns a prefix that is used to set Azure SDK User-Agent to help with diagnostics.
+func AzureUserAgentPrefix() string {
+	return fmt.Sprintf("%s/%s", prefix, version)
```
vangent: For GCP we included the API name (see below) so you can distinguish between uses other than `blob` once they exist, WDYT?

blob/azureblob/azureblob.go
```diff
@@ -116,7 +118,11 @@ func ServiceURLFromAccountKey(accountName, accountKey string) (*azblob.ServiceUR
 		return nil, errors.New("azureblob: accountKey is required")
 	}
 	credential, _ := azblob.NewSharedKeyCredential(accountName, accountKey)
-	pipeline := azblob.NewPipeline(credential, azblob.PipelineOptions{})
+	pipeline := azblob.NewPipeline(credential, azblob.PipelineOptions{
+		Telemetry: azblob.TelemetryOptions{
+			Value: useragent.AzureUserAgentPrefix(),
```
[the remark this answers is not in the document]
vsaroopchand: Correct, [here ](https://github.com/Azure/azure-storage-blob-go/blob/5152f14ace1c6db66bd9cb57840703a8358fa7bc/azblob/zc_policy_telemetry.go#L24)is the factory that constructs the full User-Agent

internal/useragent/useragent.go
vsaroopchand: Let me confirm with the PG; it doesn't hurt to include it. If so, can I make userAgentString public to avoid adding another helper function?

vangent: Sure.

vsaroopchand: Actually, it doesn't make sense exporting func userAgentString; I just amended my original helper.

vsaroopchand:
PTAL

PR #1093, opened 2019-01-09
runtimevar: add a runtimevar implementation based on blob

### LINKED ISSUE #1085: runtimevar: add an implementation based on blob.Bucket
Given a `*blob.Bucket` and a key and a decoder, a `runtimevar.Variable` that polls the bucket.

vangent:
Fixes #1085.


runtimevar/blobvar/blobvar.go
```diff
@@ -0,0 +1,152 @@
+// Copyright 2018 The Go Cloud Authors
```
shantuo: s/2018/2019

vangent:
> What's the reason of having a blobvar? Azure doesn't have it?

Correct; this is what they recommend (at least in one place: https://docs.microsoft.com/en-us/azure/architecture/patterns/external-configuration-store). And having configuration in a blob doesn't seem like a crazy way to do it even on other clouds.

runtimevar/blobvar/blobvar.go
vangent: Done.

PR #1095, opened 2019-01-10
blob/s3blob: add query options to url parsing

elliots:
Adds endpoint, disableSSL, and s3ForcePathStyle query options.

There are more options in aws.Config, but these three are the ones I needed in order to use the package with a local minio instance.

At least endpoint would be required by other s3-compatible services.

### DIFF
--- a/blob/s3blob/s3blob.go
+++ b/blob/s3blob/s3blob.go
@@ -25,6 +25,9 @@
 // https://docs.aws.amazon.com/sdk-for-go/api/aws/session/.
 // The following query options are supported:
 //  - region: The AWS region for requests; sets aws.Config.Region.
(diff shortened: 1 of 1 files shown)


blob/s3blob/s3blob.go
```diff
@@ -25,6 +25,9 @@
 // https://docs.aws.amazon.com/sdk-for-go/api/aws/session/.
 // The following query options are supported:
 //  - region: The AWS region for requests; sets aws.Config.Region.
+//  - endpoint: The endpoint URL (hostname only or fully qualified URI); sets aws.Config.Endpoint.
+//  - disableSSL: Disable SSL when sending requests; sets aws.Config.DisableSSL.
```
vangent: Can you clarify that only the value of "true" has this effect?

E.g., something like 

```
//  - disableSSL: A value of "true" disables SSL when sending requests; sets ....

```

PR #1102, opened 2019-01-10
blob/azureblob: verify User-Agent matches, and re-record with new User-Agent

vangent:
The new User-Agent field was added in #1092.


blob/azureblob/testdata/TestConformance/TestAttributes.yaml
```diff
@@ -8,7 +8,7 @@ interactions:
       Content-Length:
       - "12"
       User-Agent:
-      - X-Az-Target Azure-Storage/0.3 (go1.11; linux)
+      - go-cloud/blob/0.9.0 Azure-Storage/0.3 (go1.11.2; linux)
```
vangent: Here's an example of the new User-Agent field.

internal/testing/setup/setup.go
```diff
@@ -155,13 +155,14 @@ func NewGCPgRPCConn(ctx context.Context, t *testing.T, endPoint, api string) (*g
...
 // NewAzureTestPipeline creates a new connection for testing against Azure Blob.
-func NewAzureTestPipeline(ctx context.Context, t *testing.T, accountName string, accountKey string) (pipeline pipeline.Pipeline, done func(), httpClient *http.Client) {
+func NewAzureTestPipeline(ctx context.Context, t *testing.T, api, accountName, accountKey string) (pipeline pipeline.Pipeline, done func(), httpClient *http.Client) {
 	mode := recorder.ModeReplaying
 	if *Record {
 		mode = recorder.ModeRecording
 	}
 
 	azMatchers := &replay.ProviderMatcher{
+		Headers: []string{"User-Agent"},
```
vangent: This ensures that the User-Agent header matches during replay.

internal/testing/setup/setup.go
```diff
@@ -181,23 +182,21 @@ func NewAzureTestPipeline(ctx context.Context, t *testing.T, accountName string,
...
 		panic("pipeline credential can't be nil")
 	}
 
 	f := []pipeline.Factory{
 		// sets User-Agent for recorder
 		azblob.NewTelemetryPolicyFactory(azblob.TelemetryOptions{
-			Value: "X-Az-Target",
+			Value: useragent.AzureUserAgentPrefix(api),
 		}),
-		// sets header X-Ms-Client-Request-Id, see https://msdn.microsoft.com/en-us/library/mt766820.aspx
```
vangent: @vsaroopchand I wasn't sure why this header was needed, and it seems to work fine without it, so I removed it. If there's a reason for it, please let me know and I'll put it back with a comment explaining why.

internal/testing/setup/setup.go
```diff
@@ -181,23 +182,21 @@ func NewAzureTestPipeline(ctx context.Context, t *testing.T, accountName string,
...
-func newPipeline(c azblob.Credential, r *recorder.Recorder) pipeline.Pipeline {
+func newPipeline(c azblob.Credential, api string, r *recorder.Recorder) pipeline.Pipeline {
 	if c == nil {
 		panic("pipeline credential can't be nil")
 	}
 
 	f := []pipeline.Factory{
 		// sets User-Agent for recorder
 		azblob.NewTelemetryPolicyFactory(azblob.TelemetryOptions{
-			Value: "X-Az-Target",
```
vangent: I'm not sure what the "X-Az-Target" here meant; it was getting prepended to the User-Agent header.

vsaroopchand: Yeah, unintended. And to your other point, its is not needed for the tests and can be removed all together.

internal/testing/setup/setup.go
vsaroopchand: Its safe to remove otherwise any change to Version will require re-recording these tests. If you keep it, use a static value such as go-cloud-test, etc.

internal/testing/setup/setup.go
vangent: Ah, I have to comment this out because Azure adds something with the "go" version in it to User-Agent, which varies.

PR #1111, opened 2019-01-11
internal/secrets: add a concrete type for secrets

### LINKED ISSUE #1066: internal/secrets: add a portable API

shantuo:
Fixes #1066


internal/secrets/gcpkms/kms_test.go
```diff
@@ -43,7 +43,7 @@ type harness struct {
 }
 
 func (h *harness) MakeDriver(ctx context.Context) (driver.Crypter, error) {
```
vangent: The driver tests should be updated to use the concrete type instead of the driver directly; right now your test coverage of the concrete type is 0%.

internal/secrets/driver/driver.go
```diff
@@ -29,3 +29,7 @@ type Crypter interface {
 	// Encrypt encrypts the plaintext and returns the cipher message.
 	Encrypt(ctx context.Context, plaintext []byte) ([]byte, error)
 }
+
+// CrypterOptions controls Crypter behaviors.
```
vangent: Don't think we need this; it's not used...? We want every public constructor to take `Options`, but it should probably be a different type per provider impl.

internal/secrets/gcpkms/kms.go
```diff
@@ -37,11 +38,11 @@ func Dial(ctx context.Context, ts gcp.TokenSource) (*cloudkms.KeyManagementClien
 }
 
 // NewCrypter returns a new Crypter to to encryption and decryption.
-func NewCrypter(client *cloudkms.KeyManagementClient, ki *KeyID) *Crypter {
-	return &Crypter{
+func NewCrypter(client *cloudkms.KeyManagementClient, ki *KeyID) *secrets.Crypter {
```
vangent: So add an Options struct here.

internal/secrets/secrets.go
```diff
@@ -0,0 +1,75 @@
...
+	b, err := c.Decrypt(ctx, ciphertext)
+	if err != nil {
+		return nil, wrapError(c, err)
+	}
+	return b, nil
+}
+
+// CrypterOptions controls Crypter behaviors.
+// It is provided for future extensibility.
+type CrypterOptions struct{}
```
vangent: Don't need this one either; `NewCrypter` is technically public but should only be used by provider implementations.

internal/secrets/secrets.go
```diff
@@ -0,0 +1,75 @@
...
+
+// wrappedError is used to wrap all errors returned by drivers so that users are
+// not given access to provider-specific errors.
+type wrappedError struct {
```
vangent: (OK in separate PR...)
Add a test using a stub driver implementation to verify that errors are wrapped (and not double-wrapped). See runtimevar or blob for examples.

internal/secrets/gcpkms/kms.go
```diff
@@ -22,6 +22,7 @@ import (
 
 	cloudkms "cloud.google.com/go/kms/apiv1"
 	"gocloud.dev/gcp"
+	"gocloud.dev/internal/secrets"
```
vangent: `localsecrets` should be updated in the same way.

zombiezen:
I have nothing to add to this review beyond what @vangent has already said. Removing myself from review line.

internal/secrets/driver/driver.go
clausti: @shantuo can you provide some context (a summary of in-person conversation?) for this being resolved?

shantuo:
PTAL

PR #1117, opened 2019-01-15
internal/secret: add an AWS KMS implementation

shantuo:
Update #420

### DIFF
--- a/internal/secrets/awskms/example_test.go
+++ b/internal/secrets/awskms/example_test.go
@@ -1,0 +1,86 @@
+// Copyright 2019 The Go Cloud Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


internal/secrets/awskms/example_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"gocloud.dev/internal/secrets/awskms"
+)
+
+func Example_encrypt() {
+	sess, err := session.NewSession(&aws.Config{
+		Region: aws.String("us-west-1"),
+	})
+	if err != nil {
+		panic(err)
```
vangent: Nit: Use log.Fatal instead of panic.

https://github.com/google/go-cloud/blob/master/internal/docs/design.md#coding-conventions

PR #1123, opened 2019-01-15
gcppubsub: cap the number of messages in a Publish RPC

jba:
It turns out there is a limit on the number of messages you can publish
in a single RPC. (I reached it during benchmarking.)

### DIFF
--- a/pubsub/gcppubsub/gcppubsub.go
+++ b/pubsub/gcppubsub/gcppubsub.go
@@ -88,19 +88,31 @@ func openTopic(ctx context.Context, client *raw.PublisherClient, proj gcp.Projec
 
 // SendBatch implements driver.Topic.SendBatch.
 func (t *topic) SendBatch(ctx context.Context, dms []*driver.Message) error {
-	var ms []*pb.PubsubMessage
-	for _, dm := range dms {
-		ms = append(ms, &pb.PubsubMessage{
-			Data:       dm.Body,
-			Attributes: dm.Metadata,
-		})
+	// The PubSub service limits the number of messages in a single Publish RPC.
+	const maxPublishCount = 1000
+	for len(dms) > 0 {
+		n := len(dms)
+		if n > maxPublishCount {
+			n = maxPublishCount
+		}
+		batch := dms[:n]
+		dms = dms[n:]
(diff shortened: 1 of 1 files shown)


pubsub/gcppubsub/gcppubsub.go
```diff
@@ -88,19 +88,31 @@ func openTopic(ctx context.Context, client *raw.PublisherClient, proj gcp.Projec
...
-		ms = append(ms, &pb.PubsubMessage{
-			Data:       dm.Body,
-			Attributes: dm.Metadata,
-		})
-	}
-	req := &pb.PublishRequest{
-		Topic:    t.path,
-		Messages: ms,
+	// The PubSub service limits the number of messages in a single Publish RPC.
+	const maxPublishCount = 1000
```
vangent: Is this something that could be done in the batcher instead? (in case other providers have similar restrictions)?

jba: Yes, we could add that as an option, and I think it would be easy to implement. Would you prefer I do that?

vangent: What do you think? My intuition is that it would be easier to implement there, and also more powerful since it could be used by other providers.

vangent: @ijt aren't you changing things so that the batcher is provided by the driver implementation? I think that would make the above easier, it would be something like:

```
package gcppubsub

...

func NewTopic(....) {
  drv := newTopic(....)
  return pubsub.NewTopic(drv, <batcher>)
}
````

I can't find the PR where this was happening though...?

jba: At head, you can supply an ack batcher to NewSubscription. But I don't think there's anything on the publish side.

vangent: Ah, sorry, that's what I was thinking of. OK, you're right then. We can revisit if necessary.

ijt: @vangent, we could also supply the batcher as an arg to NewTopic. We just didn't have a specific reason to do it until now.

Issue #1124, opened 2019-01-16
all: handle errors in a common way


jba:
We should try to unify treatment of errors across APIs and providers.

I propose that we have a single `Error` type in its own `errors` package. It would consist of an error code, a message string, and possibly a wrapped message (typically, the error that the driver returned). The `Error` type will have an `Unwrap` method that returns the wrapped message, to conform to the Go 2 draft proposal (implemented at https://godoc.org/golang.org/x/exp/errors). See #1125 for a proposed implementation.

The concrete type will be responsible for wrapping driver errors in the new type. A driver can assist by categorizing errors by code. Each driver interface will add the method
```
ErrorCode(error) errors.Code
```
A driver can be sure that the error passed to its `ErrorCode` method came from its own implementation. It should determine the right code for the error. For instance, a gRPC-based driver might do something like
```
func (d *driver) ErrorCode(err error) errors.Code {
    switch status.Code(err) {
    case codes.NotFound : return errors.NotFound
    case codes.PermissionDenied: return errors.PermissionDenied
    ...
    default: return errors.Unknown
}
```

The concrete type would wrap driver errors like so:
```
if err := driver.SomeMethod(); err != nil {
    return errors.Newf(driver.ErrorCode(err), err, "calling SomeMethod: %v", err)
}
```
where `Newf` takes a code, the wrapped error, and format data and returns an `*errors.Error`.

PR #1125, opened 2019-01-16
errors: a unified error type

jba:
Provide Go Cloud with an Error type that can be used for all APIs and providers.

Addresses #1124.


errors/errors.go
```diff
@@ -0,0 +1,88 @@
...
+
+import "fmt"
+
+// A Code describes the error's category. Programs should act upon an error's
+// code, not its message.
+type Code int
+
+const (
+	// The error could not be categorized.
+	Unknown Code = iota
```
vangent: Consider binding them to explicit constants to make sure they don't change?

errors/errors.go
```diff
@@ -0,0 +1,88 @@
...
+	// The resource was not found.
+	NotFound
+
+	// The resource exists, but it should not.
+	AlreadyExists
+
+	// The caller is not authorized to perform the operation.
+	PermissionDenied
+
+	// A value given to a Go Cloud API is incorrect.
```
vangent: Are we committing to actually converting underlying provider errors to each of these? So far I believe we just have `NotFound`, adding all of these others might not be needed (yet).

jba: In my docstore prototype I needed AlreadyExists. InvalidArgument is important so callers can distinguish their own errors. E.g.  "blob.NewWriter: WriterOptions.Metadata keys may not be empty strings" should be InvalidArgument. Internal (better name?) is important because it indicates where _we_ messed up—users know to file a bug if they see it. The rest we don't need for now.

errors/errors.go
```diff
@@ -0,0 +1,88 @@
...
+
+// New returns a new error with the given code, underlying error and message.
+func New(c Code, err error, msg string) *Error {
+	return &Error{
+		Code: c,
+		msg:  msg,
+		err:  err,
+	}
+}
+
```
vangent: Can you add the changes from your other PR? E.g., gerrors.ErrorCode()?

errors/errors.go
```diff
@@ -0,0 +1,88 @@
...
+	return fmt.Sprintf("%s: %s", cs, e.msg)
+}
+
+// TODO(jba): use the new errors formatting code to print the underlying error.
+
+// Unwrap returns the error underlying the receiver.
+func (e *Error) Unwrap() error {
+	return e.err
+}
+
```
vangent: IIUC, the proposed replacement for `blob.IsNotExist(err)` is `gerrors.ErrorCode(err) == gerrors.NotFound`, right?

That seems less discoverable to me, mostly because it's not obvious from the return type (error) that you should be looking at this other error package (gerrors).

Maybe we can just keep the existing `blob.IsNotFound` etc.? And/or add `blob.ErrorCode(err)` that returns the gerrors.ErrorCode?

Similarly, what's the proposed replacement for

```
var aerr aws.Error
if blob.ErrorAs(&aerr) {
  // use aerr.
}
```
?

I think something like:

```
rawerr := gerrors.Unwrap(err)
if rawerr != nil {
  if aerr, ok := rawerr.(aws.Error); ok {
    // Use aerr.
  }
}
```

Again, that seems rather unwieldy in comparison, and also doesn't match the `As` style used elsewhere. Maybe we should keep `blob.ErrorAs` and `driver.ErrorAs`...?

vangent:
@zombiezen FYI #1127 has this commit plus an extra one showing how it would be used in `blob`, useful for context.

errors/errors.go
jba: We would document the error codes returned in all functions where it mattered, and we'd have examples. People using our APIs would figure out pretty quickly that we have our own errors package with its own type and codes. Especially since that's how both GCP and AWS do it: gRPC has a `Status`, Amazon has `aws.Error`.

I'm fine with leaving a legacy boolean function in blob, but let's rename it IsNotFound for consistency. I wouldn't add new ones to other packages, though, since they're redundant.

The replacement for `ErrorAs` is `errors.As`, the general `As` function that is currently in `golang.org/x/exp/errors` but will be in the stdlib `errors` in, we think,Go 1.13.

errors/errors.go
jba: Done.

jba:
Since this seems to be moving forward, I'd like to suggest the following name changes:

Let's not call the package `errors`, because it conflicts with the stdlib one, forcing importers to use an alias. Suggestions: `gcerr` (like `awserr`; my preference), `gcerrors`, gerrors`. 

Call the type `ErrorCode` and the function `Code` (i.e. swap the names of those two entities), since users will invoke the function much more often than writing the type.

errors/errors.go
vangent: OK, that sounds fine. We should document that driver implementations should return `Unknown` as a default.

errors/errors.go
vangent: * I think if we document a few error codes in specific functions, and add something to the package documentation pointing to our errors package, that should be OK.

* I'm fine with just removing the legacy boolean function I guess; we would keep it for backwards compatibility, so there's not much point in renaming it.

* Unless you have a strong objection let's keep `ErrorAs` in our concrete types (and in our conformance tests). For the driver interface, I'm less sure it's necessary to keep `ErrorAs`, but maybe we should for flexibility (e.g., one of the reasons for `As` was to allow providers to support multiple target types). The implementation for each driver can use `golang.org/x/exp/errors` or stay the same as it is now.

jba: > let's keep `ErrorAs` in our concrete types

I don't have a strong objection now, but when `errors.As` is in the standard library, it will be the idiomatic way to do this, and we should be in the vanguard in adopting it. So as long as we remove `ErrorAs` before we go GA, I'm fine leaving it for now.

errors/errors.go
jba: Done.

jba:
I made the Code/ErrorCode name change, but haven't renamed the package yet.
PTAL

errors/errors.go
```diff
@@ -0,0 +1,91 @@
...
+	// Something unexpected happened. Internal errors always indicate
+	// bugs in Go Cloud (or possibly the underlying provider).
+	Internal = 5
+)
+
+// To get stringer: go get golang.org/x/tools/cmd/stringer
+
+//go:generate stringer -type=ErrorCode
+
+// An Error described a Go Cloud error.
```
vangent: described -> describes

errors/errors.go
```diff
@@ -0,0 +1,91 @@
...
+		return e.Code
+	}
+	return Unknown
+}
+
+// New returns a new error with the given code, underlying error and message.
+func New(c ErrorCode, err error, msg string) *Error {
+	return &Error{
+		Code: c,
+		msg:  msg,
```
vangent: Do we need the additional msg? Based on the usage you showed in the other PR, it looks like an API would be sufficient? Otherwise people are mostly going to copy err.Error() as the message.

Something like:

```
func New(c ErrorCode, api string, err error) *Error {...}


// no Newf

func (e *Error) Error() string {
  return fmt.Sprintf("%s: %s - %v", e.api, e.Code, e.err)
}
```

jba: I think we'll want to replace a lot of fmt.Errorfs in the concrete implementation with Newfs. E.g. from blob.go:

current:
```
 return nil, fmt.Errorf("blob.NewWriter: duplicate case-insensitive metadata key %q", lowerK)
```
proposed:
```
  return nil, gcerr.Newf(gcerr.InvalidArgument, nil, blob.NewWriter: duplicate case-insensitive metadata key %q", lowerK)
```

We could also consider adding more info to driver errors:

current:
```
    if w.w, err = w.b.NewTypedWriter(w.ctx, w.key, ct, w.opts); err != nil {
        return 0, wrapError(w.b, err)
    }
```
proposed:
```
    if w.w, err = w.b.NewTypedWriter(w.ctx, w.key, ct, w.opts); err != nil {
        return 0, gcerr.Newf(w.b.ErrorCode(err), err, "writing key %q: %v", w.key, err)
    }
```
In the last case we ultimately won't need to have the "%v" with `err`, because something will format the underlying error for us. I'm not quite sure how that part is going to work at the moment, though.

vangent: Your last paragraph is what I'm wondering about. The current code you have doesn't print the underlying error's message at all, so the caller will always want to provide a message that includes the underlying error's `Error()`, or it won't have a message at all. E.g., `gcerr.New(w.b.ErrorCode(err), err, "")` will not work. Maybe `*Error` could check is `msg` is empty, and if so, default to `err.Error`?

I also do like having the `blob: ` (or whatever) prefix for all of our errors; it's not obvious how to get that consistently if every error is formatted arbitrarily.

Maybe a separate "prefix" argument, and if "msg" is empty default to using `err.Error()`?

errors/errors.go
jba: Done.

errors/errors.go
jba: Okay, I finally figured out how formatting is going to work with the new error design, and made the necessary changes. 

In the new design, `%v` prints a one-line message, and `%+v` prints details over several lines. The details will include a line like `gocloud.dev/blob.NewWriter`, so there won't be any need for a `blob:` prefix.

So say you wrap a driver error with message "foo" with a code and  message: 
`New(NotFound, err, "something")`
Output with %v:
```
NotFound: something: foo
```
Output with %+v:
```
NotFound: something
  gocloud.dev/blob.NewWriter
  path/to/file/blob.go:197
  - foo
```

If you pass `""` for the message, the output is the same except `something` won't appear.

jba:
The most recent commit renames the package to 'gcerrors'.

PTAL.

gcerrors/errors.go
```diff
@@ -0,0 +1,115 @@
...
+	xfmt.FormatError(s, c, e)
+}
+
+func (e *Error) FormatError(p xerrors.Printer) (next error) {
+	p.Print(e.Error())
+	e.frame.Format(p)
+	return e.err
+}
+
+// Unwrap returns the error underlying the receiver.
```
vangent: , which may be `nil`.

vangent:
Tests are failing, you need to update the proxy with the new dependency.

gcerrors/errors_test.go
```diff
@@ -0,0 +1,121 @@
...
+			"%+v",
+			[]string{
+				`^AlreadyExists: message:$`,
+				`^\s+gocloud.dev/gcerrors.TestFormatting$`,
+				`^\s+.*/go-cloud/gcerrors/errors_test.go:\d+$`,
+				`^\s+- wrapped$`,
+			},
+		},
+		{
+			New(AlreadyExists, errors.New("wrapped"), ""),
```
vangent: How does this work? The `Error` function looks like it would just return `%v` of the `Code` in this case.

gcerrors/errors.go
jba: Done.

gcerrors/errors_test.go
jba: The `fmt` package calls the `Format` method instead of `Error`. That ends up calling `FormatError`. (Once the new error stuff has landed, the `Format` method won't be necessary: `fmt` will call `FormatError` directly.)

errors/errors.go
jba: No, `%+v`.

Would you like to include a short API name like `blob` even for `%v`?

zombiezen:
High-level question: can we push this down to an internal package, or is there end-user benefit to having new API be exported? I'm all for the `exp/errors` way of handling things, but if there's the possibility of it changing in incompatible ways, it seems like keeping it internal and then gradually exposing more of the functionality in a controlled way seems like a better approach.

jba:
We can push the construction methods to internal, but users will need the `Error` and `ErrorCode` types and the `Code` function.

Since a lot of the current discussion is around the constructors, I'll do exactly that if there are no objections.

zombiezen:
I don't follow why they need to be exported: couldn't our concrete type packages house the same info and people code against that, but then have the implementations of these things use an `internal/gcerrors` package?

jba:
We want a common set of codes, and a single function for getting the code from an error. We don't need to export the error type.

See the latest commit.

jba:
@zombiezen, just to be clear: it sounded like you were suggesting we have a `blob.NotFound`, `pubsub.NotFound`, `runtimevar.NotFound`, etc. I am saying that we should have a single package that has a `NotFound` constant of type `ErrorCode`. That same package will have a function `Code(error) ErrorCode` that extracts the code from an error.

zombiezen:
I am suggesting that there be exported symbols in each of `pubsub`, `runtimevar`, etc. The concern I have is about discoverability: how do folks know to look at this other package?

I would be fine with the symbols being aliases to the unified package or some other alleviation of the discovery aspect.

jba:
> I would be fine with the symbols being aliases to the unified package or some other alleviation of the discovery aspect.

Documentation:
```
// If the bucket doesn't exist, Attributes returns an error for which gcerrors.Code returns NotFound.
```
And also, in the package doc for each API:
```
Errors

All the errors returned from this package have an associated numeric code, listed in the gocloud.dev/gcerrors package. Use that package's Code function to retrieve the code.
```

gcerrors/errors.go
```diff
@@ -0,0 +1,60 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package gcerrors provides support for getting error codes from
+// errors returned by Go Cloud APIs.
+//
+// This package is EXPERIMENTAL and subject to change without notice.
```
vangent: Remove?

internal/gcerr/gcerr.go
```diff
@@ -0,0 +1,103 @@
...
+type Error struct {
+	Code  ErrorCode
+	msg   string
+	frame xerrors.Frame
+	err   error
+}
+
+func (e *Error) Error() string {
+	if e.msg == "" {
+		return fmt.Sprintf("%v", e.Code)
```
vangent: WDYT about including `err.Error()` here if `e.err != nil`?

internal/gcerr/gcerr.go
```diff
@@ -0,0 +1,103 @@
...
+
+// New returns a new error with the given code, underlying error and message. Pass 1
+// for the call depth if New is called from the function raising the error; pass 2 if
+// it is called from a helper function that was invoked by the original function; and
+// so on.
+func New(c ErrorCode, err error, callDepth int, msg string) *Error {
+	return &Error{
+		Code:  c,
+		msg:   msg,
+		frame: xerrors.Caller(callDepth),
```
vangent: Sorry, what is this used for? I don't see it in tests.

errors/errors.go
vangent: It would be nice, but I guess it's not critical.

internal/gcerr/gcerr.go
vangent: Sorry, I meant specifically the `callDepth`.

jba: We'll need it in blob, because the top-level functions call a helper to wrap. So we'll need to pass 2 in order for the file and line to refer to the top-level function, not the wrapper.

internal/gcerr/gcerr.go
jba: (Huh, I thought I wrote an answer to this but I guess it got eaten?)

This will happen automatically. See this test:
```
        {
            New(AlreadyExists, errors.New("wrapped"), 1, ""),
            "%v",
            []string{`^AlreadyExists: wrapped`},
        },
      ```

File: internal/gcerr/gcerr.go (hunk 11 above)
Comment (vangent) [reply internal/gcerr/gcerr.go]: OK.

**Comment by jba** [comment]
@zombiezen, any other concerns that you'd like to discuss? (Robert told me he is okay with discoverability via adding sufficient docs + examples.)

File: gcerrors/errors.go [hunk 12]
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// Code returns the ErrorCode of err if it is an *Error.
+// It returns Unknown if err is a non-nil error of a different type.
+// If err is nil, it returns the special code OK.
+func Code(err error) ErrorCode {
+	if err == nil {
+		return OK
+	}
+	if e, ok := err.(*gcerr.Error); ok {
```
zombiezen: Isn't the correct idiom here to use the `As` function?

```suggestion
	var e *gcerr.Error
	if errors.As(err, &e) {
```

jba:
>  Having a shared set likely means some not being relevant to certain APIs, which I think is okay.

Yes and yes. My experience with common error codes (e.g. [gRPC's](https://github.com/grpc/grpc-go/blob/master/codes/codes.go)) is that the codes that you haven't come across in your work don't add any cognitive clutter. 

> As far as end-user impact: this does mean that in order to use a particular API with the error codes, they now need to import two packages. 

Yes, but I think that's a minor issue. 

> OTOH, what happens if we have an error that's incredibly specific to one API, does that go here? (My recommendation is that we write something up in design.md.)

We could have a maximum value for shared codes, and API-specific errors could be defined above that. I will add something in in `design.md`.

PR #1129, opened 2019-01-16
pubsub/doc: update docstrings to be consistent with other packages

vangent:


pubsub/gcppubsub/gcppubsub.go
```diff
@@ -72,8 +75,8 @@ func SubscriberClient(ctx context.Context, conn *grpc.ClientConn) (*raw.Subscrib
 // TopicOptions will contain configuration for topics.
 type TopicOptions struct{}
 
-// OpenTopic opens the topic on GCP PubSub for the given projectID and
-// topicName.
+// OpenTopic creates a *pubsub.Topic backed by GCP PubSub for the given projectID
```
jba: This is a little misleading, because no backend resources are created. Does blob use the same phrase?

pubsub/pubsub.go
```diff
@@ -14,6 +14,23 @@
 
 // Package pubsub provides an easy and portable way to interact with publish/
 // subscribe systems.
+//
+// Subpackages contain distinct implementations of pubsub for various providers,
+// including Cloud and on-prem solutions. For example, "gcspubsub" supports
+// Google Cloud Pub/Sub. Your application should import one of these
+// provider-specific subpackages and use its exported function(s) to create a
```
jba: "functions": there will always be two.

jba: Also, that word "create" again.

pubsub/gcppubsub/gcppubsub.go
vangent: Good catch; `blob` uses "returns". Updated to use "returns" and also to clarify that the GCP pubsub topic must already exist.

pubsub/pubsub.go
vangent: Done, and done.

blob/fileblob/fileblob.go
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 // Package fileblob provides a blob implementation that uses the filesystem.
-// Use OpenBucket to construct a blob.Bucket.
+// Use OpenBucket to construct a *blob.Bucket.
```
ijt: This doesn't make sense. The thing being constructed is a blob.Bucket. "Constructing" the pointer is trivial.

vangent: I'm updating this for consistency with other package docstrings. It feels more accurate because that's what the function is actually returning, no?

ijt: Okay, I guess no one will really be confused by it.

pubsub/rabbitpubsub/doc.go
```diff
@@ -23,7 +25,7 @@
 // A Pub/Sub subscription is an AMQP queue. The queue should be bound to the exchange
 // that is the topic of the subscription. See the package example for details.
 //
-// This package exposes the following types for As:
-// Topic: *amqp.Connection
-// Subscription: *amqp.Connection
+// As
```
ijt: This should probably be replaced with a complete sentence, like "The As method can be called on the following generic types to get the following provider-specific types."

vangent: There's a full example of how to use `As` on `Topic.As`.

Issue #1126, opened 2019-01-16
all: add OpenCensus tracing


jba:
We should use the OpenCensus API to add tracing to our APIs. (This proposal covers only tracing, not metrics.)

We should trace the methods of the concrete type that make RPCs. If we trace the concrete type, we only have to write the code once to cover all drivers, and it means new driver authors don't have to do anything.

An exception would be the driver.OpenXXX methods, which should have tracing if they make RPCs.

We should write a small `internal/trace` package to assist. Each traced method will need two lines at the top, to start the trace and defer its end. This is an example from `cloud.google.com/go/datastore`:
```
func (c *Client) Get(ctx context.Context, key *Key, dst interface{}) (err error) {
    ctx = trace.StartSpan(ctx, "cloud.google.com/go/datastore.Get")
    defer func() { trace.EndSpan(ctx, err) }()
```

Note that the error return value is named and passed to the `EndSpan` function so it can be turned into an OpenCensus [Status](https://godoc.org/go.opencensus.io/trace#Status). To do that right, we need to be able to extract an error code, so this proposal is dependent on #1124.

vangent:
See also #629 which overlaps with this. Maybe file a separate issue for metrics and close #629?

jba:
Tracing has been added to all four APIs.

PR #1136, opened 2019-01-17
pubsub: add tracing

jba:
See #1126.

### DIFF
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -43,6 +43,7 @@ import (
 	"github.com/googleapis/gax-go"
 	"gocloud.dev/internal/batcher"
 	"gocloud.dev/internal/retry"
+	"gocloud.dev/internal/trace"
 	"gocloud.dev/pubsub/driver"
 )
 
@@ -97,13 +98,16 @@ type msgErrChan struct {
 // Send publishes a message. It only returns after the message has been
 // sent, or failed to be sent. Send can be called from multiple goroutines
 // at once.
-func (t *Topic) Send(ctx context.Context, m *Message) error {
+func (t *Topic) Send(ctx context.Context, m *Message) (err error) {
+	ctx = trace.StartSpan(ctx, "gocloud.dev/pubsub.Topic.Send")
+	defer func() { trace.EndSpan(ctx, err) }()
+
 	// Check for doneness before we do any work.
 	if err := ctx.Err(); err != nil {
 		return err
 	}
 	t.mu.Lock()
-	err := t.err
+	err = t.err
 	t.mu.Unlock()
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


pubsub/pubsub.go
```diff
@@ -208,7 +217,10 @@ type Subscription struct {
 // concurrently from multiple goroutines. The Ack() method of the returned
 // Message has to be called once the message has been processed, to prevent it
 // from being received again.
-func (s *Subscription) Receive(ctx context.Context) (*Message, error) {
+func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
+	ctx = trace.StartSpan(ctx, "gocloud.dev/pubsub.Receive")
```
ijt: Shouldn't it be pubsub.Subscription.Receive?

pubsub/pubsub.go
```diff
@@ -338,7 +353,9 @@ func defaultAckBatcher(ctx context.Context, s *Subscription) driver.Batcher {
 	const maxHandlers = 1
 	handler := func(items interface{}) error {
 		ids := items.([]driver.AckID)
-		err := retry.Call(ctx, gax.Backoff{}, s.driver.IsRetryable, func() error {
+		err := retry.Call(ctx, gax.Backoff{}, s.driver.IsRetryable, func() (err error) {
+			ctx = trace.StartSpan(ctx, "gocloud.dev/pubsub.SendAcks")
```
ijt: pubsub/driver.Subscription.SendAcks?

pubsub/pubsub.go
```diff
@@ -175,8 +182,10 @@ func newTopic(d driver.Topic) *Topic {
 			dms = append(dms, dm)
 		}
 
-		return retry.Call(callCtx, gax.Backoff{}, d.IsRetryable, func() error {
-			return d.SendBatch(callCtx, dms)
+		return retry.Call(callCtx, gax.Backoff{}, d.IsRetryable, func() (err error) {
+			ctx := trace.StartSpan(callCtx, "gocloud.dev/pubsub.SendBatch")
```
ijt: pubsub/driver.Topic.SendBatch?

vangent: This is a driver call, not a concrete type call, do we even need to trace it? The end user doesn't know that this API exists...

Or at least distinguish it from the concrete type traces somehow, e.g. `pubsub.driver.SendBatch`.

pubsub/pubsub.go
vangent: I don't feel strongly, but I also found it a bit weird that some had `Topic`/`Subscription` and others didn't. Might be better to be consistent even if it's not necessary to be unique.

jba: OK, wrote method path names out in full.

jba:
PTAL.

PR #1138, opened 2019-01-17
secrets: add tracing

jba:
See #1126.


internal/docstore/docstore.go
```diff
@@ -0,0 +1,4 @@
+package docstore
```
clausti: accidental inclusion?

jba: oops. Should be gone.

PR #1139, opened 2019-01-17
secrets: add a Vault implementation

shantuo:
Fixes: #420


vangent:
Tests are failing, you may need to [update the proxy](https://github.com/google/go-cloud/blob/master/internal/proxy/README.md) for new dependencies.

secrets/vault/vault.go
```diff
@@ -0,0 +1,93 @@
...
+	"context"
+	"encoding/base64"
+	"errors"
+	"path"
+
+	"github.com/hashicorp/vault/api"
+	"gocloud.dev/secrets"
+)
+
+// Config is the authentication configurations of the Vault server.
```
vangent: Consider dropping this type and just putting the two values into `Dial` args.

Document the args (e.g., it looks like `Token` is optional?). Add a link to the public documentation.

secrets/vault/vault.go
```diff
@@ -0,0 +1,93 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if cfg.Token != "" {
+		c.SetToken(cfg.Token)
+	}
+	return c, nil
+}
+
+// NewKeeper returns a new Keeper to do encryption and decryption.
```
vangent: Refresh to match others. E.g.,

```
// NewKeeper returns a *secrets.Keeper that uses Hashicorp's Vault.
// See the package documentation for an example.
```

vangent: And document what exactly `keyID` is....

secrets/vault/vault_test.go
```diff
@@ -0,0 +1,87 @@
...
+	apiAddress = "http://127.0.0.1:0"
+)
+
+type harness struct {
+	client *api.Client
+	close  func()
+}
+
+func (h *harness) MakeDriver(ctx context.Context) (driver.Keeper, error) {
+	return &keeper{
```
vangent: Consider adding a `newKeeper` function that returns the driver, and using it here instead of creating the struct explicitly.

secrets/vault/vault_test.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (h *harness) Close() {
+	h.close()
+}
+
+func newHarness(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
+	// Start a new test server.
+	c, cleanup := testVaultServer(t)
+	// Enable the transit API.
```
vangent: What is the transit API?

shantuo: The transit secrets engine, which allows us to call the encryption service of vault. Updated comment.

secrets/vault/vault.go
shantuo: The value of combining them together is that it may grow in future. The vault's api.Config includes only a subset of the configurations that api.Client can set. We may need to add them in future so this provides the flexibility. I've added some comments talking about the token.

secrets/vault/vault_test.go

internal/contributebot/go.mod
```diff
@@ -7,12 +7,11 @@ require (
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/google/go-cmp v0.2.0
 	github.com/google/go-github v17.0.0+incompatible
-	github.com/google/go-querystring v1.0.0 // indirect
 	github.com/google/wire v0.2.0
-	go.opencensus.io v0.17.0
+	go.opencensus.io v0.18.1-0.20181204023538-aab39bd6a98b
```
vangent: Why are there ContributeBot changes in here?

PR #1143, opened 2019-01-17
blob: add tracing

jba:
Add tracing to methods that (a) make RPCs to the service, and (b) are
unlikely to be called in an inner loop. Specifically, do NOT trace:

- SignedURL, because it doesn't make an RPC
- Read and Write, because they are likely to be called frequently,
  so the trace information would be voluminous.


blob/blob.go
```diff
@@ -401,7 +402,10 @@ func (b *Bucket) As(i interface{}) bool {
 
 // ReadAll is a shortcut for creating a Reader via NewReader with nil
 // ReaderOptions, and reading the entire blob.
-func (b *Bucket) ReadAll(ctx context.Context, key string) ([]byte, error) {
+func (b *Bucket) ReadAll(ctx context.Context, key string) (_ []byte, err error) {
+	ctx = trace.StartSpan(ctx, "gocloud.dev/blob.ReadAll")
+	defer func() { trace.EndSpan(ctx, err) }()
+
```
vangent: Add to `List` as well.

blob/blob.go
```diff
@@ -401,7 +402,10 @@ func (b *Bucket) As(i interface{}) bool {
 
 // ReadAll is a shortcut for creating a Reader via NewReader with nil
 // ReaderOptions, and reading the entire blob.
-func (b *Bucket) ReadAll(ctx context.Context, key string) ([]byte, error) {
```
vangent: I don't think we need to trace this one explicitly, as it calls `NewReader`.

blob/blob.go
```diff
@@ -495,7 +505,10 @@ func (b *Bucket) NewRangeReader(ctx context.Context, key string, offset, length
 }
 
 // WriteAll is a shortcut for creating a Writer via NewWriter and writing p.
-func (b *Bucket) WriteAll(ctx context.Context, key string, p []byte, opts *WriterOptions) error {
+func (b *Bucket) WriteAll(ctx context.Context, key string, p []byte, opts *WriterOptions) (err error) {
```
vangent: Same, don't think we need this one as it calls `NewWriter`.

blob/blob.go
```diff
@@ -522,7 +535,10 @@ func (b *Bucket) WriteAll(ctx context.Context, key string, p []byte, opts *Write
 //
 // The caller must call Close on the returned Writer, even if the write is
 // aborted.
-func (b *Bucket) NewWriter(ctx context.Context, key string, opts *WriterOptions) (*Writer, error) {
+func (b *Bucket) NewWriter(ctx context.Context, key string, opts *WriterOptions) (_ *Writer, err error) {
```
vangent: The traces for `NewRangeReader` and `NewWriter` probably aren't that useful, as they don't usually make RPCs either. Could we start a trace here, and close it when the `Reader`/`Writer` is closed?

blob/blob.go
jba: List makes no RPCs, and the returned ListIterator has no Stop or Close method, so we don't know when to end the span.

blob/blob.go
```diff
@@ -571,6 +567,7 @@ func (b *Bucket) NewWriter(ctx context.Context, key string, opts *WriterOptions)
 		dopts.Metadata = md
 	}
 	ctx, cancel := context.WithCancel(ctx)
+	tctx := trace.StartSpan(ctx, "gocloud.dev/blob.NewWriter")
```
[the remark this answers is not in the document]
jba: PTAL.

blob/blob.go
```diff
@@ -498,11 +498,17 @@ func (b *Bucket) NewRangeReader(ctx context.Context, key string, offset, length
...
+	tctx := trace.StartSpan(ctx, "gocloud.dev/blob.NewRangeReader")
+	defer func() {
+		if err != nil {
+			trace.EndSpan(tctx, err)
+		}
+	}()
 	r, err := b.b.NewRangeReader(ctx, key, offset, length, dopts)
 	if err != nil {
-		return nil, wrapError(b.b, err)
+		err = wrapError(b.b, err)
```
vangent: Is this changed needed? Isn't err always set to the return value?
https://play.golang.org/p/SbrHLsHoGoD

PR #1144, opened 2019-01-18
pubsub: add benchmark for Receive

jba:
A simple benchmark that measures how many messages we can receive per second.
For now, just using the GCP driver.

The current result is about 100 msgs/sec. cloud.google.com/go/pubsub
achieves about 8000 with default settings, so we have a way to go.


pubsub/benchmarks_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	"gocloud.dev/gcp"
+	"gocloud.dev/pubsub"
+	"gocloud.dev/pubsub/gcppubsub"
+	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/semaphore"
+)
+
+var (
+	projectID = flag.String("benchmark-project", "", "project ID")
```
vangent: Can't we get the projectID from the creds?

https://github.com/google/go-cloud/blob/master/gcp/gcp.go#L87

pubsub/benchmarks_test.go
```diff
@@ -0,0 +1,120 @@
...
+	topicName = flag.String("benchmark-topic", "", "topic name")
+)
+
+func BenchmarkReceive(b *testing.B) {
```
vangent: Why not do this as a conformance benchmark like the one I added for `blob`?
https://github.com/google/go-cloud/blob/master/blob/drivertest/drivertest.go#L202

pubsub/benchmarks_test.go
```diff
@@ -0,0 +1,120 @@
...
+	topic, sub, cleanup, err := openGCPTopicAndSub()
+	if err != nil {
+		b.Fatal(err)
+	}
+	defer cleanup()
+	b.Run("gcp", func(b *testing.B) {
+		b.ResetTimer()
+		for i := 0; i < b.N; i++ {
+			b.StopTimer()
+			if err := publishN(topic, nMessages, 100, attrs, body); err != nil {
```
vangent: Instead of 100, use `RunParallel`?
https://golang.org/pkg/testing/#B.RunParallel

pubsub/benchmarks_test.go
```diff
@@ -0,0 +1,120 @@
...
+// Call function f n times concurrently, with a maximum number of maxg goroutines.
+// Wait until all calls complete. If any fail, cancel the remaining ones.
+func runConcurrently(n, maxg int, f func(context.Context) error) error {
+	g, ctx := errgroup.WithContext(context.Background())
+	sem := semaphore.NewWeighted(int64(maxg))
+	for i := 0; i < n; i++ {
+		if err := sem.Acquire(ctx, 1); err != nil {
+			break // Don't return the context cancelation, return the error that caused it.
+		}
+		g.Go(func() error {
```
vangent: This also spins up a goroutine for each message, which seems like it might add significant overhead. Wouldn't it be better to do a more traditional threadpool and let each goroutine handle `n / nGoRoutines` messages?

pubsub/benchmarks_test.go
jba: That's a good idea. I was focussing on GCP performance now because I've been comparing us to cloud.google.com/go/pubsub. But I will do that in a followup PR.

jba: #1147

pubsub/benchmarks_test.go
jba: I don't quite see how I'd use it here. This is setup code; I want to publish a lot of messages so I can receive them in the timed part of the benchmark, and I want to do it quickly. I don't want to run the timed part of the benchmark in parallel, at least not here.

pubsub/benchmarks_test.go
jba: That's a fair point, though in this case the RPC time will dominate. Nevertheless, I changed it just to make sure.

pubsub/benchmarks_test.go
jba: Doesn't work for me. My default creds come from ~/.config/gcloud/application_default_credentials.json, which doesn't have a project. @zombiezen any idea how to fix?

vangent: I think you need to set it using `gcloud config set`:
https://cloud.google.com/sdk/gcloud/reference/config/set

jba:
PTAL.

pubsub/benchmarks_test.go
vangent: Sorry, I meant more for the receiving part. Something like this:

```
    // publish b.N * nMessages messages
    b.ResetTimer
	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {
			// receive nMessages messages
        }
    }
```

Maybe this doesn't work because it requires all `b.N * nMessages` messages to be published up front. But it lets the degree of concurrency/parallelism be determined by standard `go benchmark` flags, which is nice.

pubsub/benchmarks_test.go
ijt: I just ran a little benchmark to find out how long it takes to spin up a goroutine. It's something less than 200 nanoseconds, so probably not worth worrying about.

```
[ ~/src/gr ] go run main.go
Each goroutine took about 196ns to run.
[ ~/src/gr ] cat main.go 
package main

import (
	"fmt"
	"sync"
	"time"
)

func main() {
	var wg sync.WaitGroup
	n := int(1e6)
	wg.Add(n)
	t0 := time.Now()
	for i := 0; i < n; i++ {
		go func() { wg.Done() }()
	}
	wg.Wait()
	dt := time.Now().Sub(t0)
	dtPerGoroutine := time.Duration(float64(dt) / float64(n))
	fmt.Printf("Each goroutine took about %v to run.\n", dtPerGoroutine)
}
```

pubsub/benchmarks_test.go
jba: No, I have a project set in that way. Doesn't matter.

pubsub/benchmarks_test.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	attrs := map[string]string{"label": "value"}
+	body := []byte("hello, world")
+	const nMessages = 10000
+	topic, sub, cleanup, err := openGCPTopicAndSub()
+	if err != nil {
+		b.Fatal(err)
+	}
+	defer cleanup()
+	b.Run("gcp", func(b *testing.B) {
```
ijt: Why only GCP? Shouldn't we be able to run the benchmarks for all the different providers?

ijt: Maybe add a todo or an issue?

pubsub/benchmarks_test.go
jba: Still doesn't make sense. `RunParallel` "creates multiple goroutines and distributes b.N iterations among them. " But `b.N` will aways be 1, because it takes long enough to run one iteration.

pubsub/benchmarks_test.go
jba: Sure, that's #1147.

pubsub/benchmarks_test.go
zombiezen: @jba is right: explicit project ID is the right way to go here. IIRC there are examples of this around the Go Cloud repo.

jba:
Hopefully I fixed the breakage. I also added doc.
PTAL at the doc.

jba:
Ignore, wrong PR.

PR #1156, opened 2019-01-23
blob/fileblob: create tempfile near target

EvyBongers:
When using `blob.NewWriter` on a fileblob that uses a mounted volume as backing filesystem, `writer.Close()` errors, because `os.Rename` can't rename files across mounts.

Creating the tempfile near the target file, ensures that the tempfile will be on the same filesystem as the target file.


vangent:
Thanks!

PR #1157, opened 2019-01-23
all: mention postgres in features list, add package names

ijt:

PR #1166, opened 2019-01-23
internal/docs/design.md: incorporate gcerrors

jba:
Rewrite the advice on errors to incorporate the new gcerrors package
and error codes.

all: mention postgres in features list, add package names (#1157)

### DIFF
--- a/internal/docs/design.md
+++ b/internal/docs/design.md
@@ -208,22 +208,44 @@ Concrete types should:
     end users, so that users can't peek into provider-specific error details
     without using `As`. Make sure not to double-wrap.
 
--   Prefix errors with the package name (e.g., `blob: fail`). For most errors,
-    this can be done as part of the wrapping.
+- Use `internal/gcerr.New` when wrapping driver errors, like so:
+    ```
+    if err := driver.Call(xyz); err != nil {
+        return gcerr.New(code, err, 1, "blob")
+    }
+    ```
+    The first argument is an error code. See below for advice on choosing the
+    appropriate code.
 
--   Prefer to keep details of returned `error`s unspecified. The most common
-    case is that the caller will only care whether an operation succeeds or not.
+    The third argument is the distance in stack frames from the function whose
+    location should be associated with the error. It should be `1` if you are
+    calling `New` from the same function that made the driver call, `2` if you
+    are calling new from a helper function, and son on. The fourth argument is
+    an additional string that will display with the error. You should pass
+    the API name.
 
--   If certain `error`s are interesting for callers to distinguish, expose
-    predicate functions like
-    [`os.IsNotExist`](https://golang.org/pkg/os/#IsNotExist). This allows the
-    internal representation of the `error` to change over time while being
-    simple to use. In most cases, these predicate functions will translate into
-    a driver interface function to do the check. The concrete type must pass
-    such functions the raw error that the driver returned (not a wrapped error).
+-   By default, choose the code `Unknown`, keeping details of returned `error`s
+    unspecified. The most common case is that the caller will only care whether
+    an operation succeeds or not.
 
--   Expose an `ErrorAs` function to allow users to access provider-specific
-    error types.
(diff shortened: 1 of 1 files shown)


internal/docs/design.md
```diff
@@ -208,22 +208,44 @@ Concrete types should:
...
-    predicate functions like
-    [`os.IsNotExist`](https://golang.org/pkg/os/#IsNotExist). This allows the
-    internal representation of the `error` to change over time while being
-    simple to use. In most cases, these predicate functions will translate into
-    a driver interface function to do the check. The concrete type must pass
-    such functions the raw error that the driver returned (not a wrapped error).
-
--   Expose an `ErrorAs` function to allow users to access provider-specific
-    error types.
+- Use `internal/gcerr.New` when wrapping driver errors, like so:
```
vangent: Nit: can you run `mdformat` on this for consistency before merging?

internal/docs/design.md
```diff
@@ -208,22 +208,44 @@ Concrete types should:
...
+        return gcerr.New(code, err, 1, "blob")
+    }
+    ```
+    The first argument is an error code. See below for advice on choosing the
+    appropriate code.
+
+    The third argument is the distance in stack frames from the function whose
+    location should be associated with the error. It should be `1` if you are
+    calling `New` from the same function that made the driver call, `2` if you
+    are calling new from a helper function, and son on. The fourth argument is
```
vangent: `son on` -> so on

internal/docs/design.md
```diff
@@ -208,22 +208,44 @@ Concrete types should:
...
+    - If more than one error code makes sense, choose the most specific one.
+    - If none make sense, choose `Unknown`.
+    - If none make sense but you want something more specific than `Unknown`:
+      - If you can generalize your code to make it applicable to more than just your
+        API, add it to `gcerrors.ErrorCode`. Look at the [gRPC error
+        codes](https://github.com/grpc/grpc-go/blob/master/codes/codes.go) for inspiration.
+      - Otherwise, you can define a custom code in your API's concrete package.
+        Your code should use a negative integer.
+
+-   If your package already exposes an `ErrorAs` function to allow users to
```
vangent: I am not sure I buy this one yet.

One of the reasons we went for `As` instead of just type casts is that it gives providers flexibility to support multiple types, even if the underlying error only has a single type. For example, a provider might choose to support `aws.Error` for `As` even though that's not what the underlying provider uses inherently (e.g., you might do this to provide full interoperability between your provider and AWS).

I'm not sure how realistic a use case that is, but `errors.As` probably won't cover it. Given that all of our APIs have `ErrorAs` already, it probably makes sense to keep with that pattern even if all the providers use `errors.As` as the implementation.

jba: `errors.As` does indeed handle that case. You can define your own `As` method to override the default behavior. https://godoc.org/golang.org/x/exp/errors#As

internal/docs/design.md
jba: Done.

internal/docs/design.md
jba: Done.

internal/docs/design.md
jba: Language softened.

internal/docs/design.md
```diff
@@ -141,9 +141,9 @@ concrete type backed by GCS.
 -   Order arguments that are less likely to change across multiple calls to the
     constructor before ones that are likely to change. For example, connection
     and authorization related arguments should go before names, so
-    ```OpenBucket(ctx, client, "mybucket")``` instead of ```OpenBucket(ctx,
-    "mybucket", client)```.
--   All public constructors should take an `Options` struct (see next section).
+    `OpenBucket(ctx, client, "mybucket")` instead of `OpenBucket(ctx,
+    "mybucket", client)`. - All public constructors should take an `Options`
```
vangent: It looks like `mdformat` pulled this bullet point up incorrectly.

jba: Done.

PR #1170, opened 2019-01-23
internal/imports: improve index.html

jba:
Fix the message you see before redirected.

While we're at it, use a separate file for index.html instead
of trying to twist the template file into the right shape.

### DIFF
--- a/internal/imports/index.html
+++ b/internal/imports/index.html
@@ -6,6 +6,7 @@
     <meta http-equiv="refresh" content="0; url=https://github.com/google/go-cloud">
   </head>
   <body>
-    Redirecting you to our GitHub home page.
+    You will be redirected to the project's Github page shortly. If that doesn't
+    happen, <a href="https://github.com/google/go-cloud">click here</a>.
   </body>
 </html>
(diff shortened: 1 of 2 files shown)


internal/imports/index.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta name="go-import" content="gocloud.dev git https://github.com/google/go-cloud">
+    <meta name="go-source" content="gocloud.dev https://github.com/google/go-cloud https://github.com/google/go-cloud/tree/master{/dir} https://github.com/google/go-cloud/tree/master{/dir}/{file}#L{line}">
+    <meta http-equiv="refresh" content="0; url=https://github.com/google/go-cloud">
+  </head>
+  <body>
+    Redirecting you to our GitHub home page.
```
eliben: Could redirects fail? (maybe some browser configuration, IIRC browsers can disable this). What I've commonly seen done is somethiing like:

You will be redirected to the project's Github page shortly. If that doesn't happen, [click here][link]

jba: Done.

jba:
@eliben, our convention is that all assignees must approve before we can merge. (No pressure, just letting you know.)

PR #1179, opened 2019-01-24
pubsub: wrap errors with gcerrors

jba:
Wrap driver errors using the gcerrors package.

### DIFF
--- a/internal/gcerr/gcerr.go
+++ b/internal/gcerr/gcerr.go
@@ -55,6 +55,8 @@ const (
 	FailedPrecondition ErrorCode = 7
 )
 
+// TODO(jba) call stringer after it's fixed for modules
+
 // Call "go generate" whenever you change the above list of error codes.
 // To get stringer:
 //   go get golang.org/x/tools/cmd/stringer
--- a/internal/gcerr/gcerr.go
+++ b/internal/gcerr/gcerr.go
@@ -50,6 +50,9 @@ const (
 
 	// The feature is not implemented.
 	Unimplemented ErrorCode = 6
+
+	// The system was in the wrong state.
+	FailedPrecondition ErrorCode = 7
 )
 
 // Call "go generate" whenever you change the above list of error codes.
--- a/pubsub/pub_test.go
+++ b/pubsub/pub_test.go
@@ -18,6 +18,8 @@ func (t *funcTopic) SendBatch(ctx context.Context, ms []*driver.Message) error {
 	return t.sendBatch(ctx, ms)
 }
 
+func (t *funcTopic) IsRetryable(error) bool { return false }
+
 func TestTopicShutdownCanBeCanceledEvenWithHangingSend(t *testing.T) {
 	dt := &funcTopic{
 		sendBatch: func(ctx context.Context, ms []*driver.Message) error {
--- a/pubsub/pub_test.go
+++ b/pubsub/pub_test.go
@@ -18,8 +18,6 @@ func (t *funcTopic) SendBatch(ctx context.Context, ms []*driver.Message) error {
 	return t.sendBatch(ctx, ms)
 }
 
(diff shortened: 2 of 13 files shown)


pubsub/pub_test.go
```diff
@@ -19,8 +20,6 @@ func (t *funcTopic) SendBatch(ctx context.Context, ms []*driver.Message) error {
 
 func (s *funcTopic) IsRetryable(error) bool { return false }
```
vangent: Nit: could be deleted too, no?

pubsub/pubsub.go
```diff
@@ -369,3 +372,22 @@ func defaultAckBatcher(ctx context.Context, s *Subscription) driver.Batcher {
...
+
+func wrapError(ec errorCoder, err error) error {
+	if err == nil {
+		return nil
+	}
+	// Don't wrap context errors.
+	if _, ok := err.(*retry.ContextError); ok || err == context.Canceled || err == context.DeadlineExceeded {
+		return err
+	}
+	// Don't double-wrap.
```
vangent: When would this happen? It seems like it would just be hiding a bug.

pubsub/pubsub.go
```diff
@@ -369,3 +372,22 @@ func defaultAckBatcher(ctx context.Context, s *Subscription) driver.Batcher {
 	}
 	return batcher.New(reflect.TypeOf([]driver.AckID{}).Elem(), maxHandlers, handler)
 }
+
+type errorCoder interface {
+	ErrorCode(error) gcerrors.ErrorCode
+}
+
+func wrapError(ec errorCoder, err error) error {
+	if err == nil {
```
vangent: When would this happen? Seems like it's hiding a bug.

pubsub/sub_test.go
```diff
@@ -44,8 +45,6 @@ func (s *scriptedSub) SendAcks(ctx context.Context, ackIDs []driver.AckID) error
 
 func (s *scriptedSub) IsRetryable(error) bool { return false }
```
vangent: Remove?

pubsub/pub_test.go
jba: Done.

jba:
PTAL.
I still need to do `ErrorAs`, but I'll do that in a separate PR if that's OK.

internal/gcerr/gcerr.go
```diff
@@ -50,6 +50,9 @@ const (
 
 	// The feature is not implemented.
 	Unimplemented ErrorCode = 6
+
+	// The system was in the wrong state.
+	FailedPrecondition ErrorCode = 7
```
vangent: Doesn't `gcerr_string.go` need to be updated for this?

pubsub/pubsub.go
vangent: `if err == nil` wasn't deleted?

internal/gcerr/gcerr.go
jba: The `stringer` tool is currently broken in module mode. I'm going to work with the tools team in New York to work around that. I'll address in a followup PR.

pubsub/pubsub.go
vangent: Ping on this one.

internal/gcerr/gcerr.go
vangent: Add a TODO?

pubsub/pubsub.go
jba: Sorry, missed it. Done now.

internal/gcerr/gcerr.go
jba: Done.

jba:
I also had to add back an ErrorCode method on a test driver. It failed during the MacOS travis run. Not sure what that's about...

PTAL

PR #1182, opened 2019-01-25
blob/fileblob: convert remaining / characters in URL to os.PathSeparator; add tests

### LINKED ISSUE #1154: blob/fileblob: URL opening does not change slashes to OS file separator
### Describe the bug

When calling `blob.OpenURL` with a `file://` URL on Windows, the constructed paths will use slash-separated paths instead of backslash-separated paths.

### To Reproduce

```go
blob.OpenURL("file:///C:/foo.txt")
```

## Expected behavior

Bucket is opened with `C:\foo.txt`, but instead it will be opened with `C:/foo.txt`.

### Version

master

### Additional context

Problematic lines here:

https://github.com/google/go-cloud/blob/5dcea9a7c5454dd0811d61ca18d798b852aeae19/blob/fileblob/fileblob.go#L74-L78

vangent:
Fixes #1154.


blob/fileblob/fileblob.go
```diff
@@ -72,12 +76,18 @@ const defaultPageSize = 1000
...
 	blob.Register("file", func(_ context.Context, u *url.URL) (driver.Bucket, error) {
-		path := u.Path
-		if os.PathSeparator != '/' && strings.HasPrefix(path, "/") {
+		return openBucket(mungeURLPath(u.Path, os.PathSeparator), nil)
+	})
+}
+
+func mungeURLPath(path string, pathSeparator uint8) string {
+	if pathSeparator != '/' {
+		if strings.HasPrefix(path, "/") {
```
zombiezen: `strings.TrimPrefix`?

blob/fileblob/fileblob.go
```diff
@@ -72,12 +76,18 @@ const defaultPageSize = 1000
...
 			path = path[1:]
 		}
-		return openBucket(path, nil)
-	})
+		path = strings.Replace(path, "/", string(pathSeparator), -1)
```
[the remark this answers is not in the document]
vangent: Nice, done.

blob/fileblob/fileblob.go

Issue #1177, opened 2019-01-24
all: update non-local examples to compile but not run instead of expecting failure


ijt:
The examples should ideally be fairly realistic and easy to copy and paste into real apps without too much modification. The output should show things working, not failures due to not being able to assume the example can access the network.

That's not the current situation though. For example, in `blob/gcsblob/example_test.go` the [expected output](https://github.com/google/go-cloud/blob/d22bb640872cc897bb38b124f1ce8c55197fda27/blob/gcsblob/example_test.go#L70) is 

> ReadAll failed due to invalid credentials.

One possible solution would be to put a line like this at the top of the example_test.go files:
```
// +build !no_network
```
and then pass `-tag=no_network` to `go test` within Travis. A downside would be that these examples would not be built or run from Travis, though presumably we would find out quickly if they are broken since people would try them and report any issues they find. 

This solution isn't perfect, but I think it beats the current situation where our examples don't really make sense.

ijt:
@vangent, @zombiezen what do you think?

jba:
Remove `// Output`?

ijt:
@jba, that makes it so the example doesn't run under any circumstances when `go test` is invoked.

zombiezen:
Right, but examples will always be compiled, which I think is more of the assurance we need. I'm not seeing the benefit to running the examples versus the cost to clarity that you cite in your problem statement.

ijt:
Fair enough. Let's go with that for now.

vangent:
Please wait, I have an opinion :-).

ijt:
It seems like there's a lot of copy/pasted boilerplate within option 2. I don't see a way around it without adding something like `gcp.FakeDefaultCredentials` but I doubt whether anyone would be willing to approve adding that.

vangent:
Yea, generally I think it's preferable to have copy/paste boilerplate in examples than to factor out to helper functions, unless the helper functions are really not useful, because the helper code aren't included in the example code. `gcp.FakeInvalidCredentials` might be OK though, as we'd presumably have a comment telling the user that they should use `gcp.DefaultCredentials()` instead.

jba:
> because the helper code aren't included in the example code.

Note this sentence from https://godoc.org/testing:

> The entire test file is presented as the example when it contains a single example function, at least one other function, type, variable, or constant declaration, and no test or benchmark functions.

jba:
Not that I'm necessarily advocating for helper functions.

[...]

PR #1184, opened 2019-01-25
all: add gcp.FakeDefaultCredentials helper, use it

ijt:
This is related to #1177.

### DIFF
--- a/blob/gcsblob/example_test.go
+++ b/blob/gcsblob/example_test.go
@@ -25,24 +25,9 @@ import (
 	"golang.org/x/oauth2/google"
 )
 
-// jsonCreds is a fake GCP JSON credentials file.
-const jsonCreds = `
-{
-  "type": "service_account",
-  "project_id": "my-project-id"
-}
-`
-
 func Example() {
 	ctx := context.Background()
-
-	// Get GCP credentials.
-	// Here we use a fake JSON credentials file, but you could also use
-	// gcp.DefaultCredentials(ctx) to use the default GCP credentials from
-	// the environment.
-	// See https://cloud.google.com/docs/authentication/production
-	// for more info on alternatives.
-	creds, err := google.CredentialsFromJSON(ctx, []byte(jsonCreds))
+	creds, err := gcp.FakeDefaultCredentials(ctx)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/gcp/gcp.go
+++ b/gcp/gcp.go
@@ -75,6 +75,30 @@ func DefaultCredentials(ctx context.Context) (*google.Credentials, error) {
 	return adc, nil
 }
 
+// FakeDefaultCredentials gets GCP credentials from a fake JSON credentials
+// file in memory. This function is meant for use as a helper in example code,
+// to be replaced by gcp.DefaultCredentials(ctx) or google.CredentialsFromJSON
+// in non-example code.
+// See https://cloud.google.com/docs/authentication/production
+// for more info on alternatives.
+func FakeDefaultCredentials(ctx context.Context) (*google.Credentials, error) {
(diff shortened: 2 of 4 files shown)


ijt:
PTAL

blob/gcsblob/example_test.go
```diff
@@ -22,27 +22,11 @@ import (
...
-{
-  "type": "service_account",
-  "project_id": "my-project-id"
-}
-`
-
 func Example() {
 	ctx := context.Background()
-
-	// Get GCP credentials.
```
vangent: Can we leave the comment?

gcp/gcp.go
```diff
@@ -75,6 +76,29 @@ func DefaultCredentials(ctx context.Context) (*google.Credentials, error) {
 	return adc, nil
 }
 
+// FakeDefaultCredentials gets GCP credentials from a fake JSON credentials
+// file in memory. This function is meant for use as a helper in example code,
+// to be replaced by gcp.DefaultCredentials(ctx) or google.CredentialsFromJSON
+// in non-example code.
+// See https://cloud.google.com/docs/authentication/production
```
vangent: Don't think this reference is useful here.

gcp/gcp.go
```diff
@@ -75,6 +76,29 @@ func DefaultCredentials(ctx context.Context) (*google.Credentials, error) {
...
+// for more info on alternatives.
+func FakeDefaultCredentials(ctx context.Context) (*google.Credentials, error) {
+	// Here we use a fake JSON credentials file, but you could also use
```
vangent: Or this comment.

runtimevar/runtimeconfigurator/example_test.go
```diff
@@ -30,24 +30,9 @@ type MyConfig struct {
...
-{
-  "type": "service_account",
-  "project_id": "my-project-id"
-}
-`
-
 func ExampleNewVariable() {
 	ctx := context.Background()
-
-	// Get GCP credentials and dial the server.
```
vangent: Same.

gcp/gcp.go
ijt: Removed it.

blob/gcsblob/example_test.go
ijt: Done.

gcp/gcp.go
ijt: I'm not sure what you're asking for here.

runtimevar/runtimeconfigurator/example_test.go
ijt: Done.

ijt:
PTAL

gcp/gcp.go
vangent: I think you can remove the "Here we use a fake JSON credentials..." comment. Or reword it. It was written to be in the example code itself.

blob/gcsblob/example_test.go
vangent: Not just this line, the whole thing. It explains that we're using fake credentials here, but that they should use `DefaultCredentials` or one of the other alternatives from the link.

gcp/gcp.go
ijt: Removed it.

blob/gcsblob/example_test.go
ijt: Part of why I wanted this change was to get rid of the copy/pasted comments and have that text in one place, right above the definition of `gcp.FakeDefaultCredentials`.

ijt:
Added comments. PTAL.

zombiezen:
It seems unfortunate that this is exported API surface, since it means that it's a function that intentionally returns junk data. I was going to propose a different idea on #1177.

PR #1185, opened 2019-01-25
pubsub: turn benchmark into a conformance benchmark

vangent:
Some interesting notes:
*  The benchmark concurrently calls `Receive` with no concurrency; I left this unchanged.
*  I did try it locally with concurrency of 100, and the benchmark got slower for all 3 providers. Hmm. Warrants investigation....
*  The Rabbit provider is too slow for `nMessages=10000`; it reads about a few messages a second, so the benchmark takes a really long time to run. Not sure why it is so slow. Warrants investigation....


pubsub/drivertest/drivertest.go
```diff
@@ -389,3 +397,67 @@ func testAs(t *testing.T, newHarness HarnessMaker, st AsTest) {
...
+		nMessages          = 10000
+		concurrencySend    = 100
+		concurrencyReceive = 1
+	)
+	if nMessages%concurrencySend != 0 || nMessages%concurrencyReceive != 0 {
+		b.Fatal("nMessages must be divisible by # of sending/receiving goroutines")
+	}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		b.StopTimer()
```
ijt: Why is this here instead of at the bottom of the loop?

pubsub/drivertest/drivertest.go
```diff
@@ -389,3 +397,67 @@ func testAs(t *testing.T, newHarness HarnessMaker, st AsTest) {
...
+func publishNConcurrently(topic *pubsub.Topic, nMessages, nGoroutines int, attrs map[string]string, body []byte) error {
+	return runConcurrently(nMessages, nGoroutines, func(ctx context.Context) error {
+		return topic.Send(ctx, &pubsub.Message{Metadata: attrs, Body: body})
+	})
+}
+
+func receiveNConcurrently(sub *pubsub.Subscription, nMessages, nGoroutines int) error {
+	return runConcurrently(nMessages, nGoroutines, func(ctx context.Context) error {
+		m, err := sub.Receive(ctx)
+		if err == nil {
```
ijt: This is a bit unconventional. Why not this:
```
if err != nil {
    return err
}
m.Ack()
return nil
```

pubsub/rabbitpubsub/rabbit_test.go
```diff
@@ -66,6 +66,22 @@ func TestConformance(t *testing.T) {
 	drivertest.RunConformanceTests(t, harnessMaker, asTests)
 }
 
+func BenchmarkRabbit(b *testing.B) {
+	ctx := context.Background()
+	h := &harness{conn: mustDialRabbit(b)}
```
ijt: Why does this create a harness but the GCP one doesn't?

pubsub/drivertest/drivertest.go
vangent: FYI, this code is copied from the previous benchmark, which was GCPPubsub-specific.

The timer is stopped during the publish phase, because the benchmark is specifically timing `Receive` and not `Send`. So we `StopTimer` before publishing, and `StartTimer` again before receive.

pubsub/rabbitpubsub/rabbit_test.go
vangent: The GCP harness does record/replay, which we don't want for this.

PR #1197, opened 2019-01-26
pubsub: add ErrorAs

### LINKED ISSUE #957: pubsub: add ErrorAs()

jba:
- Add an ErrorAs method to Topic and Subscription.
- Add it to drivers as well.
- Write test.

Fixes #957.


pubsub/drivertest/drivertest.go
```diff
@@ -94,6 +99,10 @@ func (verifyAsFailsOnNil) SubscriptionCheck(s *pubsub.Subscription) error {
 	return nil
 }
 
+func (verifyAsFailsOnNil) ErrorCheck(t *pubsub.Topic, err error) error {
+	return nil
```
vangent: This should call `t.ErrorAs(nil)` and verify it returns false (like the other tests in the struct.

pubsub/pubsub.go
```diff
@@ -166,6 +166,27 @@ func (t *Topic) As(i interface{}) bool {
 	return t.driver.As(i)
 }
 
+// ErrorAs converts err to provider-specific types.
+// See Topic.As for more details.
+// ErrorAs panics if target is nil or not a pointer.
```
vangent: Be consistent with `blob`... I don't think panics in those scenarios does it?

pubsub/rabbitpubsub/rabbit.go
```diff
@@ -340,6 +346,21 @@ func (t *topic) As(i interface{}) bool {
 	return true
 }
 
+// ErrorAs implements driver.Topic.ErrorAs
+func (*topic) ErrorAs(err error, target interface{}) bool {
+	return errorAs(err, target)
+}
+
+func errorAs(err error, target interface{}) bool {
+	if aerr, ok := err.(*amqp.Error); ok {
```
vangent: Support `MultiError` here as well?

pubsub/pubsub.go
jba: Could I change blob? Both should be consistent with `xerrors.As`. See https://godoc.org/golang.org/x/exp/xerrors#As. Note that there is now an [official proposal](https://github.com/golang/go/issues/29934) for this.

vangent: Sure. And `runtimevar`.

pubsub/rabbitpubsub/rabbit.go
jba: Done.

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -137,6 +139,21 @@ func (t *topic) As(i interface{}) bool {
 	return true
 }
 
+// ErrorAs implements driver.Topic.ErrorAs
+func (*topic) ErrorAs(err error, target interface{}) bool {
+	return errorAs(err, target)
+}
+
+func errorAs(err error, target interface{}) bool {
+	if s, ok := status.FromError(err); ok {
```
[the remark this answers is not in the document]
jba: Done.

pubsub/rabbitpubsub/rabbit.go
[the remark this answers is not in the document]
jba: Rewritten with a switch.

pubsub/drivertest/drivertest.go
jba: Replaced with a panic check.

jba:
PTAL.

pubsub/rabbitpubsub/doc.go
```diff
@@ -28,4 +28,5 @@
 // As
 //  - Topic: *amqp.Connection
 //  - Subscription: *amqp.Connection
+//  - Error: *amqp.Error
```
vangent: and ampq.MultiError.

jba:
PTAL @ijt

jba:
@ijt ping

Issue #2, opened 2018-05-11
blob: no way to distinguish non-existent blob


zombiezen:
For both Delete and NewReader, there's no way to distinguish blob-not-found from a normal error. This is useful for serving 404 instead of 500.

/cc @shantuo

Issue #691, opened 2018-11-13
pubsub: dynamically adjust number of messages fetched by Receive


ijt:
It could be that multiple machines are needed to handle the messages coming in on some subscription. It would be best if the concrete API would dynamically adjust the max number of messages being pulled in ReceiveBatch, which would then have to take that as a parameter.

vangent:
We may also want to proactively call `ReceiveBatch` before there are 0 messages left.

We may also want to call `ReceiveBatch` concurrently (today, the driver interface says it must be callable concurrently, but we never do).

jba:
Here is one way to pick the maxMessages value to pass to `ReceiveBatch`.

We want to keep a few messages in memory at all times, so callers of `Receive` can get a message right away. But we don't want to keep too many around: if it takes too long to start processing the messages, they might expire (their ack deadline might be exceeded); and there might be other subscribers who could be handling those waiting messages.

So we want to aim for the in-memory queue of messages to be a certain size. What is a good size? If you think in terms of number of messages, that is hard question. Is 10 a good size? If messages are processed in a millisecond, it's too small—those ten messages will probably be consumed before an RPC to get more will return. If it takes two minutes to process a message, then 10 is too large—the last message will be sitting around for 18 minutes, which feels like along time (and exceeds the maximum ack deadline on Google Cloud Pub/Sub).

It makes more sense to measure the size of the queue in *time*. A queue that is 2 seconds long will give the next RPC 2 seconds to refill it, will make messages wait at most 2 seconds before being processed, and will result in only 2 seconds of lost work in a crash.

To measure the queue in time, we need to know how long it takes to process one message. So we keep track of the average time it takes to do so. Since the value can change over time, we keep a running average over some interval, say one minute.

We could measure processing time as the interval between the return of `Receive` and the call of `Ack`.  But not every message will be acked. It is perfectly reasonable for a subscriber to nack (or fail to ack) a significant fraction of the messages it receives. Processing time for those unacked messages should still be included in the average.

So we should measure processing time from the return of `Receive` until the next call to `Receive`. That calculation will be confused by concurrent calls to `Receive`, but only in the short term. In the longer term, it should provide a pretty good sense of the average processing time. 

With the average processing time in hand, and a desired queue length in time, we can calculate how many messages to ask from `ReceiveBatch` when the queue is empty:
```
messagesToRequest = desiredQueueLengthInTime / processTimePerMessage
```
[...]

PR #1200, opened 2019-01-27
pubsub: dynamically choose the number of messages for ReceiveBatch

jba:
To decide how many messages to pull at a time, we aim for the
in-memory queue of messages to be a certain size. That gives us a
buffer of messages to draw from, ensuring high throughput, without
pulling so many messages that the unconsumed ones languish.

We measure the size by time instead of message count. Time is more
relevant, because ack deadlines are expressed in time, and it's easier
to think about lost work (in the event of a crash) in terms of time
lost rather than messages lost.

We keep track of the average time it takes to process a message. Then
we can convert a queue size in time to a number of messages.

We compute processing time by measuring the time between when Receive
returns and when it is next called. Although this is incorrect in the
short term, because multiple goroutines may call Receive at the same time,
in the long run it is accurate enough.

We rejected the obvious alternative, measuring time from Receive to
Ack, because not every message will be acked. It is perfectly
reasonable for a subscriber to nack (or fail to ack) a significant
fraction of the messages it receives, but processing time for those
unacked messages should still be included in the calculation of how
many messages to pull.

This change significantly improves the Receive benchmark -- messages
per second is more than quadrupled. But there is more work to do. We
should pre-emptively pull messages when the queue size gets low, and
we should issue multiple ReceiveBatch calls concurrently.

Besides performance, this change also improves behavior over current
master at very low processing rates. Currently we pull a constant 10
messages per ReceiveBatch. If it takes a long time to process one
message, then the other 9 will sit in RAM and may expire. With this
change, we will pull just one message at a time if need be.

Addresses #691.


pubsub/averager.go
```diff
@@ -0,0 +1,81 @@
+// Copyright 2019 The Go Cloud Authors
```
vangent: "Go Cloud Development Kit Authors" please.

pubsub/averager.go
```diff
@@ -0,0 +1,81 @@
...
+// limitations under the License.
+
+package pubsub
+
+import (
+	"sync"
+	"time"
+)
+
+// An averager keeps track of an average value over a time interval.
```
vangent: Why not

https://github.com/VividCortex/ewma ?

pubsub/averager_test.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2019 The Go Cloud Authors
```
vangent: Same.

pubsub/pubsub.go
```diff
@@ -255,12 +286,31 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
 		}
 		// No messages are available and there are no calls to ReceiveBatch in flight.
 		// Make a call.
+		//
+		// Ask for a number of messages that will give us the desired queue length.
+		// Unless we don't have information about process time (at the beginning), in
+		// which case just get one message.
+		nMessages := 1
```
vangent: Why 1? Seems like a very conservative estimate. It "guesses" that it will take 2 seconds to process a message.

pubsub/pubsub.go
```diff
@@ -255,12 +286,31 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
...
+		if !math.IsNaN(avgProcessTime) {
+			// Using Ceil guarantees at least one message.
+			n := math.Ceil(desiredQueueLength.Seconds() / avgProcessTime)
+			// Cap nMessages at some non-ridiculous value.
+			// Slight hack: we should be using a larger cap, like MaxInt32. But
+			// that messes up replay: since the tests take very little time to ack,
+			// n is very large, and since our averaging process is time-sensitive,
+			// values can differ slightly from run to run. The current cap happens
+			// to work, but we should come up with a more robust solution.
+			// (Currently it doesn't matter for performance, because gcppubsub can't
```
vangent: Comment sentence is fragmented ("can't caps").

pubsub/pubsub.go
```diff
@@ -255,12 +286,31 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
...
+			// caps to 1000 anyway.)
```
vangent: 1000 seems like a reasonable cap to me. How hard is it going to be to fetch multiple batches in parallel?

vangent: (a smaller cap like 1000 will be more reasonable if we can do multiple in parallel)

pubsub/pubsub.go
```diff
@@ -255,12 +286,31 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
...
+			nMessages = int(math.Min(n, 10000))
+		}
 		s.waitc = make(chan struct{})
 		s.mu.Unlock()
 		// Even though the mutex is unlocked, only one goroutine can be here.
 		// The only way here is if s.waitc was nil. This goroutine just set
 		// s.waitc to non-nil while holding the lock.
-		msgs, err := s.getNextBatch(ctx)
+		msgs, err := s.getNextBatch(ctx, nMessages)
```
vangent: Test is failing:

```
 drivertest.go:271: pubsub (code=Unknown): replayer: request not found: subscription:"projects/go-cloud-test-216917/subscriptions/TestConformance_TestSendReceiveTwo-subscription-1" max_messages:936
```

That looks likely to be flaky....

pubsub/averager.go
jba: Well, I didn't know about it. (How did you find it?) It looks much more sophisticated than what I have, which is nice. But it says

> Current implementations assume an implicit time interval of 1.0 between every sample added. That is, the passage of time is treated as though it's the same as the arrival of samples. If you need time-based decay when samples are not arriving precisely at set intervals, then this package will not support your needs at present.

which means we can't use it.

But you raise the question whether time-based decay is what we want. I thought about implementing it, but decided the complexity wasn't worth it. I'm not sure it's going to matter much whether the contribution of 1-minute-old points is reduced or not. It's more important that their contribution eventually goes to zero, which my implementation achieves rather abruptly.

pubsub/averager.go
jba: Done.

pubsub/averager_test.go
jba: Done.

pubsub/averager.go
vangent: I found it like this:
https://www.google.com/search?q=golang+moving+average

pubsub/averager.go
```diff
@@ -0,0 +1,81 @@
...
+type averager struct {
+	mu             sync.Mutex
+	buckets        []bucket // always nBuckets of these, last is most recent
+	bucketInterval time.Duration
+	end            time.Time // latest time we can currently record (end of the last bucket's interval)
+}
+
+type bucket struct {
+	count float64 // number of points in the bucket
+	avg   float64 // average value of the bucket
```
vangent: Are you keeping `avg` instead of `total` to avoid overflow? You seem to always multiply `avg` by `count`. If so, add a comment?

pubsub/pubsub.go
jba: 2 or more seconds. Say it takes three minutes to process a message. Then we never want to fetch more than one at a time. If we started by fetching two, then the second would lie around for a while and might expire. 

I think it's better to be conservative at the beginning. Applications with long processing times will be happy that their messages don't expire. Everyone else will ramp up very quickly (since after that first message is serviced in 10ms, the next ReceiveBatch call will ask for 200 messages).

pubsub/averager.go
vangent: >  which means we can't use it.

Is that true? I am convinced by your argument about using the predicted time-to-process of the queue to determine the batch size, but I'm not sure that that implies that we need to use time-based bucketing for the estimate of how long each message takes to process.

I'm not sure that that library will do better than what you have here, but it's worth thinking about. Are there scenarios where exponential decay will do the wrong thing? (probably).

What about an app that gets bursts of message periodically (e.g., every 15m it gets 1000 messages). IIUC, your impl will start at a constant (currently 1) for each burst and ramp up (because after 1m it forgets all about previous history). The decaying moving average wouldn't forget anything during the idle 15m. Which is better? (not obvious, TBH).

pubsub/pubsub.go
jba: Done.

pubsub/pubsub.go
jba: Not sure how hard the coding will be yet. I have an idea of how to think about it, though. But I want this to land first.

pubsub/averager.go
jba: Just being stupid. Rewrote.

pubsub/averager.go
jba: I see your point. It seemed obvious to me that we wanted the number to decay in proportion to elapsed time, but you're right, there's no strong reason for that (aside from a hand-wavy argument about things generally having some time locality). 

`ewma` will forget in proportion to the number of messages processed, rather than time. Maybe that's fine. If it took you 100ms to process a message three days ago, why wouldn't it take the same time now? On the other hand, system behavior tends to be spiky. You want to ride the spikes when they happen, then quickly forget about them. But I'm really just speculating here. 

I'm fine doing any of the following:
1. Keeping my code.
2. Switching to `ewma`
3. Doing a better job of time-based moving average. I think doing it exactly requires saving every point, but we could combine bucketing with a decay factor.

In any case, we should add this discussion to the issue, or a new issue.

pubsub/averager.go
```diff
@@ -49,10 +49,10 @@ func newAverager(dur time.Duration, nBuckets int) *averager {
 func (a *averager) average() float64 {
```
vangent: Nit: consider making this `Average` since it's an "external" function? (even though the whole struct isn't). Not sure what typical Go style for this is, but I think it can be useful to be clear about what the expected API of internal objects is even if it's not enforced.

This code could also be moved into the `internal/batcher` package.

pubsub/pubsub.go
vangent: `9364`? Why? (why isn't this flaky?) Seems like it's dependent on timing, no?

pubsub/averager.go
vangent: Let's discuss at stand-up tomorrow.

pubsub/pubsub.go
jba: I got the number 9364 from the actual travis log. I think you dropped a digit when you copied it. The relevant point is that it's > 1000.

It is time-dependent. But with a cap of 1000, it would have to take longer than 1ms between calls to Receive on average, and considering that we do nothing but call Ack, that's extremely unlikely.

I agree, though, that this solution isn't great.

pubsub/averager.go
jba: Removed the whole thing.

pubsub/pubsub.go
```diff
@@ -211,12 +213,38 @@ type Subscription struct {
...
+	// In other words, if it takes 100ms to process a message on average, and we want
+	// 2s worth of queued messages, then we need 2/.1 = 20 messages.
+	//
+	// If desiredQueueLength is too small, then there won't be a large enough buffer
+	// of messages to handle fluctuations in processing time, and the queue is likely
+	// to become empty, reducing throughput. If desiredQueueLength is too large, then
+	// messages will wait in memory for a long time, possibly timing out (that is,
+	// their ack deadline will be exceeded). Those messages could have been handled
+	// by another process receiving from the same subscription.
+	desiredQueueLength = 2 * time.Second
```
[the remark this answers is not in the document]
jba: Are you okay with desiredQueueDuration?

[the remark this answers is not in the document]
ijt: Sure, as long as the comment explains what it means. I think it's how long we want a message to spend in the queue, although I find this idea a bit strange since I don't have a preference about how long messages stay in the queue so long as it's a lot less than the ack deadline.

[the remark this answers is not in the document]
jba: > although I find this idea a bit strange

Exactly, that's why I don't think that's the right way to think about it. It's more like, we want to keep a few messages around in case Receive speeds up and starts chewing through them faster than it has been. In other words, a buffer. How many messages do we want to keep in the buffer? No, that's the wrong question: how much runway (in time) do we want before we run out of messages?

Yes, if Receive always takes the same time, then a message will spend all that time in the queue. But that's a side effect, the price we pay for having messages available on demand to improve throughput.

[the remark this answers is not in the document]
ijt: I see, so to write it in Javanese, it would be more like `desiredAmountOfTimeBeforeQueueProbablyRunsOut`. Makes sense.

vangent:
There's a lot going on here (not just on this PR), and we need a path forward.

1.  How to estimate "time to process" and "time to fetch"? Time-bucket-based moving average, decaying average, pick a constant, ....
2.  For "time to process" in particular, which timepoints to use (only `Receive`, or `Receive` -> `Ack`).
3.  When to fetch (wait till buffer is empty, predicted-buffer-empty-time minus "time to fetch" estimate, that plus some padding, multiple concurrent fetches makes it even more complicated).
4.  How many to fetch.

We don't have to solve all of these at once. Let's start with simple changes that make big improvements and see how far that gets us. I propose:

1.  Something like this PR that uses a simple estimate for "time to process" to decide how many to fetch. I do think we need to resolve #2 above (which timepoints to use), and I think I agree with @ijt that using `Receive` + `Ack` is simpler.

2.  Next step IMHO would be pre-fetching (#3) with a constant padding (e.g., 1s before we predict our buffer will be empty, make the RPC).

3. After that we can try different models for "time to process", try modeling "time to fetch" instead of using a constant, and making multiple fetches concurrently.

Thoughts?

jba:
Totally agree.

Let me fix the name of the constant as @ijt requested, and change the measurement to end at ack. Then I think this is a workable and useful PR.

jba:
PTAL.

pubsub/pubsub.go
```diff
@@ -314,15 +364,24 @@ func (s *Subscription) getNextBatch(ctx context.Context) ([]*Message, error) {
...
-		})
+		}
+		m2.ack = func() {
+			t := float64(time.Since(m2.processingStartTime).Seconds())
+			// Note: m2's mutex is locked here as well. Deadlock will
+			// result if Message.Ack is ever called with s.mu held. That currently
+			// cannot happen, but we should be careful if/when implementing features
+			// like auto-ack.
+			s.mu.Lock()
+			s.avgProcessTime = s.avgProcessTime*(1-decay) + t*decay
```
vangent: Consider putting this in a function on `s`?

vangent: Won't this understimate at startup? E.g., if the first message takes 100ms to start, `avgProcessTime` is 0, and you'll set it to `0* (1-0.05) + .1 * 0.05 = .005` instead of `.1`.

Maybe if `s.avgProcessTime == 0` it should just take `t`?

jba: Done.

jba: Done.

jba:
PTAL.

vangent:
Separately, it might be worth writing a fake `Subscription` driver implementation that can simulate timing for getting batches of messages. E.g., give it a minimum RPC duration, an incremental duration-per-message-requested, a distribution, etc. Then we could run benchmarks against it at various settings.

Similarly for a fake Receiver that simulates how long it takes to process messages.

PR #1202, opened 2019-01-28
runtime, blob: make ErrorAs panic on nil or non-pointer target

jba:
Fixes # 1201.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -67,6 +67,7 @@ import (
 	"mime"
 	"net/http"
 	"net/url"
+	"reflect"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


blob/blob.go
```diff
@@ -420,6 +421,9 @@ func (b *Bucket) As(i interface{}) bool {
 // ErrorAs converts i to provider-specific types.
 // See Bucket.As for more details.
```
vangent: Add the comment here about it panicking if i is nil or not a pointer.

runtimevar/runtimevar.go
```diff
@@ -180,9 +180,12 @@ func (w *wrappedError) Error() string {
 // ErrorAs converts i to provider-specific types.
 // See Snapshot.As for more details.
 func ErrorAs(err error, i interface{}) bool {
```
vangent: And here.

blob/blob.go
jba: Done.

runtimevar/runtimevar.go
jba: Done.

PR #1203, opened 2019-01-29
secrets: use gcerrors

### LINKED ISSUE #1124: all: handle errors in a common way

jba:
Fixes #1124.

### DIFF
--- a/secrets/awskms/kms.go
+++ b/secrets/awskms/kms.go
@@ -79,7 +79,7 @@ func (k *keeper) Encrypt(ctx context.Context, plaintext []byte) ([]byte, error) 
 
 // ErrorCode implements driver.ErrorCode.
 func (k *keeper) ErrorCode(error) gcerrors.ErrorCode {
-	// TODO(shantuo): try to classify vault error codes
+	// TODO(shantuo): try to classify aws error codes
 	return gcerrors.Unknown
 }
 
--- a/secrets/awskms/kms.go
+++ b/secrets/awskms/kms.go
@@ -23,6 +23,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/service/kms"
+	"gocloud.dev/gcerrors"
 	"gocloud.dev/secrets"
 )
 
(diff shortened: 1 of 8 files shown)


secrets/awskms/kms.go
```diff
@@ -76,6 +77,12 @@ func (k *keeper) Encrypt(ctx context.Context, plaintext []byte) ([]byte, error)
 	return result.CiphertextBlob, nil
 }
 
+// ErrorCode implements driver.ErrorCode.
+func (k *keeper) ErrorCode(error) gcerrors.ErrorCode {
+	// TODO(shantuo): try to classify vault error codes
```
vangent: Nit: AWS KMS error codes, not vault.

secrets/localsecrets/localsecrets_test.go
```diff
@@ -25,7 +25,7 @@ import (
 type harness struct{}
 
 func (h *harness) MakeDriver(ctx context.Context) (driver.Keeper, error) {
-	return NewKeeper(ByteKey("very secret secret")), nil
+	return &keeper{secretKey: ByteKey("very secret secret")}, nil
```
shantuo: Let's keep using NewKeeper so that it got covered by tests.

secrets/vault/vault.go
```diff
@@ -95,6 +96,12 @@ func (k *keeper) Encrypt(ctx context.Context, plaintext []byte) ([]byte, error)
 	return []byte(secret.Data["ciphertext"].(string)), nil
 }
 
+// ErrorCode implements driver.ErrorCode.
+func (k *keeper) ErrorCode(error) gcerrors.ErrorCode {
+	// TODO(shantuo): try to classify vault error codes
```
shantuo: Filed #1206

secrets/localsecrets/localsecrets_test.go
jba: We can't: `NewKeeper` returns a `*secrets.Keeper`, and `MakeDriver` needs to return a `driver.Keeper`. (It worked before because the concrete and driver interfaces happened to be the same.)

Issue #1162, opened 2019-01-23
secrets: get provider code-coverage up to 85%


vangent:
Most of the uncovered lines (https://coveralls.io/github/google/go-cloud) look like they are things like public constructors and `Dial`. You can probably get coverage for these by making the examples run; see `runtimevar` for examples.

shantuo:
- [ ] Portable API
- [ ] GCP
- [ ] AWS
- [ ] Vault
- [ ] local

PR #1205, opened 2019-01-29
secrets: improve test coverages by testing fail connections

shantuo:
Updates: #1162

### DIFF
--- a/secrets/awskms/kms_test.go
+++ b/secrets/awskms/kms_test.go
@@ -37,7 +37,10 @@ type harness struct {
 }
 
 func (h *harness) MakeDriver(ctx context.Context) (driver.Keeper, error) {
-	return NewKeeper(h.client, keyID, nil), nil
+	return &keeper{
+		keyID:  keyID,
+		client: h.client,
+	}, nil
 }
 
 func (h *harness) Close() {
--- a/secrets/awskms/kms_test.go
+++ b/secrets/awskms/kms_test.go
@@ -58,11 +58,13 @@ func TestConformance(t *testing.T) {
 
 // KMS-specific tests.
 
-func TestNoConnectionError(t *testing.T) {
+func TestNoSessionProvidedError(t *testing.T) {
 	if _, err := Dial(nil); err == nil {
 		t.Error("got nil, want no AWS session provided")
 	}
+}
 
+func TestNoConnectionError(t *testing.T) {
 	os.Setenv("AWS_ACCESS_KEY", "myaccesskey")
 	os.Setenv("AWS_SECRET_KEY", "mysecretkey")
 	os.Setenv("AWS_REGION", "us-east-1")
@@ -70,12 +72,14 @@ func TestNoConnectionError(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	ctx := context.Background()
+
 	client, err := Dial(sess)
-	plaintext := []byte("test")
+	if err != nil {
+		t.Fatal(err)
+	}
 	keeper := NewKeeper(client, keyID, nil)
 
-	if _, err := keeper.Encrypt(ctx, plaintext); err == nil {
+	if _, err := keeper.Encrypt(context.Background(), []byte("test")); err == nil {
 		t.Error("got nil, want UnrecognizedClientException")
 	}
 }
--- a/secrets/awskms/kms_test.go
+++ b/secrets/awskms/kms_test.go
@@ -16,11 +16,12 @@ package awskms
 
 import (
 	"context"
+	"os"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


shantuo:
The only exception is localsecrets (84%), which has two lines of return err's not tested. I think we might be ok with those?

secrets/awskms/kms_test.go
```diff
@@ -57,3 +55,27 @@ func newHarness(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
 func TestConformance(t *testing.T) {
 	drivertest.RunConformanceTests(t, newHarness)
 }
+
+// KMS-specific tests.
+
+func TestNoConnectionError(t *testing.T) {
+	if _, err := Dial(nil); err == nil {
```
vangent: Nit: this could be a separate test.

secrets/awskms/kms_test.go
```diff
@@ -57,3 +55,27 @@ func newHarness(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
...
+
+	os.Setenv("AWS_ACCESS_KEY", "myaccesskey")
+	os.Setenv("AWS_SECRET_KEY", "mysecretkey")
+	os.Setenv("AWS_REGION", "us-east-1")
+	sess, err := session.NewSession()
+	if err != nil {
+		t.Fatal(err)
+	}
+	ctx := context.Background()
+	client, err := Dial(sess)
```
vangent: Should check err here.

secrets/awskms/kms_test.go
```diff
@@ -57,3 +55,27 @@ func newHarness(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
...
+	}
+
+	os.Setenv("AWS_ACCESS_KEY", "myaccesskey")
+	os.Setenv("AWS_SECRET_KEY", "mysecretkey")
+	os.Setenv("AWS_REGION", "us-east-1")
+	sess, err := session.NewSession()
+	if err != nil {
+		t.Fatal(err)
+	}
+	ctx := context.Background()
```
vangent: Nit: `ctx` and `plaintext` are only used once, so just use them inline, it's easier to read.

secrets/gcpkms/kms_test.go
```diff
@@ -76,3 +74,27 @@ func newHarness(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
 func TestConformance(t *testing.T) {
 	drivertest.RunConformanceTests(t, newHarness)
 }
+
+// KMS-specific tests.
+
```
vangent: Same comments here as above.

secrets/vault/vault_test.go
```diff
@@ -89,3 +82,32 @@ func testVaultServer(t *testing.T) (*api.Client, func()) {
 func TestConformance(t *testing.T) {
 	drivertest.RunConformanceTests(t, newHarness)
 }
+
+// Vault-specific tests.
+
+func TestNoConnectionError(t *testing.T) {
```
vangent: Same as above.

vangent:
> The only exception is localsecrets (84%), which has two lines of return err's not tested. I think we might be ok with those?

The second one looks like it could be covered by a conformance test: try to decrypt something that wasn't encrypted with the key; it should fail. This can be a separate PR though.

secrets/awskms/kms_test.go
shantuo: Done.

secrets/awskms/kms_test.go
shantuo: Done.

secrets/awskms/kms_test.go
shantuo: Done.

secrets/gcpkms/kms_test.go
shantuo: Done.

secrets/vault/vault_test.go

shantuo:
I need to update some tests after jba's PR.

shantuo:
> > The only exception is localsecrets (84%), which has two lines of return err's not tested. I think we might be ok with those?
> 
> The second one looks like it could be covered by a conformance test: try to decrypt something that wasn't encrypted with the key; it should fail. This can be a separate PR though.

Except AWS's decryption relies on the key information inside the encryption body, and they don't take into a key in decryption. This is not conformance across all providers.

vangent:
> Except AWS's decryption relies on the key information inside the encryption body, and they don't take into a key in decryption. This is not conformance across all providers.

Per discussion, let's add a conformance test that calls `Decrypt` with some random bytes (e.g., `[]byte("hello world")`). It should fail on all providers, even AWS.

PR #1208, opened 2019-01-29
runtimevar: use gcerrors

### LINKED ISSUE #1124: all: handle errors in a common way

jba:
Fixes #1124.

### DIFF
--- a/runtimevar/blobvar/blobvar.go
+++ b/runtimevar/blobvar/blobvar.go
@@ -145,7 +145,11 @@ func (w *watcher) ErrorAs(err error, i interface{}) bool {
 	return w.bucket.ErrorAs(err, i)
 }
 
-// IsNotExist implements driver.IsNotExist.
-func (*watcher) IsNotExist(err error) bool {
-	return gcerrors.Code(err) == gcerrors.NotFound
+// ErrorCode implements driver.ErrorCode.
+func (*watcher) ErrorCode(err error) gcerrors.ErrorCode {
+	// err might have come from blob, in which case use its code.
+	if c := gcerrors.Code(err); c != gcerrors.Unknown {
+		return c
+	}
+	return gcerrors.Unknown
 }
--- a/runtimevar/blobvar/blobvar.go
+++ b/runtimevar/blobvar/blobvar.go
@@ -148,8 +148,5 @@ func (w *watcher) ErrorAs(err error, i interface{}) bool {
 // ErrorCode implements driver.ErrorCode.
 func (*watcher) ErrorCode(err error) gcerrors.ErrorCode {
 	// err might have come from blob, in which case use its code.
-	if c := gcerrors.Code(err); c != gcerrors.Unknown {
-		return c
-	}
-	return gcerrors.Unknown
+	return gcerrors.Code(err)
 }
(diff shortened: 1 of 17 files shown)


runtimevar/blobvar/blobvar.go
```diff
@@ -145,7 +145,11 @@ func (w *watcher) ErrorAs(err error, i interface{}) bool {
 	return w.bucket.ErrorAs(err, i)
 }
 
-// IsNotExist implements driver.IsNotExist.
-func (*watcher) IsNotExist(err error) bool {
-	return gcerrors.Code(err) == gcerrors.NotFound
+// ErrorCode implements driver.ErrorCode.
+func (*watcher) ErrorCode(err error) gcerrors.ErrorCode {
+	// err might have come from blob, in which case use its code.
+	if c := gcerrors.Code(err); c != gcerrors.Unknown {
```
vangent: Nit: this is just `return gcerrors.Code(err)`, isn't it?

runtimevar/blobvar/blobvar_test.go
```diff
@@ -83,7 +83,7 @@ func (verifyAs) SnapshotCheck(s *runtimevar.Snapshot) error {
 
 func (verifyAs) ErrorCheck(w driver.Watcher, err error) error {
 	var perr *os.PathError
-	if !runtimevar.ErrorAs(err, &perr) {
+	if !runtimevar.New(w).ErrorAs(err, &perr) {
```
vangent: It would probably be better to change `ErrorCheck` to take the concrete type instead of the driver.

runtimevar/constantvar/constantvar.go
```diff
@@ -26,13 +26,14 @@ import (
 	"errors"
 	"time"
 
+	"gocloud.dev/gcerrors"
 	"gocloud.dev/runtimevar"
 	"gocloud.dev/runtimevar/driver"
 )
 
-// ErrNotExist is a sentinel error that can be used with NewError to return an
+// errNotExist is a sentinel error that can be used with NewError to return an
```
vangent: Comment is stale; maybe just delete it as it's private now.

runtimevar/blobvar/blobvar_test.go
jba: Done.

runtimevar/constantvar/constantvar.go
jba: Done.

jba:
PTAL

Issue #1174, opened 2019-01-24
all: add open-by-name functions to all APIs


zombiezen:
(This grew out of #931 and is now a larger effort, so opening a bigger-scoped issue.)

A very common use case when developing a portable application is to want to use a local implementation for development and then use the real implementation in production. Commonly, developers don't want to have two different build configurations to do this: they just want to be able to change by command-line flag. Right now, this is difficult to do: every application using Go Cloud would have to develop their own mechanism for specifying which provider resource they want to use. (This would be similar to SQL connection strings, but I think we want to keep credentials explicitly passed.)

I'm going to iterate on the blob URL-opening API, and once that gets to a place we're happy with, then I'd like to duplicate the API for our other portable APIs (pub/sub, secrets, runtimevar, and SQL).

ijt:
Can you give a concrete example of what this would look like in a particular case?

zombiezen:
I do not have a design yet, but #1145 was one iteration of this idea. @ijt, can you be more specific about what information you're aiming to glean?

ijt:
Right, #1145 looks like a useful example of this.

ijt:
I didn't have anything specific in mind, just wanted to see an example.

jba:
I'm against doing this now, for several reasons:
- It's just a convenience feature; it adds no new functionality.
- No one has asked for it. The only known use is in Hugo, and that is just for blob. (I don't object to blob URLs.)
- We'll spend a lot of time figuring out how to shoehorn everything into strings, avoiding global registries, and other design issues. (Evidence: we've already done so.) Ross, I think your time is better spent on Go Cloud Init, which has a much higher potential to get us users.
- It will be a never-ending drag on our productivity. If we do it, then every time we add an option to any provider, we will have to figure out how to destringify it and what the flag/query-param name for it will be, and we'll have to hook it up.

zombiezen:
Unavailability of this feature blocks users who want to create a [Twelve-Factor Application](https://12factor.net/) (specifically [IV. Backing Services](https://12factor.net/backing-services)). They would need to roll their own solution, which I think would make supporting infrastructure around Go CDK-based applications hard to work with. I think Go CDK having an opinion helps here.

vangent:
Per discussions, @zombiezen is doing this for `blob`, and @shantuo will do it for `runtimevar`, `secrets`, `pubsub`.

PR #1209, opened 2019-01-29
internal/docs: add section about URLs

zombiezen:
Updates #931
Updates #1174

### DIFF
--- a/internal/docs/design.md
+++ b/internal/docs/design.md
@@ -105,13 +105,16 @@ https://godoc.org/github.com/google/go-cloud/runtimevar#Variable
 https://godoc.org/github.com/google/go-cloud/blob#Bucket.NewWriter
 [`database/sql`]: https://godoc.org/database/sql
 
-## No Global State
+## Minimize Global State
 
 As a library, the Go CDK should not introduce global state. Global state is
 difficult to reason about in large codebases, where it can be necessary for
 different parts of the application to use different states. Instead of adding
 global state, push responsibility to the application to inject the state where
 it is needed.
+
+The exception we permit is URL scheme registration as documented under
+[URLs][#urls].
 
 ## Package Naming Conventions
 
@@ -183,6 +186,100 @@ with `foo.New(...)` and later add `foo.NewWithOptions(..., opts *Options)` if
 needed). However, this would result in inconsistent names over time (e.g., some
 packages would expose `New` with an `Options`, while others would expose
 `NewWithOptions`).
+
+### URLs
+
+To enable the [Backing services factor][] of a Twelve-Factor Application, Go
+Cloud includes the ability to construct each of its API objects using
+identifying URLs. The concrete type should have API like this:
+
+```go
+package foo
+
+// A type that implements WidgetURLOpener can open widgets based on a URL.
+// The opener must not modify the URL argument. OpenWidgetURL must be safe to
+// call from multiple goroutines.
+//
+// For maximum flexibility, WidgetURLOpeners should not assume that the URL
+// has a particular scheme.
+type WidgetURLOpener interface {
+  OpenWidgetURL(ctx context.Context, u *url.URL) (*Widget, error)
+}
+
+// A type that implements NamedWidgetURLOpener has a preferred scheme for
+// the URLs it opens.
+type NamedWidgetURLOpener interface {
+  WidgetURLOpener
+  WidgetURLScheme() string
+}
+
+// URLMux is a URL opener multiplexer. It matches the scheme of the URLs
+// against a set of registered schemes and calls the opener that matches the
+// URL's scheme.
+//
+// The zero value is a URLMux with no registered schemes.
+type URLMux struct {
+  // ...
+}
+
+// AddWidget registers the opener with its preferred scheme. If an opener
+// already exists for the scheme, AddWidget panics.
+func (mux *URLMux) AddWidget(opener NamedWidgetURLOpener) {
+  mux.RegisterWidget(opener.WidgetURLScheme(), opener)
+}
+
+// RegisterWidget registers the opener with the given scheme. If an opener
+// already exists for the scheme, RegisterWidget panics.
+func (mux *URLMux) RegisterWidget(scheme string, opener WidgetURL) {
+  // ...
+}
+
+// OpenWidget calls OpenWidgetURL with the URL parsed from urlstr.
+func (mux *URLMux) OpenWidget(ctx context.Context, urlstr string) (*Widget, error) {
+  u, err := url.Parse(urlstr)
+  if err != nil {
+    return nil, fmt.Errorf("open widget: %v", err)
+  }
+  return mux.OpenWidgetURL(ctx, u)
+}
+
+// OpenWidgetURL dispatches the URL to the opener that is registered with the
+// URL's scheme.
+func (mux *URLMux) OpenWidgetURL(ctx context.Context, u *url.URL) (*Widget, error) {
+  // ...
+}
+
+// DefaultURLMux returns the URLMux used by OpenWidget.
+func DefaultURLMux() *URLMux {
+  return defaultURLMux
+}
+
+var defaultURLMux = new(URLMux)
+
+// OpenWidget opens the Widget identified by the URL given. URL openers must be
+// registered in the DefaultURLMux, which is typically done in driver
+// packages' initialization.
+func OpenWidget(ctx context.Context, urlstr string) (*Widget, error) {
+  return DefaultURLMux().OpenWidget(urlstr)
+}
+```
+
+The repetition of `Widget` in the method names permits a type to handle multiple
+resources within the API. Exporting the `URLMux` allows applications to build
+their own muxes, potentially wrapping existing.
+
+Driver packages should include their own `URLOpener` struct type which
+implements all the relevant `WidgetURLOpener` methods. The URL should only serve
+to identify which resource to open. Any credentials or other complex values
+should be taken in as struct fields, not as input from URL. If the driver
+package registers its `URLOpener` with the `DefaultURLMux`, then it should
+populate these complex fields from environment variables. If doing so is
+undesirable or expensive, then it should not register with the `DefaultURLMux`
+and instead rely on users to create their own mux. If there already exists a
+well-established URI format for the backend (like S3 URLs or database connection
+URIs), then drivers should honor them where possible.
+
+[Backing services factor]: https://12factor.net/backing-services
 
 ## Errors
 


internal/docs/design.md
```diff
@@ -184,6 +187,102 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
+// For maximum flexibility, WidgetURLOpeners should not assume that the URL
+// has a particular scheme.
+type WidgetURLOpener interface {
+  OpenWidgetURL(ctx context.Context, u *url.URL) (*Widget, error)
+}
+
+// A type that implements NamedWidgetURLOpener has a preferred scheme for
+// the URLs it opens. WidgetURLScheme must be safe to call from multiple
+// goroutines.
+type NamedWidgetURLOpener interface {
```
jba: Why do you need the extra complexity of the preferred scheme (basically, this type, and AddWidget method)?

internal/docs/design.md
```diff
@@ -184,6 +187,102 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
+  WidgetURLOpener
+  WidgetURLScheme() string
```
vangent: Why not a slice of string? IIUC, this is basically allowing provider impls to provide default scheme(s), why not let the default be more than one?

internal/docs/design.md
```diff
@@ -184,6 +187,102 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
+func OpenWidget(ctx context.Context, urlstr string) (*Widget, error) {
+  return DefaultURLMux().OpenWidget(urlstr)
+}
+```
+
+The repetition of `Widget` in the method names permits a type to handle multiple
+resources within the API. Exporting the `URLMux` allows applications to build
+their own muxes, potentially wrapping existing.
+
+Driver packages should include their own `URLOpener` struct type which
```
vangent: "Driver packages" -- you mean provider implementations of driver, right? Might be good to clarify (it's not the actual `driver` package).

internal/docs/design.md
```diff
@@ -184,6 +187,102 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
+
+// OpenWidget opens the Widget identified by the URL given. URL openers must be
+// registered in the DefaultURLMux, which is typically done in driver
+// packages' initialization.
+func OpenWidget(ctx context.Context, urlstr string) (*Widget, error) {
+  return DefaultURLMux().OpenWidget(urlstr)
+}
+```
+
+The repetition of `Widget` in the method names permits a type to handle multiple
```
vangent: Ah, this is relevant to my comment above. WDYT about only including this if the API supports opening for more than one resource?

internal/docs/design.md
vangent: Can you given an example of what a driver implementation would look like? (including an example of setting up the struct to allow credentials to be passed in, as well as the `init` function registering a default implementation of URLOpener for scheme(s)).

internal/docs/design.md
vangent: I'm not sure I get the need for this extra interface this either. E.g., why not have each provider impl expose its default scheme(s) explicitly? The code setting up a `URLMux` can use them or not, so there's no loss of control. It might help me to understand when you add an example of what a provider impl would look like.

internal/docs/design.md
```diff
@@ -105,14 +105,17 @@ https://godoc.org/github.com/google/go-cloud/runtimevar#Variable
...
-## No Global State
+## Minimize Global State
 
 As a library, the Go CDK should not introduce global state. Global state is
 difficult to reason about in large codebases, where it can be necessary for
 different parts of the application to use different states. Instead of adding
 global state, push responsibility to the application to inject the state where
 it is needed.
 
+The exception we permit is URL scheme registration as documented under
```
ijt: Why do we permit it?

internal/docs/design.md
```diff
@@ -184,6 +187,103 @@ needed). However, this would result in inconsistent names over time (e.g., some
 packages would expose `New` with an `Options`, while others would expose
 `NewWithOptions`).
 
+### URLs
+
+To enable the [Backing services factor][] of a Twelve-Factor Application, Go
```
ijt: Does the Go CDK have a policy of adhering to Twelve-Factor or some subset of it?

internal/docs/design.md
ijt: I agree with @vangent on this. Prevent stuttering.

internal/docs/design.md
```diff
@@ -184,6 +187,103 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
+  return mux.OpenWidgetURL(ctx, u)
+}
+
+// OpenWidgetURL dispatches the URL to the opener that is registered with the
+// URL's scheme. OpenWidget is safe to call from multiple goroutines.
+func (mux *URLMux) OpenWidgetURL(ctx context.Context, u *url.URL) (*Widget, error) {
+  // ...
+}
+
+// DefaultURLMux returns the URLMux used by OpenWidget.
```
vangent: Do you think we'd want `SetDefaultURLMux`?

E.g., if the user creates their own mux, but doesn't want to have to pass it around everywhere.

internal/docs/design.md
vangent: I am envisioning something like this (names are fungible):

*  the concrete type exposes a way to register an `URLOpener` to a scheme -- your `RegisterWidget` -- and nothing else about schemes/naming.
*  each provider impl has a public function `RegisterOpener` that takes a mux and an `URLOpener` and registers the opener for the default scheme(s).
*  each provider impl has a package `init` that calls its `RegisterOpener` with the default mux and provider-default `URLOpener`.

Each driver impl can define its default schemes and default Opener behavior to be exposed via the default mux. Users can choose to create their own mux, create openers with custom settings (e.g., creds) but still register for the default schemes (by calling `provider.Register(mymux, myopener)`, or choose their own schemes (by calling `foo.Register(scheme, opener)` directly).

internal/docs/design.md
```diff
@@ -184,6 +187,103 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
+// against a set of registered schemes and calls the opener that matches the
+// URL's scheme.
+//
+// The zero value is a URLMux with no registered schemes.
+type URLMux struct {
+  // ...
+}
+
+// AddWidget registers the opener with its preferred scheme. If an opener
+// already exists for the scheme, AddWidget panics.
```
vangent: WDYT about allowing re-registrations for the same scheme instead of panic? (last one wins).

This would allow a user to create a custom `URLOpener` for a specific provider and re-register it on the default mux (instead of having to create their own mux).

I'm imagining a scenario where user has multiple providers installed, and is happy with everything but wants to change one default for one provider. Without re-registration, they are forced to create a new mux and register all of the existing providers, then pass the new mux around everywhere -- a big code change for a small semantic change.

internal/docs/design.md
zombiezen: It also stops a type from implementing multiple interfaces. Again, imagine an opener that wants to be able to open buckets and blobs.

internal/docs/design.md
zombiezen: Panicking is the more reasonable behavior. The library can't distinguish from the usage you describe and two packages that have a naming collision. Package initialization order is well-defined wrt dependencies, but not well-defined across an arbitrary program. Seemingly random restructurings of the program can affect which one gets registered.

(See golang/go#20478 for a related issue of where this can go wrong.)

internal/docs/design.md
zombiezen: @vangent, I'm having trouble coming up with a crisp rationale here, but I think @ijt's sentiment indicates we should have one written here. I'll try to talk around it and see if we stumble upon the right phrasing:

We found that the setup code for a non-Wire user wanting to use multiple providers with URLs was pretty painful. The concession here is that we permit registration for the 90% use-case, but any level of customization requires creating the registry from scratch. URL-opening is built on top of other constructors, so using this logic does not require global state.

internal/docs/design.md
```diff
@@ -184,6 +187,102 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
+Cloud includes the ability to construct each of its API objects using
+identifying URLs. The concrete type's package should include APIs like the
+following:
+
+```go
+package foo
+
+// A type that implements WidgetURLOpener can open widgets based on a URL.
+// The opener must not modify the URL argument. OpenWidgetURL must be safe to
+// call from multiple goroutines.
```
shantuo: Is this required at this design doc level?

internal/docs/design.md
zombiezen: @vangent How necessary is this given that we now have a global registry? There's enough going on in this design as is: I'm not seeing your solution as saving the user much effort as just registering the openers with the schemes explicitly and it loses a lot of the clarity.

internal/docs/design.md
zombiezen: The concurrency guarantee? Yes it is, since URL muxes may be shared at a global level.

internal/docs/design.md
```diff
@@ -184,6 +187,102 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
+### URLs
+
+To enable the [Backing services factor][] of a Twelve-Factor Application, Go
+Cloud includes the ability to construct each of its API objects using
+identifying URLs. The concrete type should have API like this:
+
+```go
+package foo
+
+// A type that implements WidgetURLOpener can open widgets based on a URL.
```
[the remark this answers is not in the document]
vangent: OK.

internal/docs/design.md
vangent: IIUC, your original proposal had each provider impl defining a `NamedWidgetURLOpener` per scheme. This is replacing those structs-implementing-interface with a single function, which seems simpler to me.

We could also drop this idea entirely, and require users that don't want the default/global mux to create a new mux and explicitly register specific schemes for it. We'd have to expose the default scheme as a public constant, because most users won't have any reason to change the scheme. I guess I'm OK with that; IMHO my proposal is clearly better if there are multiple schemes (e.g., aliases), but maybe we just don't do that.

internal/docs/design.md
vangent: IMHO it will be useful to include such a snippet as an example. Something like:

```
const Scheme = "gs"

type URLOpener struct {
	Client  *gcp.HTTPClient
	Options Options
}

    ...
	return OpenBucket(ctx, o.Client, u.Host, &o.Options)
}

func init() {
    blob.DefaultURLMux().Register(Scheme, URLOpener{})
}
```

as well as example of the two ways of using it:

```
// Using default mux. Import provider package, then:
bucket, err := blob.DefaultURLMux().OpenURL(ctx, "<url>")

// Custom mux.
myMux := blob.URLMux{}
myMux.Register(gcsblob.Scheme, gcsblob.URLOpener{Client: client})
myMux.Register(fileblob.Scheme, fileblob.URLOpener{})
...
bucket, err := myMux.OpenURL(ctx, "<url>")
```

internal/docs/design.md
vangent: I guess the rationale is that it just makes URLs so easy to use. There's no way to achieve that without letting the package initializers set things up, which requires global state. And, we're providing an alternative for more advanced users who want to avoid it.

internal/docs/design.md
vangent: OK. I guess they can always create their own global variable if they want to shoot themselves in their own foot.

internal/docs/design.md
vangent: We could have a separate method, `Reregister` or some such, that would allow it; package initializers would never use it but users could.

internal/docs/design.md
ijt: I was just thinking the shorter version would be what we use most of the time, and in situations like you're pointing out such as `blob` we could do use the longer names.

internal/docs/design.md
zombiezen: Done.

internal/docs/design.md
```diff
@@ -184,6 +190,158 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
+func (mux *URLMux) OpenWidget(ctx context.Context, urlstr string) (*Widget, error) {
+  u, err := url.Parse(urlstr)
+  if err != nil {
+    return nil, fmt.Errorf("open widget: %v", err)
+  }
+  return mux.OpenWidgetURL(ctx, u)
+}
+
+// OpenWidgetURL dispatches the URL to the opener that is registered with the
+// URL's scheme. OpenWidget is safe to call from multiple goroutines.
```
vangent: Nit: OpenWidgetURL here.

internal/docs/design.md
```diff
@@ -184,6 +190,158 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
+  Client  *gcp.HTTPClient
+  Options Options
+}
+
+func (o *URLOpener) OpenWidgetURL(ctx context.Context, u *url.URL) (*foo.Widget, error) {
+  // ...
+  return OpenWidget(ctx, o.Client, u.Host, &o.Options)
+}
+
+type lazyURLOpener struct {
```
vangent: Could this be merged into `URLOpener`?

```
type URLOpener struct {
  Client  *gcp.HTTPClient
  Options Options
  init sync.Once
  err error
}

func (o *URLOpener) OpenWidgetURL(ctx context.Context, u *url.URL) (*foo.Widget, error) {
  o.init.Once(func() {
    if o.creds == nil {
      creds, err := gcp.DefaultCredentials(ctx)
      if err != nil {
        o.err = err
      } else {
        o.o.Client, _ = gcp.NewHTTPClient(http.DefaultTransport, creds.TokenSource)
      }
    }
  })
  if o.err != nil {
    return nil, o.err
  }
  return OpenWidget(ctx, o.Client, u.Host, &o.Options)
}
```

internal/docs/design.md
```diff
@@ -184,6 +190,158 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
+  o.init.Once(func() {
+    creds, err := gcp.DefaultCredentials(ctx)
+    if err != nil {
+      o.err = err
+      return
+    }
+    o.o = new(URLOpener)
+    o.o.Client, _ = gcp.NewHTTPClient(http.DefaultTransport, creds.TokenSource)
+  })
+  return o.OpenWidgetURL(ctx, u)
```
vangent: I think you meant `o.o.OpenWidgetURL` here, but see above.

internal/docs/design.md
zombiezen: Done.

internal/docs/design.md
zombiezen: It makes the end-user semantics harder to reason about. Generally, fields that are protected by some synchronization mechanism should not be exported.

I'm doing it this way for the globally registered one because I don't want to force an I/O operation during package initialization (the user can't change anything about package initialization). Server initialization code can block on I/O in reasonable ways and then construct a `URLOpener`.

internal/docs/design.md
zombiezen: Fixed with better names.

zombiezen:
To @shantuo and @clausti: the friendliest of pings.

internal/docs/design.md
```diff
@@ -184,6 +187,102 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
 `NewWithOptions`).
 
+### URLs
+
+To enable the [Backing services factor][] of a Twelve-Factor Application, Go
+Cloud includes the ability to construct each of its API objects using
+identifying URLs. The concrete type should have API like this:
+
+```go
+package foo
```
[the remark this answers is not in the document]
clausti: I think the foo/widget/which-one-means-blob confusion is intense enough to warrant some comment near the example package name.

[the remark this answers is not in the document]
zombiezen: Added commentary.

PR #1212, opened 2019-01-30
all: where appropriate update examples to compile but not run, and add tests where needed for coverage

### LINKED ISSUE #869: pubsub: each cloud driver implementation should have examples
The examples don't have to demonstrate how to use the concrete type; the concrete type examples should show that. The per-provider examples are for demonstrating how to use the provider-specific constructor.

For example, `gcppubsub` has:

```
func OpenTopic(ctx context.Context, client *raw.PublisherClient, proj gcp.ProjectID, topicName string, opts *TopicOptions) (*pubsub.Topic, error) {
```

An example would make it easier for a new user to figure out how to create a `raw.PublisherClient`.

### LINKED ISSUE #1177: all: update non-local examples to compile but not run instead of expecting failure

vangent:
Fixes #1177.
Fixes #869.


blob/azureblob/example_test.go
```diff
@@ -42,232 +39,71 @@ func Example() {
...
-				}
-			}
-		}
-	}
-
-	// Output:
-	// ReadAll failed due to invalid credentials
+	b, err = blob.Open(ctx, "azblob://mycontainer?cred_path=replace-with-path-to-credentials-file")
+	_ = b
+	_ = err
```
shantuo: nit: err can be moved before `_ = b` or ignored entirely.

pubsub/mempubsub/example_test.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2018 The Go Cloud Development Kit Authors
```
shantuo: s/2018/2019

pubsub/rabbitpubsub/example_test.go
```diff
@@ -0,0 +1,60 @@
+// Copyright 2018 The Go Cloud Development Kit Authors
```
shantuo: s/2018/2019

blob/azureblob/example_test.go
vangent: Sorry, I didn't get this. I need to touch both b and err or else there's a compile error that they aren't used.

pubsub/mempubsub/example_test.go
vangent: Done.

pubsub/rabbitpubsub/example_test.go
vangent: Done.

PR #1214, opened 2019-01-30
blob: support for OpenCensus metrics

jba:
First, we combine tracing and metrics into a single package,
internal/oc.

The metrics support includes functions for creating the standard
measure and standard views.

Second, we combine tracing and metric recording into one pair of calls.
We define a Tracer type to hold common values across an API resource.

We use this machinery in blob:
- Every bucket holds a Tracer.
- We call the Tracer's Start and End methods where previously we
  used the interal/trace package.

The internal/trace package is obsolete. We'll remove it after
we replace it in the other APIs.

TODO: testing.


internal/oc/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+	return []*view.View{
+		{
+			Name:        pkg + "/completed_calls",
+			Measure:     latencyMeasure,
+			Description: "Count of method calls by provider, method and status.",
+			TagKeys:     []tag.Key{ProviderKey, MethodKey, StatusKey},
+			Aggregation: view.Count(),
+		},
+		{
+			Name:        "gocloud.dev/pubsub/roundtrip_latency",
```
vangent: Nit: round-trip may not be accurate since these are concrete type measurements. E.g., a single write may result in multiple RPCs.

vangent: `pubsub` is wrong here, I think you want `pkg + /something`.

blob/blob.go
```diff
@@ -172,6 +173,7 @@ func (a *Attributes) As(i interface{}) bool {
 type Writer struct {
 	b          driver.Bucket
 	w          driver.Writer
+	tracer     *oc.Tracer
```
vangent: `tctx` should probably move up here; it's currently in the section that says "These fields exist only when w is not yet created."

internal/oc/metrics.go
```diff
@@ -0,0 +1,69 @@
...
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+)
+
+// LatencyMeasure returns the measure for method call latency used
+// by Go CDK APIs.
+func LatencyMeasure(pkg string) *stats.Float64Measure {
+	return stats.Float64(
+		pkg+"/roundtrip_latency",
```
vangent: Same as other comment, not sure `roundtrip` is right here.

internal/oc/metrics.go
jba: Just call it latency?

jba: Done.

internal/oc/metrics.go
jba: Done.

internal/oc/metrics.go
jba: Done.

internal/oc/trace.go
```diff
@@ -0,0 +1,82 @@
...
+	startTime := ctx.Value(startTimeKey{}).(time.Time)
+	elapsed := time.Since(startTime)
+	code := gcerrors.Code(err)
+	span := trace.FromContext(ctx)
+	if err != nil {
+		span.SetStatus(trace.Status{Code: int32(code), Message: err.Error()})
+	}
+	span.End()
+	stats.RecordWithTags(ctx, []tag.Mutator{tag.Upsert(StatusKey, fmt.Sprint(code))},
+		t.LatencyMeasure.M(elapsed.Seconds()*1000)) // milliseconds
```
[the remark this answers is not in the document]
jba: Done.

blob/blob.go
jba: I replaced both `tctx` and `tracer` with a closure.

jba: I did the same for Reader.

jba:
PTAL.

blob/blob.go
```diff
@@ -172,7 +172,8 @@ func (a *Attributes) As(i interface{}) bool {
 type Writer struct {
 	b          driver.Bucket
 	w          driver.Writer
-	cancel     func() // cancels the ctx provided to NewTypedWriter if contentMD5 verification fails
+	end        func(error) // called to end the OpenCensus span in Close
```
vangent: nit: use same comment as above, "called at Close to finish trace and metric collection"

blob/blob.go
```diff
@@ -598,10 +612,12 @@ func (b *Bucket) NewWriter(ctx context.Context, key string, opts *WriterOptions)
 		dopts.Metadata = md
 	}
 	ctx, cancel := context.WithCancel(ctx)
-	tctx := trace.StartSpan(ctx, "gocloud.dev/blob.NewWriter")
+	tctx := b.tracer.Start(ctx, "NewWriter")
+	end := func(err error) { b.tracer.End(tctx, err) }
+
 	defer func() {
 		if err != nil {
```
vangent: Is `end` only called with non-nil err here because you don't want to count writes that failed right away (e.g., bad ContentType)? Is it OK to `Start` the span but then never `End` it?

blob/oc_test.go
```diff
@@ -0,0 +1,64 @@
...
+	if _, err := b.ReadAll(ctx, "noSuchKey"); err == nil {
+		t.Fatal("got nil, want error")
+	}
+	// TODO(jba): after #1213, use DiffSpans.
+
+	data := <-te.Stats
+	for i, r := range data.Rows {
+		fmt.Printf("%d: %s %T\n", i, r, r.Data)
+	}
+
```
vangent: Why 3x?

blob/oc_test.go
```diff
@@ -0,0 +1,64 @@
...
+	data = <-te.Stats
+	for i, r := range data.Rows {
+		fmt.Printf("%d: %s %T\n", i, r, r.Data)
```
vangent: Can we verify these? E.g., the counts should be predictable.

blob/blob.go
jba: Done.

blob/blob.go
jba: I _do_ want to count writes that failed right away.

I'm sure nothing will break horribly if I don't close a span, but why not close them all? This call to `end` plus the one in `Close` should do that.

BTW, this is old code (except for the change to `end`) that was already there. Not that we can't revisit, of course.

blob/oc_test.go
jba: This is WIP and should not have been in the PR. Removing the file.

blob/oc_test.go
jba: ditto

jba:
PTAL.

jba:
I added `oc_test.go` back in a working state. Ready to review.

blob/blob.go
vangent: Sorry, I was misreading the code, this is (still) fine.

Can you add a comment here like 

```
// If err == nil, we handed the end closure off to the
// returned *Writer; it will be called when the Writer
// is Closed.
```

to make this more obvious? Thanks!

blob/oc_test.go
```diff
@@ -45,20 +51,61 @@ func TestOpenCensus(t *testing.T) {
...
+		diff := cmp.Diff(data.Rows, want, cmpopts.SortSlices(lessRow))
+		if diff != "" {
+			t.Errorf("metrics: %s", diff)
+		}
+		break
 	}
+}
+
+// The order doesn't matter; we just need a deterministic ordering of rows.
+func lessRow(r1, r2 *view.Row) bool { return rowKey(r1) < rowKey(r2) }
```
vangent: Shouldn't this be in `oc` somewhere? Seems like you'll need it for every test.

vangent:
(note two small comments)

blob/oc_test.go
jba: Yes, I'm planning on factoring out the testing code  at some point. Probably with my next metrics integration.

jba:
What two comments are you referring to? Are they:

- add tests first, then integrate with other packages: done
- factor out common test stuff: TBD

Am I missing something?

vangent:
My last two comments:

*  https://github.com/google/go-cloud/pull/1214#discussion_r252755057
*  https://github.com/google/go-cloud/pull/1214#discussion_r252756554

The second is fine to refactor later. The first one is a request to add a comment near the code that I misinterpreted.

blob/blob.go
jba: Done.

PR #1215, opened 2019-01-30
gcerrors: add PermissionDenied and ResourceExhausted

jba:
These both came up as useful codes in the review of #1207.

Also, run stringer.

### DIFF
--- a/internal/gcerr/gcerr.go
+++ b/internal/gcerr/gcerr.go
@@ -61,7 +61,7 @@ const (
 	ResourceExhausted ErrorCode = 9
 )
 
-// When adding a new error code, try to use the names define in google.golang.org/grpc/codes.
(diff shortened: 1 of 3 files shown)


internal/gcerr/gcerr.go
```diff
@@ -52,9 +52,18 @@ const (
...
+	// The caller does not have permission to execute the specified operation.
+	PermissionDenied ErrorCode = 8
+
+	// Some resource has been exhausted, typically because a service resource limit
+	// has been reached.
+	ResourceExhausted ErrorCode = 9
 )
 
-// TODO(jba) call stringer after it's fixed for modules
+// When adding a new error code, try to use the names define in google.golang.org/grpc/codes.
```
vangent: nit: define -> defined.

jba: Done.

PR #1217, opened 2019-01-31
secrets/tests: add a test using multiple keys

### LINKED ISSUE #919: internal/secrets: Add more tests
Per a [discussion](https://github.com/google/go-cloud/pull/899#discussion_r239570657)  on the `localsecrets` implementation PR, we should add additional conformance tests for `secrets`. 

At minimum: 
- Verify that encrypting the same message with the same key twice doesn't result in the same encrypted bytes.
- Verify that different keys work (and again, don't result in the same encrypted bytes).

vangent:
I also added/updated some comments about how `Decrypt` is not required to use the key provider to `Keeper`; I tried to add a test that showed that decrypting a ciphertext with a different key than the one used to encrypt it fails, but it (somewhat surprisingly) didn't fail for AWSKMS. It looks like they encode the key as part of the ciphertext. They do return the keyID used for decryption in the result, so I thought maybe we could return an artificial error if the keys didn't match, but they return the KeyID ARN, whereas we accept various aliases, so that doesn't work (at least not easily).

I also updated the record files for AWS + GCP to use our shared projects so that it's easy for other team members to run them.

Fixes #919.


secrets/drivertest/drivertest.go
```diff
@@ -28,8 +29,8 @@ import (
 // Harness descibes the functionality test harnesses must provide to run
 // conformance tests.
 type Harness interface {
-	// MakeDriver returns a driver.Keeper for use in tests.
-	MakeDriver(ctx context.Context) (driver.Keeper, error)
+	// MakeDriver returns a pair of driver.Keeper, each backed by a different key.
+	MakeDriver(ctx context.Context) (driver.Keeper, driver.Keeper, error)
```
shantuo: Do you think returning two keepers would be useful for more general cases? Otherwise how about calling makedriver multiple times when needed?

vangent: The problem is that we need two separate keys, and `MakeDriver` would need to know which one to return. I did consider adding an argument to `MakeDriver` to choose which key, but it's awkward; e.g., would the argument be a boolean? An integer only allowed to be 0 or 1? A name that every provider test needs to check for? I ended up deciding that just returning two drivers was simpler. WDYT?

shantuo: Does random generating a key id each time affect replay tests? If not we can do that in MakeDriver, otherwise I think what you have now probably the simplest.

vangent: We can't randomly generate a keyID for record/replay, the key has to be provisioned on the server. Even if we could, yes, it would affect the replay.

Another option would be to change the test to provision the key when the test starts up (for providers that need that), instead of creating it manually. That's what `pubsub` does. But that seems like a bigger change we could do separately if you think it's better.

vangent:
PTAL.

PR #1219, opened 2019-02-01
pubsub: add metrics

jba:
- Adapt pubsub concrete type to use the new oc.Tracer type.
- Add instrumentation for ReceiveBatch.
- Fix bug in which a Context variable was changed instead of using a fresh one.
- Factor out common testing code to the octest package.
- Combine comparison of traces and metric counts.
- Use new testing code in blob.

### DIFF
--- a/internal/testing/octest/diff.go
+++ b/internal/testing/octest/diff.go
@@ -18,14 +18,20 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
 	"go.opencensus.io/trace"
 	"gocloud.dev/gcerrors"
+	"gocloud.dev/internal/oc"
 )
 
-// Span holds the expected contents of a span obtained from tracing.
-type Span struct {
-	Name string
-	Code gcerrors.ErrorCode
+// Call holds the expected contents of a measured call.
+// It is used for both metric and trace comparison.
+type Call struct {
+	Method string
+	Code   gcerrors.ErrorCode
 }
 
 func formatSpanData(s *trace.SpanData) string {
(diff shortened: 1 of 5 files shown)


internal/testing/octest/diff.go
```diff
@@ -35,21 +41,33 @@ func formatSpanData(s *trace.SpanData) string {
...
+	return fmt.Sprintf("<Name: %q, Code: %d>", c.Method, c.Code)
 }
 
-// DiffSpans compares the list of spans obtained from OpenCensus tracing
+// Diff compares the list of spans and metric counts obtained from OpenCensus instrumentation
 // (using the TestExporter in this package, or similar) with an expected
-// list of spans. Only the name and code of the spans are compared.
-// Order matters.
-func DiffSpans(got []*trace.SpanData, want []Span) string {
+// list of call. Only the name and code are compared.
```
ijt: list of calls?

jba: Done.

vangent:
Tests are failing due to data race.

jba:
> Tests are failing due to data race.

Working on it.

I'm sorry you keep having to diagnose my test failures. That's really my job. In the future, if a test is failing, don't even waste your time commenting on the PR. I will either fix the failure or comment as to why it's not relevant.

jba:
Tests are passing. PTAL.

(Test for secrets will be addressed in a follow-up PR today.)

internal/testing/octest/exporter.go
```diff
@@ -38,7 +40,11 @@ func NewTestExporter(views []*view.View) *TestExporter {
 
 	// Register for metrics.
 	view.RegisterExporter(te)
-	view.SetReportingPeriod(time.Millisecond)
+	// The reporting period will affect how long it takes to get stats (view.Data).
+	// We want it short so tests don't take too long, but long enough so that all
```
vangent: This looks flaky so it makes me a bit nervous. Can you clarify "long enough so that all the work gets done"? It looks like `Counts` waits until data arrives...?

jba: I'm not fond of it either. I attempted to clarify. PTAL.

vangent: I'll approve for now but it might be worth thinking about how to make this less racy. For example, can `diffCounts` retry until it matches expectation (with a relatively long cap like 10s so actually failing tests don't hang forever)?

PR #1223, opened 2019-02-01
pubsub: add metrics

jba:
- Adapt pubsub concrete type to use the new oc.Tracer type.
- Add instrumentation for ReceiveBatch.
- Fix bug in which a Context variable was changed instead of using a fresh one.
- Factor out common testing code to the octest package.
- Combine comparison of traces and metric counts.
- Use new testing code in blob.

* doc tweak

* fix race; reorder test operations to remove nondeterminism

* fix metric tests in the presence of -count

* temporarily remove the tracing test from secrets, until this PR goes in.

* doc tweak

* secrets: add metrics

Also, restore and expand trace test.

* .


jba:
oops, ignore

PR #1224, opened 2019-02-01
secrets: add metrics

jba:
Also, restore and expand trace test.


jba:
> Can you add the documenation in `secrets.go`, matching what you did in the other PR for `blob` & `pubsub`?

Done.

PR #1225, opened 2019-02-01
runtimevar: add metrics

jba:


runtimevar/runtimevar.go
```diff
@@ -123,8 +138,8 @@ func newVar(w driver.Watcher) *Variable {
 // If the variable does not exist, Watch returns an error for which
 // gcerrors.Code will return gcerrors.NotFound.
 func (c *Variable) Watch(ctx context.Context) (_ Snapshot, err error) {
-	ctx = trace.StartSpan(ctx, "gocloud.dev/runtimevar.Watch")
-	defer func() { trace.EndSpan(ctx, err) }()
+	ctx = c.tracer.Start(ctx, "Watch")
```
vangent: Hmm. Frankly, I'm not sure this metric is worth exporting, given that we expect this method to block almost all of the time.

Even the driver method `WatchVariable` is dicey -- for some providers that poll, it might be useful for tracking the RPC, but other providers are expected to block (see the driver interface documentation for more info).

Maybe a metric exporting how often the variable changes?

jba: > a metric exporting how often the variable changes

Can you think of a use case for that? I can't. It's also pretty easy for an application to track that number if it wants to.

I'd rather have nothing for now, and add metrics as people request them.

vangent: The use case is for correlating config variable changes with other behaviors in the app. The number itself isn't interesting, but the times when it changes are.

jba: Makes sense. 

As I started to implement it, I realized that there is no string that identifies a Variable in the API. That means there's no way to tag the measurements to distinguish one variable's changes from another. I could count the aggregate changes over all variables. What do you suggest?

vangent: Hmm, good point. That does reduce the value somewhat. But, yeah, let's try doing the aggregate changes, it's better than nothing. I expect that it'll be common for programs to have a small number of runtime.Variables (likely 1, with a struct value, rather than a bunch of independent ones).

jba:
PTAL.

PR #1227, opened 2019-02-01
docs: add a section to design.md about escaping strings

vangent:
I propose we use an `Escaper` as described herein for:

*  blob keys (filenames)
   * including replacing the existing escaping in `fileblob`
*  blob metadata keys
*  pubsub metadata keys

I can't think of other places it's currently needed, but let me know if I'm wrong.

I also propose we have a shared testcase set of "weird characters" that driver conformance tests could use to verify behavior. https://github.com/google/go-cloud/blob/master/blob/drivertest/drivertest.go#L1604-L1627 is a starting point but we can add more. We'd extend the blob and pubsub conformance tests with these.

Updates #905.


internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+*   Azure Blob only
+    [accepts C# identifiers](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-properties-metadata)
+    as metadata keys.
+*   S3 drops double slashes in blob names (e.g., `foo//bar` will end up being
+    saved as `foo/bar`).
+
+These differences lead to a loss of portability and predictability for users.
+
+We considered restricting Go CDK's APIs to strings that all providers support.
+For example, we asserting that Go CDK's blob only supports ASCII plus `/` for
```
jba: s/we asserting/?/

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+
+// Escaper supports escaping and unescaping of strings.
+type Escaper interface {
+    // Escape returns the string with unsupported characters escaped.
+    Escape(string) string
+    // Unescape reverses the escaping done by Escape.
+    Unescape(string) string
+}
+
+// NewNoopEscaper returns an Escaper that doesn't change the string at all.
```
jba: s/Noop/Nop/ (prior art: https://godoc.org/io/ioutil#NopCloser)

jba: Also, `var NopEscaper`.

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+*   Go CDK APIs have visibility into all existing strings for all providers.
+    *   A slight caveat -- strings that look like Go CDK-escaped strings will be
+        unescaped by Go CDK. To mitigate this, we'll choose unusual-looking
+        escape mechanisms.
+*   Go CDK APIs are consistent in that you can write any string to any provider
+    and read it back.
+*   Some strings that were written through the Go CDK will appear in their
+    escaped form when viewed outside of Go CDK.
+
+The Go CDK providers helpers for such escaping:
```
jba: s/providers/provides/

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+
+```
+package stringescape
```
jba: Maybe just `escape`? That's a long package name.

I propose we put this and similar packages in a `support` subdir to avoid cluttering the root dir.

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+    // Escape returns the string with unsupported characters escaped.
+    Escape(string) string
+    // Unescape reverses the escaping done by Escape.
+    Unescape(string) string
+}
+
+// NewNoopEscaper returns an Escaper that doesn't change the string at all.
+func NewNoopEscaper() Escaper { ... }
+
+// NewUnderscoreEscaper returns an Escaper that escapes each character for which
```
jba: In this paragraph you're incorrectly conflating characters and bytes.

Either:
   s/character/rune/
   s/__0xXX__/__0xXXXX__/
   s/byte value/numeric value/

Or make it clear that this works at the byte level, which is actually a better idea (doesn't assume strings are valid UTF-8).

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+```
+
+To see how `Escaper` is used, let's use Azure's blob metadata keys as an
+example. As stated above, only valid `C#` identifiers are legal. The Azure blob
+provider would include code like this:
+
+```
+// shouldEscapeMetadataKey returns true if s[i] needs to be escaped in order
+// to produce a valid C# identifier.
+func shouldEscapeMetadataKey(s string, i int) {
```
jba: Forgot `bool` return.

I suggest you actually implement this, then cut-n-paste back.

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+    return !isValidCSharpIdentifier(s[i])
```
jba: This is harder if C# allows non-Latin-8 characters in identifiers. You might need `utf8.DecodeRuneInString`.

Also, I'd call the function `isValidCSharpIdentifierByte`.

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+}
+
+// When reading metadata keys, we unescape them:
+for k, v := range azureMetadata {
+    gcdkMetadata[mdKeyEscaper.Unescape(k)] = v
+}
+```
+
+When a user writes a metadata key like `foo.bar` through Go CDK, it is written
+as `foo__0x2e__`. When it is read back via Go CDK, it is unescaped back to
```
jba: You left off the `bar`.

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+either lose data (if previously-existing invalid keys were dropped) or get an
+error. Overall, we decided that this approach is unacceptable.
+
+Instead, we assert that Go CDK can handle any string, and force provider
+implementations to handle strings that the underlying provider can't handle
+using escaping mechanisms. We enforce provider compliance with conformance
+tests. With this approach
+
+*   Go CDK APIs have visibility into all existing strings for all providers.
+    *   A slight caveat -- strings that look like Go CDK-escaped strings will be
```
jba: It's a bit worse than that: you can have data loss if a set of data (e.g. blob metadata) contains two keys, one of which escapes to the other.

At a minimum, you should mention that here.

You could also talk about how to fix that, at least on write, by further escaping (in a different way) such conflicting keys. Or at least recognizing that case and returning an error.

Another problem you don't address is that escaping may push a long key over the provider's max length.

I don't think these are showstoppers, but we should acknowledge them.

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+// to produce a valid C# identifier.
+func shouldEscapeMetadataKey(s string, i int) {
+    return !isValidCSharpIdentifier(s[i])
+}
+
+var mdKeyEscaper = stringescaper.NewUnderscoreEscaper(shouldEscapeMetadataKey)
+
+// When writing metadata keys, we escape the keys:
+for k, v := range gcdkMetadata {
+    azureMetadata[mdKeyEscaper.Escape(k)] = v
```
jba: Maybe:
```
e := mdKeyEscaper.Escape(k)
if _, ok := azureMetadata[e]; ok {
  return fmt.Errorf("duplicate keys after escaping: %q => %q", k, e)
}
azureMetadata[e] = v
```

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+    gcdkMetadata[mdKeyEscaper.Unescape(k)] = v
+}
+```
+
+When a user writes a metadata key like `foo.bar` through Go CDK, it is written
+as `foo__0x2e__`. When it is read back via Go CDK, it is unescaped back to
+`foo.bar`.
+
+We could expose the `Escaper` used by providers in their `Options` structs, but
+we will probably will wait to see if there's demand for that.
```
jba: s/probably/definitely/!

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+*   Some strings that were written through the Go CDK will appear in their
+    escaped form when viewed outside of Go CDK.
+
+The Go CDK providers helpers for such escaping:
+
+```
+package stringescape
+
+// Escaper supports escaping and unescaping of strings.
+type Escaper interface {
```
zombiezen: There isn't a need for the indirection yet, so let's avoid it as per https://golang.org/wiki/CodeReviewComments#interfaces

internal/docs/design.md
zombiezen: I don't see a need to export this as a whole package yet. Especially given the rollout plan of "wait until need demands to expose", I would rather the escaping start as fixed functions in each of the driver implementations, and then export a pattern once we find it.

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+using escaping mechanisms. We enforce provider compliance with conformance
+tests. With this approach
+
+*   Go CDK APIs have visibility into all existing strings for all providers.
+    *   A slight caveat -- strings that look like Go CDK-escaped strings will be
+        unescaped by Go CDK. To mitigate this, we'll choose unusual-looking
+        escape mechanisms.
+*   Go CDK APIs are consistent in that you can write any string to any provider
+    and read it back.
+*   Some strings that were written through the Go CDK will appear in their
```
zombiezen: Suggestion to add nuance here: there can and will be strings that are valid in the underlying provider implementation that when passed to the Go CDK, will result in a different string being used for the name.

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+*   S3 drops double slashes in blob names (e.g., `foo//bar` will end up being
+    saved as `foo/bar`).
+
+These differences lead to a loss of portability and predictability for users.
+
+We considered restricting Go CDK's APIs to strings that all providers support.
+For example, we asserting that Go CDK's blob only supports ASCII plus `/` for
+blob names (and no `//`!). However, such a rule would mean that we couldn't
+handle existing strings that were created through some mechanism other than
+through Go CDK APIs. For example, an existing blob in S3 with a unicode name
```
zombiezen: I don't know if this example sells this approach: wouldn't this still be the case after this design? An existing blob in S3 with a Unicode name still would not be visible or writable.

internal/docs/design.md
```diff
@@ -322,7 +322,8 @@ across providers. Some theoretical examples using [`blob.Bucket`][]:
 1.  **Data fields**. Some providers may support key/value metadata associated
     with a blob, others may not.
 1.  **Naming rules**. Different providers may allow different name lengths, or
-    allow/disallow unicode characters.
+    allow/disallow unicode characters. See [Strings](#strings) below for more on
```
ijt: non-ASCII unicode characters?

internal/docs/design.md
```diff
@@ -374,6 +375,100 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+blob names (and no `//`!). However, such a rule would mean that we couldn't
+handle existing strings that were created through some mechanism other than
+through Go CDK APIs. For example, an existing blob in S3 with a unicode name
+wouldn't be visible in Go CDK. Or maybe it would be visible, but you wouldn't be
+able to write to it. For metadata keys, rules like this might mean that you
+could read metadata attributes via the Go CDK and then write them back, and
+either lose data (if previously-existing invalid keys were dropped) or get an
+error. Overall, we decided that this approach is unacceptable.
+
+Instead, we assert that Go CDK can handle any string, and force provider
```
ijt: How about "insist" instead of "assert"?

internal/docs/design.md
vangent: @zombiezen -- why? I'm intending to use the exact same escaping code for every provider. Because, why not? They differ in which bytes need escaping, not how to do the escaping.

I was planning on putting this code in `internal/`, so it's not really exported outside, but usable from multiple providers.

internal/docs/design.md

internal/docs/design.md
vangent: Removed.

internal/docs/design.md
vangent: Thanks, done. I intended bytes.

internal/docs/design.md

internal/docs/design.md
vangent: Don't think it does. Anyway, that's an exercise for the driver :-).

Done on rename.

internal/docs/design.md

internal/docs/design.md

internal/docs/design.md

vangent:
@eliben , @clausti , @shantuo please take a look, thanks!

internal/docs/design.md
```diff
@@ -374,6 +375,106 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+blob names (and no `//`!). However, such a rule would mean that we couldn't
+handle existing strings that were created through some mechanism other than
+through Go CDK APIs. For example, an existing blob in S3 with a unicode name
+wouldn't be visible in Go CDK. Or maybe it would be visible, but you wouldn't be
+able to write to it. For metadata keys, rules like this might mean that you
+could read metadata attributes via the Go CDK and then write them back, and
+either lose data (if previously-existing invalid keys were dropped) or get an
+error. Overall, we decided that this approach is unacceptable.
+
+Instead, we insist that Go CDK can handle any string, and force provider
```
eliben: Do we have some less ambiguous terms to distinguish between "provider implementations" and "underlying provider"?

internal/docs/design.md
```diff
@@ -374,6 +375,106 @@ b, err := blob.NewBucket(d, blob.FeatureUnicodeNames)
...
+
+// New returns an escaper that escapes each byte in s for which
+// shouldEscape returns true with "__0xXX__", where XX is the hex representation
+// of the byte value.
+// shouldEscape takes the whole string and an index instead of a single
+// byte because some escape decisions require context. For example, we
+// might want to escape the second "/" in "//" but not the first one.
+// Unescape unescapes a string returned from Escape.
+func New(shouldEscape func(s string, i int) bool) *escaper {...}
+
```
eliben: Nit: just one newline here?

internal/docs/design.md
zombiezen: Yeah, it's a prose organization issue. This section buries the lede. Most other sections in this doc will present the guidance first, and then at the very end conclude with the rationale.

internal/docs/design.md
zombiezen: Summarizing offline remarks: IMO the prose would be easier to read in this order:

1.  Brief problem statement
2.  Normative declaration of how we want this problem solved (all the immutable principles)
3.  Non-normative example of how to solve it for a particular set of constraints (e.g. C# identifiers)
4.  Rationale/Alternatives Considered

internal/docs/design.md

internal/docs/design.md
vangent: I think "driver" instead of "provider implementation" is more consistent with the rest of the doc; fixed.

vangent:
@zombiezen PTAL.

PR #1230, opened 2019-02-04
blob: add metrics for bytes read and written

### LINKED ISSUE #1229: blob: add metrics for bytes read/written
Add metrics to `Read` and `Write` to capture the number of bytes read and written.

jba:
Fixes #1229.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -66,6 +66,10 @@
 //   are included because they call NewRangeReader.)
 // - NewWriter, from creation until the call to Close.
 //
+// It also collects the following metrics:
+// - gocloud.dev/blob/bytes_read: the total number of bytes read, by provider.
+// - gocloud.dev/blob/bytes_written: the total number of bytes written, by provider.
+//
 // To enable trace collection in your application, see "Configure Exporter" at
 // https://opencensus.io/quickstart/go/tracing.
 // To enable metric collection in your application, see "Exporting stats" at
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -90,6 +90,9 @@ import (
 	"sync"
 	"time"
 
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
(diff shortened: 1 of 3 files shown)


blob/blob.go
```diff
@@ -66,6 +66,10 @@
 //   are included because they call NewRangeReader.)
 // - NewWriter, from creation until the call to Close.
 //
+// It also collects the following metrics:
```
vangent: Not part of this PR, but should the section above say what the name of the metric is for each method?

blob/blob.go
```diff
@@ -99,14 +106,17 @@ import (
...
+	b        driver.Bucket
+	r        driver.Reader
+	end      func(error) // called at Close to finish trace and metric collection
+	provider string      // for metric collection
 }
 
 // Read implements io.Reader (https://golang.org/pkg/io/#Reader).
 func (r *Reader) Read(p []byte) (int, error) {
 	n, err := r.r.Read(p)
+	stats.RecordWithTags(context.Background(), []tag.Mutator{tag.Upsert(oc.ProviderKey, r.provider)},
```
vangent: WDYT about keeping a counter here and exporting the metric in `Close` instead?

jba: That's less accurate (Close might not be called; or the reading might go on for minutes or hours) and less maintainable. Are you concerned about performance? Nothing is getting "exported" (sent off-box) at this point.  This line shouldn't take much time. Unless someone is foolishly reading in very small chunks, it's unlikely to be noticeable. But we could benchmark (against a real service with real buffer sizes).

vangent: OK, your arguments are convincing :-).

blob/blob.go
jba: I just did that for pubsub in #1232. I kept the wording generic so we can cut and paste. If you like that, I'll repeat it here and elsewhere.

PR #1234, opened 2019-02-04
pubsub: pass in Harness rather than Harness maker func in conformance tests

### LINKED ISSUE #1231: pubsub: pass harness into tests instead of making harnesses inside them
This is a bit of refactoring in drivertest.go called for by changes made for awspubsub.

ijt:
Fixes #1231 

Also, this fixes a bug in which TestSendReceiveTwo was being skipped for all providers in replay mode, not just awspubsub.

### DIFF
--- a/pubsub/drivertest/drivertest.go
+++ b/pubsub/drivertest/drivertest.go
@@ -20,7 +20,6 @@ import (
 	"bytes"
 	"context"
 	"errors"
-	"gocloud.dev/internal/testing/setup"
 	"strconv"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


jba:
I made this suggestion months ago and it was rejected.

ijt:
It was rejected because supposedly multiple harnesses would be created within the tests but so far we have no evidence of this.

jba:
It was rejected because *blob* creates multiple harnesses, not pubsub. But I guess @vangent should weigh in.

ijt:
I see your point about consistency. I'm closing this to work on something more likely to be useful.

Issue #1112, opened 2019-01-14
pubsub: add func that spins up a bcmills worker pool to process received messages


ijt:
It could be done like this:
```
numWorkers := 10
err := sub.ReceiveWithWorkerPool(ctx, numWorkers, func(m *Message) bool {
        fmt.Printf("Received message: %+v\n", m)
        return true
})
```
or like this:
```
import "gocloud.dev/workers"  // hypothetical new subpackage

numWorkers := 10
err := workers.RunPool(ctx, numWorkers, func() {
        m, err := sub.Receive(ctx)
        fmt.Printf("Received message: %+v\n", m)
        m.Ack()
})
```

ijt:
The code was rolled back because of a very occasional goroutine leak, not because of anything seriously wrong with it.

eliben:
@jba is this something we still plan on doing/trying at some point, or should it be closed?

jba:
I was never in favor of this.

PR #1239, opened 2019-02-05
internal: add workerpool module

ijt:
This makes progress on #1112.

### DIFF
--- a/internal/workerpool/pool_test.go
+++ b/internal/workerpool/pool_test.go
@@ -68,18 +68,12 @@ func TestRun(t *testing.T) {
 			m[i] = true
 		}
 
-		var wg sync.WaitGroup
-		wg.Add(1)
-		go func() {
-			workerpool.Run(context.Background(), 10, nextTask, func(ctx context.Context, ti interface{}) {
-				t := ti.(Task)
-				mu.Lock()
-				defer mu.Unlock()
-				delete(m, t.i)
-			})
-			wg.Done()
-		}()
-		wg.Wait()
+		workerpool.Run(context.Background(), 10, nextTask, func(ctx context.Context, ti interface{}) {
+			t := ti.(Task)
+			mu.Lock()
+			defer mu.Unlock()
+			delete(m, t.i)
+		})
 		t.Log("waiting for tasks to finish")
 		for {
 			mu.Lock()


internal/workerpool/pool.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2018 The Go Cloud Development Kit Authors
```
jba: 2109

internal/workerpool/pool.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// Run runs a worker pool with no more than limit goroutines. It gets tasks
+// from the nextTask func, which returns a task and a boolean that tells
+// whether to work on the task and continue running the pool. The doWork func
+// processes a single task. The provided context can be used to cancel
+// everything and exit the loop.
+func Run(ctx context.Context, limit int, nextTask func(context.Context) (interface{}, bool), doWork func(ctx context.Context, task interface{})) {
+	type token struct{}
+	sem := make(chan token, limit)
```
jba: It looks like this takes constant space, no matter what limit is. That's good.

internal/workerpool/pool.go
```diff
@@ -0,0 +1,57 @@
...
+		task, ok := nextTask(ctx)
+		if !ok {
+			break
+		}
+		select {
+		case <-ctx.Done():
+			break Loop
+		default:
+		}
+		sem <- token{}
```
jba: I think this should be in the select.

internal/workerpool/pool.go
```diff
@@ -0,0 +1,57 @@
...
+// them than a set limit. The idea for this package comes from Bryan C. Mills's
+// talk: https://www.youtube.com/watch?v=5zXAHh5tJqQ&t=33m13s
+package workerpool
+
+import (
+	"context"
+)
+
+// Run runs a worker pool with no more than limit goroutines. It gets tasks
+// from the nextTask func, which returns a task and a boolean that tells
```
jba: You could also just have nextTask return nil.

internal/workerpool/pool.go
```diff
@@ -0,0 +1,57 @@
...
+// whether to work on the task and continue running the pool. The doWork func
+// processes a single task. The provided context can be used to cancel
+// everything and exit the loop.
+func Run(ctx context.Context, limit int, nextTask func(context.Context) (interface{}, bool), doWork func(ctx context.Context, task interface{})) {
```
jba: It's too bad that we lose type information between nextTask and doWork. Maybe there could be one function that handles both.

internal/workerpool/pool.go
ijt: Yep.

internal/workerpool/pool.go
ijt: Livin' in the future! ;)

internal/workerpool/pool.go
ijt: I wanted to make sure it would end as soon as cancellation occurs. Select randomly chooses between possible cases, so it could happen that a task gets spawned after cancellation if I put both cases in the select.

internal/workerpool/pool.go
ijt: Good idea. Done.

internal/workerpool/pool.go
ijt: I agree, but this algorithm requires these functions to be separate.

internal/workerpool/pool.go
ijt: Hmm, thinking about it a bit more, I guess what you're proposing is fine since `Run` assumes `doWork` responds to context cancellation. The test I came up with couldn't reliably detect this situation anyway.

ijt:
PTAL

PR #1247, opened 2019-02-05
internal: simplify workerpool tests

ijt:
`workerpool.Run` should not exit before its spawned goroutines do, so the tests can be simplified based on that requirement.


ijt:
Subtle.

PR #1248, opened 2019-02-06
pubsub: convert switch statements to maps in awspubsub ErrorCode methods

ijt:
This reduces the verbosity and the amount of branching in in these methods.

### DIFF
--- a/pubsub/awspubsub/awspubsub.go
+++ b/pubsub/awspubsub/awspubsub.go
@@ -104,65 +104,31 @@ func (t *topic) ErrorCode(err error) gcerrors.ErrorCode {
 	if !ok {
 		return gcerr.Unknown
 	}
-	switch ae.Code() {
-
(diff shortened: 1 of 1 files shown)


ijt:
The number of lines of code was artificially inflated to begin with, due to a random decision on my part to write it a more verbose way. This change gets rid of the inflation. I don't think it's meaningful to exercise these low-risk code paths separately.

ijt:
Okay, I changed it so the maps aren't created each time, and changed the rationale.

Maybe I'm missing something, but I don't see any way to merge these two methods into a single function since none of the cases are shared between them.

Based on your case stacking idea, I rearranged the maps to put keys with the same values next to each other.

ijt:
PTAL

ijt:
Actually, now the two methods do almost the same thing, so I do see how to merge them. I'll try that now.

PR #1251, opened 2019-02-06
all: add a utility package for escaping/unescaping strings

### LINKED ISSUE #1250: all: add a helper for escaping/unescaping strings
As described here:
https://github.com/google/go-cloud/blob/master/internal/docs/design.md#strings

vangent:
Fixes #1250.

### DIFF
--- a/internal/escape/escape.go
+++ b/internal/escape/escape.go
@@ -22,7 +22,7 @@ package escape
 // shouldEscape takes the whole string and an index instead of a single
 // byte because some escape decisions require context. For example, we
 // might want to escape the second "/" in "//" but not the first one.
-func Escape(shouldEscape func(s string, i int) bool, s string) string {
+func Escape(s string, shouldEscape func(s string, i int) bool) string {
 	// Do a first pass to see how many bytes need escaping.
 	var toEscape []int
 	for i := 0; i < len(s); i++ {
--- a/internal/escape/escape.go
+++ b/internal/escape/escape.go
@@ -1,0 +1,119 @@
+// Copyright 2019 The Go Cloud Development Kit Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package escape includes helpers for escaping and unescaping strings.
+package escape
+
+// Escape returns s, with all bytes for which shouldEscape returns true
+// escaped to "__0xXX__", where XX is the hex representation of the byte
+// value.
+//
+// shouldEscape takes the whole string and an index instead of a single
+// byte because some escape decisions require context. For example, we
+// might want to escape the second "/" in "//" but not the first one.
+func Escape(shouldEscape func(s string, i int) bool, s string) string {
+	// Do a first pass to see how many bytes need escaping.
+	var toEscape []int
+	for i := 0; i < len(s); i++ {
+		if shouldEscape(s, i) {
+			toEscape = append(toEscape, i)
+		}
+	}
+	if len(toEscape) == 0 {
+		return s
+	}
+	// Each escaped byte turns into 8 bytes ("__0xXX__"),
+	// so there's an extra 7 bytes for each.
+	escaped := make([]byte, len(s)+7*len(toEscape))
+	n := 0 // current index into toEscape
+	j := 0 // current index into escaped
+	for i := 0; i < len(s); i++ {
+		if n < len(toEscape) && i == toEscape[n] {
+			b := s[i]
+			escaped[j] = '_'
+			escaped[j+1] = '_'
+			escaped[j+2] = '0'
+			escaped[j+3] = 'x'
+			escaped[j+4] = "0123456789ABCDEF"[b>>4]
+			escaped[j+5] = "0123456789ABCDEF"[b&15]
+			escaped[j+6] = '_'
+			escaped[j+7] = '_'
+			j += 8
+			n++
+		} else {
+			escaped[j] = s[i]
+			j++
+		}
+	}
+	return string(escaped)
+}
+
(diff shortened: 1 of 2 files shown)


internal/escape/escape.go
```diff
@@ -0,0 +1,119 @@
...
+package escape
+
+// Escape returns s, with all bytes for which shouldEscape returns true
+// escaped to "__0xXX__", where XX is the hex representation of the byte
+// value.
+//
+// shouldEscape takes the whole string and an index instead of a single
+// byte because some escape decisions require context. For example, we
+// might want to escape the second "/" in "//" but not the first one.
+func Escape(shouldEscape func(s string, i int) bool, s string) string {
```
jba: Put `s` first. It will make it easier to read calls that inline the shouldEscape function.

internal/escape/escape.go
```diff
@@ -0,0 +1,119 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package escape includes helpers for escaping and unescaping strings.
+package escape
+
+// Escape returns s, with all bytes for which shouldEscape returns true
```
jba: What's going to happen if the input is non-ASCII UTF-8? You should at least talk about that somewhere.

internal/escape/escape_test.go
```diff
@@ -0,0 +1,88 @@
...
+func TestEscape(t *testing.T) {
+	always := func(_ string, _ int) bool { return true }
+
+	for _, tc := range []struct {
+		description, s, want string
+		should               func(string, int) bool
+	}{
+		{
+			description: "empty string",
+			should:      always,
```
jba: explicitly write s: "", want: ""

internal/escape/escape.go
vangent: Good point; done.

internal/escape/escape_test.go
```diff
@@ -0,0 +1,88 @@
...
+			t.Errorf("%s: got escaped %q want %q", tc.description, got, tc.want)
+		}
+		got = Unescape(got)
+		if got != tc.s {
+			t.Errorf("%s: got unescaped %q want %q", tc.description, got, tc.s)
+		}
+	}
+}
+
+func TestUnescape(t *testing.T) {
```
jba: call it `TestUnescapeErrors` or something like that.

internal/escape/escape_test.go
```diff
@@ -0,0 +1,88 @@
...
+		},
+		{
+			description: "bytes in middle",
+			s:           "hello  world",
+			want:        "hello__0x20____0x20__world",
+			should:      func(s string, i int) bool { return s[i] == ' ' },
+		},
+		{
+			description: "unicode",
+			s:           "☺☺",
```
jba: Yes, it works to encode a UTF-8 string byte-by-byte. But what about this: can you write a `shouldEscape` that escapes all non-ASCII characters?

internal/escape/escape.go
vangent: Does it matter what the input is? We're currently going through it byte by byte, so there's no interpretation of UTF-8 going on. Do you think it would be better to have `shouldEscape` take a rune, and iterate over the `s` using `range` so that we go rune-by-rune instead of byte-by-byte?

internal/escape/escape_test.go
vangent: Done.

internal/escape/escape_test.go
vangent: Done.

internal/escape/escape_test.go
```diff
@@ -0,0 +1,88 @@
...
+// limitations under the License.
+
+package escape
+
+import (
+	"testing"
+)
+
+func TestEscape(t *testing.T) {
+	always := func(_ string, _ int) bool { return true }
```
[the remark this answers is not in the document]
vangent: Done.

internal/escape/escape.go
```diff
@@ -0,0 +1,119 @@
...
+// unhex returns the hexadecimal value of the hexadecimal byte b.
+// For example, unhex('A') returns 10.
+func unhex(b byte) byte {
+	switch {
+	case '0' <= b && b <= '9':
+		return b - '0'
+	case 'A' <= b && b <= 'F':
+		return b - 'A' + 10
+	}
+	return 0
```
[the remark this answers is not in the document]
vangent: Done.

internal/escape/escape.go
vangent: WDYT about having `shouldEscape` take a slice of `rune` instead (and the index)? If it returns true, we could still escape the bytes independently, or do the whole rune.

internal/escape/escape_test.go
vangent: PTAL, using runes now which I think address this. Thanks for noticing this.

internal/escape/escape.go
vangent: I've done this, PTAL.

internal/escape/escape.go
```diff
@@ -15,105 +15,122 @@
...
+// Note: shouldEscape takes the whole string as a slice of runes and an
+// index. Passing it a single byte or a single rune doesn't provide
+// enough context for some escape decisions; for example, the caller might
+// want to escape the second "/" in "//" but not the first one.
+// We pass a slice of runes instead of the string or a slice of bytes
+// because some decisions will be made on a rune basis (e.g., encode
+// all non-ASCII runes).
+func Escape(s string, shouldEscape func(s []rune, i int) bool) string {
+	// Do a first pass to see which runes (if any) need escaping.
+	runes := []rune(s)
```
jba: OK, so now your problem is a string that is not UTF-8. (Maybe some providers support such strings.) In that case I think you'll get a Unicode replacement character instead of the true bytes.

Maybe it's not worth worrying about that now.

vangent:
Hmm, this still won't work if `s` has non-UTF8 bytes. I'm not sure if we need to support that.

internal/escape/escape.go
vangent: I added a note that non-UTF8 isn't supported (i.e., non-UTF8 characters are escaped to `unicode.ReplacementChar`). We can always add it later.

When I'm using this package, I'll try non-UTF8 with a few providers to see what happens. If anyone supports it, we should probably support it.

PR #1257, opened 2019-02-06
internal: simplify workerpool.Run to return funcs from nextTask, not interface{}

ijt:
With this change, there is no more lack of type safety in `workerpool.Run`.

This is based on an idea from @jba.

### DIFF
--- a/internal/workerpool/pool.go
+++ b/internal/workerpool/pool.go
@@ -22,12 +22,15 @@ import (
 	"context"
 )
 
-// Run runs a worker pool with no more than limit goroutines. It gets tasks
-// from the nextTask func. If nextTask returns nil, Run will no longer ask for
-// more tasks, and will return after waiting for the running goroutines to
-// finish. Each func returned from nextTask is a unit of work to be performed.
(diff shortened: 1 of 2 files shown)


internal/workerpool/pool.go
```diff
@@ -22,27 +22,30 @@ import (
 	"context"
 )
 
-// Run runs a worker pool with no more than limit goroutines. It gets tasks
-// from the nextTask func. If nextTask returns nil, Run will no longer ask for
-// more tasks, and will return after waiting for the running goroutines to
-// finish. The doWork func processes a single task. The provided context can
-// be used to cancel everything and exit the loop.
-func Run(ctx context.Context, limit int, nextTask func(context.Context) interface{}, doWork func(ctx context.Context, task interface{})) {
+type Task func(context.Context)
```
jba: Maybe it doesn't need a context, because it's created inside nextTask?

ijt: Technically I think you're right, but I want to make it clear that these tasks are expected to deal gracefully with cancellation.

PR #1263, opened 2019-02-07
pubsub: add azure servicebus pubsub example

vsaroopchand:
Adds to #76 and PR #1207


pubsub/azurepubsub/example_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// Construct a Service Bus Namespace from a SAS Token.
+	// See https://godoc.org/github.com/Azure/azure-service-bus-go#Namespace.
+	ns, err := azurepubsub.NewNamespaceFromConnectionString(connString)
+	logFatal(err)
+
+	// Construct a Service Bus Topic for a topicName associated with a NameSpace.
+	// See https://godoc.org/github.com/Azure/azure-service-bus-go#Topic.
+	sbTopic, err := azurepubsub.NewTopic(ns, topicName, nil)
+	logFatal(err)
```
vangent: Nit: can you do the error checking inline instead of in `logFatal`? IIUC, the example will not show the source for `logFatal` so it's a bit confusing to read.

pubsub/azurepubsub/example_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"os"
+
+	"gocloud.dev/pubsub"
+	"gocloud.dev/pubsub/azurepubsub"
+)
+
+var (
+	// See docs below on how to provision an Azure Service Bus Namespace and obtaining the connection string.
+	// https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-dotnet-get-started-with-queues
+	connString = os.Getenv("SERVICEBUS_CONNECTION_STRING")
```
vangent: Same here, moving this var and comment inline into each of the two examples so users see it.

PR #1264, opened 2019-02-07
pubsub: deflake TestReceiveReturnsErrorFromSendAcks

### LINKED ISSUE #1238: pubsub/tests: TestReceiveReturnsErrorFromSendAcks is flaky
Test failed on Travis:

```
--- FAIL: TestReceiveReturnsErrorFromSendAcks (5.02s)
    acks_test.go:341: got context deadline exceeded; last error: <nil>, want SendAcks failed
FAIL
```

I couldn't repro locally.

ijt:
Fixes #1238.

The problem was that there was a fixed-size queue of messages that was being consumed, and sometimes that queue would be consumed before the error from the ack could get through. Once the queue was consumed, the next call to `sub.Receive` would keep on getting empty batches from `ReceiveBatch` and so it would keep on trying and the test would hang. It was a subtle race condition not detected by the -race flag. This change solves the problem by replacing the subscription driver with one that keeps on returning singleton batches of messages indefinitely, eliminating the race condition.

With this change, I ran `TestReceiveReturnsErrorFromSendAcks` with a count of one million and it passed.
```
[ ~/src/gocloud.dev/pubsub ] go test -run=TestReceiveReturnsErrorFromSendAcks -count=$((1000 * 1000))
PASS
ok  	gocloud.dev/pubsub	60.014s
```

### DIFF
--- a/pubsub/acks_test.go
+++ b/pubsub/acks_test.go
@@ -34,6 +34,11 @@ type ackingDriverSub struct {
 }
 
 func (s *ackingDriverSub) ReceiveBatch(ctx context.Context, maxMessages int) ([]*driver.Message, error) {
+	if len(s.q) == 0 {
+		// The queue won't be replenished. This means the test would just time
+		// out, so fail early and loudly instead.
+		panic("s.q is empty!")
+	}
 	if len(s.q) <= maxMessages {
 		ms := s.q
 		s.q = nil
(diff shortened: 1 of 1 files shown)


ijt:
Updated the description. PTAL.

ijt:
I'm closing this as it didn't address the bug uncovered by the flaking test.

PR #1265, opened 2019-02-07
blob: Ensure all metadata keys/values are escaped by drivers as needed

### LINKED ISSUE #1252: blob: update all drivers to escape metadata keys as needed
And update the conformance tests to use a wider range of key/value pairs.

vangent:
I tried to use URL escaping when possible, since that's more of a standard. In cases where it doesn't work (largely determined through experimentation, with hints from the provider documentation), I used our custom escaping.

Fixes #1252.


blob/azureblob/azureblob.go
```diff
@@ -423,6 +435,14 @@ func (b *bucket) Attributes(ctx context.Context, key string) (driver.Attributes,
 	if err != nil {
 		return driver.Attributes{}, err
 	}
+
+	azureMd := blobPropertiesResponse.NewMetadata()
```
jba: `azureMD`

blob/azureblob/azureblob.go
```diff
@@ -565,16 +585,33 @@ type writer struct {
...
 	if opts.BufferSize == 0 {
 		opts.BufferSize = defaultUploadBlockSize
 	}
+
+	md := make(map[string]string, len(opts.Metadata))
+	for k, v := range opts.Metadata {
+		// See the package comments for more details on escaping of metadata
+		// keys & values.
+		e := escape.Escape(k, func(runes []rune, i int) bool {
+			c := runes[i]
```
jba: Is the rule that the keys must be valid _identifiers_, or that they must only contain valid identifier _characters_?

The difference would be in keys that start with a digit. Are they legal?

blob/fileblob/fileblob.go
```diff
@@ -49,6 +49,14 @@
 //  - file://localhost/c:/foo/bar
 //    -> Also passes "c:\foo\bar".
 //
+// Escaping
```
jba: This is confusing. Better to just omit.

internal/escape/escape.go
```diff
@@ -17,9 +17,24 @@ package escape
 
 import (
 	"fmt"
+	"net/url"
 	"strconv"
+	"strings"
 )
 
+// IsAlphanumeric returns true iff r is alphanumeric: a-z, A-Z, 0-9.
```
jba: Actually ASCII alphanumeric. Otherwise you should be using `unicode.IsLetter` and `unicode.IsNumber`.

blob/azureblob/azureblob.go
vangent: Done.

blob/azureblob/azureblob.go
vangent: Good catch. No, they are not. I added a test to `escape.WeirdStrings` and check for that here now.

internal/escape/escape.go
vangent: Done.

blob/fileblob/fileblob.go

vangent:
PTAL.

PR #1268, opened 2019-02-07
runtimevar: minor doc updates and example reformatting

eliben:
Some cosmetic fixes to documentation and code formatting while reading the docs of runtimevar in detail.

Also recomputing some hashes in go.sum since these became out of date and prevented tests from running.


eliben:
Good catch on the `tags` file

I updated the proxy

Travis now fails on linux [https://travis-ci.com/google/go-cloud/jobs/176198051] with

```
$ if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then go install github.com/mattn/goveralls; fi
can't load package: package github.com/mattn/goveralls: unknown import path "github.com/mattn/goveralls": cannot find module providing package github.com/mattn/goveralls
The command "if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then go install github.com/mattn/goveralls; fi" failed and exited with 1 during .
```

eliben:
> Can you separate the `go.mod` and `go.sum` changes from the minor code fixes.

There shouldn't be `go.mod` changes now. I'll separate the `go.sum` changes - tracking that problem (tests not running with clean module cache) in #1277 now

eliben:
Split go.sum changes to #1278 - testing it now

eliben:
Moved goverall stuff to #1280

eliben:
Moved all unrelated stuff to other PRs, now only the minor comment/formatting changes remain. PTAL

PR #1269, opened 2019-02-07
docs: update string escaping section of design.md

vangent:


internal/docs/design.md
```diff
@@ -553,12 +553,15 @@ names, metadata keys, etc.). A couple of specific examples:
 
 These differences lead to a loss of portability and predictability for users.
 
-To resolve this issue, we insist that Go CDK can handle any string, and force
-drivers to use escaping mechanisms to handle strings that the underlying
+To resolve this issue, we insist that Go CDK can handle any UTF-8 string, and
```
jba: I think this is a reasonable choice, but it does mean that we will fail if someone is using binary. That may be very unlikely for attribute keys but less so for values.

We should at least document that here.

Actually, encoding of a non-UTF-8 string will only fail for providers that do escaping, which is not ideal. Maybe all providers should check that their input is valid UTF-8.

internal/docs/design.md
```diff
@@ -553,12 +553,15 @@ names, metadata keys, etc.). A couple of specific examples:
...
+force drivers to use escaping mechanisms to handle strings that the underlying
 provider can't handle. We enforce driver compliance with conformance tests.
 
-As an example, a driver for a provider that only allows underscores and ASCII
-characters might escape the string `foo.bar` to `foo__0x2e__bar`.
+We try to use URL encoding as the escaping mechanism where possible; however,
+sometimes it is not and we'll use custom escaping. As an example, a driver for a
+provider that only allows underscores and ASCII characters might escape the
```
jba: s/ASCII characters/ASCII alphanumeric characters/

internal/docs/design.md
```diff
@@ -594,13 +597,22 @@ Sample code for the helper for escaping strings:
 // package escape provider helpers for escaping and unescaping strings.
 package escape
 
-// Escape returns s, with all bytes for which shouldEscape returns true
-// escaped.
+// Escape returns s, with all runes for which shouldEscape returns true
+// escaped to "__0xXXX__", where XXX is the hex representation of the rune
```
jba: Don't you need four X's?

internal/docs/design.md

internal/docs/design.md
vangent: Yes, done. Although it's not always 4 hex digits. But I think the example demonstrates that.

internal/docs/design.md
vangent: I documented that the behavior for non-UTF-8 strings is currently undefined and filed an issue to consider erroring out in the concrete type.

vangent:
PTAL.

PR #1272, opened 2019-02-07
secrets: add ErrorAs function to Keeper

shantuo:
Fixes: #1261


secrets/awskms/kms.go
```diff
@@ -14,12 +14,19 @@
...
+// awskms exposes the following type for As:
+//  - Error: awserr.Error
 package awskms // import "gocloud.dev/secrets/awskms"
 
 import (
 	"context"
 	"errors"
 
+	"github.com/aws/aws-sdk-go/aws/awserr"
+
```
vangent: Nit: no need for blank line here.

secrets/drivertest/drivertest.go
```diff
@@ -54,6 +64,16 @@ func RunConformanceTests(t *testing.T, newHarness HarnessMaker) {
 	t.Run("TestDecryptMalformedError", func(t *testing.T) {
 		testDecryptMalformedError(t, newHarness)
 	})
+	t.Run("TestAs", func(t *testing.T) {
```
vangent: You can also add a conformance `AsTest` that verifies `As` fails when passed `nil`:

https://github.com/google/go-cloud/blob/master/runtimevar/drivertest/drivertest.go#L72-L93
https://github.com/google/go-cloud/blob/master/runtimevar/drivertest/drivertest.go#L119-#L119

secrets/example_test.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"gocloud.dev/secrets/gcpkms"
+	"google.golang.org/grpc/status"
+)
+
+func ExampleKeeper_ErrorAs() {
```
vangent: We need an example for using `secrets` overall. I don't think any of our packages have examples for using `As`.

I suggest:

* Dropping this file from the PR.
* Send a separate PR adding this file, but make the example use `localsecrets` and actually work.

I filed a separate PR (#1274) to improve the documentation/examples for `As`, so we can handle the `As` example there.

secrets/localsecrets/localsecrets.go
```diff
@@ -78,5 +78,10 @@ func (k *keeper) Decrypt(ctx context.Context, message []byte) ([]byte, error) {
 	return decrypted, nil
 }
 
+// ErrorAs implements driver.Keeper.ErrorAs.
```
vangent: Please add an `As` section to the package docstring just saying that it doesn't support any types for `As`.
https://github.com/google/go-cloud/blob/master/blob/memblob/memblob.go#L27-L29

secrets/vault/vault.go
```diff
@@ -96,6 +96,11 @@ func (k *keeper) Encrypt(ctx context.Context, plaintext []byte) ([]byte, error)
 	return []byte(secret.Data["ciphertext"].(string)), nil
 }
 
+// ErrorAs implements driver.Keeper.ErrorAs.
```
vangent: Same here, add to package docstring even if it doesn't support any types.

What error type does `Vault` return? Maybe add to the `TODO` below, something like "classify vault error codes and add a type for ErrorAs".

secrets/drivertest/drivertest.go
shantuo: Done.

secrets/example_test.go
shantuo: Done.

secrets/localsecrets/localsecrets.go
shantuo: Done.

secrets/vault/vault.go
shantuo: Done.

jba:
LGTM.

PR #1273, opened 2019-02-07
pubsub: remove looping logic from getNextBatch to eliminate a race condition

### LINKED ISSUE #1238: pubsub/tests: TestReceiveReturnsErrorFromSendAcks is flaky

ijt:
Fixes #1238.

There was a race condition: If `m.Ack()` results in a permanent error, that should be reported from the next call to `sub.Receive()` but if `sub.Receive()` acquires the subscription's mutex before that permanent error occurs then the call to `getNextBatch()` goes into an infinite loop waiting for a non-empty batch. This change replaces the empty batch loop in `getNextBatch()` with just its body so that it will return and the check for unrecoverable errors within `Receive()` can be reached again even if no more messages come in.

Before:
```
[ ~/src/gocloud.dev/pubsub ] go test -run=TestReceiveReturnsErrorFromSendAcks -count=10000
--- FAIL: TestReceiveReturnsErrorFromSendAcks (5.00s)
    acks_test.go:341: got context deadline exceeded; last error: <nil>, want SendAcks failed
--- FAIL: TestReceiveReturnsErrorFromSendAcks (5.00s)
    acks_test.go:341: got context deadline exceeded; last error: <nil>, want SendAcks failed
FAIL
exit status 1
FAIL	gocloud.dev/pubsub	10.658s
```
After:
```
[ ~/src/gocloud.dev/pubsub ] go test -run=TestReceiveReturnsErrorFromSendAcks -count=100000
PASS
ok  	gocloud.dev/pubsub	5.469s
```

### DIFF
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -422,25 +422,15 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
 // getNextBatch gets the next batch of messages from the server and returns it.
 func (s *Subscription) getNextBatch(ctx context.Context, nMessages int) ([]*Message, error) {
 	var msgs []*driver.Message
-	for len(msgs) == 0 {
-		// Allow unrecoverable errors from acks to be detected and reported.
-		s.mu.Lock()
-		err := s.err
-		s.mu.Unlock()
-		if err != nil {
-			return nil, err
-		}
-
-		err = retry.Call(ctx, gax.Backoff{}, s.driver.IsRetryable, func() error {
-			var err error
-			ctx2 := s.tracer.Start(ctx, "driver.Subscription.ReceiveBatch")
-			defer func() { s.tracer.End(ctx2, err) }()
-			msgs, err = s.driver.ReceiveBatch(ctx2, nMessages)
-			return err
-		})
-		if err != nil {
-			return nil, wrapError(s.driver, err)
-		}
+	err := retry.Call(ctx, gax.Backoff{}, s.driver.IsRetryable, func() error {
+		var err error
+		ctx2 := s.tracer.Start(ctx, "driver.Subscription.ReceiveBatch")
+		defer func() { s.tracer.End(ctx2, err) }()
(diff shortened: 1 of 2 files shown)


pubsub/pubsub.go
```diff
@@ -423,7 +423,15 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
 func (s *Subscription) getNextBatch(ctx context.Context, nMessages int) ([]*Message, error) {
 	var msgs []*driver.Message
 	for len(msgs) == 0 {
-		err := retry.Call(ctx, gax.Backoff{}, s.driver.IsRetryable, func() error {
+		// Allow unrecoverable errors from acks to be detected and reported.
```
vangent: What would happen if we allowed `getNextBatch` to return an empty slice instead? Wouldn't we just loop in `Receive`?

E.g., try commenting out your new code, and also deleting the `for len(msgs) == 0` loop in this function, does it wor?

pubsub/pubsub.go
```diff
@@ -423,7 +423,15 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
...
+		s.mu.Lock()
+		err := s.err
+		s.mu.Unlock()
+		if s.err != nil {
```
jba: use `err` instead of `s.err` here and the line below.

pubsub/pubsub.go
ijt: It does appear to work. Good idea.

pubsub/pubsub.go
ijt: Obsoleted by @vangent's idea above.

ijt:
PTAL

ijt:
@vangent, done.

PR #1283, opened 2019-02-08
pubsub: add TestReceiveReturnsAckErrorOnNoMoreMessages

ijt:
This test detects the cause of #1238 without running multiple times.

### DIFF
--- a/pubsub/acks_test.go
+++ b/pubsub/acks_test.go
@@ -349,6 +349,7 @@ func TestReceiveReturnsErrorFromSendAcks(t *testing.T) {
 // monkey patching of both its ReceiveBatch and SendAcks methods.
 type callbackDriverSub struct {
 	driver.Subscription
+	mu           sync.Mutex
 	receiveBatch func(context.Context) ([]*driver.Message, error)
 	sendAcks     func(context.Context, []driver.AckID) error
 }
@@ -399,15 +400,19 @@ func TestReceiveReturnsAckErrorOnNoMoreMessages(t *testing.T) {
 
 	// Second call to receiveBatch will wait for the pull from the
 	// receiveHappened channel below, and return a nil slice of messages.
+	ds.mu.Lock()
 	ds.receiveBatch = func(context.Context) ([]*driver.Message, error) {
+		ds.mu.Lock()
 		// Subsequent calls to receiveBatch won't wait on receiveHappened,
 		// and will also return nil slices of messages.
 		ds.receiveBatch = func(context.Context) ([]*driver.Message, error) {
 			return nil, nil
 		}
+		ds.mu.Unlock()
 		receiveHappened <- struct{}{}
 		return nil, nil
 	}
+	ds.mu.Unlock()
 
 	errc := make(chan error)
 	go func() {
--- a/pubsub/acks_test.go
+++ b/pubsub/acks_test.go
@@ -342,3 +342,86 @@ func TestReceiveReturnsErrorFromSendAcks(t *testing.T) {
 		}
 	}
 }
+
+// callbackDriverSub implements driver.Subscription and allows something like
+// monkey patching of both its ReceiveBatch and SendAcks methods.
+type callbackDriverSub struct {
+	driver.Subscription
+	receiveBatch func(context.Context) ([]*driver.Message, error)
+	sendAcks     func(context.Context, []driver.AckID) error
+}
+
+func (s *callbackDriverSub) ReceiveBatch(ctx context.Context, maxMessages int) ([]*driver.Message, error) {
+	return s.receiveBatch(ctx)
+}
+
+func (s *callbackDriverSub) SendAcks(ctx context.Context, acks []driver.AckID) error {
+	return s.sendAcks(ctx, acks)
+}
+
+func (s *callbackDriverSub) IsRetryable(error) bool { return false }
+
+func (s *callbackDriverSub) ErrorCode(error) gcerrors.ErrorCode { return gcerrors.Internal }
+
+// This test detects the root cause of
+// https://github.com/google/go-cloud/issues/1238.
+// If the issue is present, this test times out.
+func TestReceiveReturnsAckErrorOnNoMoreMessages(t *testing.T) {
+	// If SendAcks fails, the error is returned via receive.
+	ctx := context.Background()
+	serr := errors.New("unrecoverable error")
+	receiveHappened := make(chan struct{})
+	ackHappened := make(chan struct{})
+	ds := &callbackDriverSub{
+		// First call to receiveBatch will return a single message.
+		receiveBatch: func(context.Context) ([]*driver.Message, error) {
+			ms := []*driver.Message{{AckID: 1}}
+			return ms, nil
+		},
+		sendAcks: func(context.Context, []driver.AckID) error {
+			ackHappened <- struct{}{}
+			return serr
+		},
+	}
+	sub := pubsub.NewSubscription(ds, nil)
+	defer sub.Shutdown(ctx)
+	m, err := sub.Receive(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	m.Ack()
+
+	// Second call to receiveBatch will wait for the pull from the receiveHappened channel below,
+	// and return a nil slice of messages.
+	ds.receiveBatch = func(context.Context) ([]*driver.Message, error) {
+		// Subsequent calls to receiveBatch won't wait on receiveHappened,
+		// and will also return nil slices of messages.
+		ds.receiveBatch = func(context.Context) ([]*driver.Message, error) {
+			return nil, nil
+		}
+		receiveHappened <- struct{}{}
+		return nil, nil
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		// Set err in the test outside this goroutine so it can be checked below.
+		_, err = sub.Receive(ctx)
+		wg.Done()
+	}()
+
+	// sub.Receive has to get into the loop and then we need to trigger the unrecoverable error.
+	<-receiveHappened
+
+	// Trigger the unrecoverable error.
+	<-ackHappened
+
+	// Wait for sub.Receive to return and set err so we can check err against serr.
+	wg.Wait()
+
+	ok := gcerrors.Code(err) == gcerrors.Internal && err.(*gcerr.Error).Unwrap() == serr
+	if !ok {
+		t.Fatalf("got %v, want %v", err, serr)
+	}
+}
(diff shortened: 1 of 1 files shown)


ijt:
Reassigning from @vangent to @zombiezen as requested by @vangent offline.

pubsub/acks_test.go
```diff
@@ -342,3 +342,86 @@ func TestReceiveReturnsErrorFromSendAcks(t *testing.T) {
 		}
 	}
 }
+
+// callbackDriverSub implements driver.Subscription and allows something like
+// monkey patching of both its ReceiveBatch and SendAcks methods.
+type callbackDriverSub struct {
```
jba: Does this generalize some of the other test subs?

pubsub/acks_test.go
```diff
@@ -342,3 +342,86 @@ func TestReceiveReturnsErrorFromSendAcks(t *testing.T) {
...
+func (s *callbackDriverSub) SendAcks(ctx context.Context, acks []driver.AckID) error {
+	return s.sendAcks(ctx, acks)
+}
+
+func (s *callbackDriverSub) IsRetryable(error) bool { return false }
+
+func (s *callbackDriverSub) ErrorCode(error) gcerrors.ErrorCode { return gcerrors.Internal }
+
+// This test detects the root cause of
+// https://github.com/google/go-cloud/issues/1238.
```
jba: In addition to referencing the issue, could you describe the root cause here? The issue itself does not.

pubsub/acks_test.go
```diff
@@ -342,3 +342,86 @@ func TestReceiveReturnsErrorFromSendAcks(t *testing.T) {
...
+	// If SendAcks fails, the error is returned via receive.
+	ctx := context.Background()
+	serr := errors.New("unrecoverable error")
+	receiveHappened := make(chan struct{})
+	ackHappened := make(chan struct{})
+	ds := &callbackDriverSub{
+		// First call to receiveBatch will return a single message.
+		receiveBatch: func(context.Context) ([]*driver.Message, error) {
+			ms := []*driver.Message{{AckID: 1}}
+			return ms, nil
```
jba: You don't need the variable.
```
return []*driver......{}, nil
```

pubsub/acks_test.go
```diff
@@ -342,3 +342,86 @@ func TestReceiveReturnsErrorFromSendAcks(t *testing.T) {
...
+	// sub.Receive has to get into the loop and then we need to trigger the unrecoverable error.
+	<-receiveHappened
+
+	// Trigger the unrecoverable error.
+	<-ackHappened
+
+	// Wait for sub.Receive to return and set err so we can check err against serr.
+	wg.Wait()
+
+	ok := gcerrors.Code(err) == gcerrors.Internal && err.(*gcerr.Error).Unwrap() == serr
```
jba: Inline the variable.

jba: after the && can be written:
```
xerrors.Unwrap(err) == serr
```

pubsub/acks_test.go
```diff
@@ -342,3 +342,86 @@ func TestReceiveReturnsErrorFromSendAcks(t *testing.T) {
...
+		// Subsequent calls to receiveBatch won't wait on receiveHappened,
+		// and will also return nil slices of messages.
+		ds.receiveBatch = func(context.Context) ([]*driver.Message, error) {
+			return nil, nil
+		}
+		receiveHappened <- struct{}{}
+		return nil, nil
+	}
+
+	var wg sync.WaitGroup
```
jba: I prefer
```
errc := make(chan error)
go func() { 
  _, err = ...
  errc <- err
}()
```
Then you don't need the WaitGroup. In place of `wg.Wait` below you'd write `err := <-errc`.

pubsub/acks_test.go
ijt: It could probably generalize all the other testing sub drivers. Not sure it's worth the effort though.

pubsub/acks_test.go
ijt: Done.

pubsub/acks_test.go
ijt: That's true. I find it slightly more readable the way I wrote it though.

pubsub/acks_test.go
ijt: I'd rather not. Inlining it makes it slightly harder for me to read. Here's what it would look like:
```
if !(gcerrors.Code(err) == gcerrors.Internal && err.(*gcerr.Error).Unwrap() == serr) {

}
```
I don't like having to search for the matching paren when reading this.

ijt: Done.

pubsub/acks_test.go
ijt: Done.

ijt:
PTAL

pubsub/acks_test.go
```diff
@@ -342,3 +344,88 @@ func TestReceiveReturnsErrorFromSendAcks(t *testing.T) {
...
+	defer sub.Shutdown(ctx)
+	m, err := sub.Receive(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	m.Ack()
+
+	// Second call to receiveBatch will wait for the pull from the
+	// receiveHappened channel below, and return a nil slice of messages.
+	ds.receiveBatch = func(context.Context) ([]*driver.Message, error) {
```
zombiezen: `ReceiveBatch` is documented as needing to be safe to call from concurrent goroutines. Assigning to `ds.receiveBatch` without synchronization is a data race IIUC.

ijt: I agree that there could be a data race in the case where multiple goroutines simultaneously call `sub.Receive`, but in this test only one call to `sub.Receive` ever occurs. Our CI setup runs the tests with -race, and that should detect this situation if it ever comes up for some reason.

jba: I agree with Issac—in the context of this test, this is ok.

ijt: Pardon me. I realize now that my reply didn't really address your concern. Have to ponder this for a minute...

pubsub/acks_test.go
jba: @zombiezen @vangent @shantuo Is anyone else bothered by these one-line temporary variables? I perceive them like little bumps in the road, not terrible but kind of annoying. If I'm the only one then I'll shut up.

pubsub/acks_test.go
```diff
@@ -342,3 +344,88 @@ func TestReceiveReturnsErrorFromSendAcks(t *testing.T) {
...
+	// sub.Receive has to get into the loop and then we need to trigger the unrecoverable error.
+	<-receiveHappened
+
+	// Trigger the unrecoverable error.
+	<-ackHappened
+
+	// Wait for sub.Receive to return so we can check the err from sub.Receive against serr.
+	err = <-errc
+
+	ok := gcerrors.Code(err) == gcerrors.Internal && xerrors.Unwrap(err) == serr
```
jba: @zombiezen @vangent @shantuo Here is another case. Here Issac makes the good point (in a review comment that somehow got swallowed) that inlining requires a large parenthesized expr:
```
if !(....) 
```
That doesn't bother me, but I see how it could bother someone, so I guess I could go either way on this one.

In general, my position is that these are OK sometimes, but if the resulting line isn't too long or complicated, it's more readable to inline the variable.

pubsub/acks_test.go
ijt: I can see how someone looking at the code could worry that the first call to `sub.Receive` could spawn a goroutine that would indirectly call `ds.receiveBatch`. Even if it doesn't now, it could later. I added a mutex so no one has to worry about it or spelunk the code to find out.

pubsub/acks_test.go
zombiezen: It did cause me pause while reading in the way you describe, @jba. Unfortunately, I can't quite vocalize why in the general case. In this particular case, this caused me to pause because I thought it was mutating a variable, but then I had to read the colon in the `:=` operator.

pubsub/acks_test.go
zombiezen: Thanks @ijt. That's more what I was worried about: even if it isn't a race now, a subtle change to how Pub/Sub operates could render the test racy.

pubsub/acks_test.go
zombiezen: So in this case, the concern could be addressed in one of two ways:

One way would be to give the individual terms of the conditional distinct names, like:

```go
isInternalError := gcerrors.Code(err) == gcerrors.Internal
isSubscriptionError := xerrors.Unwrap(err) == serr

if !(isInternalError && isSubscriptionError) {
  // ...
}
```

This approach has the benefit of more clearly labeling what the parts of the conditions are trying to test. In fact, this begins to reveal that these are two separate assertions, and would likely benefit from two separate conditionals. So what I'd actually recommend is:

```go
if got := gcerrors.Code(err); got != gcerrors.Internal {
  t.Errorf("error code = %v; want %v", got, gcerrors.Internal)
}
if got := xerrors.Unwrap(err); got == serr {
  t.Errorf("error = %v; want %v", got, serr)
}
```

(Now that I'm staring critically at this, it leaves me wondering whether the test should use `xerrors.Is(err, serr)`. Yay refactoring!)

Another way would been to apply [De Morgan's Law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) to "factor" the negation to each of the boolean terms, but IMO the above form is more readable anyway.

pubsub/acks_test.go
ijt: In that case I suppose I could write it as
```
var ms = []*driver.Message{{AckID: 1}}
return ms, nil
```

pubsub/acks_test.go
ijt: Good idea. Done.

ijt:
PTAL

pubsub/acks_test.go
jba: Please don't do that.

ijt: Too verbose?

ijt: Do I have to inline this to get the PR approved?

jba: No, sorry, I was just waiting for people to weigh in. 

I came back to approve, but please change the `var ms =` back to `ms :=`.

clausti: fwiw, it does not pain me to read one-line temporary variables like this. I'm kinda on the side of "it's more readable this way", bc it seems like the `nil` could get lost at the end of `return []*driver.Message{{AckID: 1}}, nil`.

ijt: That's why I often do it this way.

ijt:
Done.

PR #1288, opened 2019-02-11
blob: update all drivers to escape keys as needed; add conformance tests

### LINKED ISSUE #1081: blob/fileblob: repeated slashes fails test
If you add the following test case to `drivertest.testKeys`:

```go
{
  description: "repeatedfwdslashes",
  key: "///foo///bar///baz",
},
```

it fails the test.

(I'd also like to see tests for `./` and `../` while we're here. I like malicious inputs.)

### LINKED ISSUE #1253: blob: update all drivers to escape keys as needed
And update the conformance tests to use a wider range of key/value pairs.

vangent:
Fixes #1081.
Fixes #1253.


blob/azureblob/azureblob.go
```diff
@@ -108,25 +111,6 @@ type Options struct {
 	SASToken SASToken
 }
 
-// Azure does not handle backslashes in the blob key well. As a workaround, all
```
vangent: Note to reviewer: I removed this behavior because I don't think it's what we want. It basically made it so that you could create blob keys with "\\" in the name and they would be automagically munged to "/". But other providers don't do this, so it caused inconsistencies across providers.

Instead, I now escape "\\" if it's found in the raw blob key. Users should be using "/" as a delimiter with the `blob` package (and `fileblob` will convert "/" to `os.PathSeparator` if needed).

blob/drivertest/drivertest.go
```diff
@@ -161,6 +161,9 @@ func RunConformanceTests(t *testing.T, newHarness HarnessMaker, asTests []AsTest
 	t.Run("TestList", func(t *testing.T) {
 		testList(t, newHarness)
 	})
+	t.Run("TestListWeirdKeys", func(t *testing.T) {
```
vangent: Note to reviewer: `testKeys` used to test reading + writing + List of weird keys. I changed it to test everything except List, and added this test for List explicitly.

blob/drivertest/drivertest.go
```diff
@@ -581,65 +638,62 @@ func testListDelimiters(t *testing.T, newHarness HarnessMaker) {
 				listResult{Key: keyPrefix + "/f.txt"},
```
vangent: Note to reviewer: `Hide Whitespace Changes` is your friend for this section of code, I just uncommented some tests.

PR #1289, opened 2019-02-11
blob/fileblob: implement URL signing

clausti:
updates https://github.com/google/go-cloud/issues/546


blob/fileblob/fileblob.go
```diff
@@ -93,15 +97,22 @@ func mungeURLPath(path string, pathSeparator uint8) string {
 }
 
 // Options sets options for constructing a *blob.Bucket backed by fileblob.
-type Options struct{}
+type Options struct {
+	// URLSigner implements signing URLs (to allow access to a resource without
+	// further authorization) and verifying that a given string contains
+	// a signedURL produced by the URLSigner.
+	// URLSigner is only required for utilizing the SignedURL api.
```
vangent: Nit: api -> API.

blob/fileblob/fileblob.go
```diff
@@ -606,7 +617,104 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
 	return nil
 }
 
+// SignedURL takes a context, an object key, and a SignedURLOptions struct
+// and returns a string containing a URI and an error
 func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error) {
-	// TODO(Issue #546): Implemented SignedURL for fileblob.
-	return "", errNotImplemented
+	if b.opts.URLSigner == nil {
+		return "", errNotImplemented
```
vangent: Consider a different error here now that it's no longer not implemented. For example, `gcsblob` has:

```
	if b.opts.GoogleAccessID == "" || (b.opts.PrivateKey == nil && b.opts.SignBytes == nil) {
		return "", errors.New("to use SignedURL, you must call OpenBucket with a valid Options.GoogleAccessID and exactly one of Options.PrivateKey or Options.SignBytes")
	}
```

blob/fileblob/fileblob.go
```diff
@@ -606,7 +617,104 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
 	return nil
 }
 
+// SignedURL takes a context, an object key, and a SignedURLOptions struct
```
vangent: Nit: This is a function on a private type (bucket), so it won't make it to godoc. You can just say `SignedURL implements driver.SignedURL` or something succinct like that.

blob/fileblob/fileblob.go
```diff
@@ -606,7 +617,104 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+// creating and verifying a signed URL for objects
+// in a fileblob bucket. Signed URLs are typically used for
+// granting access to an otherwise-protected resource without
+// requiring further authentication, and callers should take care
+// to restrict the creation of signed URLs as is appropriate
+// for their application.
+type URLSigner interface {
+	// URLFromKey defines how the bucket's object key will be turned
+	// into a signed URL.
+	// URLFromKey takes in a context, object key and options.
```
vangent: Nit: You don't need to describe the input/output types -- the signature does that. You can mention them by name if there's something to add about how they are used. For example, in this case you might mention that the returned `string` can be used as a URL (right?) and that the same URL can be passed to `KeyFromURL` for verification.

blob/fileblob/fileblob.go
```diff
@@ -606,7 +617,104 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+	// It returns a signed URL string, and an error.
+	URLFromKey(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error)
+
+	// KeyFromURL takes in a context and a URL.
+	// It returns an object key string and a bool.
+	// e.g. (key, true) if the signature is valid (authentic && unexpired),
+	// or (nil, false) if it is invalid.
+	// KeyFromURL must be able to validate a URL from the URLFromKey function.
+	KeyFromURL(ctx context.Context, surl *url.URL) (string, bool)
```
vangent: Nit: it's a bit weird to say that the output from one function can be passed to another function, but have them be different types. Can they both be `string` or both be `url.URL`?

blob/fileblob/fileblob.go
```diff
@@ -606,7 +617,104 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+// URLSignerHMAC uses the crypto/hmac package to create a message authentication code
+// using a sha256 hash
+type URLSignerHMAC struct {
+	urlScheme string
+	urlHost   string
+	urlPath   string
+	secretKey []byte
+}
+
+func NewURLSignerHMAC(urlScheme, urlHost, urlPath, secretKey string) *URLSignerHMAC {
```
vangent: Nit: Add godoc for exported functions.

blob/fileblob/fileblob.go
```diff
@@ -606,7 +617,104 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+	sURL := &url.URL{
+		Host:   h.urlHost,
+		Scheme: h.urlScheme,
+		Path:   h.urlPath,
+	}
+	q := sURL.Query()
+	q.Set("obj", key)
+	expTime := time.Now().Add(opts.Expiry).Unix()
+	q.Set("expiry", strconv.FormatInt(expTime, 10))
+	q.Set("signature", "")
```
vangent: Nit -- is this needed? If so, it warrants a comment why.

blob/fileblob/fileblob.go
```diff
@@ -606,7 +617,104 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+	sURL.RawQuery = q.Encode()
+
+	if !h.checkMAC(sURL.RawQuery, sig) {
+		return "", false
+	}
+	return q.Get("obj"), true
+}
+
+func (h *URLSignerHMAC) checkMAC(message string, mac string) bool {
```
vangent: Nit: 'message, mac string`.

blob/fileblob/fileblob_test.go
```diff
@@ -33,27 +34,62 @@ import (
...
+	h.server = localServer
+	h.closer = func() { _ = os.RemoveAll(dir); localServer.Close() }
+
+	return h, nil
+}
+
+func (h *harness) serveSignedURL(w http.ResponseWriter, r *http.Request) {
+	// URLSignerHMAC requires the scheme, host, and path to construct the
+	// signed URL, but validating only requires the key.
+	urlSigner := NewURLSignerHMAC("", "", "", h.urlSignKey)
```
vangent: Could you save `urlSigner` on the harness instead of recreating it here?

blob/fileblob/fileblob_test.go
```diff
@@ -33,27 +34,62 @@ import (
...
+	}
+
+	bucket, err := OpenBucket(h.dir, &Options{})
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	blobject, err := bucket.ReadAll(r.Context(), objKey)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
```
vangent: `NotFound`? This could actually happen if the file was moved/deleted.

blob/fileblob/fileblob.go
clausti: I can either set the signature to an empty string here, or unset the "signature" param entirely when the MAC is being checked, since I need to remove the signature value before re-hashing it.

blob/fileblob/fileblob_test.go
clausti: I could, but I thought it might be better to show it being re-created with the signing key, since I doubt that people using it in their applications would have access to the same in-memory object.

blob/fileblob/fileblob.go
clausti: I originally had them both as `string`, since the `SignedURL` signature returns a string, but the request URL is a `url.URL` object already, so to make that one take a string you end up converting it to a string and then re-parsing it, which seems sub-optimal.

clausti: made them both `url.URL`

blob/fileblob/fileblob_test.go
```diff
@@ -33,27 +34,62 @@ import (
...
+		w.WriteHeader(http.StatusForbidden)
+		return
+	}
+
+	bucket, err := OpenBucket(h.dir, &Options{})
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	blobject, err := bucket.ReadAll(r.Context(), objKey)
```
zombiezen: (Doesn't need to be addressed for this PR, but thought I'd mention it for when this handler is exported. Perhaps add a TODO.)

This access method would load the entire blob into memory and then write it. A more efficient way to do this would be to use `blob.NewReader` to get a reader and then use `io.Copy` to copy from the reader to the `http.ResponseWriter`.

blob/fileblob/fileblob.go
```diff
@@ -503,7 +514,109 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
 	return nil
 }
 
+// SignedURL implements driver.SignedURL
 func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error) {
-	// TODO(Issue #546): Implemented SignedURL for fileblob.
-	return "", errNotImplemented
+	if b.opts.URLSigner == nil {
+		return "", errors.New("to use SignedURL, you must call OpenBucket with a valid Options.URLSigner")
```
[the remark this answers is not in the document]
zombiezen: @vangent, I think we should fix the `gcsblob` error message similarly to how I'm suggesting. (Sorry for review whiplash.)

blob/fileblob/fileblob.go
```diff
@@ -503,7 +514,109 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+	// into a signed URL.
+	URLFromKey(ctx context.Context, key string, opts *driver.SignedURLOptions) (*url.URL, error)
+
+	// KeyFromURL must be able to validate a URL returned from URLFromKey.
+	// KeyFromURL must only return the object if if the URL is
+	// both unexpired and authentic.
+	KeyFromURL(ctx context.Context, surl *url.URL) (string, bool)
+}
+
+// URLSignerHMAC uses the crypto/hmac package to create a message authentication code
```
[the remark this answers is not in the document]
clausti: I'd argue that "what is the actual crypto being used" is critically significant to users of the URL signer. (And I'm not sure that last suggested sentence is actually clear and unambiguous.)

blob/fileblob/fileblob.go
```diff
@@ -503,7 +514,109 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+// for their application.
+type URLSigner interface {
+	// URLFromKey defines how the bucket's object key will be turned
+	// into a signed URL.
+	URLFromKey(ctx context.Context, key string, opts *driver.SignedURLOptions) (*url.URL, error)
+
+	// KeyFromURL must be able to validate a URL returned from URLFromKey.
+	// KeyFromURL must only return the object if if the URL is
+	// both unexpired and authentic.
+	KeyFromURL(ctx context.Context, surl *url.URL) (string, bool)
```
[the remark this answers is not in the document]
clausti: *Should* we tell people *why* their putative signed URL is invalid?

blob/fileblob/fileblob_test.go
vangent: My guess it that people are just as likely to have the original `urlSigner` around as they are to have the key. They will also definitely want to have the original `blob.Bucket` instead of creating a new one as well.

clausti:
@zombiezen PTAL?

blob/fileblob/fileblob.go
```diff
@@ -512,7 +523,102 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+// Values of URLSignerHMAC with the same secret key will accept URLs produced by
+// others as valid.
+type URLSignerHMAC struct {
+	baseURL   *url.URL
+	secretKey []byte
+}
+
+// NewURLSignerHMAC creates a URLSignerHMAC. A secretKey is required.
+func NewURLSignerHMAC(baseURL *url.URL, secretKey []byte) *URLSignerHMAC {
+	if secretKey == nil {
```
zombiezen: Checking the length of the slice is more robust:

```suggestion
	if len(secretKey) == 0 {
```

blob/fileblob/fileblob.go
```diff
@@ -512,7 +523,102 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+		return nil
```
zombiezen: It seems like the intent in returning `nil` is to surface errors early. If so, I recommend this code call `panic` instead, to surface that the API is being used incorrectly. (Which implies you should update the doc: "If the secret key is empty, then NewURLSignerHMAC panics.")

blob/fileblob/fileblob.go
```diff
@@ -512,7 +523,102 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+// URLFromKey creates a signed URL by copying the baseURL and appending the
+// object key, expiry, and signature as a query params.
+func (h *URLSignerHMAC) URLFromKey(ctx context.Context, key string, opts *driver.SignedURLOptions) (*url.URL, error) {
+	sURL := new(url.URL)
+	*sURL = *h.baseURL
+
+	q := sURL.Query()
+	q.Set("obj", key)
+	q.Set("expiry", strconv.FormatInt(time.Now().Add(opts.Expiry).Unix(), 10))
+	q.Set("signature", h.getMAC(q))
```
zombiezen: I would somewhat expect that the signature would be hex or base-64 encoded. You can use the `encoding/hex` or `encoding/base64` packages to do this.

(Thanks for refactoring, this makes the logic easier to follow!)

blob/fileblob/fileblob.go
```diff
@@ -512,7 +523,102 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+		return "", false
+	}
+
+	if !h.checkMAC(q) {
+		return "", false
+	}
+	return q.Get("obj"), true
+}
+
+func (h *URLSignerHMAC) checkMAC(q url.Values) bool {
```
zombiezen: Optional: I don't think splitting this into a separate method adds much clarity, but I realize it's a pretty subjective call.

clausti: feels nicer to me this way 🙃

blob/fileblob/fileblob.go
clausti: (change is in `getMAC`)

blob/fileblob/fileblob_test.go
```diff
@@ -33,27 +34,62 @@ import (
...
+
+	h := &harness{dir: dir, urlSignKey: "I'm a secret key"}
+	localServer := httptest.NewServer(http.HandlerFunc(h.serveSignedURL))
+	h.server = localServer
+	h.closer = func() { _ = os.RemoveAll(dir); localServer.Close() }
+
+	return h, nil
+}
+
+func (h *harness) serveSignedURL(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
clausti: changed the description from "fixes" to "updates" #546

blob/fileblob/fileblob.go
```diff
@@ -512,7 +524,103 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+	}
+
+	if !h.checkMAC(q) {
+		return "", errors.New("retrieving blob key from URL: key cannot be retrieved")
+	}
+	return q.Get("obj"), nil
+}
+
+func (h *URLSignerHMAC) checkMAC(q url.Values) bool {
+	mac := q.Get("signature")
```
zombiezen: Also need to base64 decode now

blob/fileblob/fileblob.go
```diff
@@ -512,7 +524,103 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+
+func (h *URLSignerHMAC) getMAC(q url.Values) string {
+	signedVals := url.Values{}
+	signedVals.Set("obj", q.Get("obj"))
+	signedVals.Set("expiry", q.Get("expiry"))
+	msg := signedVals.Encode()
+
+	hsh := hmac.New(sha256.New, h.secretKey)
+	hsh.Write([]byte(msg))
+	return base64.StdEncoding.EncodeToString(hsh.Sum(nil))
```
zombiezen: For maximum no-percent-encoding-ness:

```suggestion
	return base64.RawURLEncoding.EncodeToString(hsh.Sum(nil))
```

blob/fileblob/fileblob.go
zombiezen: I foolishly misunderstood this and it's correct as-is, sorry. A comment would help here.

clausti: added a comment, lemme know if that looks clarifying or nah

zombiezen: How about: "This compares the Base-64 encoded MACs."

PR #1291, opened 2019-02-11
pubsub: simplify skip logic

jba:
Move the ShouldSkip logic into newHarness.

ShouldSkip was adding some complexity to the conformance tests, and
because it called newHarness twice per test it was causing the
recorder/replayer to report that it was using the file twice.


go.sum
```diff
@@ -119,6 +119,7 @@ github.com/chrismalek/oktasdk-go v0.0.0-20181212195951-3430665dfaa0/go.mod h1:5d
 github.com/circonus-labs/circonus-gometrics v2.2.5+incompatible/go.mod h1:nmEj6Dob7S7YxXgwXpfOuvO54S+tGdZdw9fuRZt25Ag=
 github.com/circonus-labs/circonusllhist v0.1.3/go.mod h1:kMXHVDlOchFAehlya5ePtbp5jckzBHf4XRpQvBOLI+I=
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
+github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd h1:qMd81Ts1T2OTKmB4acZcyKaMtRnY5Y44NuXGX2GFJ1w=
```
jba: I have no idea why these have changed.

vangent: Can you revert then? It will probably just work.

jba: Done.

jba:
By the way, I'll be removing all the should-skip stuff in my next PR. This change enabled me to use my replayer (cloud.google.com/go/httpreplay), which fixed the problems with the skipped tests. But this is a good intermediate state.

PR #1295, opened 2019-02-11
pubsub: add test to check for goroutine leakage in Shutdown methods

### LINKED ISSUE #972: pubsub: ensure no goroutine leakage in Topic.Shutdown and Subscription.Shutdown
We need to wait on <-c to make sure the goroutine that shuts down the batcher finishes before Shutdown exits.

ijt:
Fixes #972

### DIFF
--- a/internal/contributebot/go.sum
+++ b/internal/contributebot/go.sum
@@ -13,10 +13,16 @@ dmitri.shuralyov.com/html/belt v0.0.0-20180602232347-f7d459c86be0/go.mod h1:JLBr
 dmitri.shuralyov.com/service/change v0.0.0-20181023043359-a85b471d5412/go.mod h1:a1inKt/atXimZ4Mv927x+r7UpyzRUf4emIoiiSC2TN4=
 dmitri.shuralyov.com/state v0.0.0-20180228185332-28bcc343414c/go.mod h1:0PRwlb0D6DFvNNtx+9ybjezNCa8XF0xaYcETyp6rHWU=
 git.apache.org/thrift.git v0.0.0-20180902110319-2566ecd5d999/go.mod h1:fPE2ZNJGynbRyZ4dJvy6G277gSllfV2HJqblrnkyeyg=
+github.com/Azure/azure-amqp-common-go v1.1.3/go.mod h1:FhZtXirFANw40UXI2ntweO+VOkfaw8s6vZxUiRhLYW8=
+github.com/Azure/azure-amqp-common-go v1.1.4/go.mod h1:FhZtXirFANw40UXI2ntweO+VOkfaw8s6vZxUiRhLYW8=
 github.com/Azure/azure-pipeline-go v0.1.8/go.mod h1:XA1kFWRVhSK+KNFiOhfv83Fv8L9achrP7OxIzeTn1Yg=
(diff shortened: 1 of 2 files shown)


pubsub/pubsub_test.go
```diff
@@ -421,3 +422,56 @@ func TestOpenCensus(t *testing.T) {
...
+	wg.Wait()
+
+	// Receive the messages.
+	var wg2 sync.WaitGroup
+	for i := 0; i < n; i++ {
+		wg2.Add(1)
+		go func() {
+			m, err := sub.Receive(ctx)
+			if err != nil {
+				t.Fatal(err)
```
jba: t.Fatal doesn't work in goroutines. Use t.Error, and find a way to exit the test from the main goroutine. Or maybe this could be an errgroup, and call t.Fatal from the main goroutine on error.

pubsub/pubsub_test.go
```diff
@@ -421,3 +422,56 @@ func TestOpenCensus(t *testing.T) {
...
+
+	// The Shutdown methods each spawn a goroutine so we want to make sure those don't
+	// keep running indefinitely after the Shutdowns return.
+	cancel()
+	top.Shutdown(ctx)
+	sub.Shutdown(ctx)
+
+	// Wait for number of goroutines to return to normal.
+	// Otherwise the test hangs.
+	for {
```
jba: You might be interested in https://code.googlesource.com/gocloud/+/master/internal/leakcheck.

pubsub/pubsub_test.go
ijt: Changed to use errgroup.

pubsub/pubsub_test.go
ijt: It seems like that's focused on gRPC.

internal/contributebot/go.sum
```diff
@@ -13,10 +13,16 @@ dmitri.shuralyov.com/html/belt v0.0.0-20180602232347-f7d459c86be0/go.mod h1:JLBr
 dmitri.shuralyov.com/service/change v0.0.0-20181023043359-a85b471d5412/go.mod h1:a1inKt/atXimZ4Mv927x+r7UpyzRUf4emIoiiSC2TN4=
 dmitri.shuralyov.com/state v0.0.0-20180228185332-28bcc343414c/go.mod h1:0PRwlb0D6DFvNNtx+9ybjezNCa8XF0xaYcETyp6rHWU=
 git.apache.org/thrift.git v0.0.0-20180902110319-2566ecd5d999/go.mod h1:fPE2ZNJGynbRyZ4dJvy6G277gSllfV2HJqblrnkyeyg=
+github.com/Azure/azure-amqp-common-go v1.1.3/go.mod h1:FhZtXirFANw40UXI2ntweO+VOkfaw8s6vZxUiRhLYW8=
```
jba: Can you revert go.sum and see if everything still works?

pubsub/pubsub_test.go
jba: I don't see why. But your way is much simpler, so if it works reliably, that's great.

ijt: Haven't had any trouble with it so far.

internal/contributebot/go.sum
ijt: Done. PTAL

pubsub/pubsub_test.go
ijt: PTAL.

PR #1296, opened 2019-02-12
pubsub: add worker pool sample

### LINKED ISSUE #1112: pubsub: add func that spins up a bcmills worker pool to process received messages

ijt:
Fixes #1112


samples/pubsub/workerpool/main.go
```diff
@@ -0,0 +1,66 @@
...
+
+// publish simulates users uploading videos.
+func publish(ctx context.Context, top *pubsub.Topic) error {
+	for {
+		videoURL := fmt.Sprintf("acmestorage://videos/%d.mpeg", rand.Int())
+		if err := top.Send(ctx, &pubsub.Message{Body: []byte(videoURL)}); err != nil {
+			return fmt.Errorf("sending video upload message: %v", err)
+		}
+		// Simulate some delay between videos being uploaded on an up-and-coming video site.
+		time.Sleep(time.Duration(rand.Intn(10) * 100 * int(time.Millisecond)))
```
jba: `time.Duration(rand... * 100) * time.Millisecond`

samples/pubsub/workerpool/main.go
```diff
@@ -0,0 +1,66 @@
...
+	eg.Go(func() error { return publish(ctx, top) })
+	eg.Go(func() error { return subscribe(ctx, sub) })
+	if err := eg.Wait(); err != nil {
+		log.Printf("workerpool sample: %v", err)
+	}
+}
+
+// publish simulates users uploading videos.
+func publish(ctx context.Context, top *pubsub.Topic) error {
+	for {
```
jba: How do you ever get out of this loop?

samples/pubsub/workerpool/main.go
```diff
@@ -0,0 +1,66 @@
...
+		}
+		// Simulate some delay between videos being uploaded on an up-and-coming video site.
+		time.Sleep(time.Duration(rand.Intn(10) * 100 * int(time.Millisecond)))
+	}
+	return nil
+}
+
+// subscribe pulls from the subscription and simulates processing the videos.
+func subscribe(ctx context.Context, sub *pubsub.Subscription) error {
+	workerpool.Run(ctx, 10, func(ctx context.Context) workerpool.Task {
```
jba: I think it reduces clarity to use `workerpool.Task` here, instead of just `func(context.Context)`.  It's not much shorter, and since you can't use it when you actually declare the function, it just looks like something's mismatched.

I suggest either removing that type, or making it unexported if you want to use it inside workerpool.

samples/pubsub/workerpool/main.go
```diff
@@ -0,0 +1,66 @@
...
+		m, err := sub.Receive(ctx)
+		if err != nil {
+			log.Fatalf("receiving video upload message: %v", err)
+		}
+		videoURL := string(m.Body)
+		return func(ctx context.Context) {
+			log.Printf("Processing %v", videoURL)
+			// Simulate transcoding the video and storing it in various formats.
+			time.Sleep(time.Duration(rand.Intn(10) * int(time.Second)))
```
jba: ditto

samples/pubsub/workerpool/main.go
```diff
@@ -0,0 +1,66 @@
...
+			log.Printf("Done processing %v", videoURL)
+			m.Ack()
+		}
+	})
+	ctx := context.Background()
```
jba: I don't understand this line.

samples/pubsub/workerpool/main.go
```diff
@@ -0,0 +1,66 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"math/rand"
+	"time"
+
+	"gocloud.dev/internal/workerpool"
```
jba: Since workerpool is internal, this example can't be used outside of the repo.

I thought you were going to use workerpool to build something pubsub-specific, either in the `pubsub` package or a subpackage (`pubsub/helpers`?). Something like the callback version of Receive that you originally wanted to do.  

I think I prefer that to making workerpool public, but I could be convinced. One concern I have is that workerpool is a simple, elegant but very general piece of functionality that we'd be committing to supporting if it moved out of internal. @zombiezen, what do you think? 

Honestly, I think my favorite solution is just "inlining" workerpool.Run here. That demonstrates the bcmills pattern clearly, with no abstraction layer, and doesn't commit us to maintaining a general-purpose package or adding any API to `pubsub`.

samples/pubsub/workerpool/main.go
ijt: Done.

samples/pubsub/workerpool/main.go
ijt: Ctrl-C

samples/pubsub/workerpool/main.go
ijt: When I do that it gives the following error:
```
./main.go:67:26: cannot use func literal (type func(context.Context) func(context.Context)) as type func(context.Context) workerpool.Task in argument to workerpool.Run
```

ijt: I prefer it with the `workerpool.Task` type because it seems like a bit of a confusing stutter without that type:
```
	workerpool.Run(ctx, 10, func(ctx context.Context) func(context.Context) {
...
        })
```
Anyway, this will probably become moot within the sample once I respond to your good point below.

samples/pubsub/workerpool/main.go
ijt: Done.

samples/pubsub/workerpool/main.go
ijt: I don't know how it got there. Removed it.

samples/pubsub/workerpool/main.go
ijt: Good point. I'll try adding `pubsub.RunWorkerPool(sub, handleMessage)` and we can see how it looks.

ijt: Done. PTAL.

ijt: I disagree about the utility of inlining the worker pool code. Concurrency logic is notoriously difficult to get right and since we've managed to abstract it in this case it's a clear win to have it as a function.

ijt: The argument about a maintenance burden seems unconvincing since there is no clear vector of bit rot for a function like `workerpool.Run` with no dependencies on anything other than `context`.

samples/pubsub/workerpool/main.go
jba: I can see someone running this and being confused that it never stops. I think it should either have a timeout or maybe print out a message saying to ctrl-C it.

samples/pubsub/workerpool/main.go
jba: Yeah, I noticed that. But you could also define a task to take no args, since it is returned from a method that takes  a context, and it is called with the same context.

ijt: The reason I made the `Task` funcs take a `Context` was to make it clear that they should respect cancellation. What do you think?

pubsub/pubsub.go
```diff
@@ -534,3 +536,30 @@ func wrapError(ec errorCoder, err error) error {
...
+
+// RunWorkerPool repeatedly fetches messages from the given subscription sub, in an
+// inverted worker pool (link) with up to maxWorkers. The callback handleMessage is
+// called on each message. If handleMessage return nil for its error, then the
+// message is acked. If sub.Receive or handleMessage returns a non-nil error then
+// the workerpool is terminated via context cancellation.
+func RunWorkerPool(ctx context.Context, maxWorkers int, sub *Subscription, handleMessage func(ctx context.Context, m *Message) error) error {
+	errc := make(chan error)
+	ctx, cancel := context.WithCancel(ctx)
+	workerpool.Run(ctx, maxWorkers, func(ctx context.Context) workerpool.Task {
```
jba: Add `defer cancel()` before this line.

pubsub/pubsub.go
```diff
@@ -534,3 +536,30 @@ func wrapError(ec errorCoder, err error) error {
 	}
 	return gcerr.New(ec.ErrorCode(err), err, 2, "pubsub")
 }
+
+// RunWorkerPool repeatedly fetches messages from the given subscription sub, in an
```
jba: This is a nice thing, but I'm not sure it should be in the `pubsub` package. It's really a utility function or helper, written on top of `pubsub`. I can imagine other helpers, like one that automatically extends the ack deadline (when we implement that ability). I'm thinking that to keep the pubsub API small and simple, we should put those kinds of things in a separate package.

If we *do* keep this in this package, would it make more sense as a method of `Subscription`?

samples/pubsub/workerpool/main.go
jba: Fine as you have it.

pubsub/pubsub.go
jba: Thinking some more about this, I feel even more strongly that it should be in a different package. There are reasons we went with the `Subscription.Receive` function we have now: it's easy to understand, has no magic, and is symmetric with `Topic.Send`. This is definitely a "use at your own risk" kind of function. You have to be careful that your callback is goroutine-safe, for instance, and it can't pass the message off to another goroutine for processing and acking unless it blocks until the ack.

samples/pubsub/workerpool/main.go
ijt: Added a timeout.

pubsub/pubsub.go
```diff
@@ -534,3 +536,30 @@ func wrapError(ec errorCoder, err error) error {
...
+// inverted worker pool (link) with up to maxWorkers. The callback handleMessage is
+// called on each message. If handleMessage return nil for its error, then the
+// message is acked. If sub.Receive or handleMessage returns a non-nil error then
+// the workerpool is terminated via context cancellation.
```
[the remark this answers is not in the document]

pubsub/pubsub.go
```diff
@@ -534,3 +536,30 @@ func wrapError(ec errorCoder, err error) error {
 	}
 	return gcerr.New(ec.ErrorCode(err), err, 2, "pubsub")
 }
+
+// RunWorkerPool repeatedly fetches messages from the given subscription sub, in an
+// inverted worker pool (link) with up to maxWorkers. The callback handleMessage is
+// called on each message. If handleMessage return nil for its error, then the
+// message is acked. If sub.Receive or handleMessage returns a non-nil error then
```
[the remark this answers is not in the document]

pubsub/pubsub.go

pubsub/pubsub.go
```diff
@@ -534,3 +536,30 @@ func wrapError(ec errorCoder, err error) error {
...
+		m, err := sub.Receive(ctx)
+		if err != nil {
+			cancel()
+			errc <- fmt.Errorf("receiving message: %v", err)
+			return func(context.Context) {}
+		}
+		return func(ctx context.Context) {
+			if err := handleMessage(ctx, m); err != nil {
+				cancel()
+				errc <- fmt.Errorf("handling message: %v", err)
```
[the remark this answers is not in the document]
ijt: Changed it to have a buffer with `max+1` since there could be up to `max` task goroutines and maybe also the goroutine running `workerpool.Run` trying to send on the channel.

pubsub/pubsub.go
ijt: Sounds reasonable. Giving that a shot now.

ijt: Do you have any suggestion as far as the naming for this new package?

How about
```
import "gocloud.dev/pubsub/subs"

subs.ReceiveConcurrently(...)
```
?

jba: I'd go with `pubsub/pshelpers` or `pubsub/psutil` since I see this as a place to put other conveniences, like the aforementioned ack deadline extender. I think of the new package as something like `io/ioutil` or `github.com/google/cmp/cmpopts`.

BTW, it's a good idea to include at least a little bit of the parent package in the name (IOutil, CMPopts, etc.) since you end up using only the last component of the path in the importing file, and if everyone used `opts` or `util` or `testing`, it would be confusing.

ijt: Okay, `pubsub/psutil` sounds fine.

samples/pubsub/workerpool/main.go
zombiezen: I'm more in favor of inlining the `workerpool` run for sample code, as then it's more clear what is going on. I don't see "concurrency being difficult" as being a reason to not show what's going on for illustration, especially if the pattern is well-known.

That said, one bit of management I really would like to see is coming up with a better way of doing [what I did for Contribute Bot](https://github.com/google/go-cloud/blob/master/internal/contributebot/main.go#L63-L65) where I want a process to subscribe on a topic and also serve HTTP health checks for Kubernetes (think `server` package for subscriber services). Just having the worker pool isn't really enough to justify a library function in the Go CDK.

jba: We've moved on from the inline/noinline discussion to making a helper function.

I don't see why we need to provide a way to interleave receiving messages from a subscription with other activities like health checks. That is one of the nice things about concurrency: we can interleave any activities we want by running each in its own goroutine. That is what you do in the linked contributebot code, and I don't know what a better way of doing that would be.

pubsub/pubsub.go
```diff
@@ -534,3 +536,30 @@ func wrapError(ec errorCoder, err error) error {
...
+// the workerpool is terminated via context cancellation.
+func RunWorkerPool(ctx context.Context, maxWorkers int, sub *Subscription, handleMessage func(ctx context.Context, m *Message) error) error {
+	errc := make(chan error)
+	ctx, cancel := context.WithCancel(ctx)
+	workerpool.Run(ctx, maxWorkers, func(ctx context.Context) workerpool.Task {
+		m, err := sub.Receive(ctx)
+		if err != nil {
+			cancel()
+			errc <- fmt.Errorf("receiving message: %v", err)
+			return func(context.Context) {}
```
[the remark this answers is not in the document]

internal/workerpool/pool.go
```diff
@@ -30,7 +30,7 @@ type Task func(context.Context)
 // longer ask for more tasks, and will return after waiting for the running
 // goroutines to finish. The provided context can be used to cancel everything
 // and exit the loop.
-func Run(ctx context.Context, limit int, nextTask func(context.Context) Task) {
+func Run(ctx context.Context, limit int, nextTask func(context.Context) Task) error {
```
jba: The error-handing/cancelation pattern that you do in ReceiveConcurrently is actually more subtle that what you're doing here. Would you consider putting that logic here? So both nextTask and Task would return error, and cause Run to exit with that error after waiting for outstanding goroutines.

samples/pubsub/workerpool/main.go
```diff
@@ -0,0 +1,75 @@
...
+
+// subscribe pulls from the subscription and simulates processing the videos.
+func subscribe(ctx context.Context, sub *pubsub.Subscription) error {
+	return psutil.ReceiveConcurrently(ctx, 10, sub, func(ctx context.Context, m *pubsub.Message) error {
+		videoURL := string(m.Body)
+		log.Printf("Processing %v", videoURL)
+		// Simulate transcoding the video and storing it in various formats.
+		time.Sleep(time.Duration(rand.Intn(10)) * time.Second)
+		log.Printf("Done processing %v", videoURL)
+		return nil
```
jba: Add a comment that returning nil will ack.

pubsub/psutil/psutil.go
```diff
@@ -0,0 +1,57 @@
...
+	"gocloud.dev/pubsub"
+)
+
+// ReceiveConcurrently repeatedly fetches messages from the given subscription
+// sub, calling handleMessage on each message with up to max concurrent calls.
+// If handleMessage returns nil for its error, then the message is acked. If
+// sub.Receive or handleMessage returns a non-nil error then the processing is
+// terminated via context cancellation. Cancelling the context causes this
+// function to return.
+func ReceiveConcurrently(ctx context.Context, max int, sub *pubsub.Subscription, handleMessage func(ctx context.Context, m *pubsub.Message) error) error {
```
jba: Nit: I think of the subscription as the more "important" arg, so I feel it should go before max.

pubsub/pubsub.go
```diff
@@ -534,3 +536,30 @@ func wrapError(ec errorCoder, err error) error {
 	}
 	return gcerr.New(ec.ErrorCode(err), err, 2, "pubsub")
 }
+
+// RunWorkerPool repeatedly fetches messages from the given subscription sub, in an
+// inverted worker pool (link) with up to maxWorkers. The callback handleMessage is
+// called on each message. If handleMessage return nil for its error, then the
+// message is acked. If sub.Receive or handleMessage returns a non-nil error then
+// the workerpool is terminated via context cancellation.
+func RunWorkerPool(ctx context.Context, maxWorkers int, sub *Subscription, handleMessage func(ctx context.Context, m *Message) error) error {
```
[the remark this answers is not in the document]

pubsub/pubsub.go
```diff
@@ -534,3 +536,30 @@ func wrapError(ec errorCoder, err error) error {
 	}
 	return gcerr.New(ec.ErrorCode(err), err, 2, "pubsub")
 }
+
+// RunWorkerPool repeatedly fetches messages from the given subscription sub, in an
+// inverted worker pool (link) with up to maxWorkers. The callback handleMessage is
+// called on each message. If handleMessage return nil for its error, then the
```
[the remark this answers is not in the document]

pubsub/psutil/psutil.go
```diff
@@ -0,0 +1,57 @@
...
+	errc := make(chan error, max+1)
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	go func() {
+		errc <- workerpool.Run(ctx, max, func(ctx context.Context) workerpool.Task {
+			m, err := sub.Receive(ctx)
+			log.Printf("Receive returned %v for err", err)
```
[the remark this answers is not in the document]
ijt: Done.

samples/pubsub/workerpool/main.go
ijt: Done.

pubsub/psutil/psutil.go
ijt: Done.

internal/workerpool/pool.go
ijt: Seems reasonable. I'll try it.

samples/pubsub/workerpool/main.go
zombiezen: The code that I linked is subtly wrong in that it does not synchronize or shutdown the health check listening port. I would expected almost all pub/sub subscriber processes running in an orchestrated environment would need to write similar code.

That said, I don't have enough insight here to give any compelling advice, just lots of unorganized input. Consider my two cents added. :)

internal/workerpool/pool.go
ijt: Done. PTAL.

pubsub/psutil/psutil.go
ijt: Done.

pubsub/psutil/psutil.go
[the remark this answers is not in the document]
ijt: Done.

samples/pubsub/workerpool/main.go
ijt: At a glance I figure an errgroup would do the trick.

internal/workerpool/pool.go
```diff
@@ -20,32 +20,54 @@ package workerpool
...
 
 // Run runs a worker pool with no more than limit goroutines. It repeatedly
 // calls nextTask to get tasks to perform. These tasks are in the form of funcs
 // that do some work via side effects. If nextTask returns nil, Run will no
 // longer ask for more tasks, and will return after waiting for the running
 // goroutines to finish. The provided context can be used to cancel everything
 // and exit the loop.
-func Run(ctx context.Context, limit int, nextTask func(context.Context) Task) {
+func Run(ctx context.Context, limit int, nextTask func(context.Context) (Task, error)) error {
+	ctx, cancel := context.WithCancel(ctx)
```
jba: Sweet. Just factor out the error-setting code:
```
setErr := func(err error) { mu.Lock(); blah blah; mu.Unlock() }
```

internal/workerpool/pool.go
```diff
@@ -20,32 +20,54 @@ package workerpool
...
+			if retErr == nil {
+				retErr = ctx.Err()
+			}
+			mu.Unlock()
 			break Loop
 		case sem <- token{}:
 		}
 		go func() {
-			doTask(ctx)
+			err := doTask(ctx)
```
jba:
```
if err := doTask(ctx); err != nil {
   setErr(err)
   cancel()
}
```

internal/workerpool/pool_test.go
```diff
@@ -83,4 +95,41 @@ func TestRun(t *testing.T) {
...
+				return wantErr
+			}
+			return task, nil
+		}
+		err := workerpool.Run(ctx, 1, nextTask)
+		if err != wantErr {
+			t.Errorf(`workerpool.Run returned "%v" for error, want "%v"`, err, wantErr)
+		}
+		t.Log("waiting for goroutines to return")
+		for runtime.NumGoroutine() != ng0 {
```
jba: This might turn out to be fragile. Who knows when the runtime decides to create a long-lived goroutine?

But leave it and I guess we'll see.

pubsub/psutil/psutil.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+// ReceiveConcurrently repeatedly fetches messages from the given subscription
+// sub, calling handleMessage on each message with up to max concurrent calls.
+// If handleMessage returns nil for its error, then the message is acked. If
+// sub.Receive or handleMessage returns a non-nil error then the processing is
+// terminated via context cancellation. Cancelling the context causes this
+// function to return.
+func ReceiveConcurrently(ctx context.Context, sub *pubsub.Subscription, max int, handleMessage func(ctx context.Context, m *pubsub.Message) error) error {
+	return workerpool.Run(ctx, max, func(ctx context.Context) (workerpool.Task, error) {
```
jba: Wow, nice!

pubsub/psutil/psutil_test.go
```diff
@@ -0,0 +1,145 @@
...
+		sub := mempubsub.NewSubscription(top, time.Second)
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel()
+		err := psutil.ReceiveConcurrently(ctx, sub, 10, func(ctx context.Context, m *pubsub.Message) error {
+			return nil // ack
+		})
+		if err == nil {
+			t.Fatal("got nil, want error")
+		}
+		if !strings.Contains(err.Error(), "cancel") {
```
jba: Better to check if `err == context.Canceled`.

pubsub/psutil/psutil_test.go
```diff
@@ -0,0 +1,145 @@
...
+				if n == 0 {
+					break
+				}
+				time.Sleep(time.Millisecond)
+			}
+
+			t.Logf("stopping the worker pool")
+			cancel()
+			err := <-errc
+			if !strings.Contains(err.Error(), "canceled") {
```
jba: ditto

samples/pubsub/workerpool/main.go
```diff
@@ -0,0 +1,75 @@
...
+	"time"
+
+	"gocloud.dev/pubsub"
+	"gocloud.dev/pubsub/mempubsub"
+	"gocloud.dev/pubsub/psutil"
+	"golang.org/x/sync/errgroup"
+)
+
+func main() {
+	ctx, _ := context.WithTimeout(context.Background(), time.Duration(10)*time.Second)
```
jba: You don't need the `time.Duration` cast for a literal.

internal/workerpool/pool.go
ijt: Done.

internal/workerpool/pool.go
ijt: Done.

internal/workerpool/pool_test.go
ijt: That's why I don't call `t.Parallel()` at the top of these tests. I've never had a problem with this, but I guess if I do then I'll learn something.

pubsub/psutil/psutil_test.go
ijt: I wanted to do that, but that means I don't get to add a little bit of context using `fmt.Errorf`. Then it becomes hard to figure out where the error came from.

pubsub/psutil/psutil_test.go
ijt: Same response.

samples/pubsub/workerpool/main.go
ijt: Done.

ijt:
PTAL

internal/workerpool/pool.go
```diff
@@ -20,32 +20,46 @@ package workerpool
...
-func Run(ctx context.Context, limit int, nextTask func(context.Context) Task) {
+func Run(ctx context.Context, limit int, nextTask func(context.Context) (Task, error)) error {
+	ctx, cancel := context.WithCancel(ctx)
 	type token struct{}
 	sem := make(chan token, limit)
+	var mu sync.Mutex
+	var retErr error
+	setErr := func(err error) {
+		mu.Lock()
+		retErr = err
```
jba: What happened to `if retErr == nil`? I think you want that to capture the first error.

internal/workerpool/pool.go
```diff
@@ -20,32 +20,46 @@ package workerpool
...
+		mu.Unlock()
+	}
 Loop:
 	for {
-		doTask := nextTask(ctx)
-		if doTask == nil {
+		doTask, err := nextTask(ctx)
+		if err != nil {
+			setErr(err)
 			break
```
jba: You need to cancel here too.

internal/workerpool/pool.go
ijt: Done.

internal/workerpool/pool.go
ijt: Done.

ijt:
PTAL

ijt:
Test failure fixed. PTAL.

internal/workerpool/pool.go
```diff
@@ -39,8 +40,11 @@ func Run(ctx context.Context, limit int, nextTask func(context.Context) (Task, e
 	var retErr error
 	setErr := func(err error) {
 		mu.Lock()
-		retErr = err
+		if retErr == nil {
+			retErr = fmt.Errorf("workerpool: %v", err)
```
jba: I don't think it's valuable to wrap here -- you lose too much information for too little gain. (workerpool is an internal package, so identifying it doesn't really help much.) Just use err.

vangent:
For what it's worth, I tried using `workerpool` in #1344 and found it awkward. 

* There was no way to deal with errors in the Task (I think this PR is fixing that).
* The "next" function was hard to write -- I have a slice of things I want to operate on, and it wasn't obvious how to do that. I ended up doing something with an `atomic` index but it was confusing.
* There were too many layers of functions (I call `Run` and pass it a function that returns a function....

It ended up being shorter to just do it myself.

internal/workerpool/pool.go
ijt: Okay, reverted it.

ijt:
@vangent, yes this PR is fixing the lack of error handling. It might be worth another try after this is in. I'd be curious to know how it compares. I don't think you need an atomic index. Getting the tasks isn't done concurrently. That's straight sequential code.

ijt:
PTAL

ijt:
@vangent are you not approving this because you disagree with the whole thing and want it removed, or is there a specific part of it you would like to see changed?

pubsub/psutil/psutil.go
```diff
@@ -0,0 +1,46 @@
...
+		m, err := sub.Receive(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("receiving message: %v", err)
+		}
+		task := func(ctx context.Context) error {
+			if err := handleMessage(ctx, m); err != nil {
+				return fmt.Errorf("handling message: %v", err)
```
vangent: no ack here?

vangent:
> @vangent are you not approving this because you disagree with the whole thing and want it removed, or is there a specific part of it you would like to see changed?

TBH there is too much going on in this PR. Could it be broken up? It'd be nice to see `workerpool` used in a relatively simple example like your video one. I found it hard to use and I need convincing that it's easier to understand/use than a user just creating their own semaphore + loop.

vangent:
@ijt decided to drop this.

Issue #1206, opened 2019-01-29
secrets: classify AWS KMS and vault error codes


shantuo:

shantuo:
AWS is added and Vault doesn't currently have error codes that provides much value to us. Closing this for now.

PR #1297, opened 2019-02-12
secrets/awskms: support relevant error codes

shantuo:
Support error codes used for Decrypt and Encrypt.

Update #1206

### DIFF
--- a/secrets/awskms/kms.go
+++ b/secrets/awskms/kms.go
@@ -115,11 +115,11 @@ var errorCodeMap = map[string]gcerrors.ErrorCode{
 	kms.ErrCodeNotFoundException:          gcerrors.NotFound,
 	kms.ErrCodeDisabledException:          gcerrors.PermissionDenied,
 	kms.ErrCodeInvalidCiphertextException: gcerrors.InvalidArgument,
-	kms.ErrCodeKeyUnavailableException:    gcerrors.NotFound,
+	kms.ErrCodeKeyUnavailableException:    gcerrors.ResourceExhausted,
 	kms.ErrCodeDependencyTimeoutException: gcerrors.DeadlineExceeded,
 	kms.ErrCodeInvalidGrantTokenException: gcerrors.PermissionDenied,
 	kms.ErrCodeInternalException:          gcerrors.Internal,
-	kms.ErrCodeInvalidStateException:      gcerrors.InvalidArgument,
+	kms.ErrCodeInvalidStateException:      gcerrors.FailedPrecondition,
 	kms.ErrCodeInvalidKeyUsageException:   gcerrors.InvalidArgument,
 }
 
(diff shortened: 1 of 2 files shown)


secrets/awskms/kms.go
```diff
@@ -98,9 +99,28 @@ func (k *keeper) ErrorAs(err error, i interface{}) bool {
...
+		return gcerr.Unknown
+	}
+	return ec
+}
+
+var errorCodeMap = map[string]gcerrors.ErrorCode{
+	kms.ErrCodeNotFoundException:          gcerrors.NotFound,
+	kms.ErrCodeDisabledException:          gcerrors.PermissionDenied,
+	kms.ErrCodeInvalidCiphertextException: gcerrors.InvalidArgument,
+	kms.ErrCodeKeyUnavailableException:    gcerrors.NotFound,
```
jba: Are you sure this means it's permanently gone, not just temporarily unavailable? If the latter, then we should probably use a different error.

secrets/awskms/kms.go
```diff
@@ -98,9 +99,28 @@ func (k *keeper) ErrorAs(err error, i interface{}) bool {
...
+	kms.ErrCodeDependencyTimeoutException: gcerrors.DeadlineExceeded,
+	kms.ErrCodeInvalidGrantTokenException: gcerrors.PermissionDenied,
+	kms.ErrCodeInternalException:          gcerrors.Internal,
+	kms.ErrCodeInvalidStateException:      gcerrors.InvalidArgument,
```
jba: This one sounds more like FailedPrecondition.  It's not that the request was bad, it just assumed something about the system that isn't true.

secrets/awskms/kms.go
shantuo: Not permanently gone, it can be retried. I'm not sure which one fits it properly. Maybe ResourceExhausted?

secrets/awskms/kms.go
shantuo: Done.

shantuo:
Thanks @vangent , What nit comment?

secrets/awskms/kms.go
```diff
@@ -98,9 +99,28 @@ func (k *keeper) ErrorAs(err error, i interface{}) bool {
...
+		return gcerr.Unknown
+	}
+	ec, ok := errorCodeMap[ae.Code()]
+	if !ok {
+		return gcerr.Unknown
+	}
+	return ec
+}
+
+var errorCodeMap = map[string]gcerrors.ErrorCode{
```
vangent: Nit: can you organize these by value? So all of the `gcerrors.PermissionDenied` keys are grouped together. I think that makes it easier to read.

vangent:
> Thanks @vangent , What nit comment?

Hmm, I guess it didn't get posted. I re-added the comment.

secrets/awskms/kms.go
shantuo: Done.

secrets/awskms/kms.go
jba: Yes, exactly.

Issue #770, opened 2018-11-20
all: use a better replay tool


jba:
With the existing replay tool, order of RPCs matters, making concurrent tests unusable. The tool also doesn't have a way to capture state that may vary between runs, like times and random numbers.

Replace with `cloud.google.com/go/httpreplay` and/or `cloud.google.com/go/rpcreplay`.

vangent:
The underlying replace tool we're using, `github.com/dnaeon/go-vcr`, doesn't enforce ordering, but we [added it explicitly](https://github.com/google/go-cloud/blob/master/internal/testing/replay/replay.go#L60). The default non-ordered matching was not sufficient to write tests. For example, a test like this:

1. Read blob "foo" --> expect error
2. Write blob "foo"
3. Read blob "foo" --> expect success
4. Delete blob "foo"
5. Read blob "foo" --> expect failure

has the same HTTP request for "read blob foo" three times, but with three different responses.

We already use `cloud.google.com/go/rpcreplay` for non-HTTP replay. I am not sure why we didn't use `cloud.google.com/go/httpreplay` to begin with, but there may be a good reason.

That said, I would be delighted if we could simplify this and make it less fragile for concurrent tests.

jba:
> I am not sure why we didn't use `cloud.google.com/go/httpreplay` to begin with, but there may be a good reason.

An excellent one: it didn't exist.

jba:
> the same HTTP request for "read blob foo" three times, but with three different response

I think (hope) that my tool would look for responses sequentially. The only way to screw it up would (should) be if you wrote the same request concurrently; but then your test would have a race condition, I believe.

I'll try to replace go-vcr with httpreplay and see how it goes.

vangent:
@eliben is working on this, he's already converted GCS and S3.

vangent:
Only a couple more before we can delete the go-vcr dependency!

eliben:
TODO: factor out creation of replay directory - it's repeated in all the users of httpreplay

eliben:
This is currently blocked by #1971

eliben:
... and blocked (to some degree) by #1973

PR #1298, opened 2019-02-12
awspubsub: use new httpreplay package for record/replay

### LINKED ISSUE #1233: all: fix recorder/replayer so awspubsub doesn't have to skip conformance tests in replay mode
TestSendReceive, TestSendReceiveTwo and TestAs were hanging or panicking in replay mode on awspubsub, so they are currently being skipped in replay mode. It looks as though there is an issue in the recorder that is causing this problem.

jba:
Added NewAWSSession2 to testing/setup, which behaves just like
NewAWSSession except it uses the replayer at
cloud.google.com/go/httpreplay.

Used the new replayer for the awspubsub package.

This fixes whatever problems were occurring on replay,
so I also removed the shouldSkip logic.

Fixes #1233.
Updates #770.


internal/testing/setup/setup.go
```diff
@@ -55,24 +59,75 @@ func NewAWSSession(t *testing.T, region string) (sess *session.Session, rt http.
...
+			t.Fatal(err)
+		}
+		rec, err := httpreplay.NewRecorder(path, nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+		rec.RemoveQueryParams("X-Amz-Credential", "X-Amz-Signature")
+		rec.RemoveRequestHeaders("Authorization", "Duration", "X-Amz-Security-Token")
+		rec.ClearHeaders("X-Amz-Date")
+		rec.ClearHeaders("User-Agent") // AWS includes the Go version
```
vangent: What does this new replayer use for matching? Does it include the `User-Agent` request header and that's why this needs to be cleared?

internal/testing/setup/setup.go
```diff
@@ -55,24 +59,75 @@ func NewAWSSession(t *testing.T, region string) (sess *session.Session, rt http.
...
+			if err := rec.Close(); err != nil {
+				t.Fatal(err)
+			}
+		}
+	} else { // replay
+		rep, err := httpreplay.NewReplayer(path)
+		if err != nil {
+			t.Fatal(err)
+		}
+		c, err := rep.Client(context.Background())
```
vangent: Maybe `ctx` should be passed in...

vangent: (another one above)

internal/testing/setup/setup.go
jba: My tool take the opposite approach of go-vcr: it matches all headers unless explicitly told not to. I feel that reduces the chance for false matches. So yes, if I didn't mention `User-Agent` here it would be matched, and that would fail.

internal/testing/setup/setup.go
jba: Done.

jba:
PTAL.

jba:
@ijt ping

ijt:
Looking now.

internal/testing/setup/setup.go
```diff
@@ -55,24 +59,75 @@ func NewAWSSession(t *testing.T, region string) (sess *session.Session, rt http.
...
+		rec.RemoveQueryParams("X-Amz-Credential", "X-Amz-Signature")
+		rec.RemoveRequestHeaders("Authorization", "Duration", "X-Amz-Security-Token")
+		rec.ClearHeaders("X-Amz-Date")
+		rec.ClearHeaders("User-Agent") // AWS includes the Go version
+		c, err := rec.Client(ctx, option.WithoutAuthentication())
+		if err != nil {
+			t.Fatal(err)
+		}
+		rt = c.Transport
+		done = func() {
```
ijt: `done` sounds like a boolean. How about `cleanup`?

jba: er, done.

jba:
I found two problems with the tests:
- One of the queues in the TestSendReceiveTwo test was misconfigured with the wrong topic.
- The SubscriptionErrorCheck had the wrong error.
Hopefully they'll work better now.

PR #1308, opened 2019-02-12
secrets/vault: support ErrorAs

shantuo:
Vault recently added a type that we can expose to our user, it prints the curl command of the bad request for debugging, I think it can be useful to support it in ErrorAs.

Note the change in go.sum: I need to update this to make it compile due to some early Go 1.11 bug which computes the checksum differently.

Also, since the type is newly added, I need to update the go.mod to bump the vault version.


shantuo:
Dependencies fixed.

Tests failed because all API calls would return a non-nil error if we switched on the `OutputCurlString` field, which allowed us to cast the error type to `OutputStringError`. The workaround is to use vault-specific test to test this `ErrorAs` and keep the conformance tests as is.

WDYT @vangent ?

vangent:
Based on https://github.com/hashicorp/vault/blob/master/api/client.go#L718-L721, it looks to me like the Vault client doesn't even make an RPC at all when you set this flag. Instead, it figures out what the `curl` command for that RPC would be and returns it. That doesn't sound useful for us.

shantuo:
Correct, it appears to be an option for some debugging only usage. Let's drop this for now.

PR #1309, opened 2019-02-12
samples/tutorial: rewrite using blob.OpenBucket

zombiezen:
Cuts out about half the tutorial to get to the punchline.

Updates #1174


zombiezen:
Kool-Aid Man saying Oh, Yeahh!

Issue #1314, opened 2019-02-13
runtimevar: add support for Azure


vangent:
* Investigate Azure equivalent to AWS Param Store or GCP Runtime Configurator.
* If any are found, add `runtimevar` for at least one.

vangent:
Note: if we add something here, we should update the Azure section of the Guestbook sample to use it. Today it is using `blobvar` based on a blob in Azure Storage.

eliben:
@vsaroopchand is this something you might be interested in looking at?

Issue #1305, opened 2019-02-12
sql: add support for Azure SQL providers


vangent:
Investigate what SQL options Azure supports (SQLServer? hosted MySQL? hosted Postgres?) and add support for them similar to what we have for AWS and GCP.

For SQL support, we basically just have helpers for opening the existing `sql.DB` type.

@vsaroopchand in case you're interested.
@zombiezen who has done most of the existing support.

vangent:
Once this is fixed, we should also update the Azure part of the Guestbook sample to use a hosted SQL offering, and at that point we should also be able to run guestbook from an Azure VM (currently, it's running locally connecting to a local MySQL instance).

vsaroopchand:
Relational or Non-relational? The Azure Relational DB (AWS RDS equivalent) is Azure SQL which is a PaaS SQL service. The Non-relational service is CosmosDB which offers a MonboDb Driver Impl. along with a non-sql API. The CosmosDB sample is [here](https://github.com/Azure-Samples/azure-sdk-for-go-samples/tree/master/cosmosdb); the Azure SQL sample is [here](https://github.com/Azure-Samples/azure-sdk-for-go-samples/tree/master/sql)

zombiezen:
This is specifically referring to relational databases. Does Azure Relational DB only support MSSQL, or does it host MySQL/PostgreSQL as well?

vsaroopchand:
There are managed services for both [PostgresSQL ](https://azure.microsoft.com/en-us/services/postgresql/)and [MySQL](https://azure.microsoft.com/en-us/services/mysql/). I am going to test the cloudpostgres package with an Azure PostgresSQL instance to see if that can be refactored to support Azure. There might be some differences in ports, etc. but should be possible unless you prefer a separate package for Azure PostgresSQL?

zombiezen:
Thanks for the links. From quickly browsing the docs, it looks like Azure's offering is much closer to RDS than Cloud SQL in terms of code required: Azure serves using a [CA not in the roots](https://docs.microsoft.com/en-us/azure/mysql/howto-configure-ssl). The connection code (which should be in a separate package) should have a hook for the CA certificate fetching, and then open the connection with OpenCensus instrumentation.

vsaroopchand:
I'll take this up now that Secrets is merged.

PR #1316, opened 2019-02-13
samples/guestbook: add the ability to run Guestbook on Azure

### LINKED ISSUE #1122: samples/guestbook: add azure to the guestbook sample

vangent:
Fixes #1122.

Note that this is only a halfway step:

* We're running the binary and MySQL locally. Until we support some kind of SQL offering for Azure, we can't easily run on an VM at Azure. #1305 will fix this.

* For `runtimevar`, we're using a blob in Azure Storage, so only kind of cheating. #1314 will fix this if there's an Azure equivalent of of ParamStore/RuntimeConfigurator.

I've updated both of those bugs with a note to update Guestbook once they are fixed.

At least one reviewer (if not both) should actually follow the instructions and make sure they can successfully get it running.


samples/guestbook/README.md
```diff
@@ -185,6 +188,77 @@ directory using the same variables you entered during `terraform apply`.
...
+storage_container = [redacted]
+```
+
+### Running
+
+You will need to run a local MySQL database server, similar to what we did for
+running locally earlier.
+
+```shell
+go get ./localdb/... # Get package dependencies.
```
clausti: Minor note: having previously been directed to `cd` into the azure directory, one is in the wrong directory to run `go get ./localdb/...`. Might be worth adding an instruction to go up one level?

samples/guestbook/README.md
```diff
@@ -185,6 +188,77 @@ directory using the same variables you entered during `terraform apply`.
...
+go run localdb/main.go
+```
+
+In another terminal, add your Azure credentials to the environment and run the
+`guestbook` application:
+
+```shell
+# Enter the storage_account from the Terraform output earlier.
+AZURE_STORAGE_ACCOUNT=<your storage_account>
```
clausti: add `export` to the beginning of these lines?

vangent:
> Got it up and running! And followed the instructions successfully all the way through `terraform destroy`.

Great, thanks!

> Not changed in this PR so I can't comment on it directly, but it might be worth adding that Docker "Desktop" is the correct tool to install.

Done.

samples/guestbook/README.md
vangent: Done.

samples/guestbook/README.md
vangent: Done.

vangent:
@zombiezen PTAL

Issue #1302, opened 2019-02-12
blob: example for Open doesn't show necessary blank import


zombiezen:
### Describe the bug

`ExampleOpen` (soon to be `ExampleOpenBucket` after #1282) does not include the blank import of `gocloud.dev/blob/fileblob` that it requires.

## Expected behavior

Example should show `import _ "gocloud.dev/blob/fileblob"`, ideally in code, but a comment is okay.

### Version

At head (32d84ffde2518cb3b4d818b1d0d7ebf02fda64fe)

PR #1317, opened 2019-02-13
blob: split the openbucket example

shantuo:
This shows the necessary blank import line when using `OpenBucket` only. Also added some comments inside the example.

Fixes: #1302

### DIFF
--- a/blob/example_openbucket_test.go
+++ b/blob/example_openbucket_test.go
@@ -33,7 +33,7 @@ func ExampleOpenBucket() {
 	// memblob registers for the "mem" scheme.
 
 	ctx := context.Background()
-	b, err := blob.OpenBucket(ctx, "mem://mybucket")
+	b, err := blob.OpenBucket(ctx, "mem://")
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/blob/example_openbucket_test.go
+++ b/blob/example_openbucket_test.go
@@ -18,41 +18,27 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"net/url"
-	"path/filepath"
-	"strings"
 
 	"gocloud.dev/blob"
-	_ "gocloud.dev/blob/fileblob"
+	_ "gocloud.dev/blob/memblob"
 )
 
 func ExampleOpenBucket() {
 
 	// Connect to a bucket using a URL.
-	// This example uses "fileblob", the file-based implementation.
-	// We need to add a blank import line to register the fileblob provider's
+	// This example uses "memblob", the in-memory implementation.
+	// We need to add a blank import line to register the memblob provider's
 	// URLOpener, which implements blob.BucketURLOpener:
-	// import _ "gocloud.dev/blob/fileblob"
-	// fileblob registers for the "file" scheme.
-	dir, cleanup := newTempDir()
-	defer cleanup()
+	// import _ "gocloud.dev/blob/memblob"
+	// memblob registers for the "mem" scheme.
 
-	// Ensure the path has a leading slash; fileblob ignores the URL's
-	// Host field, so URLs should always start with "file:///". On
-	// Windows, the leading "/" will be stripped, so "file:///c:/foo"
-	// will refer to c:/foo.
-	urlpath := url.PathEscape(filepath.ToSlash(dir))
-	if !strings.HasPrefix(urlpath, "/") {
-		urlpath = "/" + urlpath
-	}
 	ctx := context.Background()
-	b, err := blob.OpenBucket(ctx, "file://"+urlpath)
+	b, err := blob.OpenBucket(ctx, "mem://mybucket")
 	if err != nil {
 		log.Fatal(err)
 	}
-	fmt.Println("Got a bucket for valid path")
 
-	// Now we can use b to read or write files to the container.
+	// Now we can use b to read or write blob to the container.
 	if err := b.WriteAll(ctx, "my-key", []byte("hello world"), nil); err != nil {
 		log.Fatal(err)
 	}
@@ -62,6 +48,5 @@ func ExampleOpenBucket() {
 	}
 	fmt.Println(string(data))
 	// Output:
-	// Got a bucket for valid path
 	// hello world
 }
--- a/blob/example_openbucket_test.go
+++ b/blob/example_openbucket_test.go
@@ -45,11 +45,23 @@ func ExampleOpenBucket() {
 	if !strings.HasPrefix(urlpath, "/") {
 		urlpath = "/" + urlpath
 	}
-	if _, err := blob.OpenBucket(context.Background(), "file://"+urlpath); err != nil {
+	ctx := context.Background()
+	b, err := blob.OpenBucket(ctx, "file://"+urlpath)
+	if err != nil {
 		log.Fatal(err)
 	}
 	fmt.Println("Got a bucket for valid path")
 
+	// Now we can use b to read or write files to the container.
+	if err := b.WriteAll(ctx, "my-key", []byte("hello world"), nil); err != nil {
+		log.Fatal(err)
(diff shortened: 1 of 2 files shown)


vangent:
@zombiezen you have a bunch of reviews going on so I'll take this one if you don't mind (if you do, feel free to add yourself back).

blob/example_openbucket_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"log"
+	"net/url"
+	"path/filepath"
+	"strings"
+
+	"gocloud.dev/blob"
+	_ "gocloud.dev/blob/fileblob"
+)
+
+func ExampleOpenBucket() {
```
vangent: Does this work with godoc given that there's no function named `OpenBucket`? Where does the example show up? Maybe you need "Example_openBucket" instead?

blob/example_openbucket_test.go
```diff
@@ -0,0 +1,64 @@
...
+	// Connect to a bucket using a URL.
```
vangent: I would merge these 4 lines with the ones down below, something like:

```
// Open a bucket using a URL.
// This example uses "fileblob", the file-based implementation.
// We need to add a blank import line to register the fileblob provider's
blob.BucketURLOpener:
// import _ "gocloud.dev/blob/fileblob"
// fileblob registers for the "file" scheme.
```

blob/example_openbucket_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	ctx := context.Background()
+
+	// The blob package is designed for using with a specific provider package.
+	// Here blob.OpenBucket needs a provider that implements blob.BucketURLOpener
+	// to work. In this example, we use blob/fileblob, so we need to add a blank
+	// import line to register the fileblob provider:
+	// import _ "gocloud.dev/blob/fileblob"
+	if _, err := blob.OpenBucket(ctx, "file:///nonexistentpath"); err == nil {
+		log.Fatal("Expected an error opening nonexistent path")
```
vangent: I would leave the negative case out (or move it to a test if we don't have coverage). Per previous feedback from the team, the examples should show code that is working and that is copy-pastable.

blob/example_openbucket_test.go
vangent: Or maybe ExampleOpenBucketURL?

blob/example_openbucket_test.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	fmt.Println("Got expected error opening a nonexistent path")
+
+	// Ensure the path has a leading slash; fileblob ignores the URL's
```
vangent: Also note that on Windows, `dir` has `\\` characters, and we need to convert them to "/" for the URL path.

blob/example_openbucket_test.go
```diff
@@ -0,0 +1,64 @@
...
+	// Host field, so URLs should always start with "file:///". On
```
vangent: I'm not sure "should always start with ///" is accurate. On Linux, for absolute paths it is true, but you could theoretically pass a relative path.

How about:

var url string
if os.PathSeparator == "/" {
  // Fileblob ignores the "Host" field of the URL; the directory
  // is the URL path. So, URLs for absolute paths like the one we're
  // using here will start with "file:///".
  url = path.Join("file://", dir)
} else {
  // On Windows, we need to convert "\\" to "/" for use in a URL, and
  // add a leading "/". Fileblob will strip the leading "/" and
  // convert the "/" back to "\\".
  url = "file:///" + filepath.ToSlash(dir)
}
```

WDYT?

File: blob/example_openbucket_test.go [hunk 6]
```diff
@@ -0,0 +1,64 @@
...
+	// Windows, the leading "/" will be stripped, so "file:///c:/foo"
+	// will refer to c:/foo.
+	urlpath := url.PathEscape(filepath.ToSlash(dir))
+	if !strings.HasPrefix(urlpath, "/") {
+		urlpath = "/" + urlpath
+	}
+	if _, err := blob.OpenBucket(ctx, "file://"+urlpath); err != nil {
+		log.Fatal(err)
+	}
+	fmt.Println("Got a bucket for valid path")
```
vangent: Considering using `WriteAll` to write something to the bucket?
https://github.com/google/go-cloud/blob/master/blob/fileblob/example_test.go#L75-L83

zombiezen:
Thanks @vangent

blob/example_openbucket_test.go
shantuo: Done.

blob/example_openbucket_test.go
shantuo: I think this should be handled inside the `OpenBucket` function, do you mean we want to add comments? There are comments at the function definition.

blob/example_openbucket_test.go
shantuo: The current way we handle this will ignore the leading `.`, `~` as we only use `URL.Path`. If we think this is something we want to support, we might want to adjust the function behavior.

blob/example_openbucket_test.go
```diff
@@ -0,0 +1,67 @@
...
+	urlpath := url.PathEscape(filepath.ToSlash(dir))
+	if !strings.HasPrefix(urlpath, "/") {
+		urlpath = "/" + urlpath
+	}
+	ctx := context.Background()
+	b, err := blob.OpenBucket(ctx, "file://"+urlpath)
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Println("Got a bucket for valid path")
```
vangent: I would drop this `Println`.

blob/example_openbucket_test.go
vangent: I mean that if you run this code on Windows, `dir` will look something like `C:\\Temp\foo`. You need to convert those `\\` characters to `/` to put it in the URL path. `OpenBucket` will then convert them back to `\\` for you (on Windows). You are doing this with `ToSlash` now, but it's worth a comment.

blob/example_openbucket_test.go
vangent: Thinking this over, it can be quite confusing and it's not really that important here for this `blob`-level example.

How about we move this discussion to the `fileblob/example_test` and `fileblob/fileblob_test`? I am not sure what you mean by the leading "." -- it might be good to add a testcase for that to be more explicit about it.

Meanwhile, we can change this example to either use `memblob` (URL will be `mem://`) or `gcsblob` (removing `// Output` line so example only compiles).

blob/example_openbucket_test.go
shantuo: Done.

blob/example_openbucket_test.go
shantuo: Got it, this is stale though. Will document it in `fileblob`'s example.

blob/example_openbucket_test.go
shantuo: Changed to use memblob since this example is to demonstrate open by URL.

Filed #1425 for more investigation.

blob/example_openbucket_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Connect to a bucket using a URL.
+	// This example uses "memblob", the in-memory implementation.
+	// We need to add a blank import line to register the memblob provider's
+	// URLOpener, which implements blob.BucketURLOpener:
+	// import _ "gocloud.dev/blob/memblob"
+	// memblob registers for the "mem" scheme.
+
+	ctx := context.Background()
+	b, err := blob.OpenBucket(ctx, "mem://mybucket")
```
vangent: Nit: leave off the `mybucket`, it's not needed.

shantuo: Done.

PR #1322, opened 2019-02-14
all: return error from concrete type for non-UTF8 strings

vangent:
Applied to the same set of strings we're escaping:

* blob key
* blob metadata key & value
* pubsub metdata key & value

Verified with conformance tests.

Fixes #1322.


blob/blob.go
```diff
@@ -662,6 +667,12 @@ func (b *Bucket) NewWriter(ctx context.Context, key string, opts *WriterOptions)
 			if k == "" {
 				return nil, errors.New("blob.NewWriter: WriterOptions.Metadata keys may not be empty strings")
 			}
+			if !utf8.ValidString(k) {
+				return nil, fmt.Errorf("blob.NewWriter: WriterOptions.Metadata keys must be valid UTF-8 strings: %q", k)
+			}
+			if !utf8.ValidString(v) {
+				return nil, fmt.Errorf("blob.NewWriter: WriterOptions.Metadata keys must be valid UTF-8 strings: %q", v)
```
jba: s/keys/values/

pubsub/pubsub.go
```diff
@@ -157,6 +158,14 @@ func (t *Topic) Send(ctx context.Context, m *Message) (err error) {
 	if err != nil {
 		return err // t.err wrapped when set
 	}
+	for k, v := range m.Metadata {
+		if !utf8.ValidString(k) {
+			return fmt.Errorf("pubsub.Send: Message.Metadata keys must be valid UTF-8 strings: %q", k)
+		}
+		if !utf8.ValidString(v) {
+			return fmt.Errorf("pubsub.Send: Message.Metadata keys must be valid UTF-8 strings: %q", v)
```
jba: s/keys/values/

vangent: Thanks, done.

blob/blob.go

vangent:
> What about keys passed elsewhere (attributes, NewRangeReader, etc.)?

For attributes, I haven't added any escaping for those yet. I'm not that worried about it since I don't think there's any expectation that user-invented strings should work (e.g., the reasonable values are defined by some spec). If it becomes a problem we can add validation of those fields, but it's likely to be much more than just "is UTF8".

For NewRangeReader, I expect that they are going to get some kind of error anyway since there's no way the blob exists with a non-UTF8-string key. Do you think it's worth adding a check (and also in Delete, etc.) to short-circuit?

jba:
> I expect that they are going to get some kind of error anyway since there's no way the blob exists with a non-UTF8-string key.

Is that true across all providers? None of them support binary key names?

vangent:
> > I expect that they are going to get some kind of error anyway since there's no way the blob exists with a non-UTF8-string key.
> 
> Is that true across all providers? None of them support binary key names?

Pretty sure. Well, memblob probably does. It's probably safer to verify, so I've added that.

vangent:
PTAL

Issue #1328, opened 2019-02-14
pubsub: subscribing queues to topics on AWS conformance tests is broken


ijt:
See https://github.com/aws/aws-sdk-go/issues/2415. That issue has a response with a possible solution.

vangent:
I've been trying to add a new test (for non-UTF8 bodies), and I can't even get the tests to create subscriptions anymore.

It created the topic, but no subscription....

ijt:
Looking into it.

vangent:
> It created the topic, but no subscription....
I was looking under SNS instead of SQS. To manually subscribe, you go to SQS and subscribe from there.

PR #1329, opened 2019-02-14
aws/test: make sure to use the same aws.Config throughout tests

vangent:
Before, I believe we could end up using the wrong region, for example.

Updates #1328.


ijt:
Were any tests failing because of having the wrong region? Which ones? Are there travis builds you can link to that show it, or is there a test case that reproduces the problem?

vangent:
On closer look, an empty `aws.Config` does nothing, so there's no point. Reverting and I'll send a separate PR to remove it.

Issue #1270, opened 2019-02-07
pubsub: awspubsub benchmark reports 10msgs/sec, so profile and fix top bottleneck


ijt:

ijt:
Here are some more numbers:

| gocdk pkg | msgs/sec | note |
|---|---|---|
|awspubsub|10|
|gcppubsub|310|
|mempubsub|286k|
|rabbitpubsub|144k|local instance|

The AWS driver appears to be making one RPC per message when sending and receiving, and that would probably explain a lot of why it's so slow. One expedient way to speed it up might be to call the AWS API funcs in multiple goroutines. There do not appear to be batch-based funcs in the AWS Go APIs for these things.

vangent:
After PR #1344, updated stats:

| gocdk pkg | send msgs/sec | receive msgs/sec | notes |
|---|---|---|---|
|awspubsub|660|411| |
|gcppubsub|247|2911| |
|mempubsub|333k|460k| |
|rabbitpubsub|116k|8.4k| local instance |

vangent:
Notes: those stats were with 

```
	const (
		nMessages          = 1000
		concurrencySend    = 100
		concurrencyReceive = 100
	)
```

and with awspubsub region set to us-west-1.

ijt:
A ~50x improvement is nothing to sneeze at. Looks fixed to me.

vangent:
I'm still working on this....

PR #1344, opened 2019-02-14
pubsub/awspubsub: Do concurrent sends for SendBatch; add a benchmark

vangent:
Also make sure to use the `WithContext` flavors of the AWS APIs.

Before:
```
$ go test -v -bench . --run Bench
BenchmarkAwsPubSub/BenchmarkSend-12   <snip>	  10.95 MB/s
```

After:

```
$ go test -v -bench . --run BenchmarkAwsPubSub/BenchmarkSend-
BenchmarkAwsPubSub/BenchmarkSend-12      <snip>	  25.48 MB/s
```

Updates #1270.


pubsub/awspubsub/awspubsub.go
```diff
@@ -72,41 +72,71 @@ func openTopic(ctx context.Context, client *sns.SNS, topicARN string) driver.Top
 var stringDataType = aws.String("String")
 
 // SendBatch implements driver.Topic.SendBatch.
-func (t *topic) SendBatch(ctx context.Context, dms []*driver.Message) (er error) {
+func (t *topic) SendBatch(ctx context.Context, dms []*driver.Message) error {
+	// Limit our concurrent RPCs via a semaphore.
+	const maxConcurrency = 10
+	sem := make(chan bool, maxConcurrency)
```
vangent: An alternative would be allocate `sem` as a field of t. This would be better at limiting overall concurrent RPCs to AWS SNS (since `SendBatch` can be called concurrently, this PR as written only limits to 10x something).

I'd then have to use `errc` as the way to wait for all goroutines to complete, but that should be feasible (can't break out of the `for` loop on `ctx.Done`, and would have to read all `len(dms)` errors from `errc`).

WDYT?

ijt: Seems reasonable, but what you have here is a good starting point. How about doing that in a later PR?

jba: If you use chan struct{}, the channel will take constant space. Not a big deal here, but worth getting into the habit.

pubsub/awspubsub/awspubsub.go
```diff
@@ -72,41 +72,71 @@ func openTopic(ctx context.Context, client *sns.SNS, topicARN string) driver.Top
...
+			errc <- t.sendMessage(ctx, dm)
+			<-sem
+		}(dm)
+	}
+	// Wait for all of the goroutines to complete.
+	for n := 0; n < maxConcurrency; n++ {
+		sem <- true
+	}
+	// Return the first non-nil error (if any).
+	for n := 0; n < len(dms); n++ {
```
jba: nit: Since you don't use `n` here, I think you can just write `for range dms`.

pubsub/awspubsub/awspubsub.go
```diff
@@ -72,41 +72,71 @@ func openTopic(ctx context.Context, client *sns.SNS, topicARN string) driver.Top
...
+			break
+		case sem <- true:
 		}
-		_, err := t.client.Publish(&sns.PublishInput{
-			Message:           aws.String(string(dm.Body)),
-			MessageAttributes: attrs,
-			TopicArn:          &t.arn,
-		})
+		go func(dm *driver.Message) {
+			errc <- t.sendMessage(ctx, dm)
```
jba: Important question: do you want to stop the other sendMessages on the first error you see? If so, that would probably tip the balance to using `errgroup`.

pubsub/awspubsub/awspubsub.go
vangent: Done.

pubsub/awspubsub/awspubsub.go
vangent: I don't think so.

pubsub/awspubsub/awspubsub.go
vangent: Done on `chan struct{}`.

@jba WDYT about the alternative described above?

jba: Not sure why you can't break out of the loop on ctx.Done. Can't you just count the number of goroutines you've launched?

vangent: I will tackle this in a separate PR.

PR #1347, opened 2019-02-15
contributebot: use Go CDK pubsub

jba:
Switch from using Google Cloud Pubsub directly, to using it
indirectly via the Go CDK pubsub API.

### DIFF
--- a/internal/contributebot/main.go
+++ b/internal/contributebot/main.go
@@ -28,8 +28,8 @@ import (
 	"sync"
 	"time"
 
-	"cloud.google.com/go/pubsub"
 	"github.com/google/go-github/github"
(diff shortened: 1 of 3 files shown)


jba:
A couple of things:
- I'm confused why wire is a new dependency. How could it not have been a dependency before?
- I did not test this change. The setup is long and complex (new GCP project, etc.) and I don't think it will be amortized over time for me. I'm hoping you can patch in the PR and test try it.

zombiezen:
- I think that I forgot to run `go mod tidy` the last time I changed Contribute Bot.
- Happy to test for you.

jba:
Thanks Ross!

PR #1346, opened 2019-02-15
gcppubsub: remove context args

jba:
Remove the context.Context argument from OpenTopic and OpenSubscription.

By design, all external work (e.g. establishing client connections)
has already happened by this point, so there is need for a context.

PR #1351, opened 2019-02-15
contributebot: fix build broken by #1346 + #1347

vangent:
The build is broken right now so I'll submit after anybody approves.


vangent:
(This our first build breakage caused by disabling the "you must sync to head and then re-run all tests before merging" flag -- the two PRs worked fine independently but caused a build breakage when put together).

This was my fault, as I merged the 2nd one :-|.

PR #1354, opened 2019-02-15
internal/docstore: structs

jba:
Support for structs:
- Getting and setting struct fields (in driver/document.go)
- Encoding an decoding values (including structs) into native representations.

Add struct support to memdocstore.

### DIFF
--- a/internal/docstore/driver/codec.go
+++ b/internal/docstore/driver/codec.go
@@ -47,7 +47,7 @@ var (
 // encodable.
 type Encoder interface {
 	// These methods all encode and store a single Go value.
-	EncodeNull()
+	EncodeNil()
 	EncodeBool(bool)
 	EncodeString(string)
 	EncodeInt(int64)
@@ -71,11 +71,10 @@ type Encoder interface {
 	EncodeList(n int) Encoder
 	ListIndex(i int)
 
-	// EncodeMap is called when a map or struct is encountered. Its argument is the
-	// number of fields in the map or struct. The encoding algorithm will call the
-	// returned Encoder that many times to encode the successive values of the map or
-	// struct. After each such call, MapKey will be called with the key of the
-	// element just encoded.
+	// EncodeMap is called when a map is encountered. Its argument is the number of
+	// fields in the map. The encoding algorithm will call the returned Encoder that
+	// many times to encode the successive values of the map. After each such call,
+	// MapKey will be called with the key of the element just encoded.
 	//
 	// For example, map[string}int{"A": 1, "B": 2} will result in these calls:
 	//     enc2 := enc.EncodeMap(2)
@@ -83,8 +82,11 @@ type Encoder interface {
 	//     enc2.MapKey("A")
 	//     enc2.EncodeInt(2)
 	//     enc2.MapKey("B")
-	// For struct{A, B int}{1, 2}, if EncodeStruct (see below) returns false, the same
-	// sequence of calls will occur.
+	//
+	// EncodeMap is also called for structs if EncodeStruct (see below) returns false; the map
+	// then consists of the exported fields of the struct.
+	// For struct{A, B int}{1, 2}, if EncodeStruct returns false, the same sequence
+	// of calls as above will occur.
 	EncodeMap(n int) Encoder
 	MapKey(string)
 
@@ -119,7 +121,7 @@ func Encode(v reflect.Value, e Encoder) error {
 
 func encode(v reflect.Value, enc Encoder) error {
 	if !v.IsValid() {
-		enc.EncodeNull()
+		enc.EncodeNil()
 		return nil
 	}
 	if v.Type().Implements(binaryMarshalerType) {
@@ -132,7 +134,7 @@ func encode(v reflect.Value, enc Encoder) error {
 	}
 	if v.Type().Implements(protoMessageType) {
 		if v.IsNil() {
-			enc.EncodeNull()
+			enc.EncodeNil()
 		} else {
 			bytes, err := proto.Marshal(v.Interface().(proto.Message))
 			if err != nil {
@@ -165,7 +167,7 @@ func encode(v reflect.Value, enc Encoder) error {
 		enc.EncodeString(v.String())
 	case reflect.Slice:
 		if v.IsNil() {
-			enc.EncodeNull()
+			enc.EncodeNil()
 			return nil
 		}
 		fallthrough
@@ -175,13 +177,13 @@ func encode(v reflect.Value, enc Encoder) error {
 		return encodeMap(v, enc)
 	case reflect.Ptr:
 		if v.IsNil() {
-			enc.EncodeNull()
+			enc.EncodeNil()
 			return nil
 		}
 		return encode(v.Elem(), enc)
 	case reflect.Interface:
 		if v.IsNil() {
-			enc.EncodeNull()
+			enc.EncodeNil()
 			return nil
 		}
 		return encode(v.Elem(), enc)
@@ -216,7 +218,7 @@ func encodeList(v reflect.Value, enc Encoder) error {
 // Encode a map.
 func encodeMap(v reflect.Value, enc Encoder) error {
 	if v.IsNil() {
-		enc.EncodeNull()
+		enc.EncodeNil()
 		return nil
 	}
 	keys := v.MapKeys()
@@ -281,6 +283,9 @@ func encodeStructWithFields(v reflect.Value, fields fields.List, e Encoder) erro
 			}
 			e2.MapKey(f.Name)
 		}
+		// if !ok, then f is a field in an embedded pointer to struct, and that embedded pointer
+		// is nil in v. In other words, the field exists in the struct type, but not this particular
+		// struct value. So we just ignore it.
 	}
 	return nil
 }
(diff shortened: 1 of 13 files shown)


internal/docstore/internal/fields/fold.go
```diff
@@ -0,0 +1,156 @@
...
+		return equalFoldRight
+	}
+	if nonLetter {
+		return asciiEqualFold
+	}
+	return simpleLetterEqualFold
+}
+
+// equalFoldRight is a specialization of bytes.EqualFold when s is
+// known to be all ASCII (including punctuation), but contains an 's',
```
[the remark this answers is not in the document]
ijt: Never mind. Found the comment about it.

internal/docstore/driver/document.go
```diff
@@ -33,12 +38,20 @@ func NewDocument(doc interface{}) (Document, error) {
...
-func (d Document) Map() map[string]interface{} {
-	return d.m
+	v := reflect.ValueOf(doc)
+	t := v.Type()
+	if t.Kind() == reflect.Ptr {
+		v = v.Elem()
+		t = t.Elem()
+	}
+	if t.Kind() != reflect.Struct {
+		return Document{}, gcerr.Newf(gcerr.InvalidArgument, nil, "expecting struct or map[string]interface{}, got %s", t)
```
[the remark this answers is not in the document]
jba: Done.

internal/docstore/driver/document.go
```diff
@@ -87,6 +105,19 @@ func (d Document) Get(fp []string) (interface{}, error) {
 	return d2.GetField(fp[len(fp)-1])
 }
 
+func (d Document) structField(name string) (reflect.Value, error) {
+	f := d.fields.Match(name)
```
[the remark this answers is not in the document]
jba: For compatibility with encoding/json. And I think the original reason there was to make it easy to go between exported struct fields, which are capitalized, and JSON keys, which often aren't. E.g. with encoding/json you can use
```
struct { Name string, Score int }
```
to read the JSON
```
{"name": "Joe", "score": 10}
```

internal/docstore/memdocstore/mem.go
```diff
@@ -148,18 +159,22 @@ func deleteAtFieldPath(m map[string]interface{}, fp []string) {
 }
 
 // Copy the fields of src to dest.
-func copyFields(dest, src driver.Document, fps [][]string) error {
+func copyFields(dest driver.Document, src map[string]interface{}, fps [][]string) error {
```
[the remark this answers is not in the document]
jba: It's used in the code for Get.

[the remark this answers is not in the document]
jba: Never mind, I should be using decodeDoc instead. Deleted.

internal/docstore/driver/document.go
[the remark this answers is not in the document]
ijt: Okay.

internal/docstore/driver/codec.go
```diff
@@ -0,0 +1,701 @@
...
+// values using interface{}. An Encoder is responsible for storing the value
+// it is encoding.
+//
+// Because all providers must support the same set of values, the encoding methods
+// (with the exception of EncodeStruct) do not return errors. EncodeStruct is special
+// because it is an escape hatch for arbitrary structs, not all of which may be
+// encodable.
+type Encoder interface {
+	// These methods all encode and store a single Go value.
+	EncodeNull()
```
vangent: Nit: Why not `EncodeNil`?

internal/docstore/driver/codec.go
```diff
@@ -0,0 +1,701 @@
...
+	EncodeBool(bool)
+	EncodeString(string)
+	EncodeInt(int64)
+	EncodeUint(uint64)
+	EncodeFloat(float64)
+	EncodeComplex(complex128)
```
vangent: Nit: consider dropping the `Encode` part of all these function names?

internal/docstore/driver/codec.go
```diff
@@ -0,0 +1,701 @@
...
+	// EncodeList is called when a slice or array is encountered (except for a
+	// []byte, which is handled by EncodeBytes). Its argument is the length of the
+	// slice or array. The encoding algorithm will call the returned Encoder that
+	// many times to encode the successive values of the list. After each such call,
+	// ListIndex will be called with the index of the element just encoded.
+	//
+	// For example, []string{"a", "b"} will result in these calls:
+	//     enc2 := enc.EncodeList(2)
+	//     enc2.EncodeString("a")
+	//     enc2.ListIndex(0)
```
vangent: It seems a bit strange to call `ListIndex(0)` after the 0th element has been encoded. Why not before?

internal/docstore/driver/codec.go
```diff
@@ -0,0 +1,701 @@
...
+	//     enc2.EncodeString("b")
+	//     enc2.ListIndex(1)
+	EncodeList(n int) Encoder
+	ListIndex(i int)
+
+	// EncodeMap is called when a map or struct is encountered. Its argument is the
```
vangent: This is a bit misleading for structs, since it's actually only called for structs if `EncodeStruct` returns false.

I would consider just having this paragraph/example reference the map use case, and have a sentence at the end that says something like "EncodeMap is also called for structs if EncodeStruct returns false; the map then then consists of the public fields of the struct."

internal/docstore/driver/codec.go
```diff
@@ -0,0 +1,701 @@
...
+
+// Encode a map.
+func encodeMap(v reflect.Value, enc Encoder) error {
+	if v.IsNil() {
+		enc.EncodeNull()
+		return nil
+	}
+	keys := v.MapKeys()
+	enc2 := enc.EncodeMap(len(keys))
+	for _, k := range keys {
```
vangent: Isn't there a way to do `range k,v`?

internal/docstore/driver/codec.go
```diff
@@ -0,0 +1,701 @@
...
+		return err
+	}
+	fields, err := fieldCache.Fields(v.Type())
+	if err != nil {
+		return err
+	}
+	return encodeStructWithFields(v, fields, e)
+}
+
+func encodeStructWithFields(v reflect.Value, fields fields.List, e Encoder) error {
```
vangent: Nit: not sure this is worth breaking out to a separate function....?

internal/docstore/driver/codec.go
```diff
@@ -0,0 +1,701 @@
...
+	e2 := e.EncodeMap(len(fields))
+	for _, f := range fields {
+		fv, ok := fieldByIndex(v, f.Index)
+		if ok {
+			if err := encode(fv, e2); err != nil {
+				return err
+			}
+			e2.MapKey(f.Name)
+		}
```
vangent: What if `!ok`?

internal/docstore/memdocstore/codec_test.go
```diff
@@ -0,0 +1,142 @@
...
+}
+
+func TestEncodeDoc(t *testing.T) {
+	var b bool = true
+	for _, test := range []struct {
+		in   interface{}
+		want map[string]interface{}
+	}{
+		{
+			map[string]interface{}{
```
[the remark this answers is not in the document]
jba: Done.

jba:
> Can you point me to a couple of providers that would need to implement Encoder/Decoder? (Probably naively), I had hoped that encoding/decoding would be something that would be handled by providers.

Even if a provider supports encoding a struct into its native format, there's no assurance that they will support the same set of types, and they certainly won't use a `docstore` tag. 

For example, the underlying value type for dynamodb is called [`AttributeValue`](https://godoc.org/github.com/aws/aws-sdk-go/service/dynamodb#AttributeValue). A second package supports [Marshaling of structs](https://godoc.org/github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute#Marshal), but it uses the tag `dynamodbav` and has support for types like intset that we might not want to support.

internal/docstore/driver/codec.go
jba: Done.

internal/docstore/driver/codec.go
jba: Seems weird to have methods named `Int`, `Float`, etc., even if they are in a type called `FooEncoder`.

Leaving it for now, will change if it continues to bother you.

internal/docstore/driver/codec.go
jba: Because we want to make life easy for encoder implementations, and when they are told about list item `n`, they would like to have the encoded item available so they can write
```
encodedSlice[n] = itemIJustEncoded
```
If you look at `memdocstore/codec.go` I think you'll see what I mean.

internal/docstore/driver/codec.go
jba: Done.

internal/docstore/driver/codec.go
jba: Yes, `reflect.MapRange`. But it's new in Go 1.12, so we can't use it yet.

internal/docstore/driver/codec.go
jba: It's used by `Document.Encode`, because `Document` already has the list of fields.

internal/docstore/driver/codec.go
jba: Subtle point. Clarified in a comment.

jba:
PTAL.

vangent:
Let's chat about this next week when you're here. I want to make sure I understand how this is going to work relative to existing data.

PR #1359, opened 2019-02-15
blob/s3blob: support URL query parameters to override aws.Config fields

### LINKED ISSUE #1357: blob/s3blob: URL opening doesn't work with "MissingRegion" error
When I create a `blob.Bucket` with URL `s3://mybucket`, reads and writes fail with an error like `MissingRegion: could not find region configuration`.

vangent:
Fixes #1357.


blob/gcsblob/gcsblob_test.go
```diff
@@ -294,35 +294,50 @@ func TestURLOpenerForParams(t *testing.T) {
 	tests := []struct {
 		name     string
 		query    url.Values
+		prev     Options
 		wantOpts Options
 		wantErr  bool
 	}{
 		{
-			name: "AccessID",
+			name: "Invalid query parameter",
```
zombiezen: I've carried these changes forward in #1376. Once that's submitted, let's revert this file.

blob/s3blob/s3blob.go
```diff
@@ -120,32 +125,82 @@ const Scheme = "s3"
...
+		}
 	}
-	return OpenBucket(ctx, o.ConfigProvider, u.Host, &o.Options)
+	var cfg aws.Config
+	override := false
+	if region := q["region"]; len(region) > 0 {
+		cfg.Region = aws.String(region[0])
+		override = true
+	}
+	if endpoint := q["endpoint"]; len(endpoint) > 0 {
```
zombiezen: After careful studying of these parameters, I don't see a need yet for any other than `region`. Is there a specific use case that you are targeting? `disableSSL` in particular seems dangerous.

blob/s3blob/s3blob.go
```diff
@@ -120,32 +125,82 @@ const Scheme = "s3"
...
+	return opts, nil
 }
 
 // Options sets options for constructing a *blob.Bucket backed by fileblob.
-type Options struct{}
+type Options struct {
+	// Cfgs holds configuration overrides for the S3 client. For example, you
+	// may want to use the same *session.Session for buckets across multiple
+	// regions; you can override the aws.Config.Region here.
+	Cfgs []*aws.Config
```
zombiezen: I don't think adding an option is necessary. See my diff https://github.com/google/go-cloud/compare/master...zombiezen:s3-region-url for a variant that simplifies the logic here in a more generalized way.

blob/gcsblob/gcsblob_test.go
vangent: Done.

blob/s3blob/s3blob.go
vangent: These were [added by an external contributor](https://github.com/google/go-cloud/pull/1095). "these three are the ones I needed in order to use the package with a local minio instance". Apparently there are other service providers that have an S3-compatible API.

blob/s3blob/s3blob.go
vangent: Cool; done.

vangent:
PTAL

vangent:
I can also update this to set [`SharedConfigEnable`](https://godoc.org/github.com/aws/aws-sdk-go/aws/session#SharedConfigEnable), LMK.

blob/s3blob/s3blob.go
zombiezen: Gotcha... what do you think about this: `Endpoint` and `DisableSSL` are somewhat redundant with each other: we could parse `Endpoint` as a URL and if it uses "http", then set `DisableSSL` automatically. If `Endpoint` is set at all, `s3ForcePathStyle` also needs to be set.

vangent: I think it's simpler if we just map directly onto what `aws.Config` provides. I think we'll just shoot ourselves in the foot and confuse our users if we start doing non-obvious things.

vangent:
Gentle ping @zombiezen ...

blob/s3blob/s3blob.go
zombiezen: Discussed offline. I agree with the code as-is.

zombiezen:
Decided offline: we'll make the `SharedConfigEnable` in a separate PR. We should also change the `*Session` that's in the default provider sets. @vangent, can you file an issue for that?

vangent:
> Decided offline: we'll make the `SharedConfigEnable` in a separate PR. We should also change the `*Session` that's in the default provider sets. @vangent, can you file an issue for that?

Done: 
https://github.com/google/go-cloud/issues/1416

PR #1360, opened 2019-02-16
pubsub/awspubsub: limit overall send RPCs, not send RPCs per SendBatch call

### LINKED ISSUE #1270: pubsub: awspubsub benchmark reports 10msgs/sec, so profile and fix top bottleneck

vangent:
Fixes #1270.

I don't think this makes any difference now, but will protect us from having too many in-flight RPCs once the concrete type calls `SendBatch` concurrently.

We may eventually want to expose the constant as an Option, or raise it.


vangent:
> Hold on. Where's the test?

It's a refactor, no specific test. I've run the conformance tests and benchmarks.

vangent:
PTAL @jba and @ijt

ijt:
Well the coverage hasn't changed much so I guess you're right.

pubsub/awspubsub/awspubsub.go
```diff
@@ -135,26 +143,24 @@ func (t *topic) SendBatch(ctx context.Context, dms []*driver.Message) error {
 		select {
 		case <-ctx.Done():
 			errc <- ctx.Err()
-			break
-		case sem <- struct{}{}:
+		case t.sem <- struct{}{}:
```
jba: How about 
```
return ctx.Err()
```
on the above line? When the context is done, you want to terminate immediately.

pubsub/awspubsub/awspubsub.go
```diff
@@ -56,13 +56,19 @@ import (
...
 
-// Message Attribute key used to flag that the message body is base64
-// encoded.
-const base64EncodedKey = "base64encoded"
+const (
+	// base64EncodedKey is the Message Attribute key used to flag that the
+	// message body is base64 encoded.
+	base64EncodedKey   = "base64encoded"
+	// maxPublishConcurrency limits the number of Publish RPCs to SNS in flight
+	// at once.
```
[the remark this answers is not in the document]
vangent: done.

pubsub/awspubsub/awspubsub.go
vangent: I don't think that's correct, because it will exit the function while some goroutines that it started may still be running. I'd rather explicitly wait for all of the goroutines to exit.

jba: OK, I guess since they take the same context they will stop quickly.

But it seems silly to start more goroutines after the context is Done (possibly waiting for the semaphore to do so). You should find a way to return ASAP, while waiting for outstanding goroutines to finish.

vangent: OK, now short-circuiting if ctx is done.

vangent:
> I'm approving, but I still don't love the logic in SendBatch. Can you add a TODO there to consider using errgroup? I think that can provide a more elegant solution.

PTAL, trying errgroup now.

pubsub/awspubsub/awspubsub.go
```diff
@@ -135,41 +136,25 @@ var stringDataType = aws.String("String")
...
 		case <-ctx.Done():
 			ctxErr = ctx.Err()
-			errc <- ctxErr
+			break
 		case t.sem <- struct{}{}:
-			go func(dm *driver.Message) {
-				errc <- t.sendMessage(ctx, dm)
-				<-t.sem
-			}(dm)
+			dm := dm // https://golang.org/doc/faq#closures_and_goroutines
```
jba: I actually prefer this to passing args to closures in all cases, so I'm glad you were forced into it here :)

I don't think you need the comment -- it's a standard Go technique -- but I'm fine with it.

pubsub/awspubsub/awspubsub.go
```diff
@@ -135,41 +136,25 @@ var stringDataType = aws.String("String")
...
-		// started complete.
-		if ctxErr != nil {
-			errc <- ctxErr
-			continue
-		}
 		select {
 		case <-ctx.Done():
 			ctxErr = ctx.Err()
-			errc <- ctxErr
+			break
```
jba: This is a no-op here. It just breaks you out of the select. https://golang.org/ref/spec#Break_statements. You'll need to label the loop.

vangent: Ooh, good catch. Done.

PR #1365, opened 2019-02-19
pubsub: make tests for nonexistent topics & subs optional

### LINKED ISSUE #1363: pubsub: make conformance tests for nonexistent topics and subscriptions optional
These tests don't make sense for NATS and Redis pubsub since they have a dynamic notion of topics and subscriptions. They can be removed from the map of tests that get run for all drivers, and made public.

ijt:
Fixes #1363


ijt:
Done. PTAL

pubsub/drivertest/drivertest.go
```diff
@@ -150,6 +159,13 @@ func RunConformanceTests(t *testing.T, newHarness HarnessMaker, asTests []AsTest
 		"TestMetadata":                                            testMetadata,
 		"TestNonUTF8MessageBody":                                  testNonUTF8MessageBody,
 	}
+	if opts.SkipTestsOfNonExistentThings {
```
zombiezen: It would be more clear to have the tests be added on `!opts.SkipTestsOfNonExistentThings`, instead of pattern matching on the name.

pubsub/drivertest/drivertest.go
```diff
@@ -33,6 +34,11 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+// Options contains settings for RunConformanceTests.
+type Options struct {
+	SkipTestsOfNonExistentThings bool
```
zombiezen: Instead of naming this in terms of what it functionally does, it would be clearer to name it in terms of what the underlying system behavior is. Perhaps:

```suggestion
	// DynamicTopics must be set to true if the underlying pub/sub mechanism does
	// not have an explicit "create topic" operation (i.e. publication and subscription occur
	// dynamically).
	DynamicTopics bool
```

pubsub/drivertest/drivertest.go
```diff
@@ -33,6 +34,11 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+// Options contains settings for RunConformanceTests.
```
zombiezen:
```suggestion
// Options control which conformance tests are run.
```

pubsub/drivertest/drivertest.go
```diff
@@ -33,6 +34,11 @@ import (
...
+type Options struct {
+	SkipTestsOfNonExistentThings bool
+}
+
 // Harness descibes the functionality test harnesses must provide to run
```
zombiezen: (While we're here:)

```suggestion
// Harness describes the functionality test harnesses must provide to run
```

pubsub/drivertest/drivertest.go
ijt: Done.

pubsub/drivertest/drivertest.go
ijt: Done-ish. (DynamicTopicsAndSubscriptions).

PR #1366, opened 2019-02-19
docs: Migrate landing page content and hugo scaffold into the Go CDK repo

eliben:


internal/website/README.md
```diff
@@ -0,0 +1,9 @@
+Source for the TODO website. Powered by [Hugo](https://gohugo.io/).
```
ijt: Why TODO? Don't we know it's gocloud.dev?

internal/website/themes/hyde/CHANGELOG.md
```diff
@@ -0,0 +1,7 @@
+# Changelog
```
ijt: It's annoying to have all this theme stuff in here. Isn't there something we can do to only have the real content?

internal/website/README.md
eliben: Most likely, yes. I figured we can fill this in later when we're ready to push it out.

internal/website/themes/hyde/CHANGELOG.md
eliben: I'm not sure what. We could extract only the essential parts of the theme - that would reduce the line count but would require significant time investment. The theme was modified for our needs so it's not something we can just link to outside, and it's essential for Hugo to generate the HTML.

internal/website/layouts/index.html
```diff
@@ -0,0 +1,4 @@
+{{ define "main" -}}
+{{ .Content }}
+</div>
```
zombiezen: Mismatched div?

eliben: Nice spot! Fixed

PR #1369, opened 2019-02-21
pubsub: send and receive concurrently in conformance benchmark

### LINKED ISSUE #1364: pubsub: no longer assume persistence in benchmarks
Some providers such as NATS, Redis and ZeroMQ don't persist messages.

ijt:
Fixes #1364

### DIFF
--- a/pubsub/drivertest/drivertest.go
+++ b/pubsub/drivertest/drivertest.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"sort"
 	"strconv"
 	"testing"
@@ -453,7 +454,7 @@ func testNonUTF8MessageBody(t *testing.T, newHarness HarnessMaker) {
 	defer cleanup()
 
 	// Sort the WeirdStrings map for record/replay consistency.
-	var weirdStrings [][]string  // [0] = key, [1] = value
+	var weirdStrings [][]string // [0] = key, [1] = value
 	for k, v := range escape.WeirdStrings {
 		weirdStrings = append(weirdStrings, []string{k, v})
 	}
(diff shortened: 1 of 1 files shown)


pubsub/drivertest/drivertest.go
```diff
@@ -583,27 +584,37 @@ func benchmark(b *testing.B, topic *pubsub.Topic, sub *pubsub.Subscription, time
...
-		if timeSend {
-			b.StopTimer()
-		} else {
-			b.StartTimer()
-		}
-		if err := receiveNConcurrently(sub, nMessages, concurrencyReceive); err != nil {
-			b.Fatalf("receiving: %v", err)
+		var eg errgroup.Group
+		eg.Go(func() error {
+			if !timeSend {
```
jba: Isn't the logic here backwards? Doesn't `timeSend` mean "time the publish goroutine"?

zombiezen:
This change looks fine, but the issue you link seems unrelated.

pubsub/drivertest/drivertest.go
```diff
@@ -583,27 +584,37 @@ func benchmark(b *testing.B, topic *pubsub.Topic, sub *pubsub.Subscription, time
...
+				b.StartTimer()
+				defer b.StopTimer()
+			}
+			b.Logf("publishing %d messages", nMessages)
+			if err := publishNConcurrently(topic, nMessages, concurrencySend, attrs, body); err != nil {
```
vangent: Doesn't this still assume persistence, since this could happen before `Receive` is called below?

pubsub/drivertest/drivertest.go
ijt: Yes. Done.

pubsub/drivertest/drivertest.go
ijt: Now that you mention it I have to admit this wasn't a test-driven change. I suspect the way Derek wrote the NATS driver it will work as long as the local subscription has been created, with messages queueing up locally, but need to check this...

ijt:
Closing this per our chat today.

PR #1374, opened 2019-02-21
pubsub: receive and send concurrently in conformance tests

### LINKED ISSUE #1370: pubsub: no longer assume persistence in conformance tests
NATS core doesn't work this way. The client side subscriptions have to be set up before msgs are sent to the topics in order for the client subscriptions to get the messages.

ijt:
Fixes #1370

### DIFF
--- a/pubsub/drivertest/drivertest.go
+++ b/pubsub/drivertest/drivertest.go
@@ -236,8 +236,13 @@ func testSendReceive(t *testing.T, newHarness HarnessMaker) {
 	}
 	defer cleanup()
 
-	want := publishN(ctx, t, top, 3)
-	got := receiveN(ctx, t, sub, len(want))
+	wantChan := make(chan []*pubsub.Message)
+	gotChan := make(chan []*pubsub.Message)
+	n := 3
+	go func() { gotChan <- receiveN(ctx, t, sub, n) }()
+	go func() { wantChan <- publishN(ctx, t, top, n) }()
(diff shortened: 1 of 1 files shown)


pubsub/drivertest/drivertest.go
```diff
@@ -236,8 +236,13 @@ func testSendReceive(t *testing.T, newHarness HarnessMaker) {
 	}
 	defer cleanup()
 
-	want := publishN(ctx, t, top, 3)
-	got := receiveN(ctx, t, sub, len(want))
+	wantChan := make(chan []*pubsub.Message)
+	gotChan := make(chan []*pubsub.Message)
+	n := 3
+	go func() { gotChan <- receiveN(ctx, t, sub, n) }()
+	go func() { wantChan <- publishN(ctx, t, top, n) }()
```
jba: You say on the issue that something has to be receiving while the messages are being published. But here it's possible that the publish goroutine starts, and maybe even completes, before Receive is ever called.

zombiezen:
I'm not sure I understand enough of the context to review. @jba, can I assign to you?

pubsub/drivertest/drivertest.go
ijt: Fair point. Anyway, I'm removing this PR since we agreed that we're not accepting the NATS pubsub driver for now and more likely supporting it in a different API.

Issue #1378, opened 2019-02-21
all: make a new subpackage for at-most-once pubsubs like NATS core, Redis, ZeroMQ


ijt:
The existing `pubsub` package is an awkward fit for NATS core, Redis pubsub, and ZeroMQ for the following reasons:

1. At-most-once pubsub providers do not require or directly support acknowledgment. Acking would have to be a no-op, which could be misleading.
2. At-most-once pubsub providers allow freely publishing messages to topics with any string for their name, like `connection.Publish("topic1", "message")`. To support this on at-least-once pubsub providers such as GCP pubsub, we would have to add support for dynamically creating topics and then possibly add some way to keep them from accumulating on the server. Such a garbage collection scheme seems well beyond the scope of what the Go CDK is trying to do.
3. At-most-once pubsub providers allow dynamically creating lightweight subscriptions. Supporting equivalent functionality within at-least-once providers would require a way to prevent accumulation of zombie subscriptions, similar to the issue with topics in reason 2.

Having a subpackage that supports the full functionality of at-most-once pubsub systems would make it possible to write portable apps that can dynamically publish to sets of topics that change over time, and have a varying number of clients subscribing. For example, this could be used to write an IRC-style app or client/server games.

The package could be called something like `dynpubsub` or `pubsublite`.

Here is a sketch of what using it could look like, based on a subset of the `go-nats` client example:

```go
import (
    "gocloud.dev/pubsublite"
    "gocloud.dev/pubsublite/natspsl"  // or redispsl, etc.
)

con, err := natspsl.Connect(natspsl.DefaultURL)
defer con.Close()
con.Publish("foo-topic", []byte("hello world")
con.Subscribe("foo-topic", func(m *natspsl.Message) {
   fmt.Println("got message: ", string(m.Body))
})
```

/cc @vangent, @zombiezen, @jba, @derekcollison.

derekcollison:
I like the general idea. I think that these systems, at least NATS, don't only do pub/sub but layer RPC on top of it which is useful to developers, so may consider adding that as a general API.

ijt:
If it can be layered on top, do you think it would make sense to put this RPC functionality into a utils subpackage, like `pubsublite/pslutils`?

derekcollison:
Feels core to me, but maybe others will have an opinion.

vangent:
We're also tracking this in #1368, so closing. Creating a subpackage is one option but not one that we're currently leaning toward.

jba:
[...]

PR #1380, opened 2019-02-22
pubsublite: add new package for at-most-once pubsub

ijt:
Progress towards #1378

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -106,6 +106,7 @@ require (
 	github.com/mitchellh/hashstructure v1.0.0 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.1 // indirect
+	github.com/nats-io/go-nats v1.6.0
 	github.com/ory-am/common v0.4.0 // indirect
 	github.com/ory/dockertest v3.3.4+incompatible // indirect
 	github.com/pborman/uuid v1.2.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -106,7 +106,7 @@ require (
 	github.com/mitchellh/hashstructure v1.0.0 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.1 // indirect
-	github.com/nats-io/go-nats v1.6.0
+	github.com/nats-io/go-nats v1.7.2
 	github.com/ory-am/common v0.4.0 // indirect
 	github.com/ory/dockertest v3.3.4+incompatible // indirect
 	github.com/pborman/uuid v1.2.0 // indirect
--- a/pubsublite/driver/driver.go
+++ b/pubsublite/driver/driver.go
@@ -4,6 +4,6 @@ import "context"
 
 type Pubsubber interface {
 	Publish(ctx context.Context, topic string, msg []byte) error
-	Subscribe(ctx context.Context, callback func(msg []byte)) error
+	Subscribe(ctx context.Context, topic string, callback func(msg []byte)) error
 	Close()
 }
--- a/pubsublite/driver/driver.go
+++ b/pubsublite/driver/driver.go
@@ -1,0 +1,9 @@
+package driver
+
+import "context"
+
+type Pubsubber interface {
+	Publish(ctx context.Context, topic string, msg []byte) error
+	Subscribe(ctx context.Context, callback func(msg []byte)) error
+	Close()
+}
--- a/pubsublite/psl.go
+++ b/pubsublite/psl.go
@@ -1,3 +1,8 @@
+// Package pubsublite is an API for pubsub providers that guarantee
+// at-most-once delivery, and can dynamically publish to topics with names that
+// can be chosen at runtime, and can add or remove any number of subscriptions
+// dynamically. These providers do not have a notion of message
+// acknowledgement.
 package pubsublite
 
 import (
--- a/pubsublite/psl.go
+++ b/pubsublite/psl.go
@@ -1,0 +1,21 @@
+package pubsublite
+
+import (
+	"gocloud.dev/pubsublite/driver"
+)
+
+type Connection struct {
+	ps driver.Pubsubber
+}
+
+func (c *Connection) Publish(ctx context.Context, topic string, msg []byte) error {
+	return ps.Publish(ctx, topic, msg)
+}
+
+func (c *Connection) Subscribe(ctx context.Context, callback func(msg []byte)) error {
+	return ps.Subscribe(ctx, callback)
+}
+
+func (c *Connection) Close() {
+	ps.Close()
+}
--- a/pubsublite/psl.go
+++ b/pubsublite/psl.go
@@ -1,21 +1,27 @@
 package pubsublite
 
 import (
+	"context"
+
 	"gocloud.dev/pubsublite/driver"
 )
 
-type Connection struct {
+type Conn struct {
 	ps driver.Pubsubber
 }
 
-func (c *Connection) Publish(ctx context.Context, topic string, msg []byte) error {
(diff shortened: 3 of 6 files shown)


pubsublite/driver/driver.go
```diff
@@ -0,0 +1,9 @@
+package driver
+
+import "context"
+
+type Pubsubber interface {
+	Publish(ctx context.Context, topic string, msg []byte) error
+	Subscribe(ctx context.Context, callback func(msg []byte)) error
```
derekcollison: Although I like callbacks, we should offer options here as our Go client does. Channels seem to be the Go thing to do, so possibly something like:

https://github.com/nats-io/go-nats/blob/master/nats.go#L2721

go.mod
```diff
@@ -106,6 +106,7 @@ require (
 	github.com/mitchellh/hashstructure v1.0.0 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.1 // indirect
+	github.com/nats-io/go-nats v1.6.0
```
derekcollison: Update to 1.7.2

pubsublite/driver/driver.go
derekcollison: This needs to return a handle I think *Subscription, at least for cancelling.

derekcollison: For NATS we like callbacks to have Subject, Reply and Data, which we wrap up in a nats.Msg.

Since we can subscribe to wildcards it may be important to evaluate the literal subject it was published on. Reply is for us to reply by Publishing the response.

ijt: We've generally been avoiding exposing channels in the APIs of the Go CDK. 
See https://github.com/sourcegraph/gophercon-2018-liveblog/issues/35 for background.
What would be some advantages of adding them here?

ijt: Cancellation is handled by the context. Isn't that enough?

derekcollison: Channels seem polarizing, so will leave up to you.

derekcollison: So each subscription would need a separate context to only cancel that one?

ijt: Okay, leaving that out for now. We can add it later if the need becomes clear.

go.mod
ijt: Done.

pubsublite/driver/driver.go
ijt: That would be one way, but also multiple subscriptions could share a context and all of them could be stopped by cancelling that context.

ijt: Also this is for consistency with the rest of the Go CDK.

ijt: I'm not sure such things should be exposed at this level. I would think there could be an RPC layer on top of raw pubsublite, and the RPC layer would know about Subject, Reply and Data and take care of encoding and decoding.

derekcollison: ok fair enough. CDK all in on Contexts it seems. ;)

derekcollison: Possibly, although seems a better fit here IMO. As for subject, at least for NATS that one can be important since you can subscribe to `foo.bar.*` but the messages will be delivered on literals, like `foo.bar.baz` which the user may need to know to process the message.

pubsublite/driver/driver.go
```diff
@@ -0,0 +1,9 @@
+package driver
+
+import "context"
+
+type Pubsubber interface {
+	Publish(ctx context.Context, topic string, msg []byte) error
+	Subscribe(ctx context.Context, topic string, callback func(msg []byte)) error
```
jba: How about
```
Receive(ctx, topic) (msg, error)
```

pubsublite/driver/driver.go
```diff
@@ -0,0 +1,9 @@
+package driver
+
+import "context"
+
+type Pubsubber interface {
+	Publish(ctx context.Context, topic string, msg []byte) error
```
jba: What about message metadata?

pubsublite/psl.go
```diff
@@ -0,0 +1,32 @@
+// Package pubsublite is an API for pubsub providers that guarantee
+// at-most-once delivery, and can dynamically publish to topics with names that
+// can be chosen at runtime, and can add or remove any number of subscriptions
+// dynamically. These providers do not have a notion of message
+// acknowledgement.
+package pubsublite
+
```
jba: I think we should do one of two things:
1. If we want to think of this as a kind of pubsub, then to minimize confusion we should have the same API as our `pubsub` package, minus `Ack`. 
2. Maybe despite everyone calling it pubsub, it's really something else—message broadcasting—and we should have a new, top-level package with different names for everything.

pubsublite/driver/driver.go
ijt: It's not supported as a first-class concept in any of the at-most-once providers that I know of.  

NATS: https://github.com/nats-io/go-nats/blob/f3379e01130f595688283acd70cdb1fff65d9a4d/nats.go#L429:6

Redis: https://github.com/go-redis/redis/blob/bd542089bb6e776e6fced5038edac8a0f526aa53/pubsub.go#L273:6

ZeroMQ: https://github.com/pebbe/zmq4/blob/7a493a642e7acbd03045d4f5fe9516908a62f86f/zmq4.go#L983

pubsublite/driver/driver.go
ijt: If you really want this kind of synchronous API (I don't but I'm biased towards callbacks), I think it would make more sense to have something like this:
```
type Pubsubber interface {
    ...
    Subscribe(ctx context.Context, topic string) (Receiver, error)
}

type Receiver interface {
    Receive(ctx context.Context) (Msg, error)
}
```
The `Receiver` would keep a subscription open and do any necessary local buffering to make sure messages don't get dropped if the client temporarily fails to keep up.

To be more in keeping with "take interfaces, return concrete types", it could be 
```
type Pubsubber interface {
    ...
    Subscribe(ctx context.Context, topic string, h Handler) error
}

type Handler interface {
    Handler(ctx context.Context, Msg) error
}
```
but that's essentially a more verbose way to write the callback-based thing I suggested. Maybe people would find it easier to think about somehow.

pubsublite/psl.go
ijt: I'm in favor of option 2, calling it maybe `broadcast` or `messaging`.

ijt: I don't think #1 makes sense because we can no longer rely on the existence of subscriptions defined outside the API that queue up messages. The API needs to specifically do something to help prevent dropping messages on the floor, either buffering or spawning goroutines that handle the messages.

derekcollison: Depending on usage you may want to back up and reconsider the pubsub package itself. You probably need to abstract out QoS to some degree. The only reason for two packages is because of incorrect assumptions in pubsub.

ijt: About thing 2, I still think it's fine to talk about publishing and subscribing with a different quality of service (QoS) level: at-most-once. That's the standard nomenclature as far as I know. That said, I do like the idea of calling the new subpackage `broadcast` since it's a whole, reasonably short English word that suggests the transience of messaging in these providers.

pubsublite/driver/driver.go
ijt: Okay, I think we can handle this for now by just making these take an Msg struct with a Data field. Then we can add the other fields as needed.

pubsublite/psl.go
derekcollison: I disagree. The routing is the same in most packages just QoS changes. For instance NATS is pub/sub, req/reply and distributed queues. Broadcast is not a good name for that in my opinion.

ijt:
From our conversation today, this appears unlikely to be accepted.

Issue #1368, opened 2019-02-20
pubsub: support providers that don't do acks


ijt:
NATS and Redis pubsub don't have acking, and it doesn't make sense to try to shoehorn them in by making message acks a no-op. That would mean the application code would be misleading.

To support pubsub providers without acks, I propose to add the following new types:

`pubsub.FAFSubscription`, whose `Receive` method returns a `pubsub.FAFMessage`
`pubsub.FAFMessage` that is mostly like a `pubsub.Message` but has no `Ack` method or associated bookkeeping

and corresponding interfaces within `pubsub/driver`. FAF here stands for "Fire and Forget".

Drivers would either implement `driver.Subscription` or `driver.FAFSubscription`, but probably not both.

derekcollison:
Not sure I am liking FAF or that a message changes due to its Ack methods. If the message is truly at most once than Ack would be a no-op.

ijt:
Fair enough. Let's try your approach.

derekcollison:
I have implemented the driver and necessary interfaces but things of course do not pass, so will look to work through those tomorrow. Thx.

ijt:
Great. Glad to hear about this progress.

derekcollison:
Maybe a message should have knowledge if it requires an ack. pubsub.Message.RequiresAck() bool.

ijt:
I'm not sure how that would be an improvement over having just pubsub.Message.Ack() that does a no-op on some providers. I can see a downside though: having to check that before calling Ack all the time.

derekcollison:
Possibly not, but if I was reading someone's code and just saw an Ack that seemed to return no error I would probably think it did something. That make sense? I would rather see something like..

```go
if msg.RequiresAck() {
    msg.Ack()
}
```

ijt:
I don't think that works well from the perspective of app developers. They would have to scratch their heads about whether to include that logic every time they write code that receives a message.

The evidence that I see so far points towards needing a separate Go CDK subpackage to support at-most-once providers. I filed an issue for that: #1378. Closing this issue as redundant with that one.

jba:
Robert and I just revisited the ideas here. We now think it might work to support ack-less providers with the same API, like so:
- The driver interface will have a `HasAck` method (all names open to bikeshedding).
- The concrete `Subscription` type will also have a `HasAck` method that just delegates to the driver. (Note: the method is on `Subscription`, not `Message`.)
[...]

PR #1381, opened 2019-02-22
pubsub: add NATS support

derekcollison:
This adds support to the pubsub package for NATS.

Signed-off-by: Derek Collison <derek@nats.io>

/cc @ijt @jba


derekcollison:
I signed it!

derekcollison:
Not sure failed build is something I control, looks like deps changing is what failed AFAIK.

pubsub/natspubsub/nats.go
```diff
@@ -0,0 +1,263 @@
...
+
+type subscription struct {
+	nc   *nats.Conn
+	nsub *nats.Subscription
+	oerr error
+}
+
+// OpenSubscription returns a *pubsub.Subscription representing a NATS subscription.
+// TODO(dlc) - Options for queue groups?
+func OpenSubscription(nc *nats.Conn, subscriptionName string) *pubsub.Subscription {
```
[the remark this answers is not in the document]
derekcollison: Should it also be CreateTopic?

derekcollison:
I will take a look at this tomorrow. 

https://github.com/etcd-io/etcd/pull/10337

derekcollison:
I am running into a bunch of issues trying to use the newer version of the codec since etcd uses it but does not work with the new one. I will revert that back out, update the code to not use the newer features and change Open -> Create.

pubsub/natspubsub/nats.go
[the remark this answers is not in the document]
ijt: Yes, that would be more consistent.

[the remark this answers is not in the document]
derekcollison: ok will do.

pubsub/natspubsub/nats.go
```diff
@@ -0,0 +1,263 @@
...
+// openTopic returns the driver for OpenTopic. This function exists so the test
+// harness can get the driver interface implementation if it needs to.
+func openTopic(nc *nats.Conn, topicName string) driver.Topic {
+	return &topic{nc, topicName}
+}
+
+// SendBatch implements driver.Topic.SendBatch.
+func (t *topic) SendBatch(ctx context.Context, msgs []*driver.Message) error {
+	if t == nil || t.nc == nil {
+		return errNotExist
```
ijt: Wouldn't it make more sense to say it's uninitialized, rather than that it doesn't exist?

pubsub/natspubsub/nats.go
```diff
@@ -0,0 +1,263 @@
...
+	return true
+}
+
+// ErrorAs implements driver.Topic.ErrorAs
+func (*topic) ErrorAs(error, interface{}) bool {
+	return false
+}
+
+// ErrorCode implements driver.Topic.ErrorCode
+func (*topic) ErrorCode(error) gcerrors.ErrorCode { return gcerrors.Unknown }
```
ijt: This should probably return gcerrors.FailedPrecondition for the error of the topic or its connection being nil.

pubsub/natspubsub/nats.go
```diff
@@ -0,0 +1,263 @@
...
+	return true
+}
+
+// ErrorAs implements driver.Subscription.ErrorAs
+func (*subscription) ErrorAs(error, interface{}) bool {
+	return false
+}
+
+// ErrorCode implements driver.Subscription.ErrorCode
+func (*subscription) ErrorCode(error) gcerrors.ErrorCode { return gcerrors.Unknown }
```
ijt: This should most likely return gcerrrors.FailedPrecondition for the error of the subscription or its connection being nil.

pubsub/natspubsub/nats.go
```diff
@@ -0,0 +1,263 @@
...
+
+var errNotExist = errors.New("natspubsub: topic does not exist")
+
+type topic struct {
+	nc   *nats.Conn
+	subj string
+}
+
+// For encoding we use msgpack from github.com/ugorji/go.
+// it was already imported in go-cloud and is reasonably performant.
```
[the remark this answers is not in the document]
derekcollison: Will fix.

pubsub/natspubsub/nats.go
```diff
@@ -0,0 +1,263 @@
...
+// We define our own version of message here for encoding that
+// only encodes Body and Metadata. Otherwise we would have to
+// add codec decorations to driver.Message.
+type encMsg struct {
+	Body     []byte            `codec:",omitempty"`
+	Metadata map[string]string `codec:",omitempty"`
+}
+
+// OpenTopic returns a *pubsub.Topic for use with NATS.
+// We delay checking for the proper syntax here.
```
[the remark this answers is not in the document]
derekcollison: For NATS subjects are comprised of tokens and optionally wildcards.

https://nats.io/documentation/writing_applications/subjects/

pubsub/natspubsub/nats.go
```diff
@@ -0,0 +1,263 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package natspubsub provides a pubsub implementation for NATS.io.
+// Use OpenTopic to construct a *pubsub.Topic, and/or OpenSubscription
+// to construct a *pubsub.Subscription.
```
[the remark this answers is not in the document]
derekcollison: ok will add.

pubsub/natspubsub/nats.go
```diff
@@ -0,0 +1,263 @@
...
+		if err := enc.Encode(em); err != nil {
+			return err
+		}
+		if err := t.nc.Publish(t.subj, b); err != nil {
+			return err
+		}
+	}
+	// Per specification this is supposed to only return after
+	// a message has been sent. Normally NATS is very efficient
+	// at sending messages in batches on its on and also handles
```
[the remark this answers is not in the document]
derekcollison: Will fix.

pubsub/natspubsub/nats.go
```diff
@@ -0,0 +1,263 @@
...
+	return false
+}
+
+// ErrorCode implements driver.Topic.ErrorCode
+func (*topic) ErrorCode(error) gcerrors.ErrorCode { return gcerrors.Unknown }
+
+type subscription struct {
+	nc   *nats.Conn
+	nsub *nats.Subscription
+	oerr error
```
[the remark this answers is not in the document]
derekcollison: We capture this when creating a Subscription since that signature does not return an err.

pubsub/natspubsub/nats.go
derekcollison: Possibly, copied what others were doing here. Will fix.

pubsub/natspubsub/nats.go
derekcollison: ok will take a look.

pubsub/natspubsub/nats.go
derekcollison: Will take a look.

pubsub/natspubsub/nats_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+type harness struct {
+	s  *server.Server
+	nc *nats.Conn
+}
+
+func newHarness(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
+	opts := gnatsd.DefaultTestOptions
+	opts.Port = TEST_PORT
+	s := gnatsd.RunServer(&opts)
```
ijt: Nice.

pubsub/natspubsub/nats_test.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	return ds, cleanup, nil
+}
+
+func (h *harness) MakeNonexistentSubscription(ctx context.Context) (driver.Subscription, error) {
+	return (*subscription)(nil), nil
+}
+
+func (h *harness) Close() {
+	if h == nil {
```
ijt: Why are there all these nil checks? Seems fine to just let them panic if they are nil since it's a legitimate bug.

derekcollison: Can remove.

pubsub/natspubsub/nats.go
```diff
@@ -13,8 +13,9 @@
 // limitations under the License.
 
 // Package natspubsub provides a pubsub implementation for NATS.io.
-// Use OpenTopic to construct a *pubsub.Topic, and/or OpenSubscription
-// to construct a *pubsub.Subscription.
+// Use CreateTopic to construct a *pubsub.Topic, and/or CreateSubscription
+// to construct a *pubsub.Subscription. This package will use msgPack
```
ijt: s/will use/uses/

derekcollison:
ok sounds good. Can you restart the one failed build on linux or is it something I need to change?
Should I squash all commits into one?

ijt:
There should be a button to retry the build. If it's not there for you, lmk
and I'll try it when I'm back, before 2pm. No need to squash.

ijt:
Looks like the conflicts have to be resolved for Travis to run again. Not sure, but those have to be resolved anyway.

derekcollison:
Ok will take a look


________________________________

derekcollison:
ok so listdeps returns different answers atm on MacOSX vs Linux, so the compare with alldeps will always fail on one. 

Let me know if you need anything more from me.

pubsub/natspubsub/nats.go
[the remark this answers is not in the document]
ijt: Okay, please add the link.

ijt:
Looks like we need some additional test coverage to get this past the target 76.24%.

eliben:
> ok so listdeps returns different answers atm on MacOSX vs Linux, so the compare with alldeps will always fail on one.
> 
> Let me know if you need anything more from me.

To generate the correct `alldeps` file, you have to run the same version of Go (1.11) that Travis runs. It shouldn't depend on the OS, hopefully, only on the Go version.

derekcollison:
I have go 1.11.5 on both my Mac and my Docker image and they generate different things, regardless of travis. And I can get Travis to pass on one or the other but not both due to linux generating an additional dependency..

derekcollison:
I can up the coverage, will take a look.

eliben:
> I have go 1.11.5 on both my Mac and my Docker image and they generate different things, regardless of travis.

Ah, that's unfortunate.

> And I can get Travis to pass on one or the other but not both due to linux generating an additional dependency..

Note that Travis only runs the dependency check on Linux (https://github.com/google/go-cloud/blob/master/internal/testing/runchecks.sh#L45)

derekcollison:
I will check again. Thanks..

derekcollison:
Coverage is up and I added examples. I have run the tests on linux and the dependency check and it passes here but still fails on Travis. Also locally the tests seem to hang on etcdvar.

Any pointers would be helpful. From our end we are good to go, but would like to see both tests pass on Travis.

internal/testing/alldeps
```diff
@@ -3,31 +3,31 @@ contrib.go.opencensus.io/exporter/aws
...
 contrib.go.opencensus.io/exporter/stackdriver
 contrib.go.opencensus.io/integrations/ocsql
-github.com/aws/aws-sdk-go
 github.com/Azure/azure-amqp-common-go
 github.com/Azure/azure-pipeline-go
 github.com/Azure/azure-service-bus-go
 github.com/Azure/azure-storage-blob-go
 github.com/Azure/go-autorest
+github.com/GoogleCloudPlatform/cloudsql-proxy
+github.com/aws/aws-sdk-go
```
eliben: It's strange that your listdeps run reordered the file - this could be why Travis is complaining. For example ,if I rerun `sort | uniq` on this file on my Linux machine, `github.com/aws/aws-sdk-go` sorts before the `Azure` modules, but in your PR it's moved after them. 

FWIW I'm using sort v8.28 on bash v4.4.12 running on a x64-64-pc-linux-gnu Debian machine

derekcollison: Thanks will take a look.

derekcollison: That did not seem to help..

derekcollison: https://travis-ci.com/google/go-cloud/jobs/180423337

ijt:
Hi Derek, I talked this over with @jba and @vangent. @jba and I agree that support for NATS should happen in a new subpackage of the Go CDK, something like what I outlined in #1380, possibly with a different name like `broadcast`, `messaging`, or `msg`. @jba is going to write up a design for this and there will be internal review etc.

The temporarily bad news is that we're not planning to accept this driver (at least for now) since the NATS model doesn't really fit with our `pubsub` API's at-least-once assumption. Acking is baked in.

The good news is that in a few weeks we may have a new subpackage of the Go CDK that will support NATS and friends more directly, in a way that should highlight their performance and flexibility.

I appreciate all the work you've done so far on this and for helping me and the team get a better understanding of this area. Let me know if you'd like to chat more about it.

derekcollison:
Well that was a colossal waste of my time then..

derekcollison:
We could redo it to uses NATS Streaming which does fit your current model that requires Acks etc..

ijt:
Yes, NATS streaming looks like a good fit for the existing pubsub API. Would that be a large change from the driver you've already implemented?

derekcollison:
Not too much.

derekcollison:
Let me know if this is something I should look into and spend time on. Want to avoid what happened with the first part of this PR if possible.

jba:
I don't think it was necessarily a waste of time. Even if we do go with another API, most of the code should transfer over.

But we should think through the next step carefully. Let's take some time to ponder whether we should implement this pubsub API with NATS streaming, or come up with a new API that makes more sense for pure NATS.

The first question in my mind is how close a fit NATS streaming is to our API. I'll do some research on that.

derekcollison:
NATS Streaming is an At-Least-Once persisted log based technology, similar to Kafka, so IMO very applicable to this API here which assumes At-Least-Once, persistence and redelivery characteristics.

jba:
Good news @derekcollison, we've decided having a single API is the right choice after all. We'll take this PR as is.

gedw99:
Thank Derek... NATS streaming is a huge win. 
I am very curious how the security model will work considering that it has to support the other implementations from the same interface.
If there is any more info on this aspect it would be great to add this to a readme somewhere ...

jba:
@gedw99 The plan is: first land this PR, then adapt it to our plan for at-most-once systems (#1368). Then we hope that Derek will add NATS Streaming as a pair of additional constructors (e.g. `CreateStreamingTopic`, `CreateStreamingSubscription`).

derekcollison:
Thanks. Let me know next steps and timing when you can. NATS Streaming uses NATS as its transport and adopts its authz and authn model from NATS core. The security story has changed in a good way with the upcoming NATS 2.0 release. You can read a bit about what is coming here: https://synadia.com/ngs/how-it-works/

derekcollison:
@jba Could also use a bit of guidance on getting tests to pass, failing around alldeps issues which I have updated but seem to be different for MacOSX vs Linux now, so one test will always fail it seems.

eliben:
> @jba Could also use a bit of guidance on getting tests to pass, failing around alldeps issues which I have updated but seem to be different for MacOSX vs Linux now, so one test will always fail it seems.

This recently landed: https://github.com/google/go-cloud/commit/08c1778a911edb8a0f16f9b27bc11de175b541ee should make these tests a bit more robust

derekcollison:
ok I built deps on linux which I though closely matched travis but still fails.. If someone from the core team could take a quick look that would be appreciated to get them to pass.

vangent:
> ok I built deps on linux which I though closely matched travis but still fails.. If someone from the core team could take a quick look that would be appreciated to get them to pass.

I believe if you merge from HEAD (and resolve conflicts) and generate alldeps again you should be OK. There was an issue with inconsistent sorting (some platforms were case sensitive sorting, others were not), and that's what it looks like is happening for your diffs.

derekcollison:
ok, might not have time to get to this until mid next week.

pubsub/natspubsub/nats.go
```diff
@@ -0,0 +1,292 @@
...
+	var raw [1024]byte
+	b := raw[:0]
+	enc := codec.NewEncoderBytes(&b, &mh)
+
+	for _, m := range msgs {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		enc.ResetBytes(&b)
+		em.Body, em.Metadata = m.Body, m.Metadata
```
vangent: How about sending `m.Body` if `len(m.Metadata) == 0`?

That way, if I have an existing system using NATS, I can convert my senders to use Go CDK and the bytes sent over the wire will stay the same, so my existing readers can still read them.

pubsub/natspubsub/nats.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+	}
+	// Per specification this is supposed to only return after
+	// a message has been sent. Normally NATS is very efficient
+	// at sending messages in batches on its own and also handles
+	// disconnected buffering during a reconnect event. We will
+	// let NATS handle this for now. If needed we could add a
+	// FlushWithContext() call which ensures the connected server
+	// has processed all the messages.
+	return nil
```
vangent: Does `FlushWithContext` return an error? I'm a bit concerned that we're returning success here when messages may not have actually been sent, which is misleading. I'd rather start off correct and slow....

pubsub/natspubsub/nats.go
```diff
@@ -0,0 +1,292 @@
...
+	return ms, nil
+}
+
+// Convert NATS msgs to *driver.Message.
+func decode(msg *nats.Msg) (*driver.Message, error) {
+	if msg == nil {
+		return nil, nats.ErrInvalidMsg
+	}
+	var dm driver.Message
+	// Everything is in the msg.Data
```
vangent: Similarly here, what if the message was sent from a native NATS client? We'd want it to decode to a `driver.Message` with the `Body` set and no `Metadata`?

Consider adding tests for `encode` and `decode` that verify that encoding a `driver.Message` with empty `Metadata` is just `Message.Body`, and that decoding an arbitrary `[]byte` (that doesn't decode to a `driver.Message`) decodes to a `driver.Message` with the `Body` set to the bytes.

pubsub/natspubsub/nats_test.go
```diff
@@ -0,0 +1,302 @@
...
+	return nil
+}
+
+func TestConformance(t *testing.T) {
+	asTests := []drivertest.AsTest{natsAsTest{}}
+	drivertest.RunConformanceTests(t, newHarness, asTests, nil)
+}
+
+// These are natspubsub specific to increase coverage.
+func TestSimplePubSub(t *testing.T) {
```
vangent: Does this test actually add coverage? It looks like functionality that should be covered by the conformance tests.

pubsub/natspubsub/nats_test.go
```diff
@@ -0,0 +1,302 @@
...
+	m, err := sub.Receive(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !bytes.Equal(m.Body, body) {
+		t.Fatalf("Body did not match. %q vs %q\n", m.Body, body)
+	}
+}
+
+func TestCanceledContext(t *testing.T) {
```
vangent: Same here.

jba:
@derekcollison, Now that #1467 is submitted, you should add an argument `ack func()` at the end of `CreateSubscription`, panic if it's nil, and return it from the `AckFunc` method on `Subscription`.

derekcollison:
ok will rebase and see what is what.. Thanks..

jba:
@derekcollison  reset those two files from current master.

pubsub/natspubsub/nats.go
derekcollison: Will take a look, thanks.

pubsub/natspubsub/nats.go
derekcollison: Flush in the NATS system sends a PING to the server and waits for a PONG, meaning all protocol messages sent will have been processed by the server. Makes things very slow and we usually discourage its use since NATS core is an at-most-once system. But even in STAN (at-least-once) we do not use it and wait on a reply from the stan-server.

pubsub/natspubsub/nats.go
derekcollison: I can add a test for this.

pubsub/natspubsub/nats_test.go
derekcollison: It did, not sure if you changed tests but when I added them yes they increased coverage.

derekcollison:
Do we still want to have this require and ackFunc for CreateSubscription? These seem implementation specific AFAIK..

pubsub/natspubsub/nats.go
derekcollison: Let me know if we need to iterate some more, but again for a system that is at-most-once this makes sense.

derekcollison:
Builds are green now so we are close. I have two outstanding items with ackFunc and the Flush call at the end of the SendBatch.

derekcollison:
They CreateSubscription are implementation specific based on the provider IIRC.

pubsub/natspubsub/nats.go
vangent: Agreed.

vangent:
Thanks again @derekcollison !

The PR looks like it's in a good place; I'll go ahead and add the "ack func" in a separate PR.

derekcollison:
Thanks..

PR #1382, opened 2019-02-22
internal/website: add a secrets page

shantuo:


internal/website/content/pages/secrets.md
```diff
@@ -0,0 +1,22 @@
+---
+title: "Secrets"
+---
+
+Package secrets provides an easy and portable way to encrypt and decrypt
```
eliben: `secrets` (with backticks)

internal/website/content/pages/secrets.md
```diff
@@ -0,0 +1,22 @@
...
+messages.
```
eliben: Can we add a bit more details here? From reading just this one line it's not clear why we need a portable cloud API for this as opposed to using Go's bultin crypto packages

internal/website/content/pages/secrets.md
shantuo: Done.

internal/website/content/pages/secrets.md
shantuo: Added some, PTAL.

Issue #1274, opened 2019-02-08
all: improve documentation and examples for "As"


vangent:
@eliben reported that it was confusing.

* The documentation probably shouldn't link to design.md, unless it's for a rationale about why we've implemented it this way.
* Need example code.
* ...?

eliben:
I was thinking where to move the provider/API-agnostic part of the documentation of `As`. One option is the landing page, but a better option may be just the big package comment in `doc.go`. This seems like too low-level detail-y for the landing page, and has the advantage of not having to link to non godoc/github resources from the code.

The `docs.go` package comment could have a section named "Accessing provider-specific types" that would explain what blob's bucket.As now explains, perhaps with relevant details from design.md; then we won't have to repeat this same documentation blurb in all the places in the code, instead just linking to the package comment. And we'll also remove the relevant parts from design.md

WDYT?

vangent:
Per meatspace discussion, I think it will be useful to have an explicit example per concrete API, ideally for each type of 'As' supported by that API.

eliben:
OK, next steps:

1. Move user-facing `As` documentation from `design.md` into `doc.go`
2. Add examples of `As` usage for concrete types

eliben:
Added a bunch of examples for sub-packages. Pubsub remaining. Will address that next

eliben:
I believe this is now fixed. If something specific is found missing, let's open new issues for that

PR #1388, opened 2019-02-25
blob: add As example

eliben:
Adding an example of As where the As call actually succeeds (the fileblog example in the main blob package shows how it can fail). Also renaming to make examples appear in the right places on godoc

Part of addressing #1274

### DIFF
--- a/blob/example_test.go
+++ b/blob/example_test.go
@@ -288,8 +288,7 @@ func ExampleBucket_As() {
 		log.Fatal(err)
 	}
 
-	// Bucket exposes the internal storage.Client type through the As method. Use
-	// it to access a non-portable method of storage.Client.
+	// Access storage.Client fields via gcsClient here.
 	var gcsClient *storage.Client
 	if b.As(&gcsClient) {
 		email, err := gcsClient.ServiceAccount(ctx, "project-name")
@@ -298,7 +297,7 @@ func ExampleBucket_As() {
 		}
 		_ = email
 	} else {
-		log.Fatal("Unable to access storage.Client through Bucket.As")
+		log.Println("Unable to access storage.Client through Bucket.As")
 	}
 }
 
--- a/blob/example_test.go
+++ b/blob/example_test.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"cloud.google.com/go/storage"
 	"gocloud.dev/blob"
 	"gocloud.dev/blob/fileblob"
 )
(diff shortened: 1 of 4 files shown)


runtimevar/runtimevar.go
```diff
@@ -84,7 +84,7 @@ type Snapshot struct {
 // provider-specific, and using them will make that part of your application
 // non-portable, so use with care.
 //
-// See the documentation for the subpackage used to instantiate Variable to see
+// See the documentation for the subpackage used to instantiate Snapshot to see
```
vangent: I think `Variable` is correct here. They got `Snapshot` from the concrete type, not a sub-package. Maybe "you used" instead of "used"?

blob/gcsblob/example_test.go
```diff
@@ -18,12 +18,13 @@ import (
...
 	"log"
 
+	"cloud.google.com/go/storage"
 	"gocloud.dev/blob"
 	"gocloud.dev/blob/gcsblob"
 	"gocloud.dev/gcp"
 )
 
-func Example() {
+func Example_read() {
```
vangent: Where does this put the example? It was supposed to be a package-level example before.

blob/gcsblob/example_test.go
```diff
@@ -64,3 +65,27 @@ func Example_open() {
 	_ = b
 	_ = err
 }
+
+func Example_As() {
```
vangent: So what's the overall strategy you're proposing here? So far, we've been only including examples for how to open a bucket along with each driver, and putting examples of how to use the concrete type functions in the concrete type package. `As` is a weird one because it's kind of a mix. Are you suggesting we should add a similar example to every driver? Just for top-level `As`, or for each subtype? (both of those feel like overkill to me, I'd rather have a good example in the concrete type).

runtimevar/runtimevar.go
eliben: Done

blob/gcsblob/example_test.go
eliben: I believe it still puts it at the package level and just names it more explictly as "Example (Read)".

If you look at https://godoc.org/gocloud.dev/blob/gcsblob, at the `As` function, this example comes immediately after it, which is confusing because it's not an example of `As`. Now that we have an actual `As` example, it may be not as important though.

blob/gcsblob/example_test.go
eliben: I'm not proposing a strategy, just adding an example where none existed before. I agree repeating this many times is an overkill - but having it in one place gives us something working to point to. So far the example in `blob` only shows something where `As` returns `false`; I figured that at least one place where it returns `true` is useful.

blob/gcsblob/example_test.go
vangent: OK

blob/gcsblob/example_test.go
vangent: Why not add this in the concrete type then?

eliben: I don't mind strongly either way. Wanted to keep provider-specific stuff from the generic `blob` package doc, but if it's not a concern I can move it there

vangent: I suggest replacing the existing `As` example in the concrete type (which, as you point out, it poor because it is expected to fail) with this one, which is a better example. It might also be good to extend it a bit to include `As` for write, read, error, and maybe even list.

Be sure to point out that the specific types used for each are documented in the driver-specific package doc (e.g., link to https://godoc.org/gocloud.dev/blob/gcsblob#hdr-As as an example).

eliben: PTAL. Moved it.
I actually think a single example is sufficient since they're all going to be pretty much the same.
The only one different is using `ErrorAs`, but this will likely require to use the s3 blob - I don't think gcsblob actually returns the gcs-specific error type anywhere

blob/example_test.go
```diff
@@ -273,50 +274,32 @@ func ExampleBucket_List_withDelimiter() {
...
-		if asFunc(&mutableProviderSpecific) {
-			fmt.Println("fileblob supports the `*string` type for WriterOptions.BeforeWrite")
-			// Use mutableProviderSpecific.
-		} else {
-			fmt.Println("fileblob does not support the `*string` type for WriterOptions.BeforeWrite")
+	// Bucket exposes the internal storage.Client type through the As method. Use
+	// it to access a non-portable method of storage.Client.
+	var gcsClient *storage.Client
+	if b.As(&gcsClient) {
+		email, err := gcsClient.ServiceAccount(ctx, "project-name")
```
vangent: Not sure you need this much detail here; maybe just a comment like 

```
// Access storage.Client fields via gcsClient here.

File: blob/example_test.go [hunk 5]
```diff
@@ -273,50 +274,32 @@ func ExampleBucket_List_withDelimiter() {
...
+			log.Fatal(err)
 		}
-		return nil
-	}
-	ctx := context.Background()
-	if err := bucket.WriteAll(ctx, "foo.txt", []byte("Go Cloud Development Kit"), &blob.WriterOptions{BeforeWrite: fn}); err != nil {
-		log.Fatal(err)
+		_ = email
+	} else {
+		log.Fatal("Unable to access storage.Client through Bucket.As")
```
vangent: In the example, I would avoid using log.Fatal for this.

blob/example_test.go
eliben: Done

blob/example_test.go
eliben: Done

eliben:
Thanks for the review! Next steps outlined in #1274

PR #1399, opened 2019-02-25
internal/website: fix base URL

zombiezen:


zombiezen:
I have confirmed locally that the output generated by Hugo links to the correct CSS now.

PR #1403, opened 2019-02-26
website: switch favicon to use the Go logo

eliben:


internal/website/themes/hyde/layouts/partials/head.html
```diff
@@ -23,7 +23,7 @@
 
   <!-- Icons -->
   <link rel="apple-touch-icon-precomposed" sizes="144x144" href="/apple-touch-icon-144-precomposed.png">
```
zombiezen: Optional: This icon probably does not match with your Go icon. Probably best to remove it for the time being.

eliben: Good idea, will do

PR #1406, opened 2019-02-26
internal/website: fix formatting and spelling

zombiezen:


internal/website/content/pages/pubsub.md
```diff
@@ -1,10 +1,10 @@
 ---
-title: "Pubsub"
+title: "Pub/Sub"
 ---
 
-Pubsub refers to implementations of the [publish-subscribe
+Pub/Sub refers to implementations of the [publish-subscribe
```
eliben: Should it be publish-subscribe or publisher subscriber (change in another file in this PR)? The wikipedia link does the former

internal/website/content/pages/runtimevar.md
```diff
@@ -1,27 +1,27 @@
 ---
-title: "Runtimevar"
+title: "Runtime Configuration"
```
eliben: Thinking aloud about this... Do we want the names to be package/component names or more descriptive sentences? We use "blob" but not "storage", and "secrets" is also quite terse

zombiezen: I'm mildly in favor of having the sidebar links be more descriptive and then in the page describe more of what they do (which is the direction this PR goes toward), but I can be convinced the other way. If we use package names directly, then we should lowercase the page titles to match the package names.

internal/website/content/pages/pubsub.md
zombiezen: The rule I had in my head was "Publisher/Subscriber" is a standalone noun, "publish-subscribe model" is a different noun.

internal/website/content/pages/runtimevar.md
eliben: But the others are not descriptive... "Server", "Blob", etc. 
Not sure about the casing, since a lowercase page title feels strange

zombiezen: Sure! I am willing to change more names. 🙂

eliben: I prefer to keep it as-is for now. The main page has a description for each api when linking to it, and each API page starts with a description. I feel that over time folks will use the side bar more for reference and will appreciate brief names that map to memory well. We can discuss this separately with the PMs/DevRel folks if you'd like?

zombiezen: Done. Kept as-is.

PR #1413, opened 2019-02-27
runtimevar: add infrastructure for opening Variables via URL

vangent:
Updates #1174.

I plan on adding per-driver openers in a separate PR.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -867,7 +867,7 @@ type WriterOptions struct {
 	BeforeWrite func(asFunc func(interface{}) bool) error
 }
 
-// A type that implements BucketURLOpener can open buckets based on a URL.
+// BucketURLOpener represents types that can open buckets based on a URL.
 // The opener must not modify the URL argument. OpenBucketURL must be safe to
 // call from multiple goroutines.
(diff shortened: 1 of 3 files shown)


vangent:
Note: this code is largely copied from `blob.go`, just repeating the pattern.

blob/blob.go
```diff
@@ -867,7 +867,7 @@ type WriterOptions struct {
 	BeforeWrite func(asFunc func(interface{}) bool) error
 }
 
-// A type that implements BucketURLOpener can open buckets based on a URL.
+// BucketURLOpener represents types that can open buckets based on a URL.
```
zombiezen: I'm curious why this changed. If the move is toward making the type be the first word of the sentence, it's worth interface type doc comments sometimes follow this format this exists. See [`sort.Interface`](https://golang.org/pkg/sort/#Interface)

vangent: I changed it because `golint` complains:

```
$ golint
blob.go:870:1: comment on exported type BucketURLOpener should be of the form "BucketURLOpener ..." (with optional leading article)
```

PR #1414, opened 2019-02-27
secrets: add infrastructure for opening Keepers via URL

vangent:
Updates #1174.

I'll add the per-driver openers in a separate PR.


vangent:
Note: this code is largely copied from `blob.go`, just repeating the pattern.

PR #1415, opened 2019-02-27
pubsub: add infrastructure for opening Topics/Subscriptions via URL

vangent:
Updates #1174.

I plan to add per-driver openers in a separate PR.

This code is largely lifted from `blob`.


vangent:
gentle ping @jba ?

pubsub/pubsub.go
```diff
@@ -546,3 +547,135 @@ func wrapError(ec errorCoder, err error) error {
...
+//
+// The zero value is a multiplexer with no registered schemes.
+type URLMux struct {
+	subscriptionSchemes map[string]SubscriptionURLOpener
+	topicSchemes        map[string]TopicURLOpener
+}
+
+// RegisterTopic registers the opener with the given scheme. If an opener
+// already exists for the scheme, RegisterTopic panics.
+func (mux *URLMux) RegisterTopic(scheme string, opener TopicURLOpener) {
```
jba: There's so much duplicate code here and below. Any way to factor it?

I don't care if we lose some error message context (e.g. "open topic" vs. "open subscription") and I'd even be OK with some use of `interface{}` and casting.

Like, how about a general URLMux type that had methods `OpenerFor(urlstr string) (interface{}, *url.URL)` and `OpenerForURL(*url.URL) interface{}`?

pubsub/pubsub_test.go
```diff
@@ -475,3 +476,127 @@ func TestShutdownsDoNotLeakGoroutines(t *testing.T) {
...
+		},
+		{
+			name: "fancy ps name",
+			url:  "foo:///foo/bar/baz",
+		},
+	} {
+		t.Run("topic: "+tc.name, func(t *testing.T) {
+			_, gotErr := mux.OpenTopic(ctx, tc.url)
+			if (gotErr != nil) != tc.wantErr {
+				t.Fatalf("got err %v, want error %v", gotErr, tc.wantErr)
```
jba: This going to say things like "got error blahblah, want error false" which is weird.

pubsub/pubsub_test.go
```diff
@@ -475,3 +476,127 @@ func TestShutdownsDoNotLeakGoroutines(t *testing.T) {
...
+			}
+			if gotErr != nil {
+				return
+			}
+			want, err := url.Parse(tc.url)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if diff := cmp.Diff(gotSub, want); diff != "" {
```
jba: It seems as if you're testing `url.Parse`. Can you explain what's happening here?

pubsub/pubsub.go
vangent: It is likely possible, but would make it harder to use (e.g., if it returned an `interface{}`, the user would have to cast it to the portable type).

If we could share implementation but leave the external interface the same, I'd be open to that, but it's hard because of all the different types.

pubsub/pubsub_test.go
vangent: What do you prefer? Maybe a question mark after `want error`, so "got error blahblah, want error? false"?

pubsub/pubsub_test.go
vangent: It's just verifying that the registered handler gets the right URL. Some of the tests are probably overkill since both the real code and the test use `URL.Parse`, but some of them also test other things like an unregistered scheme or a handler returning an error.

pubsub/pubsub.go
jba: Yes, I meant share implementation, not change the API.

I'll approve the PR as is but can you make an issue for this and assign it to me?

pubsub/pubsub_test.go
jba: I guess I'd prefer a colon.

pubsub/pubsub.go
vangent: Done. https://github.com/google/go-cloud/issues/1476.

pubsub/pubsub_test.go
vangent: Done. There are a few dozen other instances of this if you want to clean them up.

```
grep -R "got err" *
```

Issue #1306, opened 2019-02-12
all: run tests on windows to make sure things are working


vangent:

vangent:
@zombiezen have you run a Windows VM locally before to enable tests like this?

zombiezen:
I have run Windows VMs on GCE to check things in the past, but I don't think I've done that for this project.

vangent:
I sent a PR with some fixes for `blob`.

Additionally, `pubsub` tests are failing, I think due to timing issues with the batchers:

```
ok  	gocloud.dev/pubsub	6.980s
2019/03/01 00:03:41 ERROR: martian: failed to round trip: no matching request for &{Method:POST URL:https://sqs.us-east-2.amazonaws.com/ Proto:HTTP/1.1 ProtoMajor:1 ProtoMinor:1 Header:map[User-Agent:[aws-sdk-go/1.16.23 (go1.11.5; windows; amd64)] Content-Type:[application/x-www-form-urlencoded; charset=utf-8] X-Amz-Date:[20190301T000341Z] X-Amz-Security-Token:[FAKE_TOKEN] Accept-Encoding:[gzip] Content-Length:[177] Authorization:[AWS4-HMAC-SHA256 Credential=FAKE_ID/20190301/us-east-2/sqs/aws4_request, SignedHeaders=content-length;content-type;host;x-amz-date;x-amz-security-token, Signature=b99ef74d1510ef7280bcbac4b544580d299d61f6b4f293b75fb88379a698ddca] X-Cloud-Trace-Context:[9af66820f55577c99a1bb504f7e12bcd/1168718018569179706;o=0]] Body:{Reader:0xc0002afef0} GetBody:<nil> ContentLength:177 TransferEncoding:[] Close:false Host:sqs.us-east-2.amazonaws.com Form:map[] PostForm:map[] MultipartForm:<nil> Trailer:map[] RemoteAddr:127.0.0.1:50816 RequestURI:/ TLS:<nil> Cancel:<nil> Response:<nil> ctx:<nil>}
[...]

PR #1419, opened 2019-02-28
blob/fileblob: some fixes for windows

vangent:
Updates #1306.

Windows is super picky about readers/writers being closed, and has some additional filename restrictions.


vangent:
gentle ping @jba ?

blob/fileblob/fileblob.go
```diff
@@ -163,6 +164,9 @@ func escapeKey(s string) string {
 		// Escape the trailing slash in a key.
 		case c == '/' && i == len(r)-1:
 			return true
+		// https://docs.microsoft.com/en-us/windows/desktop/fileio/naming-a-file
+		case os.PathSeparator == '\\' && (c == '>' || c == '<' || c == ':' || c == '"' || c == '|' || c == '?' || c == '*'):
```
jba: Are you sure `os.PathSeparator` is a valid proxy for Windows? You could check `runtime.GOOS`.

vangent: Windows definitely uses `\`, and I don't think any other operating systems do. I could be wrong.

`os.PathSeparator` is used in enough places in this file that I'd rather it be consistent. If there's an OS that uses `\` other than Windows, or an OS that uses a pathSeparator other than `/` or `\`, I'd be shocked if they worked out of the box regardless.

PR #1420, opened 2019-02-28
aws: refactor aws.Config URL parsing to aws package

vangent:
This moves some code from `s3blob` to `aws`, in preparation for using it from other APIs.

Updates #1174.

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -16,6 +16,9 @@
 package aws // import "gocloud.dev/aws"
 
 import (
+	"fmt"
+	"net/url"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/aws/credentials"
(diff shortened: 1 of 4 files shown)


aws/aws.go
```diff
@@ -56,3 +59,27 @@ func (co ConfigOverrider) ClientConfig(serviceName string, cfgs ...*aws.Config)
...
+	var cfg aws.Config
+	for param, values := range q {
+		value := values[0]
+		switch param {
+		case "region":
+			cfg.Region = aws.String(value)
+		case "endpoint":
+			cfg.Endpoint = aws.String(value)
+		case "disableSSL":
+			cfg.DisableSSL = aws.Bool(value == "true")
```
zombiezen: Use `strconv.ParseBool`?

blob/s3blob/s3blob.go
```diff
@@ -129,53 +129,14 @@ func (o *URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket
...
 		return nil, fmt.Errorf("open bucket %v: %v", u, err)
 	}
-	if overrideCfg != nil {
-		configProvider.Configs = append(configProvider.Configs, overrideCfg)
-	}
+	configProvider.Configs = append(configProvider.Configs, overrideCfg)
 	return OpenBucket(ctx, configProvider, u.Host, &o.Options)
 }
 
-var legalQueryParam = map[string]bool{
```
zombiezen: Seems useful to disallow unrecognized query parameters. Consider adding a function in `aws` that would indicate whether a query parameter would be recognized by the common set.

aws/aws.go
vangent: Yes, done.

blob/s3blob/s3blob.go
vangent: There's a `default` clause that returns an error for unrecognized query parameters; this avoids having a separate `map` of them.

The expectation is that if the caller has additional query parameters that they want to process, they'll do that first and remove them from `url.Values` before calling this function. (`s3blob` doesn't have any additional query parameters though).

vangent: (added a comment to this effect)

aws/aws.go
```diff
@@ -56,3 +60,43 @@ func (co ConfigOverrider) ClientConfig(serviceName string, cfgs ...*aws.Config)
...
+//
+// It returns an error if q contains any unknown query parameters; callers
+// should remove any query parameters they know about from q before calling
+// ConfigFromURLParams.
+//
+// The following query options are supported:
+//  - region: The AWS region for requests; sets aws.Config.Region.
+//  - endpoint: The endpoint URL (hostname only or fully qualified URI); sets aws.Config.Endpoint.
+//  - disableSSL: A value of "true" disables SSL when sending requests; sets aws.Config.DisableSSL.
+//  - s3ForcePathStyle: A value of "true" forces the request to use path-style addressing; sets aws.Config.S3ForcePathStyle.
```
zombiezen: Optional: this option is only applicable to S3 requests, perhaps make it a blob-specific parameter?

blob/s3blob/s3blob.go
zombiezen: Thanks, I missed that on first read.

aws/aws.go
vangent: Interesting. Again, this seems not ideal by Amazon. But, I think I'd rather leave all of the `aws.Config` stuff together rather than try to "fix" it for them.

Issue #1423, opened 2019-03-01
pubsub: NewTopic should take a batcher constructor like NewSubscription


vangent:
And the driver tests should use predictable batchers, not the default timing-based ones.

PR #1424, opened 2019-03-01
pubsub: add a batch constructor to NewTopic

vangent:
This is basically the same thing that was already done for `NewSubscription`, but for `NewTopic` this time.

In a separate PR, I will update the `drivertest` conformance tests to pass non-nil values for the batchers, so that record/replay tests are repeatable. Currently they are flaky because they are based on timing. For example, Windows is dog slow and so the second "ReceiveBatch" call has a different `maxMessages` argument than the current goldens (1000).

Updates #1423.

### DIFF
--- a/pubsub/awspubsub/awspubsub.go
+++ b/pubsub/awspubsub/awspubsub.go
@@ -115,7 +115,7 @@ type TopicOptions struct {
 
 // OpenTopic opens the topic on AWS SNS for the given SNS client and topic ARN.
 func OpenTopic(ctx context.Context, client *sns.SNS, topicARN string, opts *TopicOptions) *pubsub.Topic {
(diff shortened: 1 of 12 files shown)


vangent:
Thinking about this more, it won't directly solve the problem, since the problem is the `maxMessages` argument being passed, not the size of the batch itself.

I still kind of think this change is good, for consistency reasons and because I think it will be relatively easy to write tests that break for record/replay without it.

For fixing the `maxMessages` issue, WDYT about adding a `MaxBatchSize() int` function to `driver.Batcher`, and changing the hard-coded cap of 1000 (https://github.com/google/go-cloud/blob/master/pubsub/pubsub.go#L396) to use `batcher.MaxBatchSize` instead? Then record/replay tests could provide a batcher with MaxBatchSize=1.

jba:
SGTM.

PR #1427, opened 2019-03-01
secrets: add support for URLOpeners to all drivers

vangent:
This is mostly mechanical, copying the pattern @zombiezen established for `blob`.

The interesting parts are probably:

* I made `localsecrets` register for 2 schemes, 'stringkey://" and "base64key://".
* `vault` is a bit interesting since the URL contains the Vault server address + token; I made it Dial but cache the result.

Updates #1174.

### DIFF
--- a/internal/testing/setup/setup.go
+++ b/internal/testing/setup/setup.go
@@ -291,10 +291,7 @@ func FakeGCPDefaultCredentials(t *testing.T) func() {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if _, err := f.Write(jsonCred); err != nil {
-		t.Fatal(err)
-	}
-	if err := f.Close(); err != nil {
+	if err := ioutil.WriteFile(f.Name(), jsonCred, 0666); err != nil {
 		t.Fatal(err)
 	}
 	oldEnvVal := os.Getenv(envVar)
--- a/internal/testing/setup/setup.go
+++ b/internal/testing/setup/setup.go
@@ -15,6 +15,7 @@
 package setup // import "gocloud.dev/internal/testing/setup"
 
 import (
+	"io/ioutil"
 	"context"
 	"flag"
 	"net"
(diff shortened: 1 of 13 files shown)


secrets/awskms/kms.go
```diff
@@ -24,17 +34,96 @@ package awskms // import "gocloud.dev/secrets/awskms"
...
+	}
+	return kms.New(p), nil
+}
+
+// URLOpener opens secrets.Keeper URLs for AWS KMS, like "awskms://keyID".
+// The keyID can be in the form of an Amazon Resource Name (ARN), alias
+// name, or alias ARN. See
+// https://docs.aws.amazon.com/kms/latest/developerguide/viewing-keys.html#find-cmk-id-arn
+// for more details.
+// The URL Host + Path are used as the keyID, to support alias names like "awskms://alias/foo".
```
shantuo: > The URL Host + Path are used as the keyID

Here the keyID can be understood as either 1) the key resource ID this provider uses in the corresponding KMS, or 2) the resource ID we, the Go CDK, defined as an identifier which contains 1). It seems to be 2) and is worth clarifying somewhere. Maybe in the package doc.

secrets/localsecrets/localsecrets.go
```diff
@@ -24,14 +32,57 @@ package localsecrets // import "gocloud.dev/secrets/localsecrets"
...
 	"gocloud.dev/secrets"
 	"golang.org/x/crypto/nacl/secretbox"
 )
 
+func init() {
+	secrets.DefaultURLMux().RegisterKeeper(SchemeString, &URLOpener{})
+	secrets.DefaultURLMux().RegisterKeeper(SchemeBase64, &URLOpener{base64: true})
+}
+
+// SchemeString is the URL scheme localsecrets registers its URLOpener under on secrets.DefaultMux.
```
shantuo: nit: Combine these two consts, and have one comment line on top. That way it would show up in the godoc together instead of side by side with the identical documentation.

internal/testing/setup/setup.go
```diff
@@ -280,3 +281,26 @@ func NewAzureTestPipeline(ctx context.Context, t *testing.T, api string, credent
 	})
 	return p, done, httpClient
 }
+
+// FakeGCPDefaultCredentials sets up the environment with fake GCP credentials.
+// It returns a cleanup function.
+func FakeGCPDefaultCredentials(t *testing.T) func() {
+	const envVar = "GOOGLE_APPLICATION_CREDENTIALS"
+	jsonCred := []byte(`{"client_id": "foo.apps.googleusercontent.com", "client_secret": "bar", "refresh_token": "baz", "type": "authorized_user"}`)
+	f, err := ioutil.TempFile("", "fake-gcp-creds")
```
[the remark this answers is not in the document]
vangent: Done.

secrets/awskms/kms.go
vangent: Sorry, what is the difference?

I intended this to be the AWS key ID, which can take several forms as discussed at the link pointed to above. They refer to it as "key ID", not "key resource ID". I'll update some comments to say "key ID" instead of "keyID" or "key resource ID" to be consistent with them.

I don't think there's a separate resource ID that we're defining.

secrets/localsecrets/localsecrets.go
vangent: Done.

secrets/awskms/kms.go
shantuo: I think I had misunderstood that, I was thinking they can mean both our definition and provider's definition of key ID. But they are essentially the same thing. Thanks for making these consistent though, that could help.

vangent:
Gentle ping @clausti ?

PR #1431, opened 2019-03-01
blob: add tests for default URL openers

vangent:


blob/fileblob/fileblob.go
```diff
@@ -88,6 +88,9 @@ type URLOpener struct{}
 //  - file://localhost/c:/foo/bar
 //    -> Also passes "c:\foo\bar".
 func (*URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket, error) {
+	for param := range u.Query() {
```
shantuo: Should unsupported query parameters just be ignored? I think that's more common than error out.

blob/memblob/memblob.go
```diff
@@ -59,10 +60,14 @@ func init() {
 const Scheme = "mem"
 
 // URLOpener opens URLs like "mem://".
+// No query parameters are supported.
 type URLOpener struct{}
 
 // OpenBucketURL returns a new in-memory bucket.
 func (*URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket, error) {
+	for param := range u.Query() {
```
shantuo: Same here.

internal/testing/setup/setup.go
```diff
@@ -280,3 +281,23 @@ func NewAzureTestPipeline(ctx context.Context, t *testing.T, api string, credent
 	})
 	return p, done, httpClient
 }
+
+// FakeGCPDefaultCredentials sets up the environment with fake GCP credentials.
+// It returns a cleanup function.
+func FakeGCPDefaultCredentials(t *testing.T) func() {
+	const envVar = "GOOGLE_APPLICATION_CREDENTIALS"
+	jsonCred := []byte(`{"client_id": "foo.apps.googleusercontent.com", "client_secret": "bar", "refresh_token": "baz", "type": "authorized_user"}`)
+	f, err := ioutil.TempFile("", "fake-gcp-creds")
```
shantuo: This can be removed, `ioutil.WriteFile` will create the file if it doesn't exist.

blob/fileblob/fileblob.go
vangent: This is consistent with how it's handled elsewhere. My first thought is that it's better to error out than just silently ignore things, but I haven't thought about it that much. If you disagree, can you file a separate issue to discuss?

internal/testing/setup/setup.go
vangent: Right, but `TempFile` is needed to get the temporary file name.

vangent:
PTAL

blob/fileblob/fileblob.go
shantuo: Filed #1435

PR #1432, opened 2019-03-01
blob: add examples for remaining As() types that gcsblob supports

eliben:
For `Error`, I'll do a followup example that uses `s3blob`.

### DIFF
--- a/blob/example_test.go
+++ b/blob/example_test.go
@@ -299,6 +299,33 @@ func ExampleBucket_As() {
 	}
 }
 
+func ExampleWriterOptions_BeforeWrite() {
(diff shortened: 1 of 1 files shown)


blob/example_test.go
```diff
@@ -329,7 +356,7 @@ func ExampleListObject_As() {
 	}
 }
 
-func ExampleListOptions_BeforeList() {
+func ExampleListOptions() {
```
eliben: I changed this because godoc currently doesn't show this example at all. I suspect this is because `BeforeList` is a struct field rather than a method/function. With the new name it will hopefully just show beneath the struct.

vangent: ok

PR #1433, opened 2019-03-02
blob: add example of ErrorAs

eliben:
For #1274

### DIFF
--- a/blob/example_test.go
+++ b/blob/example_test.go
@@ -171,6 +171,11 @@ func Example() {
 }
 
 func ExampleBucket_ErrorAs() {
+	// This example is specific to the s3blob implementation; it demonstrates
+	// access to the underlying awserr.Error type.
+	// The types exposed for ErrorAs by s3blob are documented in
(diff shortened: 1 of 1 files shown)


blob/example_test.go
```diff
@@ -169,6 +170,28 @@ func Example() {
...
 
+func ExampleBucket_ErrorAs() {
+	// This example is specific to the s3blob implementation; it demonstrates
+	// access to the underlying awserr.Error type.
+	// The types exposed for ErrorAs by s3blob are documented in
+	// https://godoc.org/gocloud.dev/blob/s3blob#hdr-As
+
+	ctx := context.Background()
+
+	b, err := blob.OpenBucket(ctx, "s3://my-bucket")
```
vangent: It might be more clear to do the empty import for s3blob and gcsblob?

vangent: (without those, the `OpenBucket` calls will fail).

eliben: Done

PR #1439, opened 2019-03-04
blob/s3blob: support v1 ListObjects

EvyBongers:
Be compatible with CEPH as blob store provider

See https://tracker.ceph.com/issues/37801 for the bug report with the CEPH project.

### DIFF
--- a/blob/s3blob/s3blob.go
+++ b/blob/s3blob/s3blob.go
@@ -146,6 +146,9 @@ func openBucket(ctx context.Context, sess client.ConfigProvider, bucketName stri
 	if bucketName == "" {
 		return nil, errors.New("s3blob.OpenBucket: bucketName is required")
 	}
+	if opts == nil {
+		opts = &Options{}
+	}
 	return &bucket{
 		name:                 bucketName,
 		client:               s3.New(sess),
--- a/blob/s3blob/s3blob.go
+++ b/blob/s3blob/s3blob.go
@@ -135,7 +135,7 @@ func (o *URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket
 
 // Options sets options for constructing a *blob.Bucket backed by fileblob.
 type Options struct {
-	UseLegacyListObjects bool
+	UseLegacyList bool
 }
 
 // openBucket returns an S3 Bucket.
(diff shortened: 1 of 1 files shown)


blob/s3blob/s3blob.go
```diff
@@ -134,19 +134,25 @@ func (o *URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket
 }
 
 // Options sets options for constructing a *blob.Bucket backed by fileblob.
-type Options struct{}
+type Options struct {
+	UseLegacyListObjects bool
```
vangent: Please add a docstring for the new option. Perhaps explain that some providers have S3-compatible APIs, but don't support the ListObjectV2 API yet.

blob/s3blob/s3blob.go
```diff
@@ -366,6 +377,89 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
 	return &page, nil
 }
 
+func (b *bucket) legacyListPaged(ctx context.Context, opts *driver.ListOptions) (*driver.ListPage, error) {
+	pageSize := opts.PageSize
```
vangent: It looks like a high percentage of the logic is shared between `ListPaged` and `legacyListPaged`. Could you try sharing more of it? For example, you could make only the `ListObjectsRequest` call conditional, and in the legacy version convert from the V2 objects to the V1 objects (and from V1 to V2 in the result).

blob/s3blob/s3blob.go
vangent: If I understand correctly, S3 itself should support both modes, right?

If so, I'd like to make the tests for S3 run in both modes so we have coverage of both paths. I could do this as a follow-up on PR from yours.

vangent: I verified that this works (i.e., I ran all of the conformance tests for S3 using `UseLegacyListObjects=true` and they passed). So, resolving this comment.

vangent: Maybe just `UseLegacyList` ?

blob/s3blob/s3blob.go
EvyBongers: Hi @vangent,

I've updated the code (see 24d73e47894c24b023fc082c5bdc8290d3c4919b). To me the previous solution looks cleaner. What do you think?

vangent: I like this one, thanks!

PR #1440, opened 2019-03-04
docs: add new logo to README

eliben:


README.md
```diff
@@ -6,6 +6,10 @@ _Write once, run on any cloud ☁️_
 [![godoc](https://godoc.org/github.com/google/go-cloud?status.svg)][godoc]
 [![Coverage](https://codecov.io/gh/google/go-cloud/branch/master/graph/badge.svg)](https://codecov.io/gh/google/go-cloud)
```
zombiezen: We previously had the logo before the first heading (see 11179ad85494f2e152628bcb79bddaba42dab4ca).

README.md
```diff
@@ -6,6 +6,10 @@ _Write once, run on any cloud ☁️_
...
 
+<p align="center">
```
zombiezen: I think I'm more okay with it not being centered.

README.md
```diff
@@ -6,6 +6,10 @@ _Write once, run on any cloud ☁️_
...
+  <img width="509" height="276" src="internal/website/static/go-cdk-logo-gopherblue.png">
```
zombiezen: Make sure to include `alt=""` to mark the image as purely decorative and not necessary for screenreaders to read.

```suggestion
  <img width="509" height="276" src="internal/website/static/go-cdk-logo-gopherblue.png" alt="">
```

README.md
eliben: Can you say why? I think it looks significantly better when centered

README.md
eliben: Yes, I'm aware, but the new location seemed like a more common place to put logos in. For example see the github repos of moby, kubernetes, go, ...

zombiezen: OK.

README.md
zombiezen: It was a grouping thing: having it be with the heading and left-aligned would more strongly associate the logo and heading. But since we're moving the logo to a position that causes a natural break anyway, centering establishes it as a separate element.

For sake of using standard markup, let's use `style` instead:

```suggestion
<p style="text-align:center;">
```

PR #1444, opened 2019-03-04
blob: fix examples to reference OpenBucket, not Open

vangent:


blob/azureblob/example_test.go
```diff
@@ -90,10 +90,10 @@ func Example_sasToken() {
 	_ = data
 }
 
-func Example_open() {
+func Example_openBucket() {
```
eliben: Shouldn't it be something like `ExampleOpenBucket` here and elsewhere?

vangent: I don't think that work, since `OpenBucket` is not a function in this package (it's a function in the concrete type). I believe this will make the example show up at the package level.

https://godoc.org/gocloud.dev/runtimevar is an example, search for `JSONVariable` and compare to https://github.com/google/go-cloud/blob/master/runtimevar/example_test.go#L26.

PR #1446, opened 2019-03-04
pubsub: add test of sending after starting receive

ijt:
A replay file is added for `gcppubsub` but not for `awspubsub` because of https://github.com/google/go-cloud/issues/1447.


pubsub/drivertest/drivertest.go
```diff
@@ -245,6 +249,52 @@ func testSendReceive(t *testing.T, newHarness HarnessMaker) {
...
+		}
+		m.Ack()
+		gotChan <- m
+	}()
+
+	// Wait for Receive to block, with any luck.
+	time.Sleep(100 * time.Millisecond)
+
+	// Send a message.
+	want := &pubsub.Message{Body: []byte(fmt.Sprintf("%d", rand.Int()))}
```
vangent: Why `rand`? Won't that make record/replay problematic?

pubsub/drivertest/drivertest.go
```diff
@@ -245,6 +249,52 @@ func testSendReceive(t *testing.T, newHarness HarnessMaker) {
...
+	if err := top.Send(ctx, want); err != nil {
+		t.Fatal(err)
+	}
+
+	// The same message should come through from gotChan.
+	got := <-gotChan
+	gotStr := string(got.Body)
```
vangent: I think you can just use `cmp.Equal` here instead of casting, then `%v` in the error.

pubsub/drivertest/drivertest.go
ijt: I like to use rand in tests to make it clear that the specific value isn't important, but I see your point. If the tests run in the same order every time then we'll get the same random numbers each time (https://github.com/golang/go/issues/16486), but their order could change or become non-deterministic if we run them in parallel.

Replacing it with something deterministic.

pubsub/drivertest/drivertest.go
ijt: Changed it to use `cmp.Equal` but `%v` would print out the byte slices literally so I'd like to leave it as `%s`.

ijt:
PTAL

pubsub/drivertest/drivertest.go
```diff
@@ -141,6 +142,7 @@ func (verifyAsFailsOnNil) MessageCheck(m *pubsub.Message) error {
 func RunConformanceTests(t *testing.T, newHarness HarnessMaker, asTests []AsTest) {
 	tests := map[string]func(t *testing.T, newHarness HarnessMaker){
 		"TestSendReceive":                                         testSendReceive,
+		"TestReceiveSend":                                         testReceiveSend,
```
vangent: Don't you need to add golden files for existing providers?

ijt: I did for gcp but the aws driver hangs on this.

vangent: OK... I think you need to fix that before you can submit this. The tests for the PR fail.

ijt: Looking into it.

ijt: Hmm, I think this has to do with the queues not getting auto-subscribed to the topics in the tests.

ijt: Done.

ijt:
PTAL

vangent:
```
--- FAIL: TestConformance (0.20s)
    --- FAIL: TestConformance/TestReceiveSend (0.11s)
        replay.go:200: Replaying from golden file testdata/TestConformance/TestReceiveSend.replay
        drivertest.go:284: pubsub (code=Unknown): replayer: request not found: topic:"projects/go-cloud-test-216917/topics/TestConformance_TestReceiveSend-topic-1" messages:<data:"TestConformance/TestReceiveSend" > 
FAIL
coverage: 10.0% of statements in ./...
FAIL	gocloud.dev/pubsub/gcppubsub	1.011s
```

ijt:
That's odd. It passes locally.

PR #1448, opened 2019-03-04
docs: move As documentation to doc.go

eliben:
This will make it easier for us to point to this documentation for more details from all the places where As is described

For #1274

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -35,5 +35,53 @@ increasing compile times or binary sizes, and avoiding any side effects from
 
 For sample applications and a tutorial, see the samples directory
 (https://github.com/google/go-cloud/tree/master/samples).
+
+
+Escaping the abstraction
+
+It is not feasible or desirable for APIs like blob.Bucket to encompass the full
+functionality of every provider. Rather, we intend to provide a subset of the
+most commonly used functionality. There will be cases where a developer wants
+to access provider-specific functionality, which might consist of:
+
+1. Top-level APIs. For example, blob does not expose a Copy, but some provider
+might.
+
+2. Data fields. For example, blob exposes a few attributes like ContentType
+and Size, but S3 and GCS both have many more.
+
+3. Errors. For example, S3 returns awserr.Error.
+
+4. Options. Different providers may support different options for functionality.
+
+As functions in the APIs provide the user a way to escape the Go CDK
+abstraction to access provider-specific types. They might be used as an interim
+solution until a feature request to the Go CDK is implemented. Or, the Go CDK
+may choose not to support specific features, and the use of As will be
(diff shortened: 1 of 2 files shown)


doc.go
```diff
@@ -35,5 +35,53 @@ increasing compile times or binary sizes, and avoiding any side effects from
...
+
+
+Escaping the abstraction
+
+It is not feasible or desirable for APIs like blob.Bucket to encompass the full
+functionality of every provider. Rather, we intend to provide a subset of the
+most commonly used functionality. There will be cases where a developer wants
+to access provider-specific functionality, which might consist of:
+
+1. Top-level APIs. For example, blob does not expose a Copy, but some provider
```
vangent: WDYT about dropping this list, or collapsing it into a parenthetical? Seems like overkill for here.

doc.go
```diff
@@ -35,5 +35,53 @@ increasing compile times or binary sizes, and avoiding any side effects from
...
+
+3. Errors. For example, S3 returns awserr.Error.
+
+4. Options. Different providers may support different options for functionality.
+
+As functions in the APIs provide the user a way to escape the Go CDK
+abstraction to access provider-specific types. They might be used as an interim
+solution until a feature request to the Go CDK is implemented. Or, the Go CDK
+may choose not to support specific features, and the use of As will be
+permanent. As an example, both S3 and GCS blobs have the concept of ACLs, but
```
vangent: Same -- consider dropping the example?

doc.go
```diff
@@ -35,5 +35,53 @@ increasing compile times or binary sizes, and avoiding any side effects from
...
+it might be difficult to represent them in a generic way (although, we have not
+tried).
+
+Using As implies that the resulting code is no longer portable; the
+provider-specific code will need to be ported in order to switch providers.
+Therefore, it should be avoided if possible. For example:
+
+	// The existing blob.Reader exposes some blob attributes, but not everything
```
vangent: Drop the example and assert that each API should have specific examples for As?

doc.go
eliben: Good points

eliben:
PTAL

doc.go
```diff
@@ -42,46 +42,21 @@ Escaping the abstraction
...
-	var s3type s3.GetObjectOutput
-	if r.As(&s3type) {
-		... use s3type...
-	}
+Therefore, it should be avoided if possible.
 
-Each provider implementation documents what type(s) it supports for each of the
-As functions.
+Each provider implementation must document what type(s) it supports for each of
+the As functions.
```
vangent: How about:

```
Each API will include examples demonstrating how to use its various As functions, and each provider implementation will document what types it supports for each.
```

?

eliben: Done

PR #1449, opened 2019-03-05
runtimevar: support OpenVariable URLs for blobvar and constantvar

vangent:
Updates #1174.

I decided to break up adding `OpenVariable` support to `runtimevar` into multiple PRs to make it easier to review.


vangent:
@zombiezen per local discussion: an alternative to the "lazy opener cache" is to invent an environment variable with the value. This would be a bit more consistent with AWS/GCP, but requires us to invent new environment variables. https://github.com/google/go-cloud/pull/1427/files#diff-f7d4c4cec11eb3091725cf02314bbb63 is probably a better example of the "lazy opener cache" than this PR (in this PR, it's used for `blobvar`, where the cache is bucket URL -> `*blob.Bucket`).

runtimevar/blobvar/blobvar.go
```diff
@@ -34,6 +51,90 @@ import (
...
+		var err error
+		o.Decoder, err = runtimevar.DecoderByName(decoderName)
+		if err != nil {
+			return nil, fmt.Errorf("open variable %q: invalid \"decoder\": %v", u, err)
+		}
+	}
+	for param, values := range q {
+		val := values[0]
+		switch param {
+		case "decoder", "bucket":
```
shantuo: "decoder" can be processed here instead of above.

runtimevar/blobvar/blobvar_test.go
```diff
@@ -17,8 +17,11 @@ package blobvar
 import (
 	"context"
 	"errors"
+	"github.com/google/go-cmp/cmp"
```
shantuo: nit: move this to the group below.

runtimevar/blobvar/blobvar_test.go
```diff
@@ -88,3 +91,66 @@ func (verifyAs) ErrorCheck(v *runtimevar.Variable, err error) error {
...
+		{"blob://filenotfound?bucket=file://" + dir + "&decoder=string", false, true, nil},
+		// Variable construction fails due to missing bucket arg.
+		{"blob://myvar.txt?decoder=string", true, false, nil},
+		// Variable construction fails due to invalid wait arg.
+		{"blob://myvar.txt?bucket=file://" + dir + "&decoder=string&wait=notaduration", true, false, nil},
+		// Variable construction fails due to invalid decoder arg.
+		{"blob://myvar.txt?bucket=file://" + dir + "&decoder=notadecoder", true, false, nil},
+		// Variable construction fails due to invalid arg.
+		{"blob://myvar.txt?bucket=file://" + dir + "&decoder=string&param=value", true, false, nil},
+		// Working example with string decoder and wait.
```
shantuo: Consider add an example with no decoder parameter so that it uses default bytes decoder.

runtimevar/blobvar/example_test.go
```diff
@@ -63,3 +63,19 @@ func Example() {
...
+	v, err := runtimevar.OpenVariable(ctx, "blob://myvar.json?bucket=file:///mypath&decoder=json")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	snapshot, err := v.Watch(ctx)
+	if err != nil {
+		log.Fatal(err)
+	}
+	_, _ = snapshot, err
```
shantuo: Don't need the `, err` part since you've already handled it.

runtimevar/constantvar/constantvar.go
```diff
@@ -24,13 +37,51 @@ package constantvar // import "gocloud.dev/runtimevar/constantvar"
...
+	var value string
+	var errVal error
+	for param, values := range q {
+		val := values[0]
+		switch param {
+		case "decoder":
+			// processed elsewhere
+		case "val":
+			value = val
+		case "err":
```
shantuo: If the URL contains both "val" and "err", it returns the error, I think we should document this.

runtimevar/blobvar/blobvar.go
```diff
@@ -16,6 +16,19 @@
 // variables read from a blob.Bucket.
 // Use NewVariable to construct a *runtimevar.Variable.
 //
+// URLs
+//
+// For runtimevar.OpenVariable URLs, blobvar registers for the scheme "blob".
```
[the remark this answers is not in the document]
vangent: The requirement is that schemes are unique per package, not unique globally.

#1453 will make it so that you can also (optionally) add the API to the scheme, e.g. `runtimevar+blob://blahblahblah`, to provide a consistent way of making URLs globally unique and more self-documenting; I think that addresses your concern.

runtimevar/blobvar/blobvar.go
vangent: It can't, at least not easily, because of the `|| o.Decoder == nil` clause above; if there is no `Decoder` provided in the `URLOpener`, we want to get the default decoder (by passing "" to `runtimevar.DecoderByName`). In this loop, if there's no decoder param, we'll never call it.

runtimevar/blobvar/blobvar_test.go
vangent: Done.

runtimevar/blobvar/blobvar_test.go
vangent: Good idea, done.

runtimevar/constantvar/constantvar.go
vangent: Done.

runtimevar/blobvar/example_test.go
vangent: Removed the handling, it's not important to the example.

runtimevar/blobvar/blobvar.go
```diff
@@ -16,6 +16,19 @@
...
+// The host+path is used as the blob key. blobvar supports the following URL
+// parameters:
+//   - bucket: The URL to be passed to blob.OpenBucket (required).
```
zombiezen: I would expect the reverse of this: I would expect the key to be in a query parameter. It would be nice if this was: `blob+file:///foo/bar&key=myvar.txt&decoder=string`

(I realize this would mean that this would mean the full scheme might be `runtimevar+blob+file`, but I think I'm okay with that.)

runtimevar/blobvar/blobvar.go
```diff
@@ -16,6 +16,19 @@
...
+//       blob.OpenBucket will be called once per unique bucket URL.
+//   - decoder: The decoder to use. Defaults to runtimevar.BytesDecoder.
+//       See runtimevar.DecoderByName for supported values.
+//   - wait: The poll interval; supported values are from time.ParseDuration.
+//       Defaults to 30s.
+// Example URL: "blob://myvar.txt?bucket=fileblob:///foo/bar&decoder=string".
```
zombiezen: Scheme is incorrect here:

```suggestion
// Example URL: "blob://myvar.txt?bucket=file:///foo/bar&decoder=string".
```

runtimevar/blobvar/blobvar.go
```diff
@@ -34,6 +51,90 @@ import (
...
+		o.buckets = map[string]*blob.Bucket{}
+	}
+	bucket := o.buckets[bucketURL]
+	if bucket == nil {
+		var err error
+		bucket, err = blob.OpenBucket(ctx, bucketURL)
+		if err != nil {
+			return nil, fmt.Errorf("open variable %q: OpenBucket failed: %v", u, err)
+		}
+		o.buckets[bucketURL] = bucket
```
zombiezen: It would be nice if these buckets would get reclaimed as their variables get closed, but I can't see this causing huge problems. Mind opening an issue to track?

runtimevar/blobvar/blobvar.go
```diff
@@ -34,6 +51,90 @@ import (
...
+// URLOpener opens Variable URLs like "blob://myblobkey?decoder=string".
+// It supports the URL parameters:
+//   - decoder: The decoder to use. Defaults to URLOpener.Decoder, or
+//       runtimevar.BytesDecoder if URLOpener.Decoder is nil.
+//       See runtimevar.DecoderByName for supported values.
+//   - wait: The poll interval; supported values are from time.ParseDuration.
+//       Defaults to 30s.
+type URLOpener struct {
+	// The Bucket to use; required.
+	Bucket *blob.Bucket
```
zombiezen: I would assume this `URLOpener` to have a `*blob.URLMux`, not a single `*blob.Bucket`.

runtimevar/blobvar/blobvar.go
```diff
@@ -34,6 +51,90 @@ import (
...
+func init() {
+	runtimevar.DefaultURLMux().RegisterVariable(Scheme, &lazyBucketOpener{})
+}
+
+// Scheme is the URL scheme blobvar registers its URLOpener under on runtimevar.DefaultMux.
+const Scheme = "blob"
+
+type lazyBucketOpener struct {
+	mu      sync.Mutex
+	buckets map[string]*blob.Bucket
```
zombiezen: Related to my suggestion below to have `URLOpener` use a `*blob.URLMux`: I would expect this sort of cache to be on the `URLOpener`, not the global default.

runtimevar/blobvar/blobvar.go
```diff
@@ -34,6 +51,90 @@ import (
...
 )
 
+func init() {
+	runtimevar.DefaultURLMux().RegisterVariable(Scheme, &lazyBucketOpener{})
+}
+
+// Scheme is the URL scheme blobvar registers its URLOpener under on runtimevar.DefaultMux.
+const Scheme = "blob"
+
+type lazyBucketOpener struct {
```
zombiezen: nit: There isn't a lazy evaluation here, really. I think that this global could probably use the same `URLOpener`.

runtimevar/constantvar/constantvar.go
```diff
@@ -24,13 +38,51 @@ package constantvar // import "gocloud.dev/runtimevar/constantvar"
...
 )
 
+func init() {
+	runtimevar.DefaultURLMux().RegisterVariable(Scheme, &urlOpener{})
+}
+
+// Scheme is the URL scheme constantvar registers its URLOpener under on blob.DefaultMux.
+const Scheme = "constant"
+
+type urlOpener struct{}
```
zombiezen: This should be exported so that Wire users can use the implementation.

runtimevar/constantvar/constantvar_test.go
```diff
@@ -17,6 +17,7 @@ package constantvar
 import (
 	"context"
 	"errors"
+	"github.com/google/go-cmp/cmp"
```
zombiezen: nit: This belongs in the second import group (non-stdlib).

runtimevar/constantvar/example_test.go
```diff
@@ -88,3 +88,21 @@ func Example_error() {
 	// Output:
 	// runtimevar (code=Unknown): my error
 }
+
+func Example_openVariable() {
```
zombiezen: Similar to #1302, it would likely be best to move this out to a separate example file.

runtimevar/blobvar/blobvar.go
vangent: I agree that would be cool. It's tricky though, because while `blob+file` is certainly one option, any `blob` provider can be used, so `blob+s3` should work as well. `blobvar` doesn't know which set of `blob` providers are registered.

We could try to fix that by having `blob` expose the set of schemes that are registered...? But even then, it will be tricky to separate the URL parameters that are for the `OpenBucket` URL from the ones that are for the `blobvar` URL.

Otherwise, given that the scheme here is `blob`, I think that the only way to encode the URL to be passed to `blob` is in a URL parameter.

runtimevar/constantvar/constantvar.go
vangent: Done.

runtimevar/constantvar/example_test.go
vangent: I'm doing that in a separate PR, although I'm only doing it for the portable API example (`runtimevar`, `blob`, etc.), not for every driver; I don't think the latter is necessary do you?

runtimevar/constantvar/constantvar_test.go
vangent: Done.

runtimevar/blobvar/blobvar.go
vangent: Done: #1485.

runtimevar/blobvar/blobvar.go
vangent: Done.

runtimevar/blobvar/blobvar.go
vangent: I made it have both; you can either provide a `Bucket`, in which case it will always use it, or you can provide a `blob.URLMux` and a `bucket` query parameter. The default opener does the latter.

runtimevar/blobvar/blobvar.go
vangent: Done.

vangent:
PTAL

runtimevar/blobvar/blobvar.go
vangent: The more I think about this, the more I think I would like to get it to work. Escaping URLs in URL parameters is icky, and there are several other providers where URLs are needed (e.g., `etcdvar` for the etcd server), and it would be nice to be able to use the main URL itself more naturally (e.g., `etcd+http://my.etcd.server:8080?var=myvar` instead of the current `etcd://myvar?client=http://my.etcd.server:8080`). I filed #1488 to follow up.

runtimevar/constantvar/example_test.go
zombiezen: I think it's nice to show the full code necessary for the example to work, but optional.

runtimevar/blobvar/blobvar.go
zombiezen: SGTM.

PR #1407, opened 2019-02-26
secrets: add Azure KeyVault 

vsaroopchand:
This PR resolves #1315 ; adds to #76

PR #1451, opened 2019-03-05
internal/testing: add flags to help match sort results

shantuo:
In #1407 we had a bunch of Travis failures due to the result of `listdeps.sh`, all of which seem to be a reorder of the origin. I need to use some flags to match the result I ran with the sort in the stretch docker image to that from Travis. I've added `-d, --dictionary-order` and `-f, --ignore-case` so far to make it work.


internal/testing/listdeps.sh
```diff
@@ -33,4 +33,4 @@ function cleanup() {
 }
 trap cleanup EXIT
 
-sort "$tmpfile" | uniq
```
eliben: Can you add a comment here explaining the added flags and why we're adding them?
I suppose it's something to the extent of different systems having different default behaviors for "sort", so we want to hardcode a very specific behavior

shantuo: Done.

shantuo:
I changed to use the native byte values to sort as it seems to be more predictable. The sort distribution seems fine but the locale value makes the difference.

Travis uses the trusty image from GCE, which the locale is set to POSIX, it gives me different default result as a vanilla trusty image I pulled from Docker. So I use `LC_ALL=C` to make it consistent.

PTAL.

eliben:
LGTM, just update the commit message to reflect the new setting

Issue #1315, opened 2019-02-13
secrets: add Azure support


vangent:
* Investigate Azure equivalent for AWS KMS or GCP KMS.
* If any, add `secrets` support.

vsaroopchand:
I can help with this implementation. I already did a review but found some interesting dependencies.  BTW, the service to wrap is the Azure KeyVault, the SDK is [here](https://github.com/Azure/azure-sdk-for-go/tree/master/services/keyvault/v7.0/keyvault)

shantuo:
Thanks @vsaroopchand ! KeyVault looks like the perfect candidate for this.

PR #1452, opened 2019-03-05
secrets: add Azure KeyVault Example

vsaroopchand:
Example test for #1315 ; adds to #76

### DIFF
--- a/secrets/azurekeyvault/example_test.go
+++ b/secrets/azurekeyvault/example_test.go
@@ -1,0 +1,60 @@
+// Copyright 2019 The Go Cloud Development Kit Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package azurekeyvault_test
+
+import (
+	"context"
+	"log"
+
+	"github.com/Azure/azure-sdk-for-go/services/keyvault/v7.0/keyvault"
+	akv "gocloud.dev/secrets/azurekeyvault"
+)
+
+func Example() {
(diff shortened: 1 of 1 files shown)


secrets/azurekeyvault/example_test.go
```diff
@@ -0,0 +1,60 @@
...
+	)
+
+	// Now we can use keeper to encrypt or decrypt.
+	ctx := context.Background()
+	plaintext := []byte("Hello, Secrets!")
+	ciphertext, err := keeper.Encrypt(ctx, plaintext)
+	if err != nil {
+		log.Fatal(err)
+	}
+	decrypted, err := keeper.Decrypt(ctx, ciphertext)
```
shantuo: I think you need to assign the `err` to blank or handle it, either way is fine.

secrets/azurekeyvault/example_test.go
```diff
@@ -0,0 +1,60 @@
...
+	// - client: *keyvault.BaseClient instance, see https://godoc.org/github.com/Azure/azure-sdk-for-go/services/keyvault/v7.0/keyvault#BaseClient
+	// - keyVaultName: string representing the KeyVault name, see https://docs.microsoft.com/en-us/azure/key-vault/common-parameters-and-headers
+	// - keyName: string representing the keyName, see https://docs.microsoft.com/en-us/rest/api/keyvault/encrypt/encrypt#uri-parameters
+	// - keyVersion: string representing the keyVersion, see https://docs.microsoft.com/en-us/rest/api/keyvault/encrypt/encrypt#uri-parameters
+	// - opts: *KeeperOptions with the desired Algorithm to use for operations. See this link for more info: https://docs.microsoft.com/en-us/rest/api/keyvault/encrypt/encrypt#jsonwebkeyencryptionalgorithm
+	keeper := akv.NewKeeper(
+		client,
+		"replace with keyVaultName",
+		"replace with keyName",
+		"replace with keyVersion",
```
shantuo: For simplicity of the example, let's use an empty keyVersion here. You could add a comment like `// replace with keyVersion if you don't want to use the default one`.

secrets/azurekeyvault/example_test.go
vsaroopchand: Ok

secrets/azurekeyvault/example_test.go
vsaroopchand: OK

PR #1457, opened 2019-03-05
doc: point to new location of "escaping abstractions" documentation in sources

eliben:
For #1274


blob/blob.go
```diff
@@ -487,9 +487,7 @@ func newBucket(b driver.Bucket) *Bucket {
 // Provider-specific types that are intended to be mutable will be exposed
 // as a pointer to the underlying type.
 //
-// See
-// https://github.com/google/go-cloud/blob/master/internal/docs/design.md#as
-// for more background.
+// See https://godoc.org/gocloud.dev for more background.
```
eliben: Not adding an anchor because it's long and may change. I'm considering, in a followup CL, to move these "As" blocks to the main doc.go too because they're all repetetive, and then point to an #As anchor there. LMK if this makes more sense to do in one go now

vangent: I support moving some form of this large comment to the existing section in main.go and removing it from here, and also adding a short anchor so it can be referenced.

I think the repeated comment for each `As` function should say something like:

```
// As converts i to provider-specific types. See https://godoc.org/gocloud.dev#As for more
// background, the "As" examples in this package for examples, and the provider-specific
// package documentation for the specific types supported for that provider.
```

eliben: Done. PTAL

pubsub/pubsub.go
```diff
@@ -192,37 +192,17 @@ func (t *Topic) Shutdown(ctx context.Context) (err error) {
...
+// See https://godoc.org/gocloud.dev#As for background information, the "As"
+// examples in this package for examples, and the provider-specific package
+// documentation for the specific types supported for that provider.
 func (t *Topic) As(i interface{}) bool {
 	return t.driver.As(i)
 }
 
-// ErrorAs converts err to provider-specific types.
-// ErrorAs panics if target is nil or not a pointer.
+// ErrorAs converts i to provider-specific types.
```
vangent: I think the changes you made on these two lines are incorrect.

pubsub/driver/driver.go
```diff
@@ -139,15 +136,15 @@ type Subscription interface {
 	// err will always be a non-nil error returned from ReceiveBatch or SendAcks.
 	IsRetryable(err error) bool
 
-	// As allows providers to expose provider-specific types.
-	//
-	// See
-	// https://github.com/google/go-cloud/blob/master/internal/docs/design.md#as
-	// for more background.
+	// As converts i to provider-specific types.
+	// See https://godoc.org/gocloud.dev#As for background information, the "As"
```
vangent: In `driver` packages, the user is the implementer of the interface, not the user of it, so the example code isn't as useful. How about something like:

```
// As converts i to provider-specific types.
// See https://godoc.org/gocloud.dev#As for background information, and
// existing driver implementations for implementation examples.
// Please also include the types that the driver supports for As functions
// in the package-level docstring; again, see existing drivers for examples.
```

blob/driver/driver.go
```diff
@@ -208,32 +208,15 @@ type Bucket interface {
...
-	//   if !ok {
-	//     return false
-	//   }
-	//   *i = foo.MyType{}  // or, more likely, the existing value
-	//   return true
-	//
-	// See
-	// https://github.com/google/go-cloud/blob/master/internal/docs/design.md#as
-	// for more background.
+	// As converts i to provider-specific types.
```
vangent: See my comment below about the docstring for driver packages.

blob/blob.go
```diff
@@ -500,7 +480,7 @@ func (b *Bucket) As(i interface{}) bool {
 // ErrorAs converts i to provider-specific types.
 // ErrorAs panics if i is nil or not a pointer.
 // ErrorAs returns false if err == nil.
-// See Bucket.As for more details.
+// See https://godoc.org/gocloud.dev#As for background information.
```
vangent: You updated `ErrorAs` to reference the main godoc, but didn't update all of the other `As` functions in this package. I don't feel strongly about which one to do (although pointing to Bucket.As is shorter, it's an extra level of indirection), but we should be consistent. Search for `Bucket.As` in this file.

doc.go
```diff
@@ -43,7 +43,10 @@ It is not feasible or desirable for APIs like blob.Bucket to encompass the full
 functionality of every provider. Rather, we intend to provide a subset of the
 most commonly used functionality. There will be cases where a developer wants
 to access provider-specific functionality, such as unexposed APIs or data
-fields, errors or options.
+fields, errors or options. This can be accomplished using As functions.
+
+
```
vangent: Nit: is this an extra CR?

pubsub/driver/driver.go
```diff
@@ -86,14 +86,11 @@ type Topic interface {
...
-	//
-	// See
-	// https://github.com/google/go-cloud/blob/master/internal/docs/design.md#as
-	// for more background.
+	// See https://godoc.org/gocloud.dev#As for background information.
 	As(i interface{}) bool
 
 	// ErrorAs allows providers to expose provider-specific types for errors.
-	// See As.
+	// See https://godoc.org/gocloud.dev#As for background information.
```
vangent: Again, I think it might be shorter and more consistent to reference the primary `As` function in this package, or duplicate the entire comment there.

pubsub/pubsub.go
```diff
@@ -192,37 +192,17 @@ func (t *Topic) Shutdown(ctx context.Context) (err error) {
...
+// ErrorAs panics if i is nil or not a pointer.
 // ErrorAs returns false if err == nil.
-// See Topic.As for more details.
+// See https://godoc.org/gocloud.dev#As for background information.
```
vangent: Same comment as above.

runtimevar/driver/driver.go
```diff
@@ -41,25 +41,10 @@ type State interface {
...
-	//   if !ok {
-	//     return false
-	//   }
-	//   *i = foo.MyType{}  // or, more likely, the existing value
-	//   return true
-	//
-	// See
-	// https://github.com/google/go-cloud/blob/master/internal/docs/design.md#as
-	// for more background.
+	// As converts i to provider-specific types.
```
vangent: Same comment about Driver types.

runtimevar/runtimevar.go
```diff
@@ -213,7 +196,7 @@ func wrapError(w driver.Watcher, err error) error {
 // ErrorAs converts i to provider-specific types.
 // ErrorAs panics if i is nil or not a pointer.
 // ErrorAs returns false if err == nil.
-// See Snapshot.As for more details.
+// See https://godoc.org/gocloud.dev#As for background information.
```
vangent: Same comment as above.

secrets/driver/driver.go
```diff
@@ -39,9 +39,7 @@ type Keeper interface {
 	// ErrorAs allows providers to expose provider-specific types for returned
 	// errors.
 	//
-	// See
-	// https://github.com/google/go-cloud/blob/master/internal/docs/design.md#as
-	// for more background.
+	// See https://godoc.org/gocloud.dev#As for background information.
```
vangent: Same comment about driver packages.

secrets/secrets.go
```diff
@@ -131,6 +131,7 @@ func (k *Keeper) Decrypt(ctx context.Context, ciphertext []byte) (plaintext []by
 //
 // ErrorAs panics if i is nil or not a pointer.
 // ErrorAs returns false if err == nil.
+// See https://godoc.org/gocloud.dev#As for background information.
```
vangent: Since this is the only `As` function in this package, it should have the full comment about seeing examples etc. in addition to background information.

pubsub/pubsub.go
eliben: Oops, good spot.

blob/blob.go
eliben: Updated to consistently point to doc.go#As

doc.go
eliben: I'm following a convention we've established in other large comment blocks, see the one on top of pubsub/pubsub.go for example.

Strictly interpreting the docstring guidelines we probably don't need 2 leading newlines, but that should be a separate change. WDYT?

pubsub/driver/driver.go
eliben: OK, I'll just shorten it to point to doc.go, since the other tips are already given there. I feel the new proposal is too lengthy for this

blob/driver/driver.go
eliben: Done

pubsub/driver/driver.go
eliben: Addressed as elsewhere

pubsub/pubsub.go
eliben: Ugh... our naming of parameters in ErrorAs is inconsistent.

i here: https://github.com/google/go-cloud/blob/master/blob/blob.go#L500
target here: https://github.com/google/go-cloud/blob/master/pubsub/pubsub.go#L222

I'll fix the inconsistency in a separate PR

runtimevar/driver/driver.go
eliben: Done

runtimevar/runtimevar.go
eliben: It's called `i` here.

secrets/driver/driver.go
eliben: Addressed (?)

secrets/secrets.go
eliben: We don't have examples for secrets ErrorAs yet... otherwise I'll update it to be more consistent

eliben:
Thanks for the review!

The main guiding principle for the changes: don't do double references - the As part in doc.go should explain everything so we can point to it from many places. The portable types provide a bit more details because they're user-facing, driver interfaces are for implementors and hence the doc is more concise. These changes are iffy to make because they touch so many places, and there are already naming inconsistencies. Making the scheme as simple as possible will reduce errors going forward

PR #1458, opened 2019-03-05
docs: add new logo to landing page

eliben:
Fiddled much with different options - hard to align nicely on both mobile and web. Ended up just adding it to the main page.


eliben:
Another idea would be to add the white version of the logo in the sidebar instead of the "Welcome to..." sub-header, WDYT?

internal/website/themes/hyde/static/css/hyde.css
```diff
@@ -70,6 +70,11 @@ html {
   }
 }
 
+figure.aligncenter img {
```
zombiezen: Are there issues with aligning the figure element? I would expect the `figure` element to be block-level, whereas `img` is typically inline.

internal/website/themes/hyde/static/css/hyde.css
```diff
@@ -70,6 +70,11 @@ html {
...
+  margin-left: auto;
+  margin-right: auto;
```
zombiezen: As per [Google CSS style](https://google.github.io/styleguide/htmlcssguide.html#Shorthand_Properties), I recommend rewriting this as `margin: 1em auto`.

internal/website/themes/hyde/static/css/hyde.css
eliben: Sorry, I'm not sure I'm following. Can you please elaborate?

For reference, the HTML created by Hugo for:

```
{{< figure src="go-cdk-logo-gopherblue-small.png" alt="Go CDK logo"
class="aligncenter" >}}
```

Is:


```
<figure class="aligncenter">
    <img src="go-cdk-logo-gopherblue-small.png"
         alt="Go CDK logo"/> 
</figure>
```

zombiezen: Sure. I'm wondering why the rule selector is targetting the `img` element and not the `figure` element (and also suggesting that perhaps the rule selector should target the `figure` element).

eliben:
Merging after IRL discussion. Ross will further hack the CSS into submission

zombiezen:
Hacked in #1468

PR #1460, opened 2019-03-05
secrets/azurekeyvault: support URL opener

vangent:
Updates #1174.


secrets/azurekeyvault/akv.go
```diff
@@ -54,6 +64,81 @@ var (
...
+
+	if u.Host == "" {
+		return nil, fmt.Errorf("open keeper %q: URL is expected to have a non-empty Host (the key vault name)", u)
+	}
+	path := u.Path
+	if path[0] == '/' {
+		path = path[1:]
+	}
+	pathParts := strings.Split(path, "/")
+	if len(pathParts) != 2 || pathParts[0] == "" || pathParts[1] == "" {
```
shantuo: The key version can be omitted if user doesn't want to specify a version, so I think it is ok to only have the first part instead of something like "key//"?

vangent: Ah, didn't know that; done.

PR #1467, opened 2019-03-06
pubsub: add AckFunc method to driver Subscription

### LINKED ISSUE #1378: all: make a new subpackage for at-most-once pubsubs like NATS core, Redis, ZeroMQ

jba:
WIP: TESTS WILL FAIL

This allows providers to communicate whether they support Ack.  If a
provider doesn't, it needs to a provide an alternate implementation of
Ack.

Fixes #1378.

### DIFF
--- a/pubsub/acks_test.go
+++ b/pubsub/acks_test.go
@@ -54,6 +54,8 @@ func (s *ackingDriverSub) IsRetryable(error) bool { return false }
 
 func (s *ackingDriverSub) ErrorCode(error) gcerrors.ErrorCode { return gcerrors.Internal }
 
+func (s *ackingDriverSub) AckFunc() func() { return nil }
+
 func TestAckTriggersDriverSendAcksForOneMessage(t *testing.T) {
 	ctx := context.Background()
 	var mu sync.Mutex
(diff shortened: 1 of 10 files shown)


vangent:
Basic idea LGTM.

jba:
PTAL. Tests should pass.

PR #1472, opened 2019-03-06
blob/fileblob: improve path munging on Windows, including tests/examples

vangent:


vangent:
(These tests pass on Windows).

blob/fileblob/example_test.go
```diff
@@ -65,9 +66,16 @@ func Example_openBucket() {
...
+	// On Windows, convert "\" to "/" and add a leading "/":
+	dirpath := filepath.ToSlash(dir)
+	if os.PathSeparator != '/' && !strings.HasPrefix(dirpath, "/") {
+		dirpath = "/" + dirpath
+	}
+
 	// OpenBucket creates a *blob.Bucket from a URL.
 	ctx := context.Background()
-	b, err := blob.OpenBucket(ctx, path.Join("file://", dir))
+	b, err := blob.OpenBucket(ctx, "file://"+dirpath)
```
vangent: `path.Join` is the wrong thing here, as it collapses multiple slashes, resulting in URLs like `file:/tmp/dir`.

vangent:
@zombiezen moving review to @clausti since I've sent a lot your way recently. Feel free to take a look if you'd like.

blob/fileblob/example_test.go
```diff
@@ -65,9 +66,16 @@ func Example_openBucket() {
 	}
 	defer os.RemoveAll(dir)
 
+	// On Unix, append the dir to "file://".
+	// On Windows, convert "\" to "/" and add a leading "/":
+	dirpath := filepath.ToSlash(dir)
```
zombiezen: Optional: Seeing this in the example directly, it kinda feels like we should have a utility function that does this, although I realize the user would probably just call the `fileblob.OpenBucket` call directly. WDYT?

blob/fileblob/example_test.go
zombiezen: Ack.

blob/fileblob/example_test.go
vangent: I think the user is more likely to be constructing the URLs by hand than to be writing code to generate them. I agree with you that if they have the directory in code, they should just call `OpenBucket` directly, and not go through the URL path.

vangent:
gentle ping @clausti ?

PR #1475, opened 2019-03-06
pubsub: rename target argument of ErrorAs to i, for consistency with As

### LINKED ISSUE #1474: all: consistent naming of parameters in ErrorAs
Our `ErrorAs` methods refer to the second parameter either as `i` (blob, runtimevar, etc.) or `target` (pubsub).

These should be made consistent. Even though `target` is less common, it seems more descriptive to me so I would prefer to align all names to `target`.

Objections?

eliben:
Fixes #1474

### DIFF
--- a/pubsub/azurepubsub/azurepubsub.go
+++ b/pubsub/azurepubsub/azurepubsub.go
@@ -40,11 +40,11 @@ import (
 
 	"pack.ag/amqp"
 
-	"github.com/Azure/azure-amqp-common-go"
+	common "github.com/Azure/azure-amqp-common-go"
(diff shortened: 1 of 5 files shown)


pubsub/azurepubsub/azurepubsub.go
```diff
@@ -40,11 +40,11 @@ import (
 
 	"pack.ag/amqp"
 
-	"github.com/Azure/azure-amqp-common-go"
+	common "github.com/Azure/azure-amqp-common-go"
```
eliben: goimports added these when I saved... looking through the file they seem genuinely needed - unless I'm missing something

Issue #1477, opened 2019-03-07
docstore: verify that a provider's docstore codec matches the native one  


jba:
A docstore provider must supply a codec (encoder/decoder) that maps a specified set of Go types to the provider's native type. (The words "marshal" and "unmarshal" are also commonly used in place of "encode" and "decode".) Most interesting is the mapping of structs to the native type.

Example: The Go client for Amazon DynamoDB has a type called [`AttributeValue`](https://godoc.org/github.com/aws/aws-sdk-go/service/dynamodb#AttributeValue) that represents values stored in the database. The [`dynamodbattribute`](https://godoc.org/github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute) package encodes Go types, including structs, into these values, and performs the reverse decoding. 

We want to make sure that data encoded using the native codec (the `dynamoattribute` package in the example) can be decoded losslessly by the docstore codec for DynamoDB, at least for the set of types that docstore supports; and similarly, that any value encoded with the docstore codec will be properly decoded by the native codec.

We will add conformance tests to verify these properties.

@vangent's suggestion for the form of these tests is:
- The driver supplies a list of native values for testing, as well as `Encode` and `Decode` functions for the docstore codec.
- For each native value, the conformance test decodes it, then encodes the result, and verifies that the resulting native value is the same as the input.

I don't think this is a good enough test. It doesn't really address the thorniest part of the story, which is how structs are handled. The native values supplied by the driver for testing would have to come from natively encoding some structs. How would the test communicate to the driver which structs it should use? And how would the driver encode them, other than by using the native codec?

The obvious solution is that the conformance test passes the structs it wants to test to a function, which then calls the native codec to produce a native value. The test could then decode this native value back into a value of the same struct type using the docstore codec, then verify that the input and output structs matched. Dually, the test would want to encode a struct with the docstore codec and decode it with the native codec to see if they were the same.

This all suggests that the right design is for the driver to expose four methods: Native{Encode/Decode} and Docstore{Encode/Decode}.

jba:
This was addressed at the driver level by #1478. I don't think we need to leave the bug open for each driver.

PR #1478, opened 2019-03-07
docstore: a conformance test for native codecs

jba:
Preliminary version of a test that checks that the docstore encoding
of values interoperates with a provider's native encoding.

See #1477.


internal/docstore/drivertest/drivertest.go
```diff
@@ -39,15 +39,23 @@ type Harness interface {
 // It is called exactly once per test; Harness.Close() will be called when the test is complete.
 type HarnessMaker func(ctx context.Context, t *testing.T) (Harness, error)
 
+type CodecTester interface {
```
vangent: Please add docstrings.

Issue #1476, opened 2019-03-06
all: see if we can share more code in the URL implementations


vangent:
* Possibly in the driver-specific code.
* Possibly in the portable API code.

There's a lot of duplication, but it's a bit tricky because all of the types are different.

@jba wants to take a look.

vangent:
Done?

jba:
Yes.

PR #1479, opened 2019-03-07
pubsub: factor out common URL code

jba:
Fixes partially #1476.

### DIFF
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -565,85 +565,71 @@ type SubscriptionURLOpener interface {
 //
 // The zero value is a multiplexer with no registered schemes.
 type URLMux struct {
-	subscriptionSchemes map[string]SubscriptionURLOpener
-	topicSchemes        map[string]TopicURLOpener
+	subscriptionSchemes urlSchemeMap
+	topicSchemes        urlSchemeMap
+}
+
+// NewURLMux returns URLMux ready for use.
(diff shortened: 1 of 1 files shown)


pubsub/pubsub.go
```diff
@@ -565,85 +565,71 @@ type SubscriptionURLOpener interface {
 //
 // The zero value is a multiplexer with no registered schemes.
 type URLMux struct {
-	subscriptionSchemes map[string]SubscriptionURLOpener
-	topicSchemes        map[string]TopicURLOpener
+	subscriptionSchemes urlSchemeMap
```
vangent: The "zero value" comment above is no longer true IIUC. It's a nice property. Maybe the name could be passed to each of (`Register`, `FromURL`, `FromString`) instead, so that the zero value of urlSchemeMap works?

vangent: We should also add a test for the zero value.

jba: Passing the name to each method.

`TestURLMux` does create a zero value. Did you mean something else?

jba:
PTAL.

pubsub/pubsub.go
vangent: I'm not sure how that test passed with your previous commit. It looks good now though.

vangent: Ah, I see, it was just missing the name; it always handled `m` being `nil`.

PR #1483, opened 2019-03-07
blob/s3blob: tests and fixes for s3 when using legacy ListObjects

vangent:
This functionality was contributed by @EvyBongers. I'm adding a conformance test that runs using the legacy list mode, and fixing some issues I found when running the tests.


blob/s3blob/s3blob.go
```diff
@@ -377,43 +364,67 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
...
+	}
 
-		return req, resp
-	} else {
-		return b.client.ListObjectsV2Request(in)
+	var nextContinuationToken *string
+	if legacyResp.NextMarker != nil {
+		nextContinuationToken = legacyResp.NextMarker
+	} else if aws.BoolValue(legacyResp.IsTruncated) {
+		nextContinuationToken = aws.String(aws.StringValue(legacyResp.Contents[len(legacyResp.Contents)-1].Key) + " ")
```
shantuo: Why is there a `" "` added?

blob/s3blob/s3blob_test.go
```diff
@@ -47,29 +47,39 @@ const (
...
 	h.closer()
 }
 
 func TestConformance(t *testing.T) {
-	drivertest.RunConformanceTests(t, newHarness, []drivertest.AsTest{verifyContentLanguage{}})
+	drivertest.RunConformanceTests(t, newHarness, []drivertest.AsTest{verifyContentLanguage{usingLegacyList: false}})
+}
+
+func TestConformanceUsingLegacyList(t *testing.T) {
+	drivertest.RunConformanceTests(t, newHarnessUsingLegacyList, []drivertest.AsTest{verifyContentLanguage{usingLegacyList: true}})
```
shantuo: This will run all the conformance tests one more time. It seems a little bit redundant. I don't have any quick suggestions over the top of my head and a lot of the conformance tests do use List in this case.

vangent: Acknowledged. I don't think there's a lot of cost running them again. If we ended up with lots of combinations of parameters we may reconsider.

blob/s3blob/s3blob.go
vangent: It was in the PR before; I thought it was needed to ensure that the next set of results doesn't include this one. But, I tried without it and it doesn't seem to be necessary, so removed it.

vangent:
PTAL

PR #1490, opened 2019-03-08
runtimevar: implement health.Checker

### LINKED ISSUE #1443: runtimevar: implement the health.Checker interface for Variable
Returning healthy only when there's a "good" value for the variable.

vangent:
Fixes #1443.

### DIFF
--- a/runtimevar/runtimevar.go
+++ b/runtimevar/runtimevar.go
@@ -36,7 +36,7 @@
 // http://github.com/google/wire useful for managing your initialization code.
 //
 // Variable implements health.Checker; it reports as healthy when Latest will
-// return a value.
+// return a value without blocking.
 //
 // OpenCensus Integration
 //
--- a/runtimevar/runtimevar.go
+++ b/runtimevar/runtimevar.go
@@ -35,6 +35,8 @@
 // deploy it to multiple Cloud providers. You may find
 // http://github.com/google/wire useful for managing your initialization code.
 //
+// Variable implements health.Checker; it reports as healthy when Latest will
+// return a value.
 //
 // OpenCensus Integration
 //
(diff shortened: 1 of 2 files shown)


runtimevar/runtimevar.go
```diff
@@ -35,6 +35,8 @@
 // deploy it to multiple Cloud providers. You may find
 // http://github.com/google/wire useful for managing your initialization code.
 //
+// Variable implements health.Checker; it reports as healthy when Latest will
+// return a value.
```
zombiezen:
```suggestion
// return a value without blocking.
```

runtimevar/runtimevar_test.go
```diff
@@ -202,6 +202,10 @@ func TestVariable_Latest(t *testing.T) {
 	if ctx2.Err() == nil {
 		t.Error("Latest with non-nil ctx but no value yet should block")
 	}
+	// And we're not healthy.
+	if v.CheckHealth() == nil {
```
zombiezen: Add a health check right after `Variable` creation, too.

runtimevar/runtimevar.go
```diff
@@ -278,6 +280,23 @@ func (c *Variable) Latest(ctx context.Context) (Snapshot, error) {
...
+	case <-c.haveGood:
+		haveGood = true
+	default:
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if haveGood && c.lastErr != ErrClosed {
+		return nil
+	}
+	return c.lastErr
```
zombiezen: Is `c.lastErr` guaranteed to be non-nil? What if the RPC hasn't completed yet?

runtimevar/runtimevar_test.go
vangent: Done.

runtimevar/runtimevar.go
vangent: `c.lastErr` is initialized to `errors.New("no value yet")` in the constructor, so it's always non-nil unless there's a good value.

Issue #1453, opened 2019-03-05
all: allow URL openers to have an API-specific prefix to the scheme


vangent:
E.g., `blob+mem://` should work as well as `mem://`.

This will be useful for determining which API an arbitrary URL belongs to.

Implementation thoughts:

* The API-specific `Register` function (e.g. `blob.RegisterBucket`) could always register the requested scheme as well as `<api name>+<scheme>`.

* Alternatively, the API-specific `Open` function could check for the known `<api name>+` prefix and strip it.

@zombiezen 
@clausti

clausti:
Tentatively: at scheme registration seems like the the correct place for this solution.

vangent:
Note: the prefix needs to include not just the API ("pubsub"), but also the type ("topic" or "subscription"). It might be fine for some APIs to allow a default type.

PR #1494, opened 2019-03-08
all: refactor SchemeMap to internal/ and use it everywhere

### LINKED ISSUE #1487: all: finish URLMux refactor
@jba started a refactor in #1479, applying it to `pubsub`; move the common utility code to `internal/` somewhere and apply it to the other packages.

vangent:
Fixes #1487.


vangent:
Note: I added the "api" and "typ" args (instead of just "name") because of https://github.com/google/go-cloud/issues/1453#issuecomment-471021551.

internal/openurl/openurl.go
```diff
@@ -0,0 +1,64 @@
...
+
+// Package openurl provides helpers for URLMux and URLOpeners in portable APIs.
+package openurl // import "gocloud.dev/internal/openurl"
+
+import (
+	"fmt"
+	"net/url"
+)
+
+// SchemeMap maps URL schemes to values. The zero value is an empty map.
```
jba: "an empty map, ready for use."

internal/openurl/openurl.go
```diff
@@ -0,0 +1,64 @@
...
+type SchemeMap map[string]interface{}
+
+// Register registers scheme for value; subsequent calls to FromString or
+// FromURL with scheme will return value.
+// api is the portable API name (e.g., "blob") and typ is the portable type
+// (e.g., "Bucket").
+// Register panics if scheme has already been registered.
```
jba: It just occurred to me that you could store api and typ in the map. That way the zero value would still be good to go, and you wouldn't have to dup api and typ in the FromXXX calls.

Also, you might as well have a single arg that is of the form api.type, since you always use them together. (Or take the two args here, but fmt.Sprintf them together before you store them.)

internal/openurl/openurl_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"testing"
+
+	"gocloud.dev/internal/openurl"
+)
+
+func TestSchemeMap(t *testing.T) {
+	const foo, bar = "foo value", "bar value"
+
+	tests := []struct {
+		URL     string
```
jba: Don't export these fields. It doesn't really matter, but it's jarring to read, since we never do export them.

internal/openurl/openurl_test.go
```diff
@@ -0,0 +1,60 @@
...
+		}
+
+		got, gotURL, gotErr := m.FromString("api", "type", test.URL)
+		if (gotErr != nil) != test.WantErr {
+			t.Errorf("%s: got error %v, want error: %v", test.URL, gotErr, test.WantErr)
+		}
+		if gotErr != nil {
+			continue
+		}
+		if gotURL.String() != test.URL {
```
jba: `if got := gotURL.String(); got != test.URL ...`

So you don't have to repeat the expression in the next line.

internal/openurl/openurl.go
jba: Oh, saw your comment above. So yeah, I would take two args in Register, build and store the right string, and take nothing in the Froms.

internal/openurl/openurl.go
vangent: Done.

internal/openurl/openurl_test.go
vangent: Done.

internal/openurl/openurl_test.go
vangent: Done.

internal/openurl/openurl.go
vangent: Do you mean store it in the map at some hardcoded-key?

I have a follow-on PR where I'm going to make it so that there's only a single `URLOpener` interface per portable API ("pubsub"), instead of one per type ("pubsub/Topic", "pubsub/Subscription"). So you'd `Register` an `URLOpener` for "pubsub", not for a specific type, and the `URLOpener` could have multiple functions (`OpenTopic`, `OpenSubscription`). The functionality ends up being the same, but less boilerplate.

I think this means that `Register` could take the "api" and do what you suggest, while the `From` functions would still need to take the "typ".

Sound reasonable? If so, I'd prefer to add a TODO here and do it in the follow-on PR.

vangent:
PTAL

internal/openurl/openurl.go
jba: Sounds reasonable, and I'm fine with a TODO.

vangent: Done.

PR #1496, opened 2019-03-08
blob: add Bucket.Close method

### LINKED ISSUE #1485: runtimevar/blobvar: buckets that are opened by URLOpener should be closed when the Variable is closed

### LINKED ISSUE #1492: blob: Bucket does not have a Close method
All of our other APIs have `Close` methods with the exception of `Bucket`. I don't think that we should constrain implementations of `Bucket` to not be able to hold long-lived connections, so I think bucket driver types and the portable type should support `Close`, even if all current implementations no-op.

(This would be a breaking change, but one that is fairly easy to correct for driver types.)

/cc @vangent for triage and assigning.

zombiezen:
/cc @vangent if you're interested

Updated sample code and conformance tests to call new `Close` method.

Fixes #1492
Fixes #1485


zombiezen:
One thing this is likely missing: I didn't update the `blobvar.URLOpener` type introduced in #1449 to use `Close`, which is actually how I came about this issue.

zombiezen:
Done. This PR now fixes #1485 as well.

blob/blob.go
```diff
@@ -774,10 +810,27 @@ func (b *Bucket) SignedURL(ctx context.Context, key string, opts *SignedURLOptio
...
+// Close releases any resources used for the bucket.
+func (b *Bucket) Close() error {
+	b.mu.Lock()
+	prev := b.closed
+	b.closed = false
+	b.mu.Unlock()
+	if prev {
+		return errClosed
+	}
+	return b.b.Close()
```
shantuo: Also set b.closed to true?

runtimevar/blobvar/blobvar.go
```diff
@@ -245,6 +275,10 @@ func (w *watcher) WatchVariable(ctx context.Context, prev driver.State) (driver.
 
 // Close implements driver.Close.
 func (w *watcher) Close() error {
+	if w.opener != nil {
+		w.opener.decBucketRef(w.bucket)
+		w.opener = nil // Ensure that we don't call multiple times.
```
shantuo: Should it also close the bucket when refs is 0?

blob/blob.go
zombiezen: Whoops. Thanks.

runtimevar/blobvar/blobvar.go
zombiezen: Yes. Done.

Issue #1495, opened 2019-03-08
pubsub/awspubsub: test is slow


zombiezen:
The AWS Pub/Sub tests (and to a lesser extent, the `pubsub` package tests) take roughly ~30 seconds to run. Example build: https://travis-ci.com/google/go-cloud/jobs/183234212#L1155-L1156

I haven't dug too much into root cause, but I'm unsure what could be taking this long in a replay test. @jba, any insight here?

vangent:
I noticed this too, and spent a little bit of time looking into it. It's the new replayer (in `httpreplay.NewReplayer`); I think it spins up a local proxy and that takes some time.

jba:
When I run locally all the tests, including the conformance tests, finish in under 6s. A significant fraction of that is starting the replayer: it varies between 160 - 700ms, and we start one for each conformance test. 

I sent #1497 to see what the startup times are on Travis.

jba:
Even with `-race` and `-cover`, it still only takes a little over 7s locally for me.

zombiezen:
Weird. I definitely could reproduce locally...

jba:
I'm running `go test -v -race -cover gocloud.dev/pubsub/awspubsub -count 1 `. Still getting around 6-7s.

#1497 travis run shows that indeed, proxy startup is most of the time.

Options:
- Do nothing (30s is still only a small part of the 10m it takes for a run).
- Only start the proxy once for all conformance tests. This would imply a single replay file instead of one per test. 
- Revert back to the vcr replayer.

vangent:
This is fixed.

PR #1497, opened 2019-03-08
testing/setup: instrument replayer startup time

jba:


zombiezen:
(This PR is for #1495.)

internal/testing/setup/setup.go
```diff
@@ -127,7 +128,9 @@ func NewAWSSession2(ctx context.Context, t *testing.T, region string) (sess *ses
 			}
 		}
 	} else { // replay
+		start := time.Now()
 		rep, err := httpreplay.NewReplayer(path)
+		t.Logf("starting http replayer took %s", time.Since(start))
```
zombiezen: I don't think this will show up in Travis output ordinarily, since we don't run with `-v`. Perhaps use the standard logging package?

jba:
This PR did its job. It shows that httpreplay proxy startup time on travis is a significant fraction of total test time.

PR #1499, opened 2019-03-08
runtimevar: fix tests for blobvar and filevar on Windows

### LINKED ISSUE #1462: blob/fileblob: address TODO about windows
There's a TODO in the code:

```
// TODO: use filepath.FromSlash instead; and remove the pathSeparator arg
// from this function. Test Windows behavior by opening a bucket on Windows.
```

Address this.

It may also be in `filevar.go`.

vangent:
Updates #1306.
Fixes #1462.

### DIFF
--- a/runtimevar/blobvar/blobvar_test.go
+++ b/runtimevar/blobvar/blobvar_test.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -105,6 +106,12 @@ func TestOpenVariable(t *testing.T) {
 	}
 	defer os.RemoveAll(dir)
 
+	// Convert dir to a URL path, adding a leading "/" if needed on Windows.
+	dirpath := filepath.ToSlash(dir)
+	if os.PathSeparator != '/' && !strings.HasPrefix(dirpath, "/") {
+		dirpath = "/" + dirpath
+	}
+
 	tests := []struct {
 		URL          string
 		WantErr      bool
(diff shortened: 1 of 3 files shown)


runtimevar/filevar/filevar.go
```diff
@@ -103,18 +103,11 @@ func (o *URLOpener) OpenVariableURL(ctx context.Context, u *url.URL) (*runtimeva
 	for param := range q {
 		return nil, fmt.Errorf("open bucket %q: invalid query parameter %q", u, param)
 	}
-	return New(mungeURLPath(u.Path, os.PathSeparator), o.Decoder, &o.Options)
-}
-
-func mungeURLPath(path string, pathSeparator byte) string {
-	if pathSeparator != '/' {
+	path := u.Path
+	if os.PathSeparator != '/' {
```
shantuo: I think you can use `filepath.Clean` to simplify this.

runtimevar/blobvar/blobvar_test.go
```diff
@@ -105,6 +106,12 @@ func TestOpenVariable(t *testing.T) {
 	}
 	defer os.RemoveAll(dir)
 
+	// Convert dir to a URL path, adding a leading "/" if needed on Windows.
+	dirpath := filepath.ToSlash(dir)
+	if os.PathSeparator != '/' && !strings.HasPrefix(dirpath, "/") {
```
shantuo: Maybe add a comment saying that on Unix the path will be resolved to start with a '/'.

runtimevar/filevar/filevar.go
vangent: I don't think that will strip the "/" prefix. It also does a bunch of other things I'm not sure I want.

runtimevar/blobvar/blobvar_test.go
vangent: Done.

PR #1503, opened 2019-03-08
all: use a single interface for URLOpeners

vangent:
I don't see a real down side to this, as the mux has context in the specific `Open` function about what kind of resource is being opened. It reduces the boilerplate significantly in provider APIs that have multiple openable types.

Drivers that only support a subset of the types can return an error for the ones they don't.

Drivers can even support different schemes for different resources (if they want, I think this is unlikely) by registering multiple `URLOpeners`.

Updates #1174.

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -101,7 +101,7 @@ const (
 )
 
 func init() {
-	blob.DefaultURLMux().RegisterBucket(Scheme, new(lazyCredsOpener))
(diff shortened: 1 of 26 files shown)


vangent:
Never mind, there's not a huge win here and it becomes less obvious how to add new types to an API in a backwards-compatible way. (With the current approach, you can just add a new interface and drivers can start registering for it; with this single-interface approach, that doesn't work).

PR #1504, opened 2019-03-08
testing: don't install the CDK and don't run tests for documentation-only changes

eliben:
Keeping it very simple for now.

For posterity, I had to resort to full bash here because `go run ...` in our repo for a "script" that's completely unrelated to the module still installs everything.


internal/testing/runchecks.sh
```diff
@@ -26,10 +26,41 @@ if [[ $# -gt 0 ]]; then
...
+fi
+
+tmpfile=$(mktemp)
+function cleanup() {
+  rm -rf "$tmpfile"
+}
+trap cleanup EXIT
+
+mergebase="$(git merge-base -- "$TRAVIS_BRANCH" "$TRAVIS_PULL_REQUEST_SHA")"
+git diff --name-only "$mergebase" "$TRAVIS_PULL_REQUEST_SHA" -- > $tmpfile
```
zombiezen: You can do without the temporary file by assigning to another variable:

```suggestion
has_code_changes=0
git diff --name-only "$mergebase" "$TRAVIS_PULL_REQUEST_SHA" -- | grep -v ^internal/website > /dev/null || has_code_changes=1
```

internal/testing/runchecks.sh
```diff
@@ -26,10 +26,41 @@ if [[ $# -gt 0 ]]; then
...
+
+# Find lines that don't start with internal/website in the diff log; if no such
+# lines are found, it means that we don't have to run tests. grep returns 1 in
+# this case.
+if grep -v ^internal/website $tmpfile; then
```
zombiezen: This line will print all the lines to stdout, which is likely undesirable. See my above suggestion where I pipe stdout to `/dev/null`.

eliben: Added an `echo` to clarify, as discussed

vangent: Do you want to repeat this for only `.md` file PRs?

internal/testing/runchecks.sh
```diff
@@ -26,10 +26,41 @@ if [[ $# -gt 0 ]]; then
...
+  echo "Running tests"
+else
+  echo "Diff doesn't affect tests; not running them"
+  exit 0
+fi
+
+# Install wire; it's needed to build the project. Moved here from the "install"
+# step because we don't need to install wire if the diff doesn't require testing
+# (see condition above).
+go install -mod=readonly github.com/google/wire/cmd/wire
```
vangent: Move this down to right before `wire` is used.

internal/testing/runchecks.sh
eliben: Keeping it more explicit with a temp file, as discussed.
If you prefer to factor the pattern out into a variable, I can do that. Or we can do it in a followup PR that may actually want to add patterns, where it will be easier to see what the commonailty is.

internal/testing/runchecks.sh
eliben: Done

internal/testing/runchecks.sh
eliben: I wanted to land the basic infrastructure first, and later we can add more patterns / exclusions as needed

vangent: OK.

internal/testing/runchecks.sh
```diff
@@ -53,6 +80,10 @@ fi
   echo "FAIL: dependencies changed; compare listdeps.sh output with alldeps" && result=1
 }
 
+# Install wire; Moved here from the "install" step because we don't need to
```
vangent: Nit: I don't think you need the comment, it's of historical interest only.

eliben: Good point, removed.

eliben: Oops, I forgot to push the commit that removes this comment, sorry. Will do in a followup

Issue #1456, opened 2019-03-05
runtimevar: update Guestbook and examples to use Latest


vangent:
Depends on #996.

vangent:
Examples are updated; Guestbook is blocked on a new release tag.

vangent:
> Guestbook is blocked on a new release tag.

Actually that's not true; for some reason I thought guestbook had a `.mod` file, but it doesn't.

PR #1505, opened 2019-03-09
runtimevar: update examples to use Latest; add one for Watch

vangent:
Updates #1456.

### DIFF
--- a/runtimevar/example_test.go
+++ b/runtimevar/example_test.go
@@ -82,29 +82,27 @@ func ExampleVariable_Watch() {
 
 	// Call Watch in a loop from a background goroutine to see all changes,
 	// including errors.
-	ctx := context.Background()
-	ch := make(chan struct{})
+	//
+	// You can use this for logging, or to trigger behaviors when the
+	// config changes.
+	//
+	// Note that Latest always returns the latest "good" config, so seeing
+	// an error from Watch doesn't mean that Latest will return one.
 	go func() {
 		for {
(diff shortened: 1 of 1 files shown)


runtimevar/example_test.go
```diff
@@ -72,3 +72,39 @@ func Example_stringVariable() {
...
+			if snapshot, err := v.Watch(ctx); err == nil {
+				fmt.Printf("New config: %v", snapshot.Value.(string))
+				close(ch)
+			} else if err != runtimevar.ErrClosed {
+				fmt.Printf("Error loading config: %v", err)
+				// Even though there's been an error loading the config,
+				// v.Latest will continue to return the latest "good" value.
+			} else {
+				// v has been closed; exit.
+				return
```
eliben: And `ch` is not closed? So what happens in the containing function?

vangent: The flow will be that first `Watch` will return, closing `ch`; that allows the containing function to get past the `<-ch` and exit, closing v.

This may be too confusing for an example. Maybe I should remove the `// Output` lines and just have it compile?

eliben: Yeah, I'm still a bit confused. What if the first call to `Watch` returns `runtimevar.ErrClosed`? There is a nontrivial protocol with the wrapping function here, could it be simplified? Or at least documented with comments

vangent: OK. I removed the channel and update the test to not run, hopefully it make more sense now.

PR #1512, opened 2019-03-11
all/tests: clean up and improve coverage for URLMux tests

vangent:
The fake Opener is more obvious, and this adds coverage for `OpenXXXURL`.

Updates #1174.

### DIFF
--- a/runtimevar/runtimevar_test.go
+++ b/runtimevar/runtimevar_test.go
@@ -431,23 +431,17 @@ var (
 
 func TestURLMux(t *testing.T) {
 	ctx := context.Background()
-	var got *url.URL
 
 	mux := new(URLMux)
-	// Register scheme foo to always return nil. Sets got as a side effect
-	mux.RegisterVariable("foo", variableURLOpenFunc(func(_ context.Context, u *url.URL) (*Variable, error) {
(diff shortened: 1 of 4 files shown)


runtimevar/runtimevar_test.go
```diff
@@ -507,21 +506,35 @@ func TestURLMux(t *testing.T) {
...
+			parsed, err := url.Parse(tc.url)
 			if err != nil {
 				t.Fatal(err)
 			}
-			if diff := cmp.Diff(got, want); diff != "" {
-				t.Errorf("got\n%v\nwant\n%v\ndiff\n%s", got, want, diff)
+			_, gotErr = mux.OpenVariableURL(ctx, parsed)
+			if (gotErr != nil) != tc.wantErr {
+				t.Fatalf("got err %v, want error %v", gotErr, tc.wantErr)
+			}
```
shantuo: continue when gotErr != nil

vangent: Actually, there shouldn't be any error here. If there was an expected error, we returned above. I cleaned up this code (here and elsewhere) to fail if there's an error.

vangent:
PTAL

PR #1513, opened 2019-03-11
doc: add documentation for URLs

vangent:
My plan is to do something similar to what Eli did for "As" -- remove most of the documentation from the actual code and point to here instead, so we can document well once instead of copypasta everywhere.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -44,10 +44,10 @@ In addition to creating portable types via provider-specific constructors
 can also be created using a URL. The scheme of the URL specifies the provider,
 and each provider implementation has code to convert the URL into the data
 needed to call its constructor. For example, calling blob.OpenBucket with
-"s3blob://my-bucket" will return a blob.Bucket created using s3blob.OpenBucket.
+s3blob://my-bucket will return a blob.Bucket created using s3blob.OpenBucket.
 
 Each portable API package will document the types that it supports opening
-by URL; for example, the "blob" package supports Buckets, while the "pubsub"
+by URL; for example, the blob package supports Buckets, while the pubsub
 package supports Topics and Subscriptions. Each provider implementation will
 document what scheme(s) it registers for, and what format of URL it expects.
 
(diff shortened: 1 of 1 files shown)


doc.go
```diff
@@ -37,6 +37,50 @@ For sample applications and a tutorial, see the samples directory
...
+
+In addition to creating portable types via provider-specific constructors
+(e.g., creating a blob.Bucket using s3blob.OpenBucket), many portable types
+can also be created using a URL. The scheme of the URL specifies the provider,
+and each provider implementation has code to convert the URL into the data
+needed to call its constructor. For example, calling blob.OpenBucket with
+"s3blob://my-bucket" will return a blob.Bucket created using s3blob.OpenBucket.
+
+Each portable API package will document the types that it supports opening
+by URL; for example, the "blob" package supports Buckets, while the "pubsub"
```
eliben: Quotations are not consistent here. It's a shame godoc doesn't let us use backticks to format fixed-width-font stuff :-/

I think the convention in godoc.org is not putting quotes... Let's stick to that?

doc.go
```diff
@@ -37,6 +37,50 @@ For sample applications and a tutorial, see the samples directory
...
+will export URLOpeners that give you more fine grained control over the
+arguments needed by the constructor. In particular, portable types opened via
+URL will often use default credentials from the environment. For example, the
+AWS URL openers use the credentials saved by "aws login" (we don't want to
+include credentials in the URL itself, since they are likely to be sensitive).
+
+  - Instantiate the provider's URLOpener with the specific fields you need.
+    For example, "s3blob.URLOpener{ConfigProvider: myAWSProvider}" using a
+    ConfigProvider that holds explicit AWS credentials.
+    - Create your own instance of the URLMux, e.g., "mymux := new(blob.URLMux)".
```
eliben: misaligned bullet/hyphen?

doc.go
vangent: Done.

doc.go
vangent: Done.

PR #1516, opened 2019-03-11
firedocstore: actions

jba:
Implement all the docstore actions for Firestore.


go.sum
```diff
@@ -698,6 +698,7 @@ google.golang.org/genproto v0.0.0-20181202183823-bd91e49a0898/go.mod h1:7Ep/1NZk
 google.golang.org/genproto v0.0.0-20190110221437-6909d8a4a91b/go.mod h1:7Ep/1NZk928CDR8SjdVbjWNpdIf6nzjE3BTgJDr2Atg=
 google.golang.org/genproto v0.0.0-20190122204518-eef12c790cc0 h1:yynHoGjeRPvjTZVcYPhCR+RPUhYHqK6LPreYBBJOwew=
 google.golang.org/genproto v0.0.0-20190122204518-eef12c790cc0/go.mod h1:L3J43x8/uS+qIUoksaLKe6OS3nUKxOKuIFz1sl2/jx4=
+google.golang.org/genproto v0.0.0-20190306222511-6e86cb5d2f12 h1:W8SETuMdLwUlhb1S+hjWLWedszQrLssC1giquyHZXVQ=
```
vangent: Was this intended?

internal/docstore/docstore.go
```diff
@@ -143,6 +143,11 @@ func (l *ActionList) Get(doc Document, fps ...FieldPath) *ActionList {
 // No field path in mods can be a prefix of another. (It makes no sense
 // to, say, set foo but increment foo.bar.)
 //
+// It is undefined whether updating a sub-field of a non-map field will succeed.
```
vangent: It kind of sucks to leaves this undefined since that's not very portable. Can we make it always an error?

internal/docstore/firedocstore/fs.go
```diff
@@ -0,0 +1,516 @@
...
+	client    *vkit.Client
+	dbPath    string
+	collPath  string
+	nameField string
+}
+
+// OpenCollection creates a *docstore.Collection representing a Firestore collection.
+// firedocstore requires that a single string field, nameField, be designated the
+// primary key. Its values must be unique over all documents in the collection, and
+// the primary key must be provided to retrieve a document.
```
vangent: What is `collPath`?

internal/docstore/firedocstore/fs.go
```diff
@@ -0,0 +1,516 @@
...
+func OpenCollection(client *vkit.Client, projectID, collPath, nameField string) *docstore.Collection {
+	return docstore.NewCollection(newCollection(client, projectID, collPath, nameField))
+}
+
+func newCollection(client *vkit.Client, projectID, collPath, nameField string) *collection {
+	dbPath := fmt.Sprintf("projects/%s/databases/(default)", projectID)
```
vangent: Is that always "(default)"? Seems weird.

go.sum
jba: Reverted.

internal/docstore/docstore.go
jba: Agreed, but we don't have the document so we can't check.

internal/docstore/firedocstore/fs.go
jba: Clarified.

internal/docstore/firedocstore/fs.go
jba: Agreed. I asked the Firestore folks a few times if they wanted the client to support multiple databases per project. They always said no.

jba:
PTAL.

internal/docstore/docstore.go
vangent: The driver could do an extra `Get` to find out, right? I.e., drivers that natively return an error wouldn't have to do anything, but drivers that don't could do extra work to make it happen. This is a tradeoff between portability/consistency and performance.

Can you add a TODO and/or file a bug to consider this further?

jba: Filed #1530.

I'm uncomfortable with the performance cost of a hidden Get.

Here's an idea: `Update` could check against the document it is given. That would work if the document is the result of a `Get`, which I think it often would be: read-modify-write cycles are typical, especially since they provide an alternative to transactions via optimistic locking (that's what the revision field is about). If the document given to `Update` didn't have the field (a blind update), then we'd just make the RPC rather than doing a Get.

internal/docstore/firedocstore/codec.go
```diff
@@ -44,12 +44,12 @@ import (
 // The Firestore proto definition for Value is a oneof of various types,
```
shantuo: The comment should also be updated.

internal/docstore/driver/util.go
```diff
@@ -14,8 +14,31 @@
...
+	"github.com/google/uuid"
+)
 
 // UniqueString generates a string that is unique with high probability.
 // Driver implementations can use it to generate keys for Create actions.
 func UniqueString() string { return uuid.New().String() }
+
+// Call when running tests that will be replayed.
+// Each seed value will result in UniqueString producing the same sequence of values.
+func MakeUniqueStringDeterministicForTesting(seed int64) {
```
shantuo: These functions seem can be moved to drivertest package.

jba: Done.

internal/docstore/firedocstore/codec.go
jba: I'm sorry, I don't understand what you mean.

jba:
PTAL.

internal/docstore/firedocstore/codec.go
shantuo: The function's return type is changed from Value to Document, the comment should say Document instead of "Value"

jba: I think I understand. PTAL.

PR #1521, opened 2019-03-12
pubsub: ensure that the recorded message processing time is > 0

### LINKED ISSUE #1525: pubsub: maxMessages stays 1 if message processing time is too fast
We compute `maxMessages` based on the average processing time of messages:
https://github.com/google/go-cloud/blob/master/pubsub/pubsub.go#L375

The average processing time is updated here:
https://github.com/google/go-cloud/blob/master/pubsub/pubsub.go#L319

However, note that if the `time.Duration` for processing a message == 0, then the average processing time stays 0, and thus `maxMessages` stays at its initial default, 1.

On Windows in particular, the granularity of `time.Time` appears to be higher, so the `time.Duration` for processing time is 0.

vangent:
Fixes #1525.

### DIFF
--- a/internal/testing/replay/replay.go
+++ b/internal/testing/replay/replay.go
@@ -202,6 +202,8 @@ func NewGCPDialOptions(t *testing.T, mode recorder.Mode, filename string) (opts 
 	if err != nil {
 		t.Fatal(err)
 	}
+	// Uncomment for more verbose logging from the replayer.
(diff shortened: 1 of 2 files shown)


vangent:
This makes the Windows build pass; it was failing previously because the `time.Duration` was 0, so `pubsub` would ask for `maxMessages=1` every time instead of `maxMessages=1` and then `maxMessages=1000` like it does on Linux.

PR #1523, opened 2019-03-12
travis: only go-get wire on linux

vangent:
We only use Wire on linux, per `runchecks.sh`, so there's no point in "go get"ing it on other platforms.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,10 @@ install:
   # converting \n to \r\n.
   - "git config --global core.autocrlf input"
   - "git checkout -- ."
-  - "go install github.com/google/wire/cmd/wire"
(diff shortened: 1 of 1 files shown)


eliben:
See also #1504 which moves this line into runchecks. I suppose I could move the `go install ...wire` there into the already existing "if linux" condition?

vangent:
> I suppose I could move the 

Or move it to right before `wire` is actually used; non-linux OSs will already have aborted by then.

Shall I move it in this PR? Or, if #1504 is going to land soon I can drop this.

eliben:
I think I'll just move it in #1504. I'll add you as a reviewer there so that you get subscribed

PR #1527, opened 2019-03-12
runtimevar/etcdvar: fix URL docstring, improve tests

vangent:
Updates #1174.


vangent:
I'm going to fold this into a larger change.

PR #1535, opened 2019-03-12
testing: don't run tests when diff has only .md files

eliben:


eliben:
@zombiezen now that there's more then one pattern here, LMK if you want me to refactor it in some way

PR #1537, opened 2019-03-12
pubsub: fix inaccuracy in example comment

eliben:


eliben:
I think the tests were failing because I was pushing it from a branch that was behind master, so the `merge-base` failed to find a merging point.

Issue #1536, opened 2019-03-12
secrets: add more conformance/provider specific tests to improve coverage


shantuo:

shantuo:
gcpkms and awskms is ~80%, the uncovered lines are mostly handling errors from failed client creation/API calling, since we have only tens of LOC in those files.

I have another PR for Vault and I think it might be ok to close this bug after that. @vangent WDYT?

vangent:
SGTM

shantuo:
I was going to add a test for vault's setup code for OpenKeeperURL with no environment variable set, but the init runs inside a `sync.Once`, the test will affect other tests for Open. I'm going to close this issue now. If we think there are other tests missing, we can re-open it or create separate issues.

PR #1539, opened 2019-03-13
secrets: add tests to increase coverage

shantuo:
Update #1536

### DIFF
--- a/secrets/azurekeyvault/akv.go
+++ b/secrets/azurekeyvault/akv.go
@@ -138,7 +138,7 @@ func (o *URLOpener) OpenKeeperURL(ctx context.Context, u *url.URL) (*secrets.Kee
 	if keyName == "" {
 		return nil, fmt.Errorf("open keeper %q: URL is expected to have a Path with 1 or 2 non-empty elements (the key name and optionally, key version)", u)
 	}
-	return NewKeeper(o.Client, u.Host, keyName, keyVersion, &o.Options), nil
+	return NewKeeper(o.Client, u.Host, keyName, keyVersion, &o.Options)
 }
 
 type (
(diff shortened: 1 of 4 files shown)


secrets/azurekeyvault/akv.go
```diff
@@ -254,6 +254,9 @@ func (k *keeper) getKeyVaultURI() string {
 }
 
 func (k *keeper) validateOptions() error {
+	if k.options == nil {
```
vangent: Instead of checking this separately, how about we replace `nil` Options with `&Options{}` in the constructor? The check for empty algorithm will still fail, but that's a better error message.

vangent: How about we remove `validateOptions` and put the code in the constructor? Giving the error at construction time is better than giving it at `Encrypt/Decrypt` time, no?

secrets/azurekeyvault/akv_test.go
```diff
@@ -127,10 +125,12 @@ func initEnv() {
 	// For Client Certificate and Azure Managed Service Identity, see doc below for help
 	// https://github.com/Azure/azure-sdk-for-go
 
-	if os.Getenv("AZURE_TENANT_ID") == "" ||
-		os.Getenv("AZURE_CLIENT_ID") == "" ||
-		os.Getenv("AZURE_CLIENT_SECRET") == "" {
-		log.Fatal("Missing environment for recording tests, set AZURE_TENANT_ID, AZURE_CLIENT_ID and AZURE_CLIENT_SECRET")
+	if *setup.Record {
```
vangent: Is this change because `AZURE_ENVIRONMENT` is required for the code to work even during replay? Why? It's not used or document anywhere in our code.

secrets/azurekeyvault/akv_test.go
```diff
@@ -186,3 +188,29 @@ func TestOpenKeeper(t *testing.T) {
 		}
 	}
 }
+
+func TestNoConnectionError(t *testing.T) {
+	client := keyvault.NewWithoutDefaults()
+	k := &keeper{
```
vangent: Nit: can you use the public constructor here?

secrets/azurekeyvault/akv_test.go
```diff
@@ -186,3 +188,29 @@ func TestOpenKeeper(t *testing.T) {
...
+		options:      &KeeperOptions{Algorithm: algorithm},
+	}
+	if _, err := k.Encrypt(context.Background(), []byte("secrets")); err == nil {
+		t.Error("Encrypt: got nil, want no connection error")
+	}
+}
+
+func TestAlgorithmNotProvided(t *testing.T) {
+	ctx := context.Background()
+	k := new(keeper)
```
vangent: Same here.

secrets/azurekeyvault/akv_test.go
shantuo: We don't need to set it for either replay/record, it is set at L119 above using a default value "AZUREPUBLICCLOUD". We do mention in the prerequisites comment at the top of the file for how to setup for record.

Do you think it is necessary to make this also read from environment variable instead of hard-coded? For reference the available values are:
```go
var environments = map[string]Environment{
	"AZURECHINACLOUD":        ChinaCloud,
	"AZUREGERMANCLOUD":       GermanCloud,
	"AZUREPUBLICCLOUD":       PublicCloud,
	"AZUREUSGOVERNMENTCLOUD": USGovernmentCloud,
}
```

vangent: No, I think `AZUREPUBLICCLOUD` is fine.

I still don't understand why you changed the code from only calling `initEnv` in `--record` mode to always calling it.

shantuo: I think I was having bad local env value and azure was using that, it is not needed, changed it back.

PR #1540, opened 2019-03-13
all: add support for API-specific scheme prefixes for URLs

### LINKED ISSUE #1453: all: allow URL openers to have an API-specific prefix to the scheme

vangent:
Fixes #1453.

### DIFF
--- a/blob/blob_test.go
+++ b/blob/blob_test.go
@@ -300,7 +300,6 @@ func TestURLMux(t *testing.T) {
 		name    string
 		url     string
 		wantErr bool
-		want    string
 	}{
 		{
 			name:    "empty URL",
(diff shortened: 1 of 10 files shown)


blob/blob_test.go
```diff
@@ -300,7 +300,6 @@ func TestURLMux(t *testing.T) {
 		name    string
 		url     string
 		wantErr bool
```
vangent: Note: I added this earlier because I thought that the prefixes were going to be stripped before being passed to the `URLOpener`, but that wasn't easy to do so I ended not doing it; at this point `test.want` == `test.url`.

PR #1541, opened 2019-03-13
pubsub: add AckOverride and AutoDelete example

### LINKED ISSUE #1442: pubsub/azure: support option to not lock/ack messages
Apparently the Azure client supports a mode where receiving a message automatically deletes it from the server (as opposed to what we've currently implemented, which is Lock + delete after Ack).

Once #1368 is done, we should support the "auto delete" mode.

vsaroopchand:
resolves #1442 , adds to #76, discussed in #1368

This proposal adds NoAck/CustomAck with a redirect handler. Demonstrates how to configure the Azure Subscription for AutoDelete and NoAck/CustomAck.

All other PubSub drivers can Opt-In to this feature; NonAcking pubsub drivers can use a similar approach for NoAcks.

I look forward to your thoughts and ideas.


pubsub/azurepubsub/example_test.go
```diff
@@ -117,3 +121,80 @@ func ExampleOpenSubscription() {
...
+	sbTopic, err := azurepubsub.NewTopic(ns, topicName, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer sbTopic.Close(ctx)
+
+	// Construct Receiver to AutoDelete messages.
+	// See https://godoc.org/github.com/Azure/azure-service-bus-go#SubscriptionWithReceiveAndDelete.
+	var opts []servicebus.SubscriptionOption
+	opts = append(opts, servicebus.SubscriptionWithReceiveAndDelete())
```
vsaroopchand: Set AutoDelete Option on the Subscripton

pubsub/azurepubsub/example_test.go
```diff
@@ -117,3 +121,80 @@ func ExampleOpenSubscription() {
...
+	sbSub, err := azurepubsub.NewSubscription(sbTopic, subscriptionName, opts)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer sbSub.Close(ctx)
+
+	// Construct a custom Ack Handler to override the driver acking.
+	// This is useful for a number of reasons including noAcks or custom acks.
+	subOpts := &azurepubsub.SubscriptionOptions{
+		AckOverride: func(ctx context.Context, ids []driver.AckID) error {
```
vsaroopchand: Supply a custom Ack handler to intercept & override the driver ack

pubsub/driver/driver.go
```diff
@@ -153,3 +153,11 @@ type Subscription interface {
 	// one of the other methods in this interface.
 	ErrorCode(error) gcerrors.ErrorCode
 }
+
+// AckHandlerFunc exposes a common contract for Acknowledgement override/suppression.
+type AckHandlerFunc func(ctx context.Context, ids []AckID) error
```
vsaroopchand: Handler signature for overrides

pubsub/azurepubsub/example_test.go
vangent: Consider adding a comment explaining the options you might want here, including panic, log, or no-op.

vangent: @jba 	maybe we should have something about this in the portable type docs that drivers can point to....?

pubsub/azurepubsub/example_test.go
```diff
@@ -117,3 +121,80 @@ func ExampleOpenSubscription() {
...
+		log.Fatal(err)
+	}
+
+	// Receive a message from the *pubsub.Subscription backed by Service Bus.
+	msg, err := s.Receive(ctx)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Acknowledge the message, this operation issues a 'Complete' disposition on the Service Bus message.
```
vangent: This looks wrong....?

pubsub/driver/driver.go
vangent: Is this needed? @jba

pubsub/azurepubsub/example_test.go
jba: https://github.com/google/go-cloud/pull/1545

vsaroopchand: Ok, I'll use the noted comment as a starter.

pubsub/azurepubsub/azurepubsub.go
```diff
@@ -165,6 +166,7 @@ func OpenSubscription(ctx context.Context, parentNamespace *servicebus.Namespace
 // openSubscription returns a driver.Subscription.
 func openSubscription(ctx context.Context, sbNs *servicebus.Namespace, sbTop *servicebus.Topic, sbSub *servicebus.Subscription, opts *SubscriptionOptions) driver.Subscription {
 	topicName := ""
+
```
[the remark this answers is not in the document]
vsaroopchand: Its simply there to guard against a nil dereference. We don't return errors on any of the other openSubscriptions. If it is indeed empty then a NotFound error will be returned in SendBatch, which is consistent with the other providers and unit test.

pubsub/azurepubsub/example_test.go
```diff
@@ -117,3 +121,80 @@ func ExampleOpenSubscription() {
...
+	defer sbTopic.Close(ctx)
+
+	// Construct Receiver to AutoDelete messages.
+	// See https://godoc.org/github.com/Azure/azure-service-bus-go#SubscriptionWithReceiveAndDelete.
+	var opts []servicebus.SubscriptionOption
+	opts = append(opts, servicebus.SubscriptionWithReceiveAndDelete())
+
+	// Construct a Service Bus Subscription which is a child to a Service Bus Topic.
+	// See https://godoc.org/github.com/Azure/azure-service-bus-go#Topic.NewSubscription.
+	sbSub, err := azurepubsub.NewSubscription(sbTopic, subscriptionName, opts)
```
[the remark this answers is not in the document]
vsaroopchand: This is AMQP. Consider this call the dialer to initialize the subscription properties such as PeakMode, URI, Fetch Preferences, etc. The connection is established if/when Receive is called on the subscription. There is no way to manage the underlining AMQP connection.

pubsub/azurepubsub/azurepubsub.go
```diff
@@ -174,12 +176,19 @@ func openSubscription(ctx context.Context, sbNs *servicebus.Namespace, sbTop *se
 		defaultTimeout = opts.ListenerTimeout
 	}
 
+	// Determine if caller will handle acks.
+	var ackOverride func(ctx context.Context, ids []driver.AckID) error
```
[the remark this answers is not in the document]
vsaroopchand: So, what should be the signature of func? Shouldn't it be the same as Subscription.SendAcks, otherwise whats the point?

pubsub/azurepubsub/example_test.go
```diff
@@ -117,3 +121,80 @@ func ExampleOpenSubscription() {
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer sbSub.Close(ctx)
+
+	// Construct a custom Ack Handler to override the driver acking.
+	// This is useful for a number of reasons including noAcks or custom acks.
```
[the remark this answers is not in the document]
vsaroopchand: Isn't that overriding? I use the term override because it clearly conveys the result behavior - there is no second guessing as to its intention. Also, I recently added bulk disposition to the Service Bus SDK, app devs may want to use this to throttle acking or custom log & ack - this approach solves both of these scenarios

pubsub/azurepubsub/example_test.go
vsaroopchand: Revised to reflect this example

pubsub/azurepubsub/azurepubsub.go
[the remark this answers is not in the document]
jba: No, it should be `func()`. A function of no arguments with no return values.

We don't want to encourage fancy behavior. This is just a signal to tell the operator or debugging engineer that ack is being called when it probably shouldn't be. (Or you can pass a no-op if you don't care about that.)

pubsub/azurepubsub/example_test.go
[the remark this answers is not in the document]
jba: It's not overriding:
- If you provide an ack func, that means you're in receive-and-delete mode, so there is no system ack function (question: what happens when you call Complete in this case)?
- The only way to get the "real" ack function is to _not_ provide your own. 

We just want to solve the problem of what to do with an at-most-once system. We don't want to provide general hooks for customizing behavior at this time. We'd rather make the client smart enough so that that isn't necessary. We want to minimize the number of options and hooks.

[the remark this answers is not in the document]
vsaroopchand: Got it - calling Complete (or any DispositionAction) in at-most-once mode is disregarded by Service Bus. Once its compete (read once), its complete - you cant revoke the state.
 
Because of this, there is no harm calling driver.Ack, just wasted round trip. I understand your approach and will refactor to solve the immediate issue.

vsaroopchand:
PTAL

pubsub/azurepubsub/azurepubsub.go
```diff
@@ -289,7 +289,8 @@ type subscription struct {
 
 // SubscriptionOptions will contain configuration for subscriptions.
 type SubscriptionOptions struct {
-	ListenerTimeout time.Duration
+	ListenerTimeout            time.Duration
+	AckFuncForReceiveAndDelete func() // AckFuncForReceiveAndDelete is used for at-most-once mode (SB Subscription with SubscriptionWithReceiveAndDelete option)
```
jba: The comment is long enough that it should go above the field instead of next to it.

It should make it clear that if you set the field, you will get ReceiveAndDelete mode. Something like:

// If nil, the subscription will be in Peek-Lock mode. The Ack method must be called on each message
// to complete it.
// If non-nil, the subscription will be in Receive-and-Delete mode, and this function will be called 
// whenever Ack is called on a message. 
// See the "At-most-once vs. At-least-once Delivery" section in the pubsub package documentation.

pubsub/azurepubsub/example_test.go
jba: I think I see what's going on. It looks like you have to create the SB subscription before opening the azurepubsub subscription, which means that we can't use AckFuncForReceiveAndDelete to choose the subscription mode.

In OpenSubscription, can you check that the subscription is in Receive-and-Delete mode if and only if the ack function is non-nil, and return an error otherwise?

pubsub/azurepubsub/example_test.go
```diff
@@ -131,3 +132,90 @@ func Example_openfromURL() {
...
+	}
+	defer sbSub.Close(ctx)
+
+	// This package accommodates both kinds of systems. If your application uses
+	// at-least-once providers, it should always call Message.Ack. If your application
+	// only uses at-most-once providers (ReceiveAndDeleteMode), it may call Message.Ack, but does not need to.
+	//
+	// For more information on Service Bus ReceiveMode, see https://godoc.org/github.com/Azure/azure-service-bus-go#SubscriptionWithReceiveAndDelete.
+	//
+	// Use the options.AckOverride to supply your own logic for acking (optional regardless of modes). The custom AckOverride function can perform logging, no-ops,
```
jba: I think most of this is obsolete now.

pubsub/azurepubsub/example_test.go
```diff
@@ -131,3 +132,90 @@ func Example_openfromURL() {
...
+	// or your own bulk disposition operation. See https://godoc.org/github.com/Azure/azure-service-bus-go#Subscription.SendBatchDisposition.
+	//
+	// Important:
+	// 1. AckOverride is optional for Service Bus Subscription in ReceiveAndDeleteMode (SubscriptionWithReceiveAndDelete option enabled).
+	//    Calling the default driver Ack (opts.AckOverride is nil) in at-most-once mode will be disregarded by Azure Service Bus.
+	// 2. AckOverride can be used for custom batch dispositions (SubscriptionWithReceiveAndDelete option not set) but not recommended.
+	noop := func() {
+	}
+	subOpts := &azurepubsub.SubscriptionOptions{
+		AckFuncForReceiveAndDelete: noop,
```
jba: just inline the function body, on one line:
```
AckFuncForReceiveAndDelete: func(){},
```

pubsub/azurepubsub/example_test.go
vsaroopchand: RE: In OpenSubscription, can you check that the subscription is in Receive-and-Delete mode if and only if the ack function is non-nil, and return an error otherwise?

No, there is no way to check the subscription mode. As noted, you can call Ack if Reveive&Delete mode, the request will be ignored or you can save a round trip by supplying a noAck fcn.

pubsub/azurepubsub/azurepubsub.go
vsaroopchand: Ok, I made a small change - in both scenario, the setting MUST match the mode (Nil for Peek-Lock, Not Nil for Receive-and-Delete)

pubsub/azurepubsub/example_test.go
vsaroopchand: Ok

pubsub/azurepubsub/example_test.go
vsaroopchand: Comments updated.

pubsub/driver/driver.go
vsaroopchand: OK

pubsub/azurepubsub/azurepubsub.go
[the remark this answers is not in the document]
vsaroopchand: OK

vsaroopchand:
I just want to point out a possible design flaw with this NoAck func; it does not short-circuit the driver ACK. In the case of Auto-Delete, how do I prevent driver ACK from being called? If clients panic in noack func, then it kills the process. 

The point of NoAck should be to short-circuit the acking - please correct me if I am wrong.

vangent:
> I just want to point out a possible design flaw with this NoAck func; it does not short-circuit the driver ACK. In the case of Auto-Delete, how do I prevent driver ACK from being called? If clients panic in noack func, then it kills the process.
> 
> The point of NoAck should be to short-circuit the acking - please correct me if I am wrong.

It is not a `NoAck` function, it a an `Ack` function that is called by the portable type (`pubsub.go`) instead of the `driver.Ack` function. So, in the case of Auto-Delete, by providing this function it does prevent `driver.Ack` from being called. Depending on what the user provides in the `ackfunc func()`, they can force the application to panic (e.g., if they don't expect their application to every call `pubsub.Subscription.Ack`), or to no-op, or log a warning. There's some discussion of this here:
https://github.com/google/go-cloud/blob/master/pubsub/pubsub.go#L40

and another example of it being provided here:
https://github.com/google/go-cloud/blob/master/pubsub/natspubsub/nats.go#L173

vsaroopchand:
So, how does the pubsub.go knows a custom ack was supplied? Nats ackFunc is on the subscription, my ackfunc is on the subscription.option. Shouldn't there be a standard pattern for all drivers?

vangent:
> So, how does the pubsub.go knows a custom ack was supplied? 

There is a `driver.Subscription` function `AckFunc` that you should return it from. If you return `nil` from there (as the current code does), `pubsub.go` will batch acks and call `driver.SendAcks` asynchronously. If you return non-nil, `pubsub.go` will call the function instead (and never call `driver.SendAcks`; you can check this if you want, e.g. `nats.Subscription.SendAcks` panics).

> Nats ackFunc is on the subscription, my ackfunc is on the subscription.option. 
> Shouldn't there be a standard pattern for all drivers?

Drivers for providers that are always "at-most-once" should have it as a required argument. Service Bus is different because it supports both modes, and we're using the `Option.AckFunc` (or whatever it's called) to toggle between them (so, there's a reasonable default value).

pubsub/azurepubsub/azurepubsub.go
```diff
@@ -289,7 +289,14 @@ type subscription struct {
 
 // SubscriptionOptions will contain configuration for subscriptions.
 type SubscriptionOptions struct {
-	ListenerTimeout time.Duration
+	ListenerTimeout            time.Duration
```
vangent: Please add a docstring for `ListenerTimeout`, including what the default is. IIUC, this is how long it will wait to try to get a full batch of messages. 1 second as the default seems a bit long to me; e.g., if `maxMessages` is 10, but only a single message arrives, we'll wait for a full second before returning it?

BTW, other drivers have something similar but haven't exposed it publicly; e.g. `rabbitpubsub` tries to get to `maxMessages` for 50ms:
https://github.com/google/go-cloud/blob/master/pubsub/rabbitpubsub/rabbit.go#L566

So you could also consider removing this option and making it a constant. We can always expose it later if needed.

pubsub/azurepubsub/azurepubsub.go
```diff
@@ -435,6 +450,11 @@ func (s *subscription) SendAcks(ctx context.Context, ids []driver.AckID) error {
 		return nil
 	}
 
+	// Invoke AckFunc is provided (at-most-once mode).
```
vangent: No, you don't have to do this, `pubsub.go` will. You can write 

```
if s.opts != nil && s.opts.AckFuncForReceiveAndDelete != nil {
  panic("unreachable")
}
```

if you want to.

pubsub/azurepubsub/azurepubsub.go
```diff
@@ -435,6 +450,11 @@ func (s *subscription) SendAcks(ctx context.Context, ids []driver.AckID) error {
...
+	if s.opts != nil && s.opts.AckFuncForReceiveAndDelete != nil {
+		s.opts.AckFuncForReceiveAndDelete()
+	}
+
```
vangent: On line 520, you should return the ackFunc.

pubsub/azurepubsub/azurepubsub.go
```diff
@@ -174,12 +176,19 @@ func openSubscription(ctx context.Context, sbNs *servicebus.Namespace, sbTop *se
 		defaultTimeout = opts.ListenerTimeout
 	}
 
+	// Determine if caller will handle acks.
+	var ackOverride func(ctx context.Context, ids []driver.AckID) error
+	if opts != nil && opts.AckOverride != nil {
```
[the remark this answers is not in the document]
vangent: Ping on this one.

pubsub/azurepubsub/azurepubsub.go
jba: Yes, let's hardcode this for now.

vsaroopchand: Option removed, btw the point of having it public is to allow users to configure if the default was non-optimal. An acceptable timeout **will** differ from environment to environment due to many factors. In any event, the setting is removed.

pubsub/azurepubsub/azurepubsub.go
vsaroopchand: done

pubsub/azurepubsub/azurepubsub.go
vsaroopchand: done

PR #1544, opened 2019-03-13
pubsub: rework URL comments for consistency

vangent:
Updates #1174.


pubsub/azurepubsub/azurepubsub.go
```diff
@@ -18,10 +18,13 @@
 //
 // URLs
 //
-// For pubsub.OpenTopic/Subscription URLs, azurepubsub registers for the scheme
-// "azuresb". The Service Bus Connection String defaults to the environment
-// variable "SERVICEBUS_CONNECTION_STRING". For details on the format of the
-// URL, see URLOpener.
+// For pubsub.OpenTopic and pubsub.OpenSubscription, azurepubsub registers
+// for the scheme "azuresb".
+// The default URL opener will use a Service Bus Connection String based on
```
shantuo: Probably in another PR, I think the package should be renamed to `servicebus` to be consistent.

pubsub/azurepubsub/azurepubsub.go
```diff
@@ -117,8 +118,10 @@ const Scheme = "azuresb"
 //   - The URL's host+path is used as the topic name.
```
shantuo: The URL above should be `azuresb://mytopic?subscription=`

pubsub/azurepubsub/azurepubsub.go
vangent: You are probably right. I'm not sure it's worth it at this point though. `awspubsub` should probably be something more like `snssqs` too.

On the other hand, I think it's nice to have `azure` in there to make it more recognizable.

Let's discuss at standup tomorrow.

pubsub/azurepubsub/azurepubsub.go
vangent: Ooh, good catch. Done.

vangent:
PTAL

PR #1545, opened 2019-03-13
pubsub: document how we handle at-most-once providers.

jba:


pubsub/pubsub.go
```diff
@@ -37,6 +37,30 @@
 // See https://godoc.org/gocloud.dev#URLs for more information.
 //
 //
+// At-most-once vs. At-least-once Delivery
+//
+// Some PubSub systems guarantee that messages received by subscribers but
+// not acknowledged are delivered again. These at-least-once systems require
+// that subscribers call a function to indicate that they have fully processed a
```
vangent: "a function" -> "an Ack function" ?

pubsub/pubsub.go
```diff
@@ -37,6 +37,30 @@
...
+// message.
+//
+// In other PubSub systems, a message will be delivered only once, if it is delivered
+// at all. These at-most-once systems do not need an Ack method.
+//
+// To accommodate both kinds of systems, this package has a Message.Ack function
```
vangent: How about:

```
// This package accommodates both kinds of systems. If your application uses
// at-least-once providers, it should always call Message.Ack. If your application
// only uses at-most-once providers, it may call Message.Ack, but does not need to.
// The constructor for at-most-once-providers will require you to supply a function
// to be called whenever the application calls Message.Ack; common implementations
// are: [... your text here ...].
//
// Providers that support both at-most-once and at-least-once semantics will include
// an optional ack function in their Options struct.
```

Feel free to take this, or merge with yours.

PR #1546, opened 2019-03-13
blob: rework URL comments for consistency

vangent:
Similar to the other PRs; this is the last one. This one has no semantic changes, just comments.

Updates #1174.


blob/gcsblob/gcsblob.go
```diff
@@ -133,7 +137,7 @@ func (o *URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket
 func (o *URLOpener) forParams(ctx context.Context, q url.Values) (*Options, error) {
 	for k := range q {
 		if k != "access_id" && k != "private_key_path" {
-			return nil, fmt.Errorf("unknown GCS query parameter %s", k)
+			return nil, fmt.Errorf("unknown query parameter %q", k)
```
shantuo: nit: s/unknown/invalid

PR #1547, opened 2019-03-13
firedocstore: implement the codec test

jba:
For Firestore, the "native" codec is implemented by the cloud.google.com/go/firestore
client. Use gRPC interceptors to obtain the protos from this client.

Add an implementation of drivertest.CodecTester, and pass it to the conformance tests.

Remove uint and complex from the test. Firestore can't handle either type.

.

.

### DIFF
--- a/internal/docstore/drivertest/drivertest.go
+++ b/internal/docstore/drivertest/drivertest.go
@@ -302,9 +302,7 @@ func testCodec(t *testing.T, ct CodecTester) {
 	type S struct {
 		N  *int
 		I  int
-		U  uint
 		F  float64
-		C  complex64
 		St string
 		B  bool
 		By []byte
@@ -316,9 +314,7 @@ func testCodec(t *testing.T, ct CodecTester) {
 	in := S{
 		N:  nil,
 		I:  1,
-		U:  2,
 		F:  2.5,
-		C:  complex(9, 10),
 		St: "foo",
 		B:  true,
 		L:  []int{3, 4, 5},
--- a/internal/docstore/firedocstore/native_codec_test.go
+++ b/internal/docstore/firedocstore/native_codec_test.go
@@ -1,0 +1,153 @@
+// Copyright 2019 The Go Cloud Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package firedocstore
+
+import (
+	"context"
+	"io"
+	"net"
+	"testing"
+
+	"cloud.google.com/go/firestore"
+	ts "github.com/golang/protobuf/ptypes/timestamp"
+	"github.com/google/go-cmp/cmp"
+	"google.golang.org/api/option"
+	pb "google.golang.org/genproto/googleapis/firestore/v1"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// A nativeCodec encodes and decodes structs using the cloud.google.com/go/firestore
+// client. Since that package doesn't export its codec, we have to go behind the
+// scenes and intercept traffic at the gRPC level. We use interceptors to do that. (A
+// mock server would have worked too.)
+
+type nativeCodec struct {
+	client *firestore.Client
+	doc    *pb.Document
+}
+
+func newNativeCodec() (*nativeCodec, error) {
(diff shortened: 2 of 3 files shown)


internal/docstore/drivertest/drivertest.go
```diff
@@ -302,9 +302,7 @@ func testCodec(t *testing.T, ct CodecTester) {
 	type S struct {
 		N  *int
 		I  int
-		U  uint
```
vangent: What is the implication of removing these from the test? I thought the idea was to have all types supported by `docstore` here, and do things in the drivers to make them support them (e.g., list of size 2 for complex).

If you're removing them because firestore doesn't have native support, then we're going to end up only testing the intersection of all types supported by all drivers; we need a better way.

internal/docstore/firedocstore/native_codec_test.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func (cs *clientStream) RecvMsg(m interface{}) error {
+	if cs.doc != nil {
+		cs.doc.CreateTime = &ts.Timestamp{}
+		cs.doc.UpdateTime = &ts.Timestamp{}
+		m.(*pb.BatchGetDocumentsResponse).Result = &pb.BatchGetDocumentsResponse_Found{cs.doc}
+		cs.doc = nil
+		return nil
+	} else {
```
shantuo: nit: this else can be removed.

internal/docstore/firedocstore/native_codec_test.go
```diff
@@ -0,0 +1,153 @@
...
+	pb "google.golang.org/genproto/googleapis/firestore/v1"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// A nativeCodec encodes and decodes structs using the cloud.google.com/go/firestore
+// client. Since that package doesn't export its codec, we have to go behind the
+// scenes and intercept traffic at the gRPC level. We use interceptors to do that. (A
+// mock server would have worked too.)
+
```
shantuo: Remove this blank line.

internal/docstore/firedocstore/native_codec_test.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	return nc, nil
+}
+
+// Intercept all unary (non-streaming) RPCs. The only one we should ever get is a Commit, for
+// the Create call in Encode.
+// If this completes successfully, the encoded *pb.Document will be in c.doc.
+func (c *nativeCodec) interceptUnary(_ context.Context, method string, req, res interface{}, _ *grpc.ClientConn, _ grpc.UnaryInvoker, _ ...grpc.CallOption) error {
+	c.doc = req.(*pb.CommitRequest).Writes[0].GetUpdate()
+	cres := res.(*pb.CommitResponse)
```
shantuo: Where is `cres` used? Why do we need to give it a WriteResult?

internal/docstore/firedocstore/native_codec_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+// clientStream is a fake client stream. It returns a single document, then terminates.
+type clientStream struct {
+	ctx context.Context
+	doc *pb.Document
+}
+
+func (cs *clientStream) RecvMsg(m interface{}) error {
+	if cs.doc != nil {
+		cs.doc.CreateTime = &ts.Timestamp{}
```
shantuo: Is it ok to leave all other fields as is? I see there are a few `XXX_` named fields, will those affect the cmp result?

internal/docstore/drivertest/drivertest.go
shantuo: I guess one way to do it is to implement a driver codec that converts these types to structs. But I think the challenge might be that the native firestore client won't recognize them and we do want both the native and the docstore codec to work?

jba: As per our discussion, I will leave these out in this PR. In a followup PR, I'll (a) have a Docstore round-trip test with a full set of types, and (b) add an "I don't support these types" method to `CodecTester` so providers can opt out of certain types for the native-docstore tests.

internal/docstore/firedocstore/native_codec_test.go
jba: Done.

internal/docstore/firedocstore/native_codec_test.go
jba: Got rid of `cres`. We need a WriteResult because the firestore client has a check for it.

internal/docstore/firedocstore/native_codec_test.go
jba: 1. It's OK to leave the other fields unset; this is the minimum required for the firestore client to complete without an error.

2. I call `cmp.Equal` on a struct I define in the test, not protos, so it should be OK.

jba:
PTAL.

internal/docstore/drivertest/drivertest.go
vangent: SGTM.

internal/docstore/firedocstore/native_codec_test.go
jba: Done.

Issue #1303, opened 2019-02-12
blob: include URLOpener in provider sets


zombiezen:
After #1282, the `blob` driver packages will include new `*URLOpener` types. It would be good to insert these into the default provider sets for the various packages so that there's not too much hassle with Wire to construct a `*URLOpener`.

PR #1550, opened 2019-03-14
aws/awscloud: add AWS portable types to the Services ProviderSet

vangent:
* Do `sqs.New` and `sns.New` belong in there?
* I'm thinking of adding a test in `awscloud` that instantiates each supported portable type (hopefully in a 1-2 arg provider function). Does this sound like a good idea?
* Tuo, is this going to work OK with your Wire proposal for structs?

Updates #1303.


aws/awscloud/awscloud.go
```diff
@@ -37,5 +43,27 @@ var AWS = wire.NewSet(
 // AWS set, does not include credentials. Individual services may require
 // additional configuration.
 var Services = wire.NewSet(
+	// gocloud.dev/blob
```
shantuo: Will it be a better idea to split these into their driver packages? For example, in `s3blob` package:
```go
var Set = wire.NewSet(
    s3blob.Options{},
    s3blob.URLOpener{})
```

vangent: Done.

PR #1553, opened 2019-03-14
pubsub/natspubsub: add an ackFunc to the Subscription constructor

vangent:
@derekcollison

### DIFF
--- a/pubsub/natspubsub/example_test.go
+++ b/pubsub/natspubsub/example_test.go
@@ -35,7 +35,10 @@ func ExampleCreateTopic() {
 	}
 	defer nc.Close()
 
-	pt := natspubsub.CreateTopic(nc, "go-cloud.example.send")
+	pt, err := natspubsub.CreateTopic(nc, "go-cloud.example.send")
+	if err != nil {
+		// Handle error....
+	}
 
 	err = pt.Send(ctx, &pubsub.Message{Body: []byte("example message")})
 }
@@ -59,7 +62,10 @@ func ExampleCreateSubscription() {
 		// function to do nothing, or panic/log a warning if your application
 		// is built for at-most-once semantics and should never call Ack.
 	}
-	sub := natspubsub.CreateSubscription(nc, "go-cloud.example.send", ackFunc)
+	sub, err := natspubsub.CreateSubscription(nc, "go-cloud.example.receive", ackFunc)
+	if err != nil {
+		// Handle error....
+	}
 
 	// Now we can use sub to receive messages.
 	msg, err := sub.Receive(ctx)
@@ -68,6 +74,7 @@ func ExampleCreateSubscription() {
 	}
 	// Handle Message
 
-	// Ack will call ackFunc above.
+	// Ack will call ackFunc above. If you're only going to use at-most-once
+	// providers, you can omit it.
 	msg.Ack()
 }
--- a/pubsub/natspubsub/example_test.go
+++ b/pubsub/natspubsub/example_test.go
@@ -52,7 +52,14 @@ func ExampleCreateSubscription() {
 	}
 	defer nc.Close()
 
-	sub := natspubsub.CreateSubscription(nc, "go-cloud.example.send")
+	ackFunc := func() {
+		// This function will be called when the application calls "Ack" on a
+		// received message.
+		// Since Ack is a meaningless no-op for NATS, you can provide an empty
+		// function to do nothing, or panic/log a warning if your application
+		// is built for at-most-once semantics and should never call Ack.
+	}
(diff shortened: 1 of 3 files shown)


pubsub/natspubsub/example_test.go
```diff
@@ -52,7 +52,14 @@ func ExampleCreateSubscription() {
 	}
 	defer nc.Close()
 
-	sub := natspubsub.CreateSubscription(nc, "go-cloud.example.send")
+	ackFunc := func() {
+		// This function will be called when the application calls "Ack" on a
+		// received message.
+		// Since Ack is a meaningless no-op for NATS, you can provide an empty
+		// function to do nothing, or panic/log a warning if your application
```
jba: This comment is fine. But FTR, I'm more concerned about applications that are build for at-*least*-once semantics and *do* call ack, but inadvertently get linked with with an at-most-once provider.

pubsub/natspubsub/nats.go
```diff
@@ -160,24 +159,30 @@ func (*topic) ErrorCode(err error) gcerrors.ErrorCode {
...
+	nsub    *nats.Subscription
+	ackFunc func()
+	err     error
 }
 
 // CreateSubscription returns a *pubsub.Subscription representing a NATS subscription.
 // TODO(dlc) - Options for queue groups?
-func CreateSubscription(nc *nats.Conn, subscriptionName string) *pubsub.Subscription {
-	return pubsub.NewSubscription(createSubscription(nc, subscriptionName), nil)
+func CreateSubscription(nc *nats.Conn, subscriptionName string, ackFunc func()) *pubsub.Subscription {
```
jba: You should move or copy your explanation of ackFunc here. It's more important that it be here.

pubsub/natspubsub/nats.go
```diff
@@ -160,24 +159,30 @@ func (*topic) ErrorCode(err error) gcerrors.ErrorCode {
...
+func createSubscription(nc *nats.Conn, subscriptionName string, ackFunc func()) driver.Subscription {
 	sub, err := nc.SubscribeSync(subscriptionName)
-	return &subscription{nc, sub, err}
+	return &subscription{nc, sub, ackFunc, err}
 }
 
 // AckFunc implements driver.Subscription.AckFunc.
-func (*subscription) AckFunc() func() { return nil }
+func (s *subscription) AckFunc() func() {
+	if s == nil {
```
jba: We generally don't write these receiver nil checks. (Otherwise we'd be writing them everywhere.) Unless there is some good reason, like it's normal for `s` to be nil.

pubsub/natspubsub/nats.go
```diff
@@ -160,24 +159,30 @@ func (*topic) ErrorCode(err error) gcerrors.ErrorCode {
...
 // CreateSubscription returns a *pubsub.Subscription representing a NATS subscription.
 // TODO(dlc) - Options for queue groups?
-func CreateSubscription(nc *nats.Conn, subscriptionName string) *pubsub.Subscription {
-	return pubsub.NewSubscription(createSubscription(nc, subscriptionName), nil)
+func CreateSubscription(nc *nats.Conn, subscriptionName string, ackFunc func()) *pubsub.Subscription {
+	return pubsub.NewSubscription(createSubscription(nc, subscriptionName, ackFunc), nil)
 }
 
-func createSubscription(nc *nats.Conn, subscriptionName string) driver.Subscription {
+func createSubscription(nc *nats.Conn, subscriptionName string, ackFunc func()) driver.Subscription {
```
jba: Check that ackFunc is non-nil. If it is nil, make an error and stick it into the subscription.

pubsub/natspubsub/example_test.go
vangent: Ack.

vangent: :-)

pubsub/natspubsub/nats.go
vangent: Good point, I meant to do that but forgot. Done.

pubsub/natspubsub/nats.go
vangent: The tests crash with out it, here: 

https://github.com/google/go-cloud/blob/master/pubsub/pubsub.go#L537

The conformance tests create a "nonexistent" subscription, which is implemented for NATS with a `nil` `*subscription`; `pubsub.NewSubscription` does get called for it though.

pubsub/natspubsub/nats.go
vangent: It seems that that error wasn't being used anywhere. I've modified the constructors to return errors explicitly so we can validate arguments.

PR #1557, opened 2019-03-14
.travis.yml: test Go 1.11 and 1.12

jba:
We should always be testing against the current and previous Go versions.

Only check deps on 1.12.

Only run 1.11 on linux.


.travis.yml
```diff
@@ -1,6 +1,13 @@
 language: go
 go_import_path: gocloud.dev
-go: "1.11.x"
+go:
+  - "1.11.x"
+  - "1.12.x"
```
eliben: I'd add a comment here saying something about tests depending on version numbers and search for versions in the test script when bumping these

internal/testing/runchecks.sh
```diff
@@ -77,9 +77,14 @@ fi
 # ./internal/testing/alldeps
 #
 # Whenever project dependencies change, rerun ./internal/testing/listdeps.sh
-./internal/testing/listdeps.sh | diff ./internal/testing/alldeps - || {
-  echo "FAIL: dependencies changed; compare listdeps.sh output with alldeps" && result=1
-}
+#
+# Run this only on the latest version of Go, since module behavior may differ
+# across versions.
+if [[ $(go version) == *1\.12* ]]; then
```
eliben: Nit: there's `TRAVIS_GO_VERSION` which may be a bit cleaner

.travis.yml
vangent: I think it may be clearer to remove the "go" line from here and move it down to "jobs", where we configure the 3 OSs. I think the "matrix" can be removed and handled there too.

internal/testing/runchecks.sh
jba: I saw that, but it's undocumented what the content of that env var is. I'd rather have an expression I can test locally.

.travis.yml
jba: Done, and I also expanded the matrix for clarity. Not sure there is a way to write this that doesn't repeat either the go version or the list of OSs.

jba:
Has anyone seen a "duplicate symbol" error before? (https://travis-ci.com/google/go-cloud/jobs/185066070).

I'll explore more tomorrow.

jba:
This is a known issue (golang/go#30374). It was supposedly fixed in 12.1, but it wasn't. I filed a repro on the bug. I'll just let this PR sit here until it's fixed.

jba:
The bug has been fixed and should be in Go 1.12.2, due around April 1.

vangent:
Can this go forward yet?

PR #1558, opened 2019-03-14
mysql: add support for Azure Database for MySQL

### LINKED ISSUE #1305: sql: add support for Azure SQL providers

vsaroopchand:
resolves #1305 , adds to #76 

Here is my proposal for Azure MySQL. As an FYI, I added an AzureCertFetcher helper to fetch (remote endpoint) or load (local FS) certs.


mysql/azuremysql/azuremysql.go
```diff
@@ -0,0 +1,127 @@
...
+			Passwd:                  c.params.Password,
+			TLSConfig:               tlsConfigName,
+			AllowCleartextPasswords: true,
+			AllowNativePasswords:    true,
+			DBName:                  c.params.Database,
+		}
+		c.dsn = cfg.FormatDSN()
+		close(c.ready)
+
+	case <-c.ready:
```
clausti: Can you update with a couple of explanatory comments, similar to https://github.com/google/go-cloud/blob/master/mysql/rdsmysql/rdsmysql.go#L121 and https://github.com/google/go-cloud/blob/master/mysql/rdsmysql/rdsmysql.go#L123, for clarity?

mysql/azuremysql/azurecertfetcher.go
```diff
@@ -0,0 +1,123 @@
...
+	"strings"
+
+	"golang.org/x/net/context/ctxhttp"
+)
+
+type (
+
+	// AzureCertFetcher loads the CA certificates from remote endpoint or file.
+	AzureCertFetcher struct {
+		client *http.Client
```
zombiezen: `Client` should be exported so as to permit the user to make changes.

mysql/azuremysql/azurecertfetcher.go
```diff
@@ -0,0 +1,123 @@
...
+		// certLocation can be a remote endpoint or a file path
+		certLocation string
+		useHTTP      bool
+	}
+)
+
+// NewAzureCertFetcher constructs a new *AzureCertFetcher.
+// See https://docs.microsoft.com/en-us/azure/mysql/howto-configure-ssl.
+func NewAzureCertFetcher(caBundleLocation string) (*AzureCertFetcher, error) {
```
zombiezen: Is there a reason to have a different location in Azure? Someone who wants it from a different location or a file should be able to provide their own implementation of an interface like `rds.CertPoolProvider`:

https://github.com/google/go-cloud/blob/7206007cdd0c778b597a28d615b2262c04cc7806/aws/rds/rds.go#L38-L41

mysql/azuremysql/azuremysql.go
vsaroopchand: Ok, comments added.

mysql/azuremysql/azurecertfetcher.go
```diff
@@ -0,0 +1,123 @@
...
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"golang.org/x/net/context/ctxhttp"
+)
+
+type (
```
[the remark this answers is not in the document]
vsaroopchand: Fixed.

mysql/azuremysql/azurecertfetcher.go
vsaroopchand: Can we possibly move this interface outside of aws/rds and make it more generic?

```
type CertPoolProvider interface {
	GetCertPool(context.Context) (*x509.CertPool, error)
}
```
This way, my azure cert provider can simply implement that contract?

zombiezen: I think it's helpful to have two different interfaces to avoid potentially mixing types. Conceptually, these are returning two different certificate pools. Let's introduce `azure/azuredb` for this, since likely we'll want to reuse this certificate fetcher for PostgreSQL too.

vsaroopchand: Ok, sounds good.

vsaroopchand:
PTAL

mysql/azuremysql/azurecertfetcher.go
vsaroopchand: Done

azure/azuredb/azuredb.go
```diff
@@ -0,0 +1,126 @@
...
+type AzureCertFetcher struct {
+	Client *http.Client
+	// certLocation can be a remote endpoint or a file path
+	certLocation string
+	useHTTP      bool
+}
+
+// NewAzureCertFetcher constructs a new *AzureCertFetcher.
+// See https://docs.microsoft.com/en-us/azure/mysql/howto-configure-ssl.
+func NewAzureCertFetcher(caBundleLocation string) (*AzureCertFetcher, error) {
```
zombiezen: For easier use in Wire, it would be good to have a version of this function that uses the default remote URL.

mysql/azuremysql/azuremysql_test.go
```diff
@@ -0,0 +1,72 @@
...
+		t.Skipf("Could not obtain harness info: %v", err)
+	}
+
+	serverName, _ := tfOut["serverName"].Value.(string)
+	username, _ := tfOut["username"].Value.(string)
+	password, _ := tfOut["password"].Value.(string)
+	databaseName, _ := tfOut["database"].Value.(string)
+
+	ctx := context.Background()
+	acp, _ := azuredb.NewAzureCertFetcher("https://www.digicert.com/CACerts/BaltimoreCyberTrustRoot.crt.pem")
```
zombiezen: Let's make this URL a named constant in `azuredb` (related to my other comment).

azure/azuredb/azuredb.go
vsaroopchand: Make sense, new constructor with default added.
`azuredb.NewAzureCertFetcherWithDefault()`

mysql/azuremysql/azuremysql_test.go
vsaroopchand: Test will use the newly added default constructor w/ const location
`azuredb.NewAzureCertFetcherWithDefault()`

PR #1562, opened 2019-03-14
internal/website: add info about connecting to alternative S3 APIs

zombiezen:


internal/website/content/pages/blob.md
```diff
@@ -22,6 +22,24 @@ Top-level package documentation: https://godoc.org/gocloud.dev/blob
 * [File-backed local blob](https://godoc.org/gocloud.dev/blob/fileblob) - local
   blob implementation using the file system
 
+The blob package can also interact with most any server that implements the AWS
+S3 HTTP API, like [Minio][], [Ceph][], or [SeaweedFS][]. You can change the
+endpoint used in an S3 URL using query parameters like so:
```
jba: Is there a way to do it without URLs?

zombiezen: Yes, all of the query parameters correlate 1:1 with `aws.Config` fields. I could rewrite the sample in terms of creating an AWS session, but it would be a bit longer. Do you think it helps clarity?

zombiezen:
@eliben Linked to Wikipedia section.

internal/website/content/pages/blob.md
jba: No, just mention that the same can be done with `aws.Config`. Your text gives the impression that you can only do this with URLs.

zombiezen: Done.

vangent:
As discussed in person, why put this on the website instead of in godoc? Your code snippets could be example(s).

zombiezen:
> As discussed in person, why put this on the website instead of in godoc? Your code snippets could be example(s).

My rationale is that this sort of information feels more how-to-y than reference-y. But I see your point as-is that this information is a bit strange where it is now (I would mostly consider this current page a reference page).

Since I'm planning to add more how-to guides to the site (#1576), I could hold off on this specific change for a more comprehensive solution. WDYT, @vangent?

vangent:
This info feels to me like it should be discoverable from the godoc for `s3blob`. I don't feel particularly strongly about whether it lives there, or lives elsewhere and  there's a pointer to the elsewhere in the `s3blob` godoc.

So, for this PR, my two cents is to either move it to the godoc, or add a pointer to the website from the godoc. Holding off is fine with me as well if you prefer.

zombiezen:
I'm going to incorporate this information into the how-to guide. I will link between godoc and website as appropriate when doing so. Closing this PR.

PR #1563, opened 2019-03-14
pubsub/nats: add URLOpener and Options

vangent:


pubsub/natspubsub/nats.go
```diff
@@ -39,6 +54,95 @@ import (
...
+	return CreateTopic(o.Connection, topicName, &o.TopicOptions)
+}
+
+// OpenSubscriptionURL opens a pubsub.Subscription based on u.
+func (o *URLOpener) OpenSubscriptionURL(ctx context.Context, u *url.URL) (*pubsub.Subscription, error) {
+	for param := range u.Query() {
+		return nil, fmt.Errorf("open subscription %v: invalid query parameter %s", u, param)
+	}
+	topicName := path.Join(u.Host, u.Path)
+	return CreateSubscription(o.Connection, topicName, func() {}, &o.SubscriptionOptions)
```
jba: This isn't acceptable to me. Either the ack func should panic by default, or there should not be a default.

If you choose the latter, then I'd be OK with a required query param ackfunc=nop or ackfunc=panic. Although it's weird to have a required query param, isn't it?

vangent: Defaulted to `panic`.

vangent:
@derekcollison FYI.

derekcollison:
NATS main object is a connection with which it can multiplex any subject on for publish/subscribe, RPC etc. Not sure why we are adding these that tie a connection to a single topic/subject. What objective are we trying to achieve?

vangent:
Hi @derekcollison, there's background on the use of URLs in the Go CDK [here](https://godoc.org/gocloud.dev#hdr-URLs). The main goal is to make it easier to configure connections to resources. For example, `pubsub.OpenTopic("nats://mytopic")` where the URL string could be in your config file.

Go CDK users are not using the NATS object directly; they are going through the `pubsub.Topic` or `pubsub.Subscription` portable type, which are tied to a specific topic. I understand that using NATS directly wouldn't have that requirement.

Sending/receiving messages to many different/dynamic topics is not well supported by `pubsub` right now; many providers require provisioning of topics/subscriptions. It's something we may add better portable support for later if there's demand.

derekcollison:
We should try to make sure if someone does this 10 times they do not open 10 connections to the NATS server, if that make sense.

vangent:
> We should try to make sure if someone does this 10 times they do not open 10 connections to the NATS server, if that make sense.

That should not happen. Users will either use a custom `URLOpener`, in which case they provide their own connection (see `URLOpener.Connection`), which will be reused, or they will use the default one, in which case they specify a single NATS server with via an environment variable `NATS_SERVER_URL`, which is only opened once (see `o.init.Do`).

pubsub/natspubsub/nats.go
```diff
@@ -39,6 +55,120 @@ import (
...
+
+	var ackFunc func()
+	s := q.Get("ackfunc")
+	switch s {
+	case "log":
+		ackFunc = func() { log.Printf(AckWarning) }
+	case "noop":
+		ackFunc = func() {}
+	case "", "panic":
+		ackFunc = func() { log.Fatal(AckWarning) }
```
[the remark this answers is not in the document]
vangent: Done.

PR #1565, opened 2019-03-14
pubsub: make aws and mem drivers sleep after empty messages; better cancel error checks

vangent:


vangent:
Note: the new test is failing for `awspubsub` right now, I think due to an `httpreplay` bug where it's not recording requests that end up being canceled.

pubsub/drivertest/drivertest.go
```diff
@@ -374,15 +377,32 @@ func testCancelSendReceive(t *testing.T, newHarness HarnessMaker) {
...
+	delayedCancelCtx, cancel := context.WithCancel(ctx)
+	ch := make(chan struct{})
+	go func() {
+		if _, err := sub.Receive(delayedCancelCtx); !isCanceled(err) {
+			t.Errorf("sub.Receive returned %v (%T), want context.Canceled", err, err)
+		}
+		close(ch)
+	}()
+	// Give the goroutine some time to spin up and be blocking in Receive.
+	// There's no guarantee it will get there, but tests will at least be flaky
```
jba: It took me a while to understand this sentence. How about:

"..., but if Receive doesn't handle cancelation properly, then the test will fail whenever it does get there, so at least we'll observe a flaky test."

pubsub/drivertest/drivertest.go
```diff
@@ -374,15 +377,32 @@ func testCancelSendReceive(t *testing.T, newHarness HarnessMaker) {
...
+	// if it gets there sometimes.
+	time.Sleep(50 * time.Millisecond)
+	cancel()
+	<-ch
```
jba: Add a `<-time.After(10*time.Second)` here. If Receive ignores cancelation, then this test will fail by timing out, which will take 10 minutes if you don't make it shorter.

pubsub/drivertest/drivertest.go
```diff
@@ -374,15 +377,32 @@ func testCancelSendReceive(t *testing.T, newHarness HarnessMaker) {
...
+	go func() {
+		if _, err := sub.Receive(delayedCancelCtx); !isCanceled(err) {
+			t.Errorf("sub.Receive returned %v (%T), want context.Canceled", err, err)
+		}
+		close(ch)
+	}()
+	// Give the goroutine some time to spin up and be blocking in Receive.
+	// There's no guarantee it will get there, but tests will at least be flaky
+	// if it gets there sometimes.
+	time.Sleep(50 * time.Millisecond)
```
jba: Maybe this should be longer. 50ms is nothing to Windows. (If the sleep is so short that the delay always happens before Receive blocks, then this test won't detect anything.)

jba:
> Note: the new test is failing for `awspubsub` right now, I think due to an `httpreplay` bug where it's not recording requests that end up being canceled.

So you're getting a "request not found" error on replay?

vangent:
> > Note: the new test is failing for `awspubsub` right now, I think due to an `httpreplay` bug where it's not recording requests that end up being canceled.
> 
> So you're getting a "request not found" error on replay?

```
2019/03/14 16:35:17 ERROR: martian: failed to round trip: no matching request for &{Method:POST URL:https://sqs.us-east-2.amazonaws.com/ Proto:HTTP/1.1 ProtoMajor:1 ProtoMinor:1 Header:map[Accept-Encoding:[gzip] Authorization:[AWS4-HMAC-SHA256 Credential=FAKE_ID/20190314/us-east-2/sqs/aws4_request, SignedHeaders=content-length;content-type;host;x-amz-date;x-amz-security-token, Signature=56e558b31a1951e55f5e831b31088cca077c11ae11e19587ebdd69cd469630e9] Content-Length:[183] Content-Type:[application/x-www-form-urlencoded; charset=utf-8] User-Agent:[aws-sdk-go/1.16.23 (go1.12; linux; amd64)] X-Amz-Date:[20190314T233517Z] X-Amz-Security-Token:[FAKE_TOKEN] X-Cloud-Trace-Context:[811e5f30686c2aae106d34e77eae6f9a/11468081016047235866;o=0]] Body:{Reader:0xc000612c00} GetBody:<nil> ContentLength:183 TransferEncoding:[] Close:false Host:sqs.us-east-2.amazonaws.com Form:map[] PostForm:map[] MultipartForm:<nil> Trailer:map[] RemoteAddr:127.0.0.1:39052 RequestURI:/ TLS:<nil> Cancel:<nil> Response:<nil> ctx:<nil>}
--- FAIL: TestConformance (5.44s)
    --- FAIL: TestConformance/TestCancelSendReceive (0.19s)
        setup.go:105: Replaying from golden file testdata/TestConformance/TestCancelSendReceive.replay
        drivertest.go:395: sub.Receive returned pubsub (code=Unknown): SerializationError: failed to decode query XML error response
            	status code: 502, request id: 
            caused by: EOF (*gcerr.Error), want context.Canceled
FAIL
```

pubsub/drivertest/drivertest.go
vangent: Done.

pubsub/drivertest/drivertest.go
vangent: Upped to 100ms. I don't think it matters too much though; as the previous comment observes, the test will be flaky if there's a bug and it gets to `Receive` sometimes.

pubsub/drivertest/drivertest.go
vangent: Done.

vangent:
I fixed the compile error (woops), so you should be able to see the error in the next Travis run (or patch locally and repro).

jba:
I've been working on the httpreplay issue all morning, and I'm still not sure what to do.

The fundamental problem is that your test is time-dependent. If it were deterministic, I don't think there'd be a problem.

Here is how the `http` package works: if the `http.Transport` detects that the context is done before it writes to the socket, then nothing gets transmitted. The proxy doesn't see the request, so it doesn't get logged, and the replayer can't find it. That's what you're seeing.

But wait—why does the *recorder* get canceled early, but not the *replayer*? Something about the speed of the respective code paths, I guess. In my little test I've observed:
- Recorder gets canceled early, and so does replayer, so the result is the same. (Easy to get this by canceling the context before making the request.)
- Recorder returns a canceled error, but replayer does not, and in fact the replayer returns the correct, uncanceled response. Presumably the request completed on the server, and the cancelation was noticed and reported by the client afterward. (I see this by doing `time.AfterFunc(1*time.Millsecond, cancel)` for both record and replay. YMMV.)
- Recorder gets canceled early but replayer does not, giving a "no matching request" error. I get this around half the time when the cancel delay is 500 microseconds.

I can see two ways forward: I could probably make httpreplay work in this case, but it would add a lot of complexity (there'd have to be some sort of back door from the client to the logger, which is technically across the network but is actually in the same process, so it's doable). And I'm not convinced it's the right thing. It would basically be papering over an inherently flaky test.

Or we could find a way to make the test deterministic, which I think is preferable. I haven't looked carefully at what you're doing, but I think you could probably accomplish it with a test hook:
```
var receiveTestHook func() = func(){}
...
func (Subscription) Receive... {
   ...
   receiveTestHook()
   ...
}
```
I think that bit of ugliness is preferable to a time-dependent test.

Note that this technique is used in the standard library, in the `net/http` package in fact. For an example, see https://golang.org/src/net/http/client.go, line 512.

If you're trying to make sure that the driver is context-sensitive, then maybe put the hook just before the call to ReceiveBatch.

vangent:
@jba thanks for looking into this. The test hook approach won't easily work because the tests isn't in the `pubsub` package. Let me think about it some more...

jba:
Maybe the hook could be an exported variable in an internal package.

PR #1566, opened 2019-03-14
blob: add log.Printf warnings when Reader/Writer is not closed

### LINKED ISSUE #1528: blob: consider putting a finalizer on Reader/Writer to complain if they aren't closed
See #1454 for details about the finalizer.

#1526 is an example of how this might have helped a user debug.

vangent:
Fixes #1528.


blob/blob.go
```diff
@@ -638,12 +644,18 @@ func (b *Bucket) NewRangeReader(ctx context.Context, key string, offset, length
...
+	dr, err := b.b.NewRangeReader(ctx, key, offset, length, dopts)
 	if err != nil {
 		return nil, wrapError(b.b, err)
 	}
 	end := func(err error) { b.tracer.End(tctx, err) }
-	return &Reader{b: b.b, r: r, end: end, provider: b.tracer.Provider}, nil
+	r := &Reader{b: b.b, r: dr, end: end, provider: b.tracer.Provider}
+	runtime.SetFinalizer(r, func(r *Reader) {
+		if !r.closed {
+			log.Printf("A blob.Reader reading from %q was never closed", key)
```
zombiezen: To help the user even further in debugging, you can include the caller of `NewRangeReader`. Snippet to get file/lineno which may or may not be available:

```go
// Call this inside `NewRangeReader`, not the finalizer:
_, file, lineno, ok := runtime.Caller(1)
```

You may also need to tweak `NewReader`, since that will be a common caller of `NewRangeReader`.

(You should also do this for `NewWriter`.)

PR #1568, opened 2019-03-15
pubsub: rename awspubsub->awssnssqs and azurepubsub->azuresb

### LINKED ISSUE #1561: pubsub: rename package names to be their actual product name
Carrying over the [discussion](https://github.com/google/go-cloud/pull/1544#discussion_r265356720) here. I propose to rename `awspubsub` to `awssnssqs`, `azurepubsub` to `azureservicebus` or `azuresb`. The idea is it is clear which product we implement since there might be more than one offering per provider. All of our other API's try to follow this convention. If we decide to do it we should probably do it before next release.

What do you think? @jba

vangent:
Fixes #1561.


jba:
Wait a second. This is *not* how we name other drivers:
- blob consistently uses XXXblob: azureblob, s3blob, gcsblob, etc.

vangent:
FYI, https://github.com/google/go-cloud/blob/master/internal/docs/design.md#package-naming-conventions

documents our naming conventions.

* For blob, IMHO the names are all fine; "s3" and "gcs" are short enough that appending "blob" is reasonable and avoids conflicts with the provider libraries themselves.

* For runtimevar, I think I would have picked "awsparamstore" and "gcpruntimeconfig" or something like that. IMHO, the "var" suffix doesn't add as much value as "blob".

* For secrets, I like the names.

* For pubsub, I do think "awspubsub" is not quite right, because there are almost certainly multiple pub/sub solutions supported by AWS; same for "azurepubsub" (e.g., https://docs.microsoft.com/en-us/azure/event-grid/). "gcppubsub" seems OK because that's the actual name of the product; "rabbitpubsub" and "mempubsub" seem fine to me too.

"awssnssqspubsub" seems like too much to me though, as does "azuresbpubsub".

@jba @eliben @zombiezen @shantuo @clausti for more bikeshedding.

jba:
> I think I would have picked

That's a strange tense to use for something you have control over and could change at any time.

The argument that there could be alternative products from the same company is a good one. But doesn't that apply to blob too? Why `gcsblob` and not `gcpgcsblob`? And the Azure one should technically be `azureblobblob`. 

 In any case, the argument from consistency fails because we have no consistency. (And as `azureblobblob` demonstrates, that's probably a good thing.) By all means, let's pick names we think are good, but we'll have to justify each on its own merits.

Re-approving because I guess I have no strong opinions either way about these two particular names.

vsaroopchand:
I just want to add my 2-cents, I agree with the name change for Azure Service Bus driver to azuresb ; we also have Azure EventHub and Azure Event Grid - I do plan on adding these at a later date.

jba:
Huh, github bug: I thought I unapproved, and there is a yellow dot next to my name at the top right instead of a green check, but the PR thinks I approved. Whatevs.

vangent:
> Why gcsblob and not gcpgcsblob

"gcs" stands for "Google Compute Storage", so the GCP isn't needed. I think "s3blob" was used instead of "awss3blob" just because S3 is so well known.

shantuo:
I think the proposed changes on runtimevar packages are also ok. In addition, the docstore package could be `gcpfirestore` and `awsdynamodb`, currently they are `firedocstore` and `dynamodocstore`. What do people think?

clausti:
I agree that there is value in having the portable API name be part of the provider names. Following the "servicename" + "portable API name" pattern, I find `snssqspubsub` more comprehensible than `awssnssqs` (though they're both quite an eyeful).

jba:
People tend to name packages with the whole import path in mind, but remember that maintainers and reviewers only see the last component. So `gcpfirestore` looks like it's just firestore, and similarly `awsdynamodb`. I think `firedocstore` and `dynamodocstore` uniquely identify the product and tell you that they are docstore APIs at the same time.

If I had to extract a general rule from all this, it would be to (a) identify the product exactly (so `awspubsub` doesn't work), and (b) to include the portable API suffix when that doesn't make the name too long, especially when the alternative is hard to distinguish from the provider's own package name.

vangent:
Does anyone object to the specific renames here?

vangent:
(and if so, please make a concrete suggestion).

I can send a separate PR for `runtimevar` renames.

shantuo:
I think @clausti and @jba convinced me.

vangent:
> I think @clausti and @jba convinced me.

Of what?

shantuo:
> If I had to extract a general rule from all this, it would be to (a) identify the product exactly (so `awspubsub` doesn't work), and (b) to include the portable API suffix when that doesn't make the name too long, especially when the alternative is hard to distinguish from the provider's own package name.

So `snssqspubsub` and `sbpubsub` according to these rules.

vangent:
Hmm, I don't like those names; I think that having `azure` and `aws` in there is important for discoverability. `sbpubsub` is too hard for a user to associate it with "Azure Service Bus".

My naming proposal would be something like:

`<provider><product><api>`, where:

1. `<provider>` is 'aws', 'gcp', or 'azure', or '' for 3rd party/hosted/local packages. `<provider>` can also be omitted in cases where the product name is sufficient (e.g., `s3blob` not `awss3blob`, `gcsblob` not `gcpgcsblob`).
2. `<product>` is the product/service name.
3. `<api>` is the portable API name, when it doesn't make the name too long, and when leaving it off makes it hard to distinguish from the provider's own package name.

(2 & 3 are close to @jba's rules)

Based on this rule, I would suggest the following renames:

* `runtimevar/paramstore` -> `runtimevar/awsparamstore` because "paramstore" is not sufficient to identify this as an AWS product.
* `runtimevar/runtimeconfigurator` -> `runtimevar/gcpruntimeconfig`, same reason, and leave off the "urator" just to make it a bit shorter.
* `pubsub/awspubsub` -> `pubsub/awssnssqs` because "snssqs" is not sufficient on its own to identify this as an AWS product, and because "awspubsub" does not uniquely identify the product. I don't think the "pubsub" suffix is necessary here, but could be convinced.
* `pubsub/azurepubsub` -> `pubsub/azuresb` because "azurepubsub" does not uniquely identify the product, and "servicebus" is not sufficient on its own to identify this as an Azure product. I don't think the "pubsub" suffix is necessary here, but could be convinced.


Thoughts?

jba:
SGTM.

zombiezen:
I am onboard with @vangent's latest suggestion.

vangent:
@shantuo @clausti does that sound OK?

shantuo:
SGTM.

Issue #1175, opened 2019-01-24
sql: add top-level mysql and postgres open functions


zombiezen:
While there is already an `Open` function in the standard `database/sql` package, I think we should consider including our own. @ijt has already pointed out offline that it's strange that we don't have anything in the mysql and postgres top-level directories (although he's thinking more along the lines of #1150). IMO the automatic instrumentation that we provide for AWS and GCP are useful enough to merit our own `Open` function.

(We likely want to use the design from #1174, so one can think of this as a sub-issue of that.)

Zyqsempai:
Sounds interesting;) Can i help with that?

ijt:
Sure, please feel free to send a pull request.

zombiezen:
Yes! We just finalized the [URLs design](https://github.com/google/go-cloud/blob/master/internal/docs/design.md#urls), so make sure your PR follows that pattern. See #1282 for an example of this for the blob package.

Zyqsempai:
cool, thanks, guys.

Zyqsempai:
@zombiezen @ijt  Hi guys, i have a question during the implementation, so in GCP postgres we have some extra connections params https://github.com/google/go-cloud/blob/master/postgres/cloudpostgres/cloudpostgres.go#L90 how they will be presented in the connection url? like a rawQuery?
By connection url i mean ofcourse something like that `postgresql://other@localhost/otherdb?connect_timeout=10&application_name=myapp
`

zombiezen:
What about (really rough sketch, don't take too much stock in this):

```
cloudpostgres://user:password@project/region/instance/databaseName?query
```

where `query` is passed through to [`Params.Values`](https://github.com/google/go-cloud/blob/6838e153f0aa8a29e0a6b6eb4120eabc7bbb3109/postgres/cloudpostgres/cloudpostgres.go#L52)?

zombiezen:
Also, in terms of PR sequencing, it might be good to just start with one PR that exports the URL opening functions for just one of the top-level packages (i.e. mysql or postgres), includes a plain database opener, but doesn't hook up the Cloud SQL/RDS openers yet. This would allow us to figure out the structure and then fill in the rest later, rather than trying to figure out all the details up-front.

Zyqsempai:
@zombiezen great news, thank you for merging it.
The next step will be, to implement the same for plain MySql? Or better to continue with postgreSql?

zombiezen:
Your call. I'm happy to review either of those.

[...]

PR #1571, opened 2019-03-15
postgres: RDS PostgreSQL OpenUrl implementation

Zyqsempai:
New URL design implementation for RDS PostgreSQL, related to the #1175

### DIFF
--- a/postgres/rdspostgres/rdspostgres.go
+++ b/postgres/rdspostgres/rdspostgres.go
@@ -29,6 +29,7 @@ import (
 	"contrib.go.opencensus.io/integrations/ocsql"
 	"github.com/lib/pq"
 	"gocloud.dev/aws/rds"
+	"gocloud.dev/postgres"
 )
 
 // Params specifies how to connect to an RDS database.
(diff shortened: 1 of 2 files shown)


postgres/rdspostgres/rdspostgres.go
```diff
@@ -26,6 +26,8 @@ import (
 	"net/url"
 	"time"
 
+	"gocloud.dev/postgres"
+
```
zombiezen: nit: No blank line, keep in the same group as the other imports.

postgres/rdspostgres/rdspostgres.go
```diff
@@ -53,6 +55,40 @@ type Params struct {
 	TraceOpts []ocsql.TraceOption
 }
 
+type URLOpener struct {
```
zombiezen: Missing doc comment:

> URLOpener opens RDS PostgreSQL URLs like "rdspostgres://user:password@myinstance.borkxyzzy.us-west-1.rds.amazonaws.com:5432/mydb".

postgres/rdspostgres/rdspostgres.go
```diff
@@ -53,6 +55,40 @@ type Params struct {
...
+
+func init() {
+	postgres.DefaultURLMux().RegisterPostgres(Scheme, &URLOpener{})
+}
+
+func (uo *URLOpener) OpenPostgresURL(ctx context.Context, u *url.URL) (*sql.DB, error) {
+	cf := new(rds.CertFetcher)
+	for k, _ := range u.Query() {
+		// Only permit parameters that do not conflict with other behavior.
+		if k == "user" || k == "password" || k == "dbname" || k == "host" || k == "port" || k == "sslmode" || k == "sslcert" || k == "sslkey" || k == "sslrootcert" {
```
zombiezen: As I discovered in #1560, this check makes it impossible to connect via Unix sockets. While I grant connecting over Unix sockets doesn't happen for RDS, I don't see these parameters as being overly harmful, so let's drop the check.

postgres/rdspostgres/rdspostgres.go
```diff
@@ -53,6 +55,40 @@ type Params struct {
...
+			return nil, fmt.Errorf("cloudpostgres: OpenPostgresURL: extra parameter %s not allowed; use Params fields instead", k)
+		}
+	}
+	vals := u.Query()
+	vals.Set("sslmode", "disable")
```
zombiezen: By and the large, the utility of this package is to connect to RDS databases securely using TLS. We shouldn't be disabling this (see with comment below).

postgres/rdspostgres/rdspostgres.go
```diff
@@ -53,6 +55,40 @@ type Params struct {
...
+	u.RawQuery = vals.Encode()
+
+	db := sql.OpenDB(connector{
```
zombiezen: Instead of calling `sql.OpenDB`, this code should be wrapping the `Open` function in this package, which properly handles the parameters and such.

postgres/rdspostgres/rdspostgres_test.go
```diff
@@ -60,6 +60,27 @@ func TestOpen(t *testing.T) {
 	if err := db.Close(); err != nil {
 		t.Error("Close:", err)
 	}
+
+	// Temporary test part for temporary OpenWithUrl method
```
zombiezen: So far, we've largely been leaving `URLOpener` types mostly untested because the logic is usually trivial enough to be reviewable, the cost for test in integration is high, and then the contortions to the API required to test the logic in isolation make the control flow harder to read.

Since the logic required here is mostly matching the parts of a `*URL` to the appropriate slots in the input to `rdspostgres.Open`, I think we should drop the tests here.

postgres/rdspostgres/rdspostgres.go
Zyqsempai: So, i fixed most of the comments except this one, coz it doesn't seems right to me, we pars Url struct in OpenPostgresURL to turn it into Params struct just to make it Url again in Open function, or we shoulr rewright Open function or we shouldn't use it.

zombiezen: I understand that is how it operates internally, but the MySQL driver gives us no other way to pass the data in directly other connection strings (see go-sql-driver/mysql#771 where I've tried to report a similar concern upstream). While there is a mild inefficiency here, we're trading that for simplicity of understanding: URL openers should always conceptually use the same constructor that the driver type uses. This allows us to gain test confidence in individual units rather than having to test multiple codepaths.

Zyqsempai: Ok, i get it, pushed the final version. Thanks for explanation.

zombiezen:
Ah oops, I'm mixing up PRs. I'll merge.

PR #1572, opened 2019-03-15
travis: add a check for incompatible API changes

### LINKED ISSUE #1570: all: hook up apidiff to identify breaking changes
https://github.com/golang/exp/tree/master/apidiff

Apparently @jadekler has hooked it up for the cloud.google.com/go repo.

@jba FYI

vangent:
Fixes #1570.

The checks are relatively slow; it takes about 1 second to run `apidiff` on a package, and we need to run it twice (actually, we run it 4 times, but the `-w` ones to export pkg data are the expensive ones). We have about 50 packages.

It would be nice to use `git diff --names-only master` to quickly figure out which packages have changes, and skip packages that don't have changes, but it's not obvious to me how to map file names from `git diff` to package names from `go list`.

### DIFF
--- a/internal/testing/check_api_change.sh
+++ b/internal/testing/check_api_change.sh
@@ -25,7 +25,7 @@ echo
 
 go install -mod=readonly golang.org/x/exp/cmd/apidiff
 
-MASTER_CLONE_DIR="/tmp/go-cloud"
+MASTER_CLONE_DIR="$(mktemp -d)"
 PKGINFO_BRANCH=$(mktemp)
 PKGINFO_MASTER=$(mktemp)
 
@@ -52,9 +52,7 @@ for pkg in ${PKGS}; do
   apidiff -w ${PKGINFO_BRANCH} ${pkg}
 
   # Compute export data for master@HEAD.
-  cd ${MASTER_CLONE_DIR}
-  apidiff -w ${PKGINFO_MASTER} ${pkg}
-  cd - > /dev/null
+  (cd ${MASTER_CLONE_DIR}; apidiff -w ${PKGINFO_MASTER} ${pkg})
 
   # Print all changes for posterity.
   apidiff ${PKGINFO_MASTER} ${PKGINFO_BRANCH}
--- a/internal/testing/check_api_change.sh
+++ b/internal/testing/check_api_change.sh
@@ -15,8 +15,8 @@
 
 # This script checks to see if there are any incompatible API changes on the
 # current branch relative to master@HEAD.
-# It fails if it finds any, unless there is a commit message with
-# BREAKING_CHANGE_OK in it.
+# It fails if it finds any, unless there is a commit with BREAKING_CHANGE_OK
+# in the first line of the commit message.
 
 set -euo pipefail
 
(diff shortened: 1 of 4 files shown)


internal/testing/check_api_change.sh
```diff
@@ -0,0 +1,85 @@
...
+
+if [ ${#incompatible_change_pkgs[@]} -eq 0 ]; then
+  # No incompatible changes, we are good.
+  echo "OK: No incompatible changes found."
+  exit 0;
+fi
+echo "Found breaking API change(s) in: ${incompatible_change_pkgs[@]}."
+
+# Found incompatible changes; see if they were declared as OK via a commit.
+if git cherry -v master | grep -q "BREAKING_CHANGE_OK"; then
```
jba: This only seems to display the first line of the commit. Maybe a variant of `git log`?

Or maybe the idea is just to push a trivial commit to master with BREAKING_CHANGE_OK as the only message?

internal/testing/check_api_change.sh
```diff
@@ -0,0 +1,85 @@
...
+
+incompatible_change_pkgs=()
+PKGS=$(go list ./... | grep -v internal | grep -v test | grep -v samples)
+for pkg in ${PKGS}; do
+  echo "Testing ${pkg}..."
+
+  # Compute export data for the current branch.
+  apidiff -w ${PKGINFO_BRANCH} ${pkg}
+
+  # Compute export data for master@HEAD.
```
jba: We could compute and store this file at each release.

internal/testing/check_api_change.sh
vangent: It will work as long as BREAKING_CHANGE_OK is on the first line of the commit. It seems like most people usually only have 1-line commits, so I don't think that's an onerous restriction. I updated the messages to reflect that though.

internal/testing/check_api_change.sh
vangent: We could, but if we're comparing the PR to the last release, then every PR after a breaking change would also be flagged as a break change, no?

internal/testing/check_api_change.sh
```diff
@@ -0,0 +1,85 @@
...
+# in the first line of the commit message.
+
+set -euo pipefail
+
+echo "Checking for incompatible API changes relative to master@HEAD..."
+echo
+
+go install -mod=readonly golang.org/x/exp/cmd/apidiff
+
+MASTER_CLONE_DIR="/tmp/go-cloud"
```
eliben: Any reason this isn't another `$(mktemp)`?

internal/testing/check_api_change.sh
```diff
@@ -0,0 +1,85 @@
...
+incompatible_change_pkgs=()
+PKGS=$(go list ./... | grep -v internal | grep -v test | grep -v samples)
+for pkg in ${PKGS}; do
+  echo "Testing ${pkg}..."
+
+  # Compute export data for the current branch.
+  apidiff -w ${PKGINFO_BRANCH} ${pkg}
+
+  # Compute export data for master@HEAD.
+  cd ${MASTER_CLONE_DIR}
```
eliben: A subshell would be cleaner, maybe:

`(cd ${MASTER_CLONE_DIR}; ... command)`

no need to undo

internal/testing/check_api_change.sh
jba: That's true.

internal/testing/check_api_change.sh
vangent: No. Done.

internal/testing/check_api_change.sh
vangent: Nice, done.

vangent:
PTAL.

PR #1574, opened 2019-03-15
docstore: improve codec conformance tests

jba:
Test that the docstore codec can round-trip with itself using
a large set of types (ultimately, all the types that docstore supports).

For native-docstore round trips, let the driver tell the conformance
tests what types it doesn't support, so we can skip those in a
driver-specific way.

### DIFF
--- a/internal/docstore/drivertest/drivertest.go
+++ b/internal/docstore/drivertest/drivertest.go
@@ -21,6 +21,7 @@ import (
 	"math/rand"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


internal/docstore/drivertest/drivertest.go
```diff
@@ -299,47 +317,117 @@ func testCodec(t *testing.T, ct CodecTester) {
...
+		A  [2]int
 		M  map[string]bool
+		P  *string
+		T  time.Time
 	}
-	// TODO(jba): add more fields: more basic types; pointers; structs; embedding.
+	// TODO(jba): add more fields: structs; embedding.
 
-	in := S{
+	s := "bar"
```
vangent: Consider adding a comment here, explaining that the docstore round trip should work for all docstore-supported types, regardless of native driver support.

internal/docstore/drivertest/drivertest.go
```diff
@@ -299,47 +317,117 @@ func testCodec(t *testing.T, ct CodecTester) {
...
-		enc, err := encode(in)
-		if err != nil {
-			t.Fatal(err)
-		}
-		var dec S
-		if err := decode(enc, &dec); err != nil {
-			t.Fatal(err)
+	check(dsrt, &DocstoreRoundTrip{}, ct.DocstoreEncode, ct.DocstoreDecode)
+
+	// All native codecs should support these types. If one doesn't, remove it from this
```
vangent: And then another comment here explaining that now we're testing native encode -> docstore decode and vice versa.

Or consider splitting them into separate tests?

jba: Added comment.

internal/docstore/drivertest/drivertest.go

PR #1575, opened 2019-03-15
pubsub: modify mem/aws/rabbit behavior for 0 messages in ReceiveBatch; better cancel error checks

vangent:
This PR is a bit of a grab bag now, of things I discovered trying to add a conformance test for canceling an in-flight `Receive`.

* I found that we weren't returning the `Canceled` error code correctly for AWS; fixed (although I couldn't find a constant for that error).
* The test `isCanceled` function wasn't checking for `gcerr.Canceled`.
* `awspubsub` and `mempubsub` weren't handling "got no messages" consistently with what `driver.ReceiveBatch` says.
* I updated the docstring for `driver.ReceiveBatch` to be a bit more explicit.
* I added a comment about why adding a test for canceling an in-flight `Receive` is hard.

### DIFF
--- a/pubsub/awspubsub/awspubsub.go
+++ b/pubsub/awspubsub/awspubsub.go
@@ -404,10 +404,9 @@ func openSubscription(ctx context.Context, client *sqs.SQS, qURL string) driver.
 	return &subscription{client: client, qURL: qURL}
 }
 
-const (
-	// How often ReceiveBatch should poll if no messages are available.
-	pollDuration = 250 * time.Millisecond
-)
+// How long ReceiveBatch should wait if no messages are available; controls
+// the poll interval of requests to SQS.
+const noMessagesPollDuration = 250 * time.Millisecond
 
 // ReceiveBatch implements driver.Subscription.ReceiveBatch.
 func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) (msgs []*driver.Message, er error) {
@@ -475,7 +474,10 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) (msgs 
 		ms = append(ms, m2)
 	}
 	if len(ms) == 0 {
-		time.Sleep(pollDuration)
+		// When we return no messages and no error, the portable type will call
+		// ReceiveBatch again immediately. Sleep for a bit to avoid hammering SQS
+		// with RPCs.
+		time.Sleep(noMessagesPollDuration)
 	}
 	return ms, nil
 }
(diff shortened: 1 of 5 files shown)


pubsub/awspubsub/awspubsub.go
```diff
@@ -402,6 +404,11 @@ func openSubscription(ctx context.Context, client *sqs.SQS, qURL string) driver.
 	return &subscription{client: client, qURL: qURL}
 }
 
+const (
```
jba: No need for parens.

pubsub/awspubsub/awspubsub.go
```diff
@@ -402,6 +404,11 @@ func openSubscription(ctx context.Context, client *sqs.SQS, qURL string) driver.
...
+	// How often ReceiveBatch should poll if no messages are available.
```
jba: Comment and name are not quite accurate. It's more like how long ReceiveBatch should wait before returning no messages. The polling is done in pubsub.Receive.

pubsub/rabbitpubsub/rabbit.go
```diff
@@ -592,8 +592,11 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*dr
 			}
 
 		case <-maxTime:
-			// Timed out. Return whatever we have.
-			return ms, nil
+			// Timed out. Return whatever we have, or continue waiting if
```
jba: It looks to me like this was right before — wait a little bit if there are no messages, then return — and you changed it to wait indefinitely until it gets a message. In other words, the change here is nearly the opposite of the one you made to mempubsub. 

You don't seem to mention this change in your commit description, so I'm not sure what your intent was.

pubsub/awspubsub/awspubsub.go
vangent: Done.

pubsub/awspubsub/awspubsub.go
vangent: Better?

pubsub/rabbitpubsub/rabbit.go
vangent: I believe the behavior is the same before this change and after it.

Before this change, it would return and `pubsub.go` would just call it right back and it will re-enter the `for` loop.

After this change, it will just re-enter the `for` loop directly.

This seems simpler to me, but I don't feel strongly about it, so I'll put it back if you prefer.

It's different than `mempubsub` because it already has a `for` loop, so it's easy to continue looping; by returning, `mempubsub` doesn't need the `for` loop at all.

PR #1579, opened 2019-03-15
internal/website: create new layout

### LINKED ISSUE #1578: internal/website: restructure layout to support more pages

#1469, #1576, and #1577 all propose introducing new documentation pages to gocloud.dev. The current Hyde theme is optimized around a sidebar of ~5 items, whereas we may quickly get to double-digit items. 

I propose the site layout should reduce the size of the sidebar and organize the pages using the same headings as used in the Google Cloud Platform documentation:

- Quickstarts
- How-to guides
- Reference
- Concepts

(See the [Google Cloud Storage docs](https://cloud.google.com/storage/docs/) for an example.)

We could keep the Hyde theme and show only the top-level sections. However, this forces the user to navigate a lot in order to scan the contents of the site, which I don't expect to be so large at this point that a reader couldn't scan it readily with some grouping.

zombiezen:
-  Deleted Hyde theme
-  Added from-scratch layout based on brand book elements
-  Renamed pages section to ref (for reference). Added aliases to avoid breaking existing links.
-  Added footer with contact information and other links

Fixes #1578

/cc @spf13 if you have thoughts from brand/design perspective

## Screenshots

<details>
<b>Laptop:</b>

... snip ...

<b>Mobile:</b>

... snip ...

</details>


internal/website/content/ref/blob.md
```diff
@@ -1,5 +1,8 @@
 ---
 title: "Blob"
+date: 2019-02-21T16:21:29-08:00
```
eliben: Just a question: is it necessary to specify these dates in such precision, or at all? What do they really mean for reference pages? Is this for ordering?

internal/website/content/ref/runtimevar.md
```diff
@@ -13,7 +18,7 @@ Top-level package documentation: https://godoc.org/gocloud.dev/runtimevar
 * [GCP Runtime
   Configurator](https://godoc.org/gocloud.dev/runtimevar/runtimeconfigurator)
 * [blobvar](https://godoc.org/gocloud.dev/runtimevar/blobvar) - a blob-backed
-  implementation supported by any provider that has [blob support]({{< relref "/pages/blob.md#supported-providers">}})
+  implementation supported by any provider that has [blob support]({{< relref "blob.md#supported-providers">}})
```
eliben: s/implementation/Implementation/?

internal/website/content/ref/blob.md
zombiezen: Precision not required, but comes default with Hugo. It's reflected in the RSS feed and Sitemap automatically generated for the site, which currently read "Jan 1 0001", which is bad. 🙂

internal/website/content/ref/runtimevar.md
zombiezen: Part of a larger sentence, seems correct as-is?

internal/website/content/ref/blob.md
eliben: So should we update this date/time whenever we update a page? Or post it initially? This would be good to document somewhere.

internal/website/content/ref/runtimevar.md
eliben: Sorry, missed that!

internal/website/content/ref/blob.md
zombiezen: See [Front Matter]. If you use `hugo new`, `date` is added for you. `lastmod` should be tweaked when pages get substantially new information.

Added to internal/website/README.md

Issue #1441, opened 2019-03-04
pubsub: portable type performance improvements for Receive


vangent:
#691 tracked one performance improvement for `pubsub`: dynamically setting the batch size for `Receive`.

Other ideas that were discussed on that issue included:

* Currently, we don't call `ReceiveBatch` until the message queue is empty. We could make the call earlier to avoid periods where there aren't any messages for the application to process. TBD: what does "earlier" mean?

* We could call `ReceiveBatch` concurrently. Or (possibly preferably) we could modify the driver interface to indicate that we WON'T do that, if we find that we don't need to for good performance.

ijt:
Why is this P1? Has anyone reported this as a problem?

vangent:
I've made a bunch of performance improvements attached to this issue and others. Closing.

PR #1585, opened 2019-03-16
pubsub: add a test that can be used to characterize Receive performance

vangent:
Sample output with the default settings:

```
$ go test --run TestReceivePerformance
Elapsed (sec)	Msgs/sec	Max Messages
0.000081	12318.001527	1
0.000223	7067487.437541	1000
1.000242	751985.161829	1000
2.000659	792669.396619	1000
3.001288	810490.673882	1000
4.001931	803483.069525	1000
5.002907	787231.494381	1000
6.003415	735626.551089	1000
7.003970	801555.034357	1000
8.005067	805116.952557	1000
9.005451	701730.566339	1000
PASS
ok  	gocloud.dev/pubsub	10.097s
```

Updates #1441.

My plan is to use this to experiment with different settings in pubsub.go. For example:

* What happens if `ReceiveBatch` takes 250ms every time?
* What happens if `ReceiveBatch` is fast but returns fewer than `maxMessages`?
* What happens if the application never calls `Ack`?
* What happens if message processing is very slow?
* What happens if message processing usually fast but occasionally slow?

### DIFF
--- a/pubsub/benchmark_test.go
+++ b/pubsub/benchmark_test.go
@@ -71,7 +71,7 @@ func TestReceivePerformance(t *testing.T) {
 	t.Skip("Skipping by default")
 
 	msgs := make([]*driver.Message, 1000)
-	for i := 0; i < 1000; i++ {
+	for i := range msgs {
 		msgs[i] = &driver.Message{}
 	}
 	sub := newSubscription(&fakeSub{msgs: msgs}, nil)
--- a/pubsub/benchmark_test.go
+++ b/pubsub/benchmark_test.go
@@ -1,0 +1,137 @@
+// Copyright 2019 The Go Cloud Development Kit Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pubsub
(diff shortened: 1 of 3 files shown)


vangent:
I'm probably going to end up turning this test into a suite of profiles; the test will run all the profiles and the output will be extended to include a column with the profile name.

Then I'll make a Sheet with 2 sheets, so you can easily do an A/B test by importing the output with A into sheet 1, output with B into sheet 2, and see graphs comparing them for each profile.

pubsub/benchmark_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+// TestReceivePerformance enables characterization of Receive under various
+// situations. Tune the const and var parameters above, disable the Skip,
+// and the test will output a series of timepoints with throughput and the
+// batch size being sent to ReceiveBatch.
+func TestReceivePerformance(t *testing.T) {
+	t.Skip("Skipping by default")
+
+	msgs := make([]*driver.Message, 1000)
+	for i := 0; i < 1000; i++ {
```
jba: for i := range msgs

pubsub/benchmark_test.go
```diff
@@ -0,0 +1,137 @@
...
+	for i := 0; i < nGoRoutines; i++ {
+		grp.Go(func() error {
+			// Each goroutine loops until ctx is canceled.
+			for {
+				m, err := sub.Receive(ctx)
+				if err == context.Canceled {
+					return nil
+				}
+				// TODO(rvangent): This is annoying. Maybe retry should return
+				// context.Canceled instead of wrapping if it never retried?
```
jba: Then wouldn't callers of retry have to check for both kinds of errors?

pubsub/benchmark_test.go
vangent: Done.

pubsub/benchmark_test.go
vangent: Filed #1591 to discuss.

PR #1592, opened 2019-03-18
all/travis: gofmt -s on bindata.go, and remove special case for it in gofmt check

vangent:


vangent:
> Any reason to run it on Mac/windows in addition to Linux?

I don't think so.

eliben:
> > Any reason to run it on Mac/windows in addition to Linux?
> 
> I don't think so.

Will you move it to after the OS check, where the dep checks are?

vangent:
> Will you move it to after the OS check, where the dep checks are?

Ah, done.

PR #1593, opened 2019-03-18
all: fix go vet warnings

vangent:
The remaining warnings are:

```
# gocloud.dev/internal/docstore/internal/fields
internal/docstore/internal/fields/fields_test.go:186:2: struct field unexported repeats json tag "tag" also at fields_test.go:184
internal/docstore/internal/fields/fields_test.go:186:2: struct field unexported has json tag but is not exported
internal/docstore/internal/fields/fields_test.go:192:2: struct field Z repeats json tag "Dup2" also at fields_test.go:201
```

I assume that these are on purpose?


blob/oc_test.go
```diff
@@ -55,11 +55,11 @@ func TestOpenCensus(t *testing.T) {
 	const provider = "gocloud.dev/blob/memblob"
 
 	diff := octest.Diff(te.Spans(), te.Counts(), "gocloud.dev/blob", provider, []octest.Call{
-		{"NewWriter", gcerrors.OK},
-		{"NewRangeReader", gcerrors.OK},
-		{"Attributes", gcerrors.OK},
-		{"Delete", gcerrors.OK},
-		{"NewRangeReader", gcerrors.NotFound},
+		{Method: "NewWriter", Code: gcerrors.OK},
```
jba: We control `octest`, so it shouldn't be necessary to provide struct literal keys. This is a case of vet being too picky.

internal/docstore/docstore.go
```diff
@@ -218,7 +218,7 @@ func (a *Action) toDriverAction() (*driver.Action, error) {
 			if err != nil {
 				return nil, err
 			}
-			d.Mods = append(d.Mods, driver.Mod{fp, v})
+			d.Mods = append(d.Mods, driver.Mod{FieldPath: fp, Value: v})
```
jba: ditto

internal/docstore/firedocstore/codec.go
```diff
@@ -70,31 +70,41 @@ type encoder struct {
...
 var nullValue = &pb.Value{ValueType: &pb.Value_NullValue{}}
 
-func (e *encoder) EncodeNil()            { e.pv = nullValue }
-func (e *encoder) EncodeBool(x bool)     { e.pv = &pb.Value{ValueType: &pb.Value_BooleanValue{x}} }
-func (e *encoder) EncodeInt(x int64)     { e.pv = &pb.Value{ValueType: &pb.Value_IntegerValue{x}} }
-func (e *encoder) EncodeUint(x uint64)   { e.pv = &pb.Value{ValueType: &pb.Value_IntegerValue{int64(x)}} }
-func (e *encoder) EncodeBytes(x []byte)  { e.pv = &pb.Value{ValueType: &pb.Value_BytesValue{x}} }
+func (e *encoder) EncodeNil() { e.pv = nullValue }
+func (e *encoder) EncodeBool(x bool) {
+	e.pv = &pb.Value{ValueType: &pb.Value_BooleanValue{BooleanValue: x}}
```
jba: This is a known vet bug/misfeature: proto oneof structs will only ever have one field by design, so this isn't necessary.

internal/docstore/firedocstore/fs.go
```diff
@@ -253,7 +253,7 @@ func (c *collection) actionToWrites(a *driver.Action) ([]*pb.Write, string, erro
 			docName = driver.UniqueString()
 			newName = docName
 		}
-		w, err = c.putWrite(a.Doc, docName, &pb.Precondition{ConditionType: &pb.Precondition_Exists{false}})
+		w, err = c.putWrite(a.Doc, docName, &pb.Precondition{ConditionType: &pb.Precondition_Exists{Exists: false}})
```
jba: same oneof issue.

internal/docstore/firedocstore/fs.go
```diff
@@ -299,7 +299,7 @@ func (c *collection) putWrite(doc driver.Document, docName string, pc *pb.Precon
 	}
 	pdoc.Name = c.collPath + "/" + docName
 	return &pb.Write{
-		Operation:       &pb.Write_Update{pdoc},
+		Operation:       &pb.Write_Update{Update: pdoc},
```
jba: This too.

blob/oc_test.go
vangent: Is there a bug open for `go vet` about this? If they are going to fix it, fine, but if not I'd rather err on the side of making `go vet` happy unless there's a real reason not to.

vangent: Especially since these existing warnings mask real ones; for example, I didn't notice the `Example`-related govet warnings in my editor because of all of the `octest` ones.

jba: The reason not to is that it leads to less readable code. I agree the example errors are useful, but most of these make the code worse.
Let's do what Jean did for the cloud.google.com/go repo, and grep out the messages we don't want. See https://code.googlesource.com/gocloud/+/refs/heads/master/internal/kokoro/vet.sh. 

Or we could run `go vet -composites=false`, since all the false positives are due to that one check.

vangent: Honestly, I don't find the code less readable.

I'm not planning on adding a Travis check based on `go vet`, although I guess we could. This is mostly so that I see warnings when I run it locally (my editor does it for me). I'd really prefer not to have to figure out how to configure it to run something other than the default.

If you think these warnings are invalid, that's a discussion for `go vet` owners; I poked around a bit and found https://github.com/golang/go/issues/6820 which seems to imply it's unlikely to be changed.

For example, what happens if/when we start supporting drivers in other repos; we'll likely have to make `internal/octest` public so that others can use it. Then changes to the struct could have compatibility problems.

vangent: @zombiezen 	also had some reasons why the "composites" check is legit.

zombiezen: @jba is aware of all of them AFAIK. Protobufs potentially changing internal layout was one that plagued the cloud.google.com/go repo. Crossing packages, even if we're in control of both packages, can be a bit dicey. If it really affects readability, then we should add a constructor function that builds the shorter parameter list.

jba:
> I assume that these are on purpose?

Yes.,

internal/docstore/firedocstore/codec.go
jba: @dsnet What's the current status of `go vet` vs. proto oneof structs? Is there an issue somewhere that discusses this?

dsnet: I'm not aware of any discussion about this, but your earlier statement that "oneof structs will only ever have one field by design" is pretty much true.

jba: @vangent can you remove the struct keys from the oneofs? I'm okay with everything else.

vangent: Done. I left the `&pb.Precondition_Exists{Exists: true}`, let me know if you think I should drop those too.

vangent:
PTAL

PR #1594, opened 2019-03-18
docstore: document revisions

jba:
Support optimistic locking (safe read-modify-write cycles) by supporting
a revision field in documents.

Like an ETag, a revision field uniquely identifies a document
version. The value of the field changes every time the document is
modified.

If a document with a revision field is passed to a Docstore write
action, then the action will return a PreconditionFailed error if the
stored document's revision field doesn't match.

Since providers use different types of values for revisions, docstore
doesn't specify a single type for the field.

Implemented for firedocstore and memdocstore.

Skipping dynamodocstore so shantuo can implement.


internal/docstore/memdocstore/mem.go
```diff
@@ -112,16 +116,28 @@ func (c *collection) runAction(a *driver.Action) error {
 		fallthrough
 
 	case driver.Replace, driver.Put:
+		if a.Kind != driver.Create {
```
shantuo: This `if` can be removed.

internal/docstore/memdocstore/mem.go
```diff
@@ -160,6 +176,33 @@ func (c *collection) update(doc map[string]interface{}, mods []driver.Mod) error
 			_ = setAtFieldPath(doc, m.FieldPath, m.Value)
 		}
 	}
+	if len(mods) > 0 {
+		c.changeRevision(doc)
```
shantuo: What's the relationship between `c.nextRevision` and `c.docs[key][docstore.RevisionField]`? It seems the collection's revision is changed along with the the docs' revision when doing replace and put, should the docs' revision also be changed here during update?

internal/docstore/docstore.go
```diff
@@ -100,20 +109,29 @@ func (l *ActionList) Create(doc Document) *ActionList {
 // The key fields must be set.
 // The document must already exist; an error for which gcerrors.Code returns NotFound
 // is returned if it does not.
+// If the document has a non-nil RevisionField, then an error for which gcerrors.Code
```
shantuo: In general, I think these comments should be left off from individual APIs, maybe it is enough to add those to the package doc, WDYT?

internal/docstore/docstore.go
```diff
@@ -50,6 +50,15 @@ func newCollection(d driver.Collection) *Collection {
 	return &Collection{driver: d}
 }
 
+// RevisionField is the name of the document field used for document revision
```
vangent: Don't some providers have explicit support for this? What's the story for those that do?

internal/docstore/docstore.go
```diff
@@ -50,6 +50,15 @@ func newCollection(d driver.Collection) *Collection {
...
+// information, to implement optimistic locking.
+// Every retrieved document will have this field set to a non-nil value
+// of unspecified type (different providers might use different types).
+// When defining a struct for storing docstore data, define the field to be of type
```
vangent: What happens if the user doesn't do this?

internal/docstore/dynamodocstore/testdata/TestConformance/Create.yaml
```diff
@@ -14,7 +14,7 @@ interactions:
       User-Agent:
       - aws-sdk-go/1.16.23 (go1.12; linux; amd64)
       X-Amz-Date:
-      - 20190311T215347Z
+      - 20190318T173857Z
```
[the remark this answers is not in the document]
jba: Done.

internal/docstore/memdocstore/mem.go
jba: Done.

internal/docstore/memdocstore/mem.go
jba: `c.nextRevision` is just a collection-wide counter, to make sure every write gets a unique value.

My intent is that in all three cases (replace, put and update), the stored document's revision (value of its RevisionField) is changed to a new value. Is that not happening somewhere?

internal/docstore/docstore.go

internal/docstore/docstore.go
jba: Many do. Drivers are obligated to follow these semantics, as verified in the new conformance tests in this PR. If you look at the changes to `fs.go` you'll see what that looks like in general: copying the native revision field into the returned document in a `Get`, and enforcing the necessary preconditions on writes.

internal/docstore/docstore.go
jba: Clarified in the text (which has moved to the top of the file).

jba:
PTAL.

internal/docstore/memdocstore/mem.go
shantuo: Ok, got it. No, I think I must have missed something, this looks correct.

jba:
@vangent ping

PR #1597, opened 2019-03-18
pubsub/tests: add a suite of benchmarks

vangent:
Instead of running a single benchmark, the test now runs a suite of them.

[Here's](https://docs.google.com/spreadsheets/d/1c3mqHJ33mN64xBQrx42DY1GpYrMUi7IEl-6FRawIv70/edit) the sample spreadsheet I have set up with the results. `out` is the output of the benchmark suite with HEAD. `out2` is the output of the benchmark suite with a single change, to modify `pubsub.desiredQueueDuration` from 2s to 50ms. Each row of the graph corresponds to a specific benchmark profile; the current columns are messages/sec, RPCs/sec, and batch size (== `maxMessages`).

So, for example, you can see that reducing desiredQueueDuration would increase RPCs/sec significantly in various scenarios.

Feedback on the spreadsheet and the specific tests in the suite is welcome!

Updates #1441.

### DIFF
--- a/pubsub/benchmark_test.go
+++ b/pubsub/benchmark_test.go
@@ -24,6 +24,7 @@ import (
 	"gocloud.dev/internal/retry"
 	"gocloud.dev/pubsub/driver"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/xerrors"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


jba:
Nice graphs.

Clearly we have to fix the no-ack case.

Do you understand why the 50ms behaves so poorly in the last graph?

This way of looking at it is valuable, but I'd suggest another way. A user is likely to have a target desired throughput, and will keep adding goroutines until they reach it. So you could ask how many goroutines you need to get N msgs/sec.

vangent:
> Clearly we have to fix the no-ack case.

Yup, working on it.

> Do you understand why the 50ms behaves so poorly in the last graph?

No.

> This way of looking at it is valuable, but I'd suggest another way. A user is likely to have a target desired throughput, and will keep adding goroutines until they reach it. So you could ask how many goroutines you need to get N msgs/sec.

That is complicated by the fact that it may not be possible to reach N msgs/sec just by adding goroutines, especially with a cap on maxMessages. I'll keep thinking about this though.

PR #1599, opened 2019-03-18
pubsub/azurepubsub: do more sanity checks in constructors; only do subscription-existence check once

### LINKED ISSUE #1569: pubsub/azure: clean up testSBSubscription
`testSBSubscription` is called from `ReceiveBatch`, I think some of these checks should move to the constructor and the expensive one should only happen once.

vangent:
The main change here is to only check for existence of the Subscription once, instead of once per `ReceiveBatch`. I also moved some other verification (e.g., that parameters are not `nil`) into the constructors.

I also add the benchmark harness so you can run:

```
$ go test -v -bench . ./... --run Benchmark
```

Finally, I tried to run the conformance tests in `servicebus.SubscriptionWithReceiveAndDelete` mode, but they do not pass. It looks like the semantics are not what we expect -- that options seems to delete the message even if there are multiple subscribers. I commented out the new test for now, and will open a separate issue to discuss.

Fixes #1569.

### DIFF
--- a/pubsub/azurepubsub/azurepubsub.go
+++ b/pubsub/azurepubsub/azurepubsub.go
@@ -227,7 +227,7 @@ func OpenTopic(ctx context.Context, sbTopic *servicebus.Topic, opts *TopicOption
 
 // openTopic returns the driver for OpenTopic. This function exists so the test
 // harness can get the driver interface implementation if it needs to.
-func openTopic(ctx context.Context, sbTopic *servicebus.Topic, opts *TopicOptions) (driver.Topic, error) {
+func openTopic(ctx context.Context, sbTopic *servicebus.Topic, _ *TopicOptions) (driver.Topic, error) {
 	if sbTopic == nil {
 		return nil, errors.New("azurepubsub: OpenTopic requires a Service Bus Topic")
 	}
--- a/pubsub/azurepubsub/azurepubsub.go
+++ b/pubsub/azurepubsub/azurepubsub.go
@@ -483,7 +483,6 @@ func (s *subscription) SendAcks(ctx context.Context, ids []driver.AckID) error {
 		"disposition-status": completedStatus,
 		"lock-tokens":        lockIds,
 	}
-
 	msg := &amqp.Message{
 		ApplicationProperties: map[string]interface{}{
 			"operation": "com.microsoft:update-disposition",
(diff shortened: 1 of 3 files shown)


pubsub/azurepubsub/azurepubsub.go
```diff
@@ -217,17 +217,21 @@ func NewSubscription(parentTopic *servicebus.Topic, subscriptionName string, opt
...
+	}
+	return pubsub.NewTopic(t, nil), nil
 }
 
 // openTopic returns the driver for OpenTopic. This function exists so the test
 // harness can get the driver interface implementation if it needs to.
-func openTopic(ctx context.Context, sbTopic *servicebus.Topic) driver.Topic {
-	return &topic{
-		sbTopic: sbTopic,
+func openTopic(ctx context.Context, sbTopic *servicebus.Topic, opts *TopicOptions) (driver.Topic, error) {
```
jba: s/opts/_/

pubsub/azurepubsub/azurepubsub.go
```diff
@@ -301,47 +307,38 @@ type SubscriptionOptions struct {
...
 }
 
 // openSubscription returns a driver.Subscription.
-func openSubscription(ctx context.Context, sbNs *servicebus.Namespace, sbTop *servicebus.Topic, sbSub *servicebus.Subscription, opts *SubscriptionOptions) driver.Subscription {
-	topicName := ""
-
-	if sbTop != nil {
-		topicName = sbTop.Name
+func openSubscription(ctx context.Context, sbNs *servicebus.Namespace, sbTop *servicebus.Topic, sbSub *servicebus.Subscription, opts *SubscriptionOptions) (driver.Subscription, error) {
+	if sbNs == nil {
```
jba: We really don't want to make a habit of these, or our code will be littered with them. Is there a particular reason you want to check here?

vangent: Mostly because using the `URLOpener` makes it easier to accidentally omit arguments (e.g., by using `URLOpener`) and I think a more specific error message will make it easier to figure out when went wrong.

pubsub/azurepubsub/azurepubsub.go
vangent: Done.

PR #1606, opened 2019-03-19
pubsub/tests: change benchmark to report on a ticker instead of callback

vangent:
Previously, having the lines emitted as part of the callbook hook meant that runs of the test would result in different numbers of output points, which makes the results difficult to graph. Now it consistently emits a line every 25ms for the first half second, and every 500ms after that.

Also, I changed the single callback hook into 2, one that is called before `ReceiveBatch` and one after; this gives clearer results for situations where the simulated `ReceiveBatch` call takes a long time.

I also now run each profile as a subtest to make it easier to run a single one.

Updates #1441.

### DIFF
--- a/pubsub/benchmark_test.go
+++ b/pubsub/benchmark_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pubsub/benchmark_test.go
```diff
@@ -57,7 +61,7 @@ func (s *fakeSub) AckFunc() func() { return func() {} }
 // TestReceivePerformance enables characterization of Receive under various
 // situations, characterized in "tests" below.
 func TestReceivePerformance(t *testing.T) {
-	//t.Skip("Skipped by default")
+	t.Skip("Skipped by default")
```
jba: You could put it in a separate file with a build tag, too. No preference at this time.

pubsub/pubsub.go
```diff
@@ -424,9 +425,12 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
...
 		// The only way here is if s.waitc was nil. This goroutine just set
 		// s.waitc to non-nil while holding the lock.
+		if s.preReceiveBatchHook != nil {
+			s.preReceiveBatchHook(nMessages)
+		}
 		msgs, err := s.getNextBatch(ctx, nMessages)
-		if s.onReceiveBatchHook != nil {
-			s.onReceiveBatchHook(len(msgs), nMessages)
+		if s.postReceiveBatchHook != nil {
+			s.postReceiveBatchHook(len(msgs))
```
jba: Maybe this should take the error too, so it can decide what it wants to do if err != nil.

pubsub/benchmark_test.go
vangent: Ack.

pubsub/pubsub.go
vangent: I'll add it if needed.

PR #1609, opened 2019-03-19
pubsub/all: add a constructor arg to disable dynamic batching

vangent:
This adds an argument to `pubsub.NewSubscription` that disables dynamic sizing of the `maxMessages` passed to `driver.ReceiveBatch`, and sets it to true everywhere except in `drivertest`, which needs to support record/replay tests for which the dynamism is a problem.

I also changed the default/initial value from 1 to 2; I can revert if you don't agree.

The constructor being changed (`NewSubscription`) is public, so this is technically a breaking change, but it's only expected to be used in drivers (and it's hidden in the public godoc).


pubsub/pubsub.go
```diff
@@ -406,8 +408,8 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
 		// Ask for a number of messages that will give us the desired queue length.
 		// Unless we don't have information about process time (at the beginning), in
 		// which case just get one message.
-		nMessages := 1
-		if s.avgProcessTime > 0 {
+		nMessages := 2
```
jba: Some subscribers may really only be able to handle one message at a time. For example, the message might be a video than requires transcoding, and that process may take minutes and use all CPUs. If we download a second message, that message will always expire.

If you want to start at 2 for testing, I'm fine with that.

vangent: Yea, good point. Done.

vangent:
PTAL

PR #1614, opened 2019-03-20
runtimevar: support decrypt/decoder in URL

### LINKED ISSUE #1290: runtimevar: add a DecryptDecoder for encrypted variables
The decoder should decrypt the result and also be able to pass to another decode (e.g. json decoder)

shantuo:
BREAK_CHANGE: add `context.Context` to `runtimevar.DecoderByName`.

The keeper URL is registered using environment variable.

Fixes #1290

### DIFF
--- a/runtimevar/filevar/filevar_test.go
+++ b/runtimevar/filevar/filevar_test.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -240,6 +241,8 @@ func TestOpenVariable(t *testing.T) {
 		{"file://" + secretsPath + "?decoder=decrypt+bytes", false, false, []byte(`{"Foo":"Bar"}`)},
 		// Working example with decrypt+json decoder.
 		{"file://" + secretsPath + "?decoder=decrypt+jsonmap", false, false, &map[string]interface{}{"Foo": "Bar"}},
+		// Working example with escaped decrypt+json decoder
+		{"file://" + secretsPath + "?decoder=" + url.QueryEscape("decrypt+jsonmap"), false, false, &map[string]interface{}{"Foo": "Bar"}},
 	}
 
 	for _, test := range tests {
--- a/runtimevar/filevar/filevar_test.go
+++ b/runtimevar/filevar/filevar_test.go
@@ -237,9 +237,9 @@ func TestOpenVariable(t *testing.T) {
 		// Working example with decrypt (default) decoder.
 		{"file://" + secretsPath + "?decoder=decrypt", false, false, []byte(`{"Foo":"Bar"}`)},
 		// Working example with decrypt+bytes decoder.
-		{"file://" + secretsPath + "?decoder=decrypt/bytes", false, false, []byte(`{"Foo":"Bar"}`)},
+		{"file://" + secretsPath + "?decoder=decrypt+bytes", false, false, []byte(`{"Foo":"Bar"}`)},
 		// Working example with decrypt+json decoder.
-		{"file://" + secretsPath + "?decoder=decrypt/jsonmap", false, false, &map[string]interface{}{"Foo": "Bar"}},
+		{"file://" + secretsPath + "?decoder=decrypt+jsonmap", false, false, &map[string]interface{}{"Foo": "Bar"}},
 	}
 
 	for _, test := range tests {
--- a/runtimevar/filevar/filevar_test.go
+++ b/runtimevar/filevar/filevar_test.go
@@ -17,7 +17,6 @@ package filevar
 import (
 	"context"
 	"errors"
-	"github.com/google/go-cmp/cmp"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -25,9 +24,13 @@ import (
 	"testing"
 	"time"
 
+	"gocloud.dev/secrets"
+
+	"github.com/google/go-cmp/cmp"
 	"gocloud.dev/runtimevar"
 	"gocloud.dev/runtimevar/driver"
 	"gocloud.dev/runtimevar/drivertest"
+	_ "gocloud.dev/secrets/localsecrets"
 )
 
 type harness struct {
@@ -178,6 +181,8 @@ func TestOpenVariable(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer os.RemoveAll(dir)
+
 	jsonPath := filepath.Join(dir, "myvar.json")
 	if err := ioutil.WriteFile(jsonPath, []byte(`{"Foo": "Bar"}`), 0666); err != nil {
 		t.Fatal(err)
@@ -187,7 +192,13 @@ func TestOpenVariable(t *testing.T) {
 		t.Fatal(err)
 	}
 	nonexistentPath := filepath.Join(dir, "filenotfound")
-	defer os.RemoveAll(dir)
+	ctx := context.Background()
+	secretsPath := filepath.Join(dir, "mysecret.txt")
+	cleanup, err := setupTestSecrets(ctx, dir, secretsPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cleanup()
 
 	// Convert paths to a URL path, adding a leading "/" if needed on Windows
 	// (on Unix, dirpath already has a leading "/").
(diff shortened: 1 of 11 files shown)


runtimevar/awsparamstore/paramstore_test.go
```diff
@@ -173,13 +173,26 @@ func TestOpenVariable(t *testing.T) {
...
+
 	for _, test := range tests {
 		_, err := runtimevar.OpenVariable(ctx, test.URL)
 		if (err != nil) != test.WantErr {
 			t.Errorf("%s: got error %v, want error %v", test.URL, err, test.WantErr)
 		}
 	}
 }
+
+func setupTestSecrets(ctx context.Context) func() {
```
vangent: Instead of doing this full test in `awsparamstore`, how about doing it in `constantvar` or `filevar` (with `localsecrets`) so that you can verify the output as well?

vangent: Oh, I see you did do it in filevar too; I would just remove from here then.

runtimevar/filevar/filevar_test.go
```diff
@@ -17,17 +17,20 @@ package filevar
...
-	"github.com/google/go-cmp/cmp"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 	"testing"
 	"time"
 
+	"gocloud.dev/secrets"
+
```
vangent: Nit: remove blank line and sort.

runtimevar/filevar/filevar_test.go
```diff
@@ -247,3 +259,24 @@ func TestOpenVariable(t *testing.T) {
 		}
 	}
 }
+
+func setupTestSecrets(ctx context.Context, dir, secretsPath string) (func(), error) {
+	keeperEnv := "RUNTIMEVAR_KEEPER_URL"
```
vangent: Nit: const

runtimevar/gcpruntimeconfig/runtimeconfigurator_test.go
```diff
@@ -248,13 +250,26 @@ func TestOpenVariable(t *testing.T) {
...
+
 	for _, test := range tests {
 		_, err := runtimevar.OpenVariable(ctx, test.URL)
 		if (err != nil) != test.WantErr {
 			t.Errorf("%s: got error %v, want error %v", test.URL, err, test.WantErr)
 		}
 	}
 }
+
+func setupTestSecrets(ctx context.Context) func() {
```
vangent: Same here, I think this is overkill for testing.

runtimevar/runtimevar.go
```diff
@@ -510,20 +512,42 @@ func DecryptDecode(ctx context.Context, k *secrets.Keeper, post Decode) Decode {
...
+func DecoderByName(ctx context.Context, decoderName string, dflt *Decoder) (*Decoder, error) {
+	var post *Decoder
 	if dflt == nil {
-		dflt = BytesDecoder
+		post = BytesDecoder
+	}
+	var k *secrets.Keeper
+	var err error
+	if strings.HasPrefix(decoderName, "decrypt") {
+		if k, err = secrets.OpenKeeper(ctx, os.Getenv("RUNTIMEVAR_KEEPER_URL")); err != nil {
```
vangent: I suggest giving an explicit error here if `RUNTIMEVAR_KEEPER_URL` is not set.

runtimevar/runtimevar.go
```diff
@@ -510,20 +512,42 @@ func DecryptDecode(ctx context.Context, k *secrets.Keeper, post Decode) Decode {
...
+			return nil, err
+		}
+		if decoderName == "decrypt" {
+			decoderName = ""
+		} else {
+			decoderName = strings.TrimPrefix(decoderName, "decrypt/")
```
vangent: "/" is not a valid character in query parameters and would need to be escaped. How about using "+" instead? E.g., `decrypt+bytes`.

runtimevar/runtimevar.go
```diff
@@ -510,20 +512,42 @@ func DecryptDecode(ctx context.Context, k *secrets.Keeper, post Decode) Decode {
...
 	case "bytes":
-		return BytesDecoder, nil
+		post = BytesDecoder
 	case "jsonmap":
 		var m map[string]interface{}
-		return NewDecoder(&m, JSONDecode), nil
+		post = NewDecoder(&m, JSONDecode)
 	case "string":
-		return StringDecoder, nil
+		post = StringDecoder
```
vangent: This took me a few times to read; the change for Keeper has affected the whole function. Could it be more localized with something like this?

```
// First part is unchanged from before.
switch {
  case decoderName == "": ...
  case decoderName == "bytes": ...
  ...
  case strings.StartsWith(decoderName, "decrypt"):
     post := strings.TrimPrefix(decoderName[7:], "+")
     postDecoder, err := DecoderByName(ctx, post, dflt)
     if err != nil { return nil, err }
     // Get env var, etc. as you have.
  }
// Rest is same as before.
```

runtimevar/runtimevar.go
```diff
@@ -510,20 +512,42 @@ func DecryptDecode(ctx context.Context, k *secrets.Keeper, post Decode) Decode {
 //   - "jsonmap": Returns a JSON decoder for a map[string]interface{};
 //       Snapshot.Value will be of type *map[string]interface{}.
 //   - "string": Returns StringDecoder; Snapshot.Value will be of type string.
-func DecoderByName(decoderName string, dflt *Decoder) (*Decoder, error) {
+// It also supports using "decrypt/<decoderName>" (or "decrypt" for default
+// decoder) to decrypt the data before decoding. It uses the secrets package to
+// open a keeper by the URL string stored in a envrionment variable
+// "RUNTIMEVAR_KEEPER_URL". See https://godoc.org/gocloud.dev/secrets#OpenKeeper
+// for more details.
+func DecoderByName(ctx context.Context, decoderName string, dflt *Decoder) (*Decoder, error) {
```
vangent: BTW, your change *is* a breaking change (as you noted); it's not failing the Travis checks because I accidentally checked in something from testing (see #1612 for me fixing it; once I merge that, if you merge it into this PR, the checks would fail unless you have a git commit with `BREAKING_CHANGE_OK` in it).

This is fine, especially because we don't intend external users to use this function (it is exported for user by drivers). WDYT about changing it to use the same trick we used for `New` above, i.e., having it exported as a variable instead of a function? The functionality is the same but it hides the signature in godoc.

runtimevar/runtimevar.go
shantuo: Oh, I missed the recursion here.

runtimevar/runtimevar.go
```diff
@@ -510,20 +514,60 @@ func DecryptDecode(ctx context.Context, k *secrets.Keeper, post Decode) Decode {
 //   - "jsonmap": Returns a JSON decoder for a map[string]interface{};
 //       Snapshot.Value will be of type *map[string]interface{}.
 //   - "string": Returns StringDecoder; Snapshot.Value will be of type string.
-func DecoderByName(decoderName string, dflt *Decoder) (*Decoder, error) {
+// It also supports using "decrypt/<decoderName>" (or "decrypt" for default
```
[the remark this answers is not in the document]
shantuo: Done.

runtimevar/runtimevar.go
```diff
@@ -510,20 +514,60 @@ func DecryptDecode(ctx context.Context, k *secrets.Keeper, post Decode) Decode {
...
+
+// decryptByName returns a *secrets.Keeper for decryption when decoderName
+// contains "decrypt".
+func decryptByName(ctx context.Context, decoderName string) (*secrets.Keeper, string, error) {
+	if !strings.HasPrefix(decoderName, "decrypt") {
+		return nil, decoderName, nil
+	}
+	keeperURL := os.Getenv("RUNTIMEVAR_KEEPER_URL")
+	if keeperURL == "" {
+		return nil, "", errors.New("environment varialbe RUNTIMEVAR_KEEPER_URL needed to open a *secrets.Keeper for decryption")
```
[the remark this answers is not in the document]
shantuo: Done.

runtimevar/runtimevar.go
```diff
@@ -510,20 +514,60 @@ func DecryptDecode(ctx context.Context, k *secrets.Keeper, post Decode) Decode {
...
+	}
+	k, err := secrets.OpenKeeper(ctx, keeperURL)
+	if err != nil {
+		return nil, "", err
+	}
+	if decoderName == "decrypt" {
+		return k, "", nil
+	}
+	// The parsed value is "decrypt <decoderName>".
+	return k, strings.TrimPrefix(decoderName, "decrypt "), nil
```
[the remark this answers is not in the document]
shantuo: Done.

PR #1616, opened 2019-03-20
docs: update squash and release instructions for breaking changes

vangent:


CONTRIBUTING.md
```diff
@@ -182,5 +182,12 @@ When squashing-and-merging:
 -   Do a final review of the one-line PR summary, ensuring that it meets the
     guidelines (e.g., "blob: add more blobbing") and accurately describes the
     change.
+-   If the PR includes a breaking change, which should be declared via a commit
+    with `BREAKING_CHANGE_OK` in it (Travis checks for this), add
```
eliben: This sentence is a mouthful... can you refactor it to mention what BREAKING_CHANGE_OK is in its own bullet?

CONTRIBUTING.md
```diff
@@ -182,5 +182,12 @@ When squashing-and-merging:
...
+    `BREAKING_CHANGE` to the PR summary (e.g., "blob: BREAKING_CHANGE remove old
```
zombiezen: Can these markers be put in the body of the commit/PR instead of the subject/title? It makes it really easy to blow out the 72 character suggestion for the first line. One could also imagine using [Git trailers](https://git-scm.com/docs/git-interpret-trailers) for this sort of metadata so that you could programmatically discover them.

CONTRIBUTING.md
vangent: Done.

CONTRIBUTING.md
vangent: I want them to be easily visible when reviewing commits for a release, e.g.:
https://github.com/google/go-cloud/compare/v0.11.0...v0.12.0

Is there a way to view that list with full commit bodies?

I haven't heard of commit trailers, but are they supported via the Github UI for squash-and-merge?

zombiezen: They are supported from GitHub UI in the sense that it is purely a commit message convention and you could enter them in the commit message when squashing.

One could write a script pretty easily that would find commits with a trailer in a given commit range. If not being able to do scan in the UI is a dealbreaker, then I'm fine with keeping the requirements the way they are. Figured I'd mention other tricks in the arsenal in case they were helpful in solving the problem.

vangent: Let's see how it goes with this and we can adjust if needed.

CONTRIBUTING.md
```diff
@@ -182,5 +182,13 @@ When squashing-and-merging:
 -   Do a final review of the one-line PR summary, ensuring that it meets the
     guidelines (e.g., "blob: add more blobbing") and accurately describes the
     change.
+-   Mark breaking changes with `BREAKING_CHANGE` in the commit message (e.g.,
+    "blob: BREAKING_CHANGE remove old blob").
+    -   If the PR includes a breaking change, it will be declared via a commit
+        with `BREAKING_CHANGE_OK` in it; Travis checks for this.
```
clausti: I'm confused about "Travis checks for this", in light of the second bullet point which says it's not actually required. Does Travis checking for breaking changes and then not finding `BREAKING_CHANGE_OK` have consequences for the build?

vangent: Travis verifies that if there's a breaking change, one of the commits in the PR has `BREAKING_CHANGE_OK` in it.

This text is talking about what we do during squash/merge, when all those commits get rolled up into one and there's free text for the commit message. There's no way to have Travis enforce this, but I want us as humans to including `BREAKING_CHANGE` in the squashed commit title if appropriate (and some judgement is required about that, thus the 2nd bullet).

HTH

clausti: Yeah, that helps. It does feel a little weird to me for the first time this is mentioned is in the past tense, like, by the time one is following the steps in the squash-and-merge section the `BREAKING_CHANGE_OK` commit is already suppose to have occurred but there is no directive in the sections about making a PR or reviewing?

vangent: Good point; added a bullet to the `Reviewer` section above.

PR #1619, opened 2019-03-20
pubsub: compute batchsize based on timing ReceiveBatch, not Ack

### LINKED ISSUE #1567: pubsub: consider dampening maxMessages
Currently `maxMessages` for receiving batches of messages starts at 1, and is subsequently set to a value that should get enough messages to process in 2 seconds (capped at 1000):
https://github.com/google/go-cloud/blob/master/pubsub/pubsub.go#L408

At startup, and with fast message processing times, this means we go from a batch size of 1 to a batch size of 1000 in one step, based on the processing time of a single message.

I think we should dampen this a bit to make it less noisy. It may take a little longer to ramp up at startup, but should be more stable once things are going.

Since the "how long does it take us to process a message" variable [already has dampening](https://github.com/google/go-cloud/blob/master/pubsub/pubsub.go#L356), one option would be to initialize it to `desiredQueueDuration` instead of to 0. I tried this using `mempubsub`'s Benchmark, and it looks like:

```
maxMessages: 1 based on avgProcessTime 2.000000
maxMessages: 2 based on avgProcessTime 1.900000
maxMessages: 2 based on avgProcessTime 1.900000
maxMessages: 2 based on avgProcessTime 1.629013
maxMessages: 2 based on avgProcessTime 1.396675
maxMessages: 2 based on avgProcessTime 1.260500
maxMessages: 2 based on avgProcessTime 1.197475
maxMessages: 2 based on avgProcessTime 1.026685
maxMessages: 3 based on avgProcessTime 0.926584
maxMessages: 3 based on avgProcessTime 0.794430
maxMessages: 3 based on avgProcessTime 0.681125
maxMessages: 4 based on avgProcessTime 0.614715
maxMessages: 5 based on avgProcessTime 0.475655
maxMessages: 6 based on avgProcessTime 0.368054
maxMessages: 8 based on avgProcessTime 0.270554
maxMessages: 12 based on avgProcessTime 0.179492
maxMessages: 21 based on avgProcessTime 0.096991
maxMessages: 61 based on avgProcessTime 0.033032
maxMessages: 1000 based on avgProcessTime 0.001523
```

We could make that more aggressive by modifying `decay` if we want.
Note that it starts fairly slow and gets exponential faster; this is because each batch contains a different number of messages; the batch with 61 messages decays the processing time 61 times.

Another option would be to explicitly dampen `maxMessages`; e.g., keep track of the last `maxMessages` size, compute a new ideal value based on `avgProcessTime`, and split the difference.

@jba thoughts?

### LINKED ISSUE #1584: pubsub: batch size never changes when application doesn't call `Ack`
We use timing info we've gleaned about how long it takes to process messages to figure out how big the `ReceiveBatch` batch size should be, but we only gather the timing info when `Message.Ack` is called. If the application never calls `Ack`, we never updating the timing info and always request batch size = 1.

@jba

vangent:
Benchmark results are shown [here](https://docs.google.com/spreadsheets/d/1c3mqHJ33mN64xBQrx42DY1GpYrMUi7IEl-6FRawIv70/edit), or [here](https://docs.google.com/a/google.com/spreadsheets/d/e/2PACX-1vQiQ-wdF8QpLrFz1gPBjv_VZCv2c1EXEFZIgHO0RRmkLKtVk2rM0J-bXNVbJ14EoBsObA2Kqwq-5srg/pubhtml?gid=312937778&single=true) for external users.

Each benchmark profile is a row of graphs, with blue being HEAD and red being this PR. Each row of graphs has a label to the far right.

Things to note:
* It's not obvious why the "baseline" is so much better with this PR. My guess is that the previous approach required locking/unlocking the `Subscription` for each message, which is not required anymore, but I'm a bit surprised by the extent of the improvement (
* The bug when "ack" is disabled (#1584) is fixed, you can see the drastic effect on the "no ack" row (50k -> 1.3 million throughput).
* The two rows "receive 100ms" and "receive 1s" show how the batch size ramps up over time, more slowly if `ReceiveBatch` takes longer.
* HEAD has a bug where it's basing batch size on the processing time of a message, but not taking into consideration that multiple messages can be processed concurrently. You can see the effect in the "process 100ms" and "process 1s" rows, where HEAD sticks with a very small batch size while PR increases it.
* The most realistic row is probably the last one, with some variability in message processing times and ReceiveBatch times; this PR gets about a 4x improvement.

I also ran the regular `Receive` benchmarks using real drivers:
* `mempubsub`: 275k -> 305k msgs/sec.
* `nats` (local server): 170k -> 200k msgs/sec
* `rabbit` (local server): no change (175k msgs/sec)
* `aws`: no change (100 msgs/sec)
* `gcppubsub`: no change (100 msgs/sec)

Fixes #1584.
Fixes #1567.
Updates #1441.


pubsub/pubsub.go
```diff
@@ -341,26 +339,74 @@ const (
...
+	maxBatchSize = 1000
 )
 
-// Add message processing time d to the weighted moving average.
-func (s *Subscription) addProcessingTime(d time.Duration) {
-	// Ensure d is non-zero; on some platforms, the granularity of time.Time
-	// isn't small enough to capture very fast message processing times.
-	if d == 0 {
-		d = 1 * time.Nanosecond
+// updateBatchSize updates the # of messages to request in ReceiveBatch based
```
jba: nit: s/#/number/

pubsub/pubsub.go
```diff
@@ -341,26 +339,74 @@ const (
 	// by another process receiving from the same subscription.
 	desiredQueueDuration = 2 * time.Second
 
-	// The factor by which old points decay when a new point is added to the moving
-	// average. The larger this number, the more weight will be given to the newest
-	// point in preference to older ones.
-	decay = 0.05
+	// The initial # of messages to request via ReceiveBatch.
```
jba: nit: s/#/number/

pubsub/pubsub.go
```diff
@@ -341,26 +339,74 @@ const (
...
+	if s.lastBatchRecv.IsZero() {
+		// First call, no data yet.
+		s.runningBatchSize = initialBatchSize
+		return initialBatchSize
 	}
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	if s.avgProcessTime == 0 {
-		s.avgProcessTime = d.Seconds()
+	if !s.dynamicBatchSizes {
```
jba: Thought: maybe a simpler and more flexible workaround for testing would be a `fixedBatchSize` field. Then at the top of this function:
```
if s.fixedBatchedSize != 0 {
  return s.fixedBatchSize
}
```

pubsub/pubsub.go
```diff
@@ -404,37 +449,22 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
...
-			s.preReceiveBatchHook(nMessages)
+			s.preReceiveBatchHook(batchSize)
 		}
-		msgs, err := s.getNextBatch(ctx, nMessages)
+		msgs, err := s.getNextBatch(ctx, batchSize)
 		if s.postReceiveBatchHook != nil {
 			s.postReceiveBatchHook(len(msgs))
 		}
 		s.mu.Lock()
+		s.lastBatchRecv = time.Now()
```
jba: I feel like you shouldn't update these values until you know err == nil.  Maybe it doesn't matter?

pubsub/pubsub.go

pubsub/pubsub.go

pubsub/pubsub.go
vangent: SGTM, will do separately.

pubsub/pubsub.go
vangent: Hm, interesting. Let's see.

If we update them after an error, the next call to `Receive` will try again to get more messages, and update the batch size based on 0 messages received over some small period of time, so the batch size will likely go down.

Same thing if this `ReceiveBatch` didn't return an error, but returned 0 messages.

If we don't update after an error, the next call to `Receive` will update the batch size based the number of messages received on the last successful `ReceiveBatch`, but over a longer period of time, so batch size will also go down, but probably by less.

So, I don't think it matters too much; always updating feels more correct to me since otherwise we're re-using the same values to adjust the batch size multiple times.

Added a comment.

vangent: BTW, may have to reconsider this when we start calling `ReceiveBatch` before there are 0 messages left.

PR #1620, opened 2019-03-20
internal/website: add new import pages

zombiezen:
I played it safe on this one: this doesn't remove any stale packages, just meta tags for new import paths.


zombiezen:
> missing awsparamstore?

Robert already added in 01b45689cad70e8eb674c272f219ec537a6ae522.

(I generated this PR from running internal/website/makeimports.sh.)

PR #1622, opened 2019-03-20
mysql+postgres: use shared urlopener

vangent:
@jba added this shared implementation that the other openers use now.


vangent:
Note that this means that the `mysql+` and `mysql+db+` scheme prefixes will now work (and similar for postgres).

mysql/mysql.go
```diff
@@ -76,44 +76,33 @@ type MySQLURLOpener interface {
...
 // RegisterMySQL registers the opener with the given scheme. If an opener
 // already exists for the scheme, RegisterMySQL panics.
 func (mux *URLMux) RegisterMySQL(scheme string, opener MySQLURLOpener) {
-	if mux.schemes == nil {
-		mux.schemes = make(map[string]MySQLURLOpener)
-	} else if _, exists := mux.schemes[scheme]; exists {
-		panic(fmt.Errorf("scheme %q already registered on mux", scheme))
-	}
-	mux.schemes[scheme] = opener
+	mux.schemes.Register("mysql", "DB", scheme, opener)
```
zombiezen: Should this be `"DB"` or `"MySQL"`? The resource naming is `MySQL`.

vangent: The first argument is the portable API.
The second is the portable type.

So for, blob it's `blob` and `Bucket`.

I thought the right value was `DB` because that's the portable type.
I think it makes more sense for the URL prefixes, too; this supports "mysql+db+foo://bar", while changing the "DB" to "MySQL" would support "mysql+mysql+foo://bar".

zombiezen: Right, but the Widget position here is `MySQL` or `Postgres`, since these are distinct portable "types" of resource, although the Go type is the same. I think we might want to rename one or the other, but I could be overthinking this.

vangent: Hmm, I'm not sure what you mean by "portable type" if it's not a Go type.

`mysql+db` and `postgres+db` feel right to me, but I'm open to suggestion if you want to change them. But I don't think `mysql+mysql` makes sense.

zombiezen: Discussed offline. I did not realize that we now have both a package distinguisher and a type distinguisher. One level of redundancy is fine, two levels is redundant. 🙂 LGTM.

PR #1624, opened 2019-03-20
aws_azure: bring Wire provider sets for AWS and Azure up to date

### LINKED ISSUE #1303: blob: include URLOpener in provider sets

vangent:
Fixes #1303.


zombiezen:
I'm amused by your use of `_` to avoid the Contribute Bot yelling. We could change the `.contributebot` pattern to permit `+`.

vangent:
> I'm amused by your use of `_` to avoid the Contribute Bot yelling. We could change the `.contributebot` pattern to permit `+`.

And it didn't even work :-).

vangent:
FYI I un-added the Azure set with creds; they are all blob-specific so don't necessarily belong here, and since there are two sub-sets (regular creds vs SAS token creds, with some shared types between them), it isn't possible to make a single set anyway.

Issue #1581, opened 2019-03-15
requestlog: include trace ID in Entry and Stackdriver export


zombiezen:
### Is your feature request related to a problem? Please describe.

Currently, when a request is logged to Stackdriver using the [`requestlog` package](https://godoc.org/gocloud.dev/requestlog), it does not include the trace or span ID information (see [LogEntry](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry) fields). These log entries then won't be able to use features of Stackdriver Logging like [Show latency details](https://cloud.google.com/logging/docs/view/overview#show-latency).

### Describe the solution you'd like

`requestlog` should use OpenCensus to obtain the trace and span IDs and store them in a new field in `requestlog.Entry`. [`StackdriverLogger`](https://godoc.org/gocloud.dev/requestlog#StackdriverLogger) should then use this information to populate the outputted JSON object in a way that the [Logging Agent](https://cloud.google.com/logging/docs/agent/) reading the object can understand.

### Additional context

This issue came from [a comment on #1426](https://github.com/google/go-cloud/issues/1426#issuecomment-473450269).

PR #1628, opened 2019-03-21
requestlog: add traceID

felipeweb:
ref: https://github.com/google/go-cloud/issues/1581


requestlog/stackdriver.go
```diff
@@ -74,6 +74,7 @@ func (l *StackdriverLogger) log(ent *Entry) error {
 			Seconds int64 `json:"seconds"`
 			Nanos   int   `json:"nanos"`
 		} `json:"timestamp"`
+		TraceID string `json:"trace"`
```
zombiezen: Unfortunately, this isn't a `LogEntry` object, this is the JSON data structure parsed by the [Fluent plugin](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/master/lib/fluent/plugin/out_google_cloud.rb), so the `trace` key isn't directly recognized. (Would you mind adding a comment to that effect somewhere? This isn't an obvious property of this code.)

Glancing at the plugin's source, it appears that `logging.googleapis.com/trace` and `logging.googleapis.com/spanId` are the two keys we're interested in ([source](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/f93046d92f7722db2794a042c3f2dde5df91a90b/lib/fluent/plugin/out_google_cloud.rb#L154-L155)).

requestlog/requestlog.go
```diff
@@ -107,6 +112,7 @@ type Entry struct {
 	ResponseHeaderSize int64
 	ResponseBodySize   int64
 	Latency            time.Duration
+	TraceID            string
 }
```
zombiezen: We also should capture the [span ID](https://opencensus.io/tracing/span/). (Similar to the above comment, let's use the `trace.SpanID` type for this field.)

requestlog/requestlog.go
```diff
@@ -107,6 +112,7 @@ type Entry struct {
 	ResponseHeaderSize int64
 	ResponseBodySize   int64
 	Latency            time.Duration
+	TraceID            string
```
zombiezen: Since this is our abstract representation of a request log entry, we are permitted to have more strong types here. Let's use OpenCensus's `TraceID` type directly to make it clear what format it is in.

```suggestion
	TraceID            trace.TraceID
```

requestlog/requestlog_test.go
```diff
@@ -75,6 +80,9 @@ func TestHandler(t *testing.T) {
 	if ent.ResponseBodySize != int64(len(responseMsg)) {
 		t.Errorf("ResponseBodySize = %d; want %d", ent.ResponseBodySize, len(responseMsg))
 	}
+	if ent.TraceID == "" {
```
zombiezen: I think we have enough information to verify the trace ID matches the one in the span, right? IIUC that would give us more confidence in this test than just testing for non-empty.

felipeweb:
@zombiezen done! Let me know if it needs something more.

internal/testing/alldeps
```diff
@@ -37,6 +37,7 @@ github.com/hashicorp/go-multierror
 github.com/hashicorp/go-retryablehttp
 github.com/hashicorp/go-rootcerts
 github.com/hashicorp/go-sockaddr
+github.com/hashicorp/golang-lru
```
zombiezen: To anyone else auditing here: this dependency is released under Mozilla Public License 2.0, which we already have cleared to use from the MySQL driver.

requestlog/requestlog_test.go
```diff
@@ -75,20 +76,41 @@ func TestHandler(t *testing.T) {
...
+	ctx, span := trace.StartSpan(r.Context(), "test")
+	defer span.End()
+	r = r.WithContext(ctx)
+	sc := trace.FromContext(ctx).SpanContext()
+	sh.spanCtx = &sc
+	sh.h.ServeHTTP(w, r)
 }
 
-func roundTrip(r *http.Request, h http.Handler) (*Entry, error) {
+func roundTrip(r *http.Request, h http.Handler) (*Entry, trace.SpanContext, error) {
```
zombiezen: nit: Seems like this should return a `*trace.SpanContext`?

felipeweb: @zombiezen https://github.com/google/go-cloud/pull/1628/commits/e8907ea7eba02b57e4a276f25619f27b534b1a58

PR #1632, opened 2019-03-21
travis: add check that doc site includes meta pages for all packages

### LINKED ISSUE #1621: internal/website: prevent future meta tag outages

We recently had an issue where a package was introduced without the corresponding `<meta>` tags being introduced to the website. This would break users trying to `go get` a particular new package. I would like to prevent future recurrences of this problem.

### Possible solutions

-  Add a Travis pre-merge check that verifies that `makeimports.sh` would no-op
-  Add a step in the deploy process that adds any missing meta pages (could be before or after Hugo site generation, but before pushing to `gh-pages`)
-  Find a new way to serve/generate the meta tags outside of Hugo

zombiezen:
**Note for reviewers:** I discovered while making this change that there are now two missing packages: azure/azuredb and mysql/azuremysql. I'm intentionally omitting them in the first version of this PR so as to test the failure mode. Once I verify that the check fails as expected and you approve, I will merge this PR and fix the missing packages.

Splits apart the searching and the creation steps of makeimports.sh into separate scripts.

Removed unused internal/imports directory to avoid future confusion.

Fixes #1621


zombiezen:
Excellent, the test failed as I expected it to. Carry on: the tests should pass now.

PR #1634, opened 2019-03-21
mongodocstore: docstore driver for MongoDB

jba:


internal/docstore/mongodocstore/mongo.go
```diff
@@ -0,0 +1,215 @@
...
+		err = c.update(ctx, action)
+
+	default:
+		err = gcerr.Newf(gcerr.Internal, nil, "bad action %+v", action)
+	}
+	return err
+}
+
+func (c *collection) get(ctx context.Context, a *driver.Action) error {
+	// TODO(jba): use Projection option to return only desired field paths.
```
shantuo: Is this something we would support at the driver level?

internal/docstore/mongodocstore/mongo.go
```diff
@@ -0,0 +1,215 @@
...
+	id, err := a.Doc.GetField(idField)
+	if err != nil {
+		return err
+	}
+	res := c.coll.FindOne(ctx, bson.D{{"_id", id}})
+	if res.Err() != nil {
+		return res.Err()
+	}
+	var m map[string]interface{}
+	// TODO(jba): Consider if we can avoid the double decode here.
```
shantuo: We might want to benchmark the double encoding/decoding first. It seems we need to call its `Decode` anyways?

internal/docstore/mongodocstore/mongo.go
```diff
@@ -0,0 +1,215 @@
...
+	// See https://docs.mongodb.com/manual/reference/method/db.collection.insertOne
+	mdoc, err := encodeDoc(a.Doc)
+	if err != nil {
+		return err
+	}
+	result, err := c.coll.InsertOne(ctx, mdoc)
+	if err != nil {
+		return err
+	}
+	if result.InsertedID != nil {
```
shantuo: These can be simplified:

```go
if result.InsertedID == nil {
  return nil
}
return a.Doc.SetField(idField, result.InsertedID)
```

internal/docstore/mongodocstore/mongo_test.go
```diff
@@ -0,0 +1,93 @@
...
+	defer cancel()
+	if err := coll.coll.Drop(tctx); err != nil {
+		return nil, err
+	}
+	return coll, nil
+}
+
+func (h *harness) Close() {}
+
+func TestConformance(t *testing.T) {
```
shantuo: Could we start a test server programmatically?

internal/testing/alldeps
```diff
@@ -29,6 +30,7 @@ github.com/google/subcommands
 github.com/google/uuid
 github.com/google/wire
 github.com/googleapis/gax-go
+github.com/googleapis/gax-go/v2
```
shantuo: Is this mongo's dependency?

internal/docstore/mongodocstore/codec.go
```diff
@@ -0,0 +1,190 @@
+// Copyright 2019 The Go Cloud Authors
```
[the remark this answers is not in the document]
jba: Done.

internal/docstore/mongodocstore/mongo.go
jba: The driver action has a list of field paths, so the user can request a subset of the whole document. But each driver has to implement that in its own way. I don't think there's going to be any common code across drivers.

internal/docstore/mongodocstore/mongo.go
jba: Good point. I changed the TODO.

There is something called `bson.Raw` which I think would cause the decoder to do nothing, and then we could decode the BSON ourselves. It would definitely be a lot of work though.

internal/docstore/mongodocstore/mongo.go

internal/docstore/mongodocstore/mongo_test.go
jba: It would require that either docker or mongo is installed on travis. I don't think we can expect that. I'm going to try to write a replayer.

internal/testing/alldeps
jba: Oops, that's because I ran the script on Go 1.12. Removed.

jba:
When I merged with master, I got the revision tests, which I'm not ready to implement yet. So I commented out the conformance tests temporarily. I'll address that in the next PR.

PTAL.

PR #1637, opened 2019-03-21
docstore/driver: add SplitActions

jba:
More than one driver will want to split a list of actions to run into
groups to optimize RPCs. Since the logic is a little tricky, factor
out to a common function.

Use it for the Firestore driver.

### DIFF
--- a/internal/docstore/firedocstore/fs.go
+++ b/internal/docstore/firedocstore/fs.go
@@ -64,7 +64,25 @@ func newCollection(client *vkit.Client, projectID, collPath, nameField string) *
 
 // RunActions implements driver.RunActions.
 func (c *collection) RunActions(ctx context.Context, actions []*driver.Action) (int, error) {
-	groups := groupActions(actions)
+	// Split the actions into groups, each of which can be done with a single RPC.
+	// - Consecutive writes are grouped together.
+	// - Consecutive gets with the same field paths are grouped together.
+	// TODO(jba): when we have transforms, apply the constraint that at most one transform per document
+	// is allowed in a given request (see write.proto).
+	groups := driver.SplitActions(actions, func(cur, new *driver.Action) bool {
+		// If the new action isn't a Get but the current group consists of Gets, split.
+		if new.Kind != driver.Get && cur.Kind == driver.Get {
+			return true
+		}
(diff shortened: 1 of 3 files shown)


internal/docstore/firedocstore/fs.go
```diff
@@ -64,7 +64,25 @@ func newCollection(client *vkit.Client, projectID, collPath, nameField string) *
 
 // RunActions implements driver.RunActions.
 func (c *collection) RunActions(ctx context.Context, actions []*driver.Action) (int, error) {
-	groups := groupActions(actions)
+	// Split the actions into groups, each of which can be done with a single RPC.
+	// - Consecutive writes are grouped together.
+	// - Consecutive gets with the same field paths are grouped together.
+	// TODO(jba): when we have transforms, apply the constraint that at most one transform per document
+	// is allowed in a given request (see write.proto).
+	groups := driver.SplitActions(actions, func(cur, new *driver.Action) bool {
```
shantuo: Consider pull the func out to its standalone function since it doesn't use `actions` inside and there are a lot of documentations with it.

internal/docstore/firedocstore/fs.go
```diff
@@ -64,7 +64,25 @@ func newCollection(client *vkit.Client, projectID, collPath, nameField string) *
...
+		// If the new action isn't a Get but the current group consists of Gets, split.
+		if new.Kind != driver.Get && cur.Kind == driver.Get {
+			return true
+		}
+		// If the new  action is a Get and either (1) the current group consists of writes, or (2) the current group
+		// of gets has a different list of field paths to retrieve, then split.
+		// (The BatchGetDocuments RPC we use for Gets supports only a single set of field paths.)
+		if new.Kind == driver.Get && (cur.Kind != driver.Get || !fpsEqual(cur.FieldPaths, new.FieldPaths)) {
```
shantuo: You could just 
```go
return new.Kind == driver.Get && (cur.Kind != driver.Get || !fpsEqual(cur.FieldPaths, new.FieldPaths))
```

internal/docstore/firedocstore/fs.go

internal/docstore/firedocstore/fs.go

jba:
PTAL.

PR #1639, opened 2019-03-21
docstore: move atomic-update test to memdocstore

jba:
The test in drivertest that verifies updates are atomic doesn't work
for all providers. So just test memdocstore.

### DIFF
--- a/internal/docstore/memdocstore/mem_test.go
+++ b/internal/docstore/memdocstore/mem_test.go
@@ -63,7 +63,6 @@ func TestUpdateAtomic(t *testing.T) {
 		"a":                    "A",
 		"b":                    "B",
 	}
-	// want should be unchanged
 	if !cmp.Equal(got, want) {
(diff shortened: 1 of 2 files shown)


internal/docstore/memdocstore/mem_test.go
```diff
@@ -38,3 +40,31 @@ func TestConformance(t *testing.T) {
...
+	if err := coll.Get(ctx, got); err != nil {
+		t.Fatal(err)
+	}
+	want := docmap{
+		drivertest.KeyField:    doc[drivertest.KeyField],
+		docstore.RevisionField: got[docstore.RevisionField],
+		"a":                    "A",
+		"b":                    "B",
+	}
+	// want should be unchanged
```
shantuo: nit: remove this comment.

jba: Done.

PR #1657, opened 2019-03-22
pubsub/awssnssqs: send multiple receive and ack RPCs concurrently

### LINKED ISSUE #1642: pubsub/awssnssqs: do multiple ReceiveMessage calls in parallel
Today, `driver.ReceiveBatch` gets called with `maxMessages`, which can be quite large (e.g., 1000), but the `awssnssqs` driver only ever gets 10 messages at a time, due to a constraint on the underlying RPC `ReceiveMessages`.

Update the driver to do multiple ReceiveMessages in parallel to improve throughput.

vangent:
Benchmark results. The number represents the value of the 3 "max" constants in the code below.

```
HEAD: 102.99 MB/s
10: 509.93 MB/s
25: 1081.00 MB/s
50: 1887.33 MB/s
100: 2493.14 MB/s
250: 2022.15 MB/s
```

Based on that, I've chosen 100 as a sweet spot.

Fixes #1642.

### DIFF
--- a/pubsub/awssnssqs/awssnssqs.go
+++ b/pubsub/awssnssqs/awssnssqs.go
@@ -88,9 +88,9 @@ const (
 	maxReceiveConcurrency = 100
 	// maxAckConcurrency limits the number of DeleteMessage RPCs (for Acks) to SQS in flight.
 	maxAckConcurrency = 100
-  // How long ReceiveBatch should wait if no messages are available; controls
-  // the poll interval of requests to SQS.
-   noMessagesPollDuration = 250 * time.Millisecond
+	// How long ReceiveBatch should wait if no messages are available; controls
+	// the poll interval of requests to SQS.
+	noMessagesPollDuration = 250 * time.Millisecond
 )
 
 func init() {
--- a/pubsub/awssnssqs/awssnssqs.go
+++ b/pubsub/awssnssqs/awssnssqs.go
@@ -543,7 +543,7 @@ Loop:
 			id := id
 			g.Go(func() error {
 				defer func() { <-s.acksem }()
-				_, err := s.client.DeleteMessage(&sqs.DeleteMessageInput{
+				_, err := s.client.DeleteMessageWithContext(ctx, &sqs.DeleteMessageInput{
 					QueueUrl:      &s.qURL,
 					ReceiptHandle: id.(*string),
 				})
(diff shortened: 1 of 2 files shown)


pubsub/awssnssqs/awssnssqs.go
```diff
@@ -411,21 +419,55 @@ func OpenSubscription(ctx context.Context, client *sqs.SQS, qURL string, opts *S
...
-		max = 10
+	// If maxMessages is larger than 10, we make up to maxReceiveConcurrency RPCs
+	// in parallel to try to get more.
+	const maxMessagesPerRPC = 10
+
+	var mu sync.Mutex
+	var ms []*driver.Message
+
+	g, ctx := errgroup.WithContext(ctx)
+	for n := 0; n < maxReceiveConcurrency && maxMessages > 0; n++ {
```
jba: This will behave well when lots of messages are available. But if there's only a few, you'll be making 100 RPCs, most of which will return nothing. Does Amazon charge per RPC, or only per message? I'm also worried about quota.

Could you add a TODO? One alternative is to remember how many messages you got the last time, and tailor the number of calls to that. We could even use a weighted moving average.

pubsub/awssnssqs/awssnssqs.go
```diff
@@ -483,27 +525,36 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) (msgs
...
-			return err
+		select {
+		case <-ctx.Done():
+			ctxErr = ctx.Err()
+			break Loop
+		case s.acksem <- struct{}{}:
+			id := id
+			g.Go(func() error {
+				defer func() { <-s.acksem }()
+				_, err := s.client.DeleteMessage(&sqs.DeleteMessageInput{
```
jba: There's no DeleteMessageWithContext? If not, note in a comment.

pubsub/awssnssqs/awssnssqs.go
vangent: Hmm, not sure I understand. This should only make `maxMessages / 10` RPCs at most, with a max of 100. So, e.g., at startup when `maxMessages = 1`, it will only make 1 RPC.

And after that `maxMessages` will only increase if the throughput of messages is high. While it's possible that 
-- we're going through a huge number of messages/sec
-- `maxMessages` ramps up to >= 1000
-- suddenly there aren't any more messages on the server (or many fewer)
and we'll make a bunch of unnecessary RPCs, that should be limited in time because `maxMessages` will immediately start to ramp down again.

In other words, I think that the `maxMessages` argument does what you suggested in the TODO.

pubsub/awssnssqs/awssnssqs.go
vangent: There is; not sure why we weren't using it. Done.

pubsub/awssnssqs/awssnssqs.go
jba: Yes, I see.

PR #1658, opened 2019-03-22
pubsub/gcppubsub: make Pull requests in parallel

### LINKED ISSUE #1643: pubsub/gcppubsub: do multiple Pull calls in parallel
Today, `driver.ReceiveBatch` gets called with `maxMessages`, which is currently capped at 1000. The `gcppubsub` driver can only get 1000 messages at a time, due to a constraint on the underlying RPC `Pull`.

Update the driver to do multiple `Pull` in parallel when `maxMessages` is > 1000 to improve throughput.

vangent:
Benchmark results. The number represents the value of the 3 "max" constants in the code below.

```
HEAD: 1725.95 MB/s
PR: 2093.29 MB/s
```

Not a huge win, but something. The numbers get better (including the advantage from this change) if you increase the number of message for the benchmark; it tends to be pretty slow at startup, with each RPC taking a full second (not sure why), and then at some point it ramps up and gets going pretty good.

Fixes #1643.

### DIFF
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -361,9 +361,10 @@ const (
 	// by more than that.
 	maxShrinkFactor = 0.75
 
-	// The maximum batch size to request.
-	// TODO(rvangent): Increase this.
-	maxBatchSize = 1000
+	// The maximum batch size to request. Setting this too low doesn't allow
+	// drivers to get lots of messages at once; setting it too small risks having
+	// drivers spend a long time in ReceiveBatch trying to achieve it.
+	maxBatchSize = 3000
 )
 
 // updateBatchSize updates the # of messages to request in ReceiveBatch based
--- a/pubsub/gcppubsub/gcppubsub.go
+++ b/pubsub/gcppubsub/gcppubsub.go
@@ -42,6 +42,7 @@ import (
 	"net/url"
 	"strings"
 	"sync"
+	"time"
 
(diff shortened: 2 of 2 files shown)


jba:
> taking a full second (not sure why)

You ruled out retries?

pubsub/pubsub.go
```diff
@@ -361,9 +361,10 @@ const (
 	// by more than that.
 	maxShrinkFactor = 0.75
 
-	// The maximum batch size to request.
-	// TODO(rvangent): Increase this.
-	maxBatchSize = 1000
+	// The maximum batch size to request. Setting this too low doesn't allow
+	// drivers to get lots of messages at once; setting it too small risks having
+	// drivers spend a long time in ReceiveBatch trying to achieve it.
+	maxBatchSize = 3000
```
jba: Where is this used?

vangent:
> > taking a full second (not sure why)
> 
> You ruled out retries?

No. But I'd be very surprised. The pattern I see is that the first 3-4 RPCs (with low `maxMessages`) all always take slightly more than 1s, and RPCs after that are much much faster.

pubsub/pubsub.go
vangent: Line 407:

```
s.runningBatchSize = math.Min(s.runningBatchSize, maxBatchSize)
```

It used to be hard-coded to 1000 with a long comment about how GCP Pub/Sub returns a max of 1000.

PR #1659, opened 2019-03-23
pubsub/driver: update some comments

vangent:


pubsub/driver/driver.go
```diff
@@ -127,12 +127,9 @@ type Subscription interface {
 	// This method should return only after all the ackIDs are sent, an
 	// error occurs, or the context is done.
 	//
-	// Only one RPC should be made to send the messages, and the returned
```
vangent: I'm not sure what the intent here was. E.g., `awssnssqs` only sends one ack at a time, but it just iterates over `ackID`s, it doesn't return an error if `len(ackIDs) > 1`. I'm pretty sure doing so would result in the driver not working well.

jba: I don't recall the original intent, but any driver that sends one ack at a time will have poor performance.

vangent: Understood. The previous comment said that implementations that send one ack at a time should return an error if len(ackIDs) != 1, which is wrong.

PR #1668, opened 2019-03-25
mongodocstore: support revisions

jba:
Preliminary support for revisions. Delete requires two RPCs in the error case, which might be OK.


internal/docstore/mongodocstore/mongo.go
```diff
@@ -121,7 +121,7 @@ func (c *collection) create(ctx context.Context, a *driver.Action) error {
 	if err != nil {
 		return err
 	}
-	mdoc[docstore.RevisionField] = int64(1)
+	mdoc[docstore.RevisionField] = 1
```
shantuo: I suggest to use the unique string as the revision. This field is an implementation detail and used only by the library, not read by human. Also I think use may implement some human-readable version if they have their specific needs so we want to avoid use a version-like revision here the prevent confuse them.

internal/docstore/mongodocstore/mongo.go
```diff
@@ -209,6 +259,29 @@ func (c *collection) update(ctx context.Context, a *driver.Action) error {
...
+// Error code for a write error when no documents match a filter.
+// (The Go mongo driver doesn't define an exported constant for this.)
+const mongoNotFoundCode = 11000
+
 func (c *collection) ErrorCode(err error) gcerrors.ErrorCode {
-	return gcerr.GRPCCode(err)
+	if c := gcerr.GRPCCode(err); c != gcerrors.Unknown {
+		return c
+	}
+	if wexc, ok := err.(mongo.WriteException); ok && len(wexc.WriteErrors) > 0 {
```
shantuo: It looks like all `err`s have `gcerrors.Unknown` codes?

internal/docstore/mongodocstore/mongo.go
```diff
@@ -142,16 +142,33 @@ func (c *collection) replace(ctx context.Context, a *driver.Action, upsert bool)
 	if err != nil {
 		return err
 	}
-	opts := options.Replace()
-	if upsert {
-		opts.SetUpsert(true) // Document will be created if it doesn't exist.
+	rev, err := a.Doc.GetField(docstore.RevisionField)
+	if err != nil && gcerrors.Code(err) != gcerrors.NotFound {
```
shantuo: Do all other error codes need to proceed?

internal/docstore/mongodocstore/mongo.go

internal/docstore/mongodocstore/mongo.go
jba: Not sure exactly what you mean, but I shouldn't be using GRPCCode here. Fixed.

internal/docstore/mongodocstore/mongo.go
jba: I think if somehow we get an error that isn't NotFound, then something is really wrong with the document and we should fail.

jba:
PTAL.

Issue #1576, opened 2019-03-15
internal/website: add how-to guides


zombiezen:
### Is your feature request related to a problem? Please describe.

The current documentation does not have many problem-focused documents. For example, if I want to store a blob, I would expect there to be a guide on how to use blob storage, perhaps with examples that include how to write a blob. Currently, one has to dig through examples in the API reference. I think there's room for improvement.

### Describe the solution you'd like

gocloud.dev should include pages on the site that focus on solving a problem (like data persistence or setting up a new server) that include step-by-step guides.

### Additional context

[What nobody tells you about documentation](https://www.divio.com/blog/documentation/)

zombiezen:
The list of guides I've brainstormed so far:

-  How to store and access blobs
-  How to store and access secret information
-  How to access dynamic configuration data
-  How to connect to a hosted SQL service
-  How to set up an HTTP server
-  How to set up a Pub/Sub subscriber program
-  How to provision resources for my server

For any of the portable API how-tos, it should include:

-  How to open the API using the constructor and the URL opener
-  How to write your own driver

zombiezen:
Another potential how-to, based on [this SO answer](https://stackoverflow.com/a/55360616/111866): how to use provider-specific attributes for common types. Such guide sections probably will show a common scenario with one provider, and then point to reference to indicate how this can be done for other providers.

zombiezen:
I've opened individual issues for the remaining bullets to permit fine-grained prioritization. Closing this issue in favor of individual issues.

PR #1669, opened 2019-03-25
internal/website/content/howto/secrets: add section

zombiezen:
Like #1630, kept as draft for now to batch how-to guide for publication. This guide has some intentional outstanding TODOs for issues that I found while writing the guide. This will need to be addressed before publishing, but should not change the structure of the guides AFAIK.

Updates #1576

/cc @clausti @shantuo for context on issues I've been opening.

### DIFF
--- a/internal/website/content/howto/secrets/_index.md
+++ b/internal/website/content/howto/secrets/_index.md
@@ -6,7 +6,17 @@ draft: true
 ---
 
 Cloud applications frequently need to store sensitive information like web
-API credentials or encryption keys in a medium that is not fully secure. The
-Go CDK provides access to key management providers in a portable way. These
-guides show how to work with secret keepers in the Go CDK.
+API credentials or encryption keys in a medium that is not fully secure. For
+example, an application that interacts with GitHub needs to store its OAuth2
+client secret and use it when obtaining end-user credentials. If this
+information was compromised, it could allow someone else to impersonate the
+application. In order to keep such information secret and secure, you can
+encrypt the data, but then you need to worry about rotating the encryption
+keys and distributing them securely to all of your application servers.
+Most cloud providers include a key management service to perform these tasks,
+usually with hardware-level security and audit logging.
+
+The Go CDK provides access to key management providers in a portable way
+called "secret keepers". These guides show how to work with secret keepers in
+the Go CDK.
 
--- a/internal/website/content/howto/secrets/_index.md
+++ b/internal/website/content/howto/secrets/_index.md
@@ -1,0 +1,12 @@
+---
+title: "Secrets"
+date: 2019-03-21T17:42:18-07:00
+showInSidenav: true
+draft: true
+---
+
+Cloud applications frequently need to store sensitive information like web
+API credentials or encryption keys in a medium that is not fully secure. The
+Go CDK provides access to key management providers in a portable way. These
+guides show how to work with secret keepers in the Go CDK.
+
--- a/internal/website/content/howto/secrets/crypt.md
+++ b/internal/website/content/howto/secrets/crypt.md
@@ -49,7 +49,7 @@ encrypting or decrypting large amounts of data is prohibitively slow (and in
 some providers not permitted). If you need your application to encrypt or
 decrypt large amounts of data, you should:
 
-1. generate a key for the encryption algorithm (16KiB chunks with
+1. Generate a key for the encryption algorithm (16KiB chunks with
    [`secretbox`][] is a reasonable approach).
 2. Encrypt the key with secret keeper.
 3. Store the encrypted key somewhere accessible to the application.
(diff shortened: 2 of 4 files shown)


internal/website/content/howto/secrets/crypt.md
```diff
@@ -0,0 +1,58 @@
...
+plainText, err := keeper.Decrypt(ctx, cipherText)
+if err != nil {
+    return err
+}
+```
+
+## Large Messages
+
+The secrets keeper API is designed to work with small messages (i.e. <10 KiB
+in length.) Cloud key management services are high enough latency that using
```
eliben: Sentence reads a bit strangely to me. Something like "Cloud key management services are high latency and using them ..." is simpler IMHO

internal/website/content/howto/secrets/crypt.md
```diff
@@ -0,0 +1,58 @@
...
+them for encrypting or decrypting large amounts of data is prohibitively slow
+(and in some cases not permitted). If you need your application to encrypt or
+decrypt large amounts of data, your application should generate a key for the
+encryption algorithm (16KiB chunks with [`secretbox`][] is a reasonable
+approach), encrypt the key with the secret keeper, and then store the
```
eliben: I suggest using a numbered list here with the steps, otherwise it's not easy to follow.

internal/website/content/howto/secrets/_index.md
```diff
@@ -0,0 +1,12 @@
+---
+title: "Secrets"
+date: 2019-03-21T17:42:18-07:00
+showInSidenav: true
+draft: true
+---
+
+Cloud applications frequently need to store sensitive information like web
```
eliben: Mind spending an extra paragraph or two here on the motivation and the setup? With my cloud newbie hat on, it took a bit to understand what these services actually do (after all, Go already has good crypto in the stdlib)

internal/website/content/howto/secrets/open-keeper.md
```diff
@@ -0,0 +1,242 @@
...
+    return err
+}
+
+// Create a KMS client.
+kmsClient, err := gcpkms.Dial(ctx, gcp.CredentialsTokenSource(creds))
+if err != nil {
+    return err
+}
+
+// Create a *blob.Bucket.
```
eliben: not a bucket?

internal/website/content/howto/secrets/open-keeper.md
```diff
@@ -0,0 +1,242 @@
...
+```go
+import (
+    "github.com/hashicorp/vault/api"
+    "gocloud.dev/secrets/vault"
+)
+
+// ...
+
+vaultClient, err := vault.Dial(ctx, &vault.Config{
+    Token: "<Client (Root) Token>",
```
jba: This is the first time you've used the angle-bracket syntax for a metavariable. Above you use, e.g `MYKEYRING`.

internal/website/content/howto/secrets/_index.md
zombiezen: Done.

internal/website/content/howto/secrets/crypt.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Large Messages
+
+The secrets keeper API is designed to work with small messages (i.e. <10 KiB
+in length.) Cloud key management services are high latency; using them for
+encrypting or decrypting large amounts of data is prohibitively slow (and in
+some providers not permitted). If you need your application to encrypt or
+decrypt large amounts of data, you should:
+
+1. generate a key for the encryption algorithm (16KiB chunks with
```
eliben: Generate

zombiezen: Done.

internal/website/content/howto/secrets/crypt.md
```diff
@@ -0,0 +1,58 @@
...
+}
+```
+
+## Decrypting data
+
+To decrypt data with a keeper, you call `Decrypt` with the byte slice you
+want to decrypt. This should be data that you obtained from a previous call
+to `Encrypt` with a keeper that uses the same secret material (e.g. two AWS
+KMS keepers created with the same customer master key ID). The `Decrypt`
+method will return an error if the input data did not come from the same
```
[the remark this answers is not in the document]
clausti: recap from private chat: It is definitely the case for `localsecrets` via [x/crypto/nacl/secretbox](https://godoc.org/golang.org/x/crypto/nacl/secretbox) and `awskms` via their [KMS Cryptographic Details](https://d1.awsstatic.com/whitepapers/KMS-Cryptographic-Details.pdf) white paper that decryption will fail if the ciphertext was not encrypted with the key you're attempting to decrypt it with.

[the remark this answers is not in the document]
clausti: It would surprise me to learn that any of the managed services helpfully supply "you're using the wrong key" as an error, but I don't have a citation for that.

PR #1673, opened 2019-03-26
internal/docstore/dynamodocstore: add revision field

shantuo:
Fixes: #1647


internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -100,13 +113,13 @@ func (c *collection) missingKeyField(m map[string]*dyn.AttributeValue) string {
...
 
-func (c *collection) put(ctx context.Context, doc driver.Document, condition *string) error {
+func (c *collection) put(ctx context.Context, k driver.ActionKind, doc driver.Document, condition *expression.ConditionBuilder) error {
 	av, err := encodeDoc(doc)
 	if err != nil {
 		return err
 	}
 	mf := c.missingKeyField(av.M)
-	if condition != &notExistsCond && mf != "" {
+	if k != driver.Create && mf != "" {
```
jba: Instead of passing the Kind, pass a boolean like `createMissingKeyField`. It seems a bit cleaner to me that way.

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -66,20 +63,36 @@ func (c *collection) KeyFields() []string {
...
 		case driver.Delete:
-			err = c.delete(ctx, a.Doc)
+			err = c.delete(ctx, a.Doc, pc)
 		case driver.Get:
 			err = c.get(ctx, a.Doc, a.FieldPaths)
 		case driver.Update:
-			err = c.update(ctx, a.Doc, a.Mods)
+			cb := expression.AttributeExists(expression.Name(c.partitionKey))
+			if pc != nil {
+				cb = cb.And(*pc)
```
jba: Nice. I see the value of the expression builder.

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -118,13 +131,22 @@ func (c *collection) put(ctx context.Context, doc driver.Document, condition *st
 		// It doesn't make sense to generate a random sort key.
 		return fmt.Errorf("missing sort key %q", c.sortKey)
 	}
+
+	if av.M[docstore.RevisionField], err = encodeValue(time.Now().UTC()); err != nil {
```
jba: I don't think we should use time alone, because it may differ across machines, and it's possible (though unlikely) that we'd get the same value twice. How about `driver.UniqueString`?

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -152,62 +174,102 @@ func (c *collection) get(ctx context.Context, doc driver.Document, fieldpaths []
...
+	if !ok {
+		return gcerr.Unknown
+	}
+	return ec
+}
+
+var errorCodeMap = map[string]gcerrors.ErrorCode{
+	dyn.ErrCodeConditionalCheckFailedException:          gcerr.FailedPrecondition,
+	dyn.ErrCodeProvisionedThroughputExceededException:   gcerr.ResourceExhausted,
+	dyn.ErrCodeResourceNotFoundException:                gcerr.FailedPrecondition,
```
jba: I think this one should be NotFound.

internal/docstore/dynamodocstore/dynamo.go
shantuo: Changed to use unique string, and that was failing the replay tests. I was doing this because I saw `firedocstore` uses timestamps. Is it feasible for us to make revision format consistent across drivers?

internal/docstore/dynamodocstore/dynamo.go
shantuo: Oh, I thought this was a more direct translation from what was here before. I think from the call site this way it is clear about what value is passed in instead of creating a new boolean.

internal/docstore/dynamodocstore/dynamo.go
shantuo: Done.

internal/docstore/dynamodocstore/dynamo.go
jba: Call MakeUniqueStringDeterministicForTesting.

jba: Firestore provides the timestamps natively, at the server.

shantuo: That is [called](https://github.com/google/go-cloud/pull/1673/files#diff-df25a5456a33858bf647bc1fb859938aR57) before running conformance tests.

internal/docstore/dynamodocstore/dynamo.go
jba: OK.

jba:
Remember to write "PTAL" in a comment when you're ready for me to look at the PR again. My dashboard won't call it to my attention without that. (I just happened to look for this PR specially, but I may forget to do that.)

PR #1680, opened 2019-03-26
pubsub/awssnssqs: use batch RPC for acks

vangent:
Because, why not?

I just happened to notice this batch version when looking at the AWS godoc. I'm not sure why we weren't using it before.


pubsub/awssnssqs/awssnssqs.go
```diff
@@ -530,25 +531,50 @@ func (s *subscription) receiveMessages(ctx context.Context, max int64) ([]*drive
...
+			case s.acksem <- struct{}{}:
+				req := req
+				g.Go(func() error {
+					return sendBatch(req)
 				})
-				return err
-			})
+			}
+			// Start the next batch.
+			req = &sqs.DeleteMessageBatchInput{QueueUrl: aws.String(s.qURL)}
```
jba: or just `req.Entries = nil`.

vangent: I don't think that works, because it leaves the `req` that we've handed to the background goroutine pointing to the same underlying struct, and we're modifying its fields.

PR #1683, opened 2019-03-26
internal/docstore/dynamodocstore: support binary sets

### LINKED ISSUE #1583: docstore/dynamodocstore: support set types
I noticed that the `dynamodbattribute` package's [Marshal](https://godoc.org/github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute#Marshal) method will encode a `[][]byte` as a "binary data set". We don't support DynamoDB's set types in docstore.

If we natively (meaning, with the `dynamoattribute` package) encode a `[][]byte` and try to decode it with the docstore codec, I believe we'll currently get an error. We could fix that with an `AsSpecial` method that recognizes this case.

If we docstore-encode a `[][]byte`, it will currently encode as a list of of `[]byte`s. I'm not sure how `dynamodbattribute.Unmarshal` will handle decoding that into a `[][]byte`.  But in any case the docstore codec should have an `EncodeSpecial` method that handles this.

shantuo:
Fixes #1583

### DIFF
--- a/internal/docstore/driver/document.go
+++ b/internal/docstore/driver/document.go
@@ -32,7 +32,8 @@ type Document struct {
 	fields fields.List            // for structs
 }
 
-// Create a new document from doc, which must be a map[string]interface{} or a struct pointer.
+// NewDocument creates a new document from doc, which must be a
+// map[string]interface{} or a struct pointer.
 func NewDocument(doc interface{}) (Document, error) {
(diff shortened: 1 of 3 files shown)


jba:
Let's wait on this, pending resolution of #1583.

shantuo:
Drop this as the conclusion in #1583 , will make another PR for that.

PR #1685, opened 2019-03-27
postgres: GCP PostgreSQL UrlOpener implementation

Zyqsempai:
UrlOpener implementation for GCP Postgres.

related to the #1175

This is a WIP PR since there are one big question, how we should pass GCPProjectID, Region and Instance id to the connection URL?


vangent:
@zombiezen will review.

zombiezen:
@Zyqsempai Brainstorming here: WDYT about a form that substitutes the colons in the connection name (e.g. `myproject:us-central1:myid`) with slashes, resulting in `cloudpostgres://myproject/us-central1/myid`?

Zyqsempai:
Hm, from first look, make sense, I will try to implement it and then i will have more clear opinion.

Zyqsempai:
@zombiezen Hi, I tried to use "/" as a split character for `project/region/instance` and it's not best approach since we also should provide DbPath in in same formate and it became something like this `cloudpostgres://myproject/us-central1/myid/mydb` and url parser consider `myproject` as a Host and `us-central1/myid/mydb` as a DBPath, but if we will use "." as a split argument like that `myproject.us-central1.instanceId/mydb` then Parser takes this part `myproject.us-central1.instanceId` as a host, and `/myDb` as a DBPath as it should be.
That's why I choosed "."

zombiezen:
Just a heads-up: I'm on vacation this coming week so I won't be able to review.

Zyqsempai:
Ok, have a nice vacation;)

postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -23,9 +23,15 @@ import (
...
 	"net"
+	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
+	"gocloud.dev/gcp"
+	"gocloud.dev/gcp/cloudsql"
+	"gocloud.dev/postgres"
+
```
zombiezen: nit: no blank line

postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -55,6 +61,49 @@ type Params struct {
...
+// OpenPostgresURL opens a new GCP database connection wrapped with OpenCensus instrumentation.
+func (uo *URLOpener) OpenPostgresURL(ctx context.Context, u *url.URL) (*sql.DB, error) {
+	vals := u.Query()
+	u.RawQuery = vals.Encode()
+	creds, err := gcp.DefaultCredentials(ctx)
+	client, err := gcp.NewHTTPClient(http.DefaultTransport, creds.TokenSource)
+	certSource := cloudsql.NewCertSource(client)
+	gcpInstance := strings.Split(u.Host, ".")
+	if len(gcpInstance) < 3 {
```
zombiezen: Having more components seems bad too.

```suggestion
	if len(gcpInstance) != 3 {
```

postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -55,6 +61,49 @@ type Params struct {
...
+		User:      u.User.Username(),
+		ProjectID: projectId,
+		Region:    region,
+		Instance:  instanceId,
+		Password:  password,
+		Database:  u.RawPath,
+		Values:    u.Query(),
+		TraceOpts: uo.TraceOpts,
+	}
+	db, err := Open(ctx, certSource, &params)
```
zombiezen: nit: You can simplify this by directly returning the result of `Open`.

```suggestion
	return Open(ctx, certSource, &params)
```

postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -55,6 +61,49 @@ type Params struct {
...
+	creds, err := gcp.DefaultCredentials(ctx)
+	client, err := gcp.NewHTTPClient(http.DefaultTransport, creds.TokenSource)
+	certSource := cloudsql.NewCertSource(client)
+	gcpInstance := strings.Split(u.Host, ".")
+	if len(gcpInstance) < 3 {
+		return nil, fmt.Errorf("cloudpostgres: OpenPostgresURL: not enough arguments in GCP host $s", u.Host)
+	}
+	projectId, region, instanceId := gcpInstance[0], gcpInstance[1], gcpInstance[2]
+	password, _ := u.User.Password()
+	params := Params{
```
zombiezen: nit: Since this will immediately be used in a pointer context, take the address of the composite literal.

```suggestion
	params := &Params{
```

postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -55,6 +61,49 @@ type Params struct {
...
+	postgres.DefaultURLMux().RegisterPostgres(Scheme, &URLOpener{})
+}
+
+// OpenPostgresURL opens a new GCP database connection wrapped with OpenCensus instrumentation.
+func (uo *URLOpener) OpenPostgresURL(ctx context.Context, u *url.URL) (*sql.DB, error) {
+	vals := u.Query()
+	u.RawQuery = vals.Encode()
+	creds, err := gcp.DefaultCredentials(ctx)
+	client, err := gcp.NewHTTPClient(http.DefaultTransport, creds.TokenSource)
+	certSource := cloudsql.NewCertSource(client)
```
zombiezen: The certificate source should be an overridable field in the `URLOpener` struct. It's fine for a `nil` value to have this behavior, but we should allow the user to bring their own.

zombiezen:
> @zombiezen Hi, I tried to use "/" as a split character for `project/region/instance` and it's not best approach since we also should provide DbPath in in same formate and it became something like this `cloudpostgres://myproject/us-central1/myid/mydb` and url parser consider `myproject` as a Host and `us-central1/myid/mydb` as a DBPath, but if we will use "." as a split argument like that `myproject.us-central1.instanceId/mydb` then Parser takes this part `myproject.us-central1.instanceId` as a host, and `/myDb` as a DBPath as it should be.
> That's why I choosed "."

I'm not sure that I understand this rationale, and I'm somewhat worried about dot being a common separator. I could imagine something like this (untested) to allow slash-separated to work:

```go
pathParts := strings.Split(u.Host + u.Path, "/")
if len(pathParts) < 4 {
  return nil, fmt.Errorf("something something bad format: %q", u.Host + u.Path)
}
projectID := pathParts[0]
region := pathParts[1]
instanceID := pathParts[2]
dbName := strings.Join(pathParts[3:], "/")
```

Zyqsempai:
@zombiezen Glad you are back, related to sepparator, since `project/region/instance` it's more lke a host and less like something else it's more logic to sepparate it and handle like a host, and the default sepparator for host is a ".", definately I can imagine voulnarables for each sepparator, but DOT make more sense to me.
Will be glad to hear your concierns and ideas.

zombiezen:
> related to sepparator, since `project/region/instance` it's more lke a host and less like something else it's more logic to sepparate it and handle like a host, and the default sepparator for host is a ".", definately I can imagine voulnarables for each sepparator, but DOT make more sense to me.

I hear what you're saying about the instance-related parameters being conceptually the "host" of the connection, but my concern is for consistency with other URL openers. Specifically, both [GCP Pub/Sub](https://godoc.org/gocloud.dev/pubsub/gcppubsub#URLOpener) and [GCP KMS](https://godoc.org/gocloud.dev/secrets/gcpkms#URLOpener) use a `project/grouping/name` slash-separated syntax for this same concept. This is also broadly consistent with how these are represented in the underlying GCP APIs, so I'd like for the Go CDK to not invent a new scheme for this. What do you think?

Zyqsempai:
I get it, consistency it's definitely can be a reason to make it in "/" way, but can also be a good reason to change Pub/Sub and KMS to "." way, since it's still new functionality and we don't risk here.
You decide;) I will be glad to work on both implementations.

zombiezen:
@vangent to break the tie: do you think GCP PostgreSQL URLs should take the form `cloudpostgres://project.region.instance/dbname` or `cloudpostgres://project/region/instance/dbname`?

vangent:
I agree with @zombiezen, let's go with `/` as the separator for consistency, not only with other code in the Go CDK, but also with GCP itself, e.g. [here](https://cloud.google.com/kms/docs/object-hierarchy#key) and [here](https://cloud.google.com/pubsub/docs/publisher) use slashes.

Zyqsempai:
Good, will rework it.

Zyqsempai:
@zombiezen Done.

postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -55,6 +58,45 @@ type Params struct {
 	TraceOpts []ocsql.TraceOption
 }
 
+// URLOpener opens GCP PostgreSQL URLs
+// like "cloudpostgres://user:password@myproject/us-central1/instanceId/mydb".
+type URLOpener struct {
+	// TraceOpts contains options for OpenCensus.
+	TraceOpts  []ocsql.TraceOption
+	CertSource *certs.RemoteCertSource
```
zombiezen: Using a non-interface type here implies a specific implementation. Instead, use `proxy.CertSource`.

```suggestion
	CertSource proxy.CertSource
```

postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -55,6 +58,45 @@ type Params struct {
...
+}
+
+// Scheme is the URL scheme cloudpostgres registers its URLOpener under on
+// postgres.DefaultMux.
```
zombiezen: To match other packages' docs, please add info about this to the package documentation. See [gcsblob](https://godoc.org/gocloud.dev/blob/gcsblob#hdr-URLs) for an example.

postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -55,6 +58,45 @@ type Params struct {
...
+	postgres.DefaultURLMux().RegisterPostgres(Scheme, &URLOpener{})
+}
+
+// OpenPostgresURL opens a new GCP database connection wrapped with OpenCensus instrumentation.
+func (uo *URLOpener) OpenPostgresURL(ctx context.Context, u *url.URL) (*sql.DB, error) {
+	vals := u.Query()
+	u.RawQuery = vals.Encode()
+	gcpInstance := strings.Split(u.Path, "/")
+	if len(gcpInstance) != 4 {
+		return nil, fmt.Errorf("cloudpostgres: OpenPostgresURL: not enough arguments in GCP host $s", u.Host)
```
zombiezen:
```suggestion
		return nil, fmt.Errorf("cloudpostgres: open url: %s is not in the form project/region/instance/dbname", u.Host + u.Path)
```

postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -55,6 +58,45 @@ type Params struct {
...
+
+func init() {
+	postgres.DefaultURLMux().RegisterPostgres(Scheme, &URLOpener{})
+}
+
+// OpenPostgresURL opens a new GCP database connection wrapped with OpenCensus instrumentation.
+func (uo *URLOpener) OpenPostgresURL(ctx context.Context, u *url.URL) (*sql.DB, error) {
+	vals := u.Query()
+	u.RawQuery = vals.Encode()
+	gcpInstance := strings.Split(u.Path, "/")
```
zombiezen: I believe you meant:

```suggestion
	gcpInstance := strings.Split(u.Host + u.Path, "/")
```

I think it would be worth it to split out the string-splitting code into its own function so we can unit test that part, which would catch these sort of trip-ups.

postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -55,6 +58,45 @@ type Params struct {
...
+const Scheme = "cloudpostgres"
+
+func init() {
+	postgres.DefaultURLMux().RegisterPostgres(Scheme, &URLOpener{})
+}
+
+// OpenPostgresURL opens a new GCP database connection wrapped with OpenCensus instrumentation.
+func (uo *URLOpener) OpenPostgresURL(ctx context.Context, u *url.URL) (*sql.DB, error) {
+	vals := u.Query()
+	u.RawQuery = vals.Encode()
```
zombiezen: This mutates the passed-in URL, which violates the `URLOpener` contract.

(Separately: I'm not sure I understand the purpose of these first two lines.)

postgres/cloudpostgres/cloudpostgres.go
zombiezen: (And this field should have a doc comment.)

postgres/cloudpostgres/cloudpostgres.go
Zyqsempai: Actually there is no need to do so, coz `u.Host` will take `ProjectId` part by default, the only reason why we can do like this it's just to avoid misunderstanding.

Zyqsempai:
@zombiezen Probably done;)

postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -14,6 +14,30 @@
...
+// To customize the URL opener, or for more details on the URL format,
+// see URLOpener.
+// See https://godoc.org/gocloud.dev#hdr-URLs for background information.
+//
+// Escaping
+//
+// Go CDK supports all UTF-8 strings; to make this work with providers lacking
+// full UTF-8 support, strings must be escaped (during writes) and unescaped
+// (during reads). The following escapes are performed for gcsblob:
+//  - Blob keys: ASCII characters 10 and 13 are escaped to "__0x<hex>__".
```
vangent: Is this relevant for SQL?

Zyqsempai: Totally not, copy/paste, sorry, my bad.

zombiezen:
Hey @Zyqsempai, I think we're close with this PR (and thus the two follow-on PRs, #1735 and #1736) but the tests that are in this PR now aren't enough where I'm confident that the parsing functions correctly ([this comment](https://github.com/google/go-cloud/pull/1685#discussion_r274492709) was one of the places where I was concerned). I trust that you've verified its behavior, but I'm concerned that a later change could cause breakage and we wouldn't detect it. I'd like to move it to a table-driven test with more clear logic, and I don't want to merge it in until that happens, especially because this has implications for the other two PRs.

However, I recognize these PRs have dragged for a while, and I don't want to add more work for you. I can commit the tests I'm thinking of into your branch and we can discuss further. Would that be alright?

Zyqsempai:
@zombiezen Sure, no worries, feel free to add everything is needed to that branch.

zombiezen:
Cool, I've made the changes I think we need. PTAL @Zyqsempai.

Zyqsempai:
LGTM

PR #1694, opened 2019-03-27
mongodocstore: codec test

jba:
Implement the codec conformance test for MongoDB.

- The BSON codec lower-cases struct field names. To match it without
  affecting the encoding for maps, I had to distinguish map and struct
  encoding in the driver codec framework.

- Times are only saved with millisecond precision, so make an
  UnsupportedType for that. Also document that Docstore only guarantees
  milli precision.


internal/docstore/drivertest/drivertest.go
```diff
@@ -492,6 +498,34 @@ func testCodec(t *testing.T, ct CodecTester) {
 		check(a, &Arrays{}, ct.DocstoreEncode, ct.NativeDecode)
 		check(a, &Arrays{}, ct.NativeEncode, ct.DocstoreDecode)
 	}
+	// Nanosecond-precision time.
+	type NT struct {
+		T time.Time
+	}
+	nt := &NT{nanoTime}
+	if unsupported[NanosecondTimes] {
+		// Expect rounding (or truncation) to the nearest millisecond.
```
shantuo: I think rounding already includes "(or truncation)".

internal/docstore/mongodocstore/codec.go
```diff
@@ -49,17 +51,27 @@ type encoder struct {
...
+func (e *encoder) EncodeString(x string)      { e.val = x }
+func (e *encoder) ListIndex(int)              { panic("impossible") }
+func (e *encoder) MapKey(string)              { panic("impossible") }
+
+var typeOfGoTime = reflect.TypeOf(time.Time{})
+
+func (e *encoder) EncodeSpecial(v reflect.Value) (bool, error) {
+	// Treat time "specially" as itself (otherwise its BinaryMarshal method will be called).
+	if v.Type() == typeOfGoTime {
+		e.val = v.Interface()
```
shantuo: Should we truncate it in the encoder to match the native encoder?

internal/docstore/mongodocstore/codec.go
```diff
@@ -118,8 +137,14 @@ func (d decoder) AsString() (string, bool) {
 }
 
 func (d decoder) AsInt() (int64, bool) {
-	i, ok := d.val.(int64)
-	return i, ok
+	switch v := d.val.(type) {
+	case int64:
+		return v, true
+	case int32:
+		return int64(v), true
```
shantuo: Can it be `int`? How about

```go
case int64:
case int32:
case int:
  return int64(v), true
```

internal/docstore/mongodocstore/mongo_test.go
```diff
@@ -73,6 +74,46 @@ func (h *harness) MakeCollection(ctx context.Context) (driver.Collection, error)
 
 func (h *harness) Close() {}
 
+type codecTester struct {
+}
```
shantuo: nit: move the `}` above.

internal/docstore/drivertest/drivertest.go
```diff
@@ -452,10 +458,10 @@ func testCodec(t *testing.T, ct CodecTester) {
 		M:  map[string]bool{"a": true, "b": false},
 		By: []byte{6, 7, 8},
 		P:  &s,
-		T:  time.Now(),
+		T:  milliTime,
 	}
 	check(nm, &NativeMinimal{}, ct.DocstoreEncode, ct.NativeDecode)
-	check(nm, &NativeMinimal{}, ct.NativeEncode, ct.DocstoreDecode)
+	//	check(nm, &NativeMinimal{}, ct.NativeEncode, ct.DocstoreDecode)
```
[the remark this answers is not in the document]
jba: Oops. Fixed.

internal/docstore/driver/codec.go
```diff
@@ -83,11 +84,11 @@ type Encoder interface {
...
-	// EncodeMap is also called for structs if EncodeStruct (see below) returns false; the map
-	// then consists of the exported fields of the struct.
-	// For struct{A, B int}{1, 2}, if EncodeStruct returns false, the same sequence
-	// of calls as above will occur.
-	EncodeMap(n int) Encoder
+	// EncodeMap is also called for structs, with a second argument of true. The map
+	// then consists of the exported fields of the struct. For struct{A, B int}{1,
+	// 2}, if EncodeStruct returns false, the same sequence of calls as above will
+	// occur.
+	EncodeMap(n int, isStruct bool) Encoder
```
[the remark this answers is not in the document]
jba: Yes, it's the same almost everywhere—mongo is the one exception (so far).

internal/docstore/drivertest/drivertest.go

internal/docstore/mongodocstore/codec.go
jba: I don't think it can: http://bsonspec.org/spec.html.

internal/docstore/mongodocstore/mongo_test.go
jba: Done.

jba:
PTAL.

internal/docstore/mongodocstore/codec.go
jba: Not necessary: the docstore encoder first creates a `map[string]interface{}`, then passes that to the native encoder. So it will get truncated there.

jba:
PTAL.

PR #1697, opened 2019-03-28
pubsub: add support for Message.Nack

### LINKED ISSUE #1455: pubsub: support Nack

vangent:
Fixes #1455.

The breaking change is for driver implementations, not end users of `pubsub`. Since we're hosting all of the driver implementations, this shouldn't really be considered a breaking change.


pubsub/azuresb/azuresb.go
```diff
@@ -450,12 +453,23 @@ func messageAsFunc(sbmsg *servicebus.Message) func(interface{}) bool {
...
+	return s.updateMessageDispositions(ctx, ids, dispositionForNack)
+}
+
+// IMPORTANT: This is a workaround to issue message dispositions in bulk which is not supported in the Service Bus SDK.
+func (s *subscription) updateMessageDispositions(ctx context.Context, ids []driver.AckID, disposition string) error {
 	if len(ids) == 0 {
 		return nil
 	}
 
+	// TODO(rvangent): Can we cache this on the subscription instead of
```
jba: Yes, please.

pubsub/awssnssqs/awssnssqs.go
```diff
@@ -553,28 +553,69 @@ func (s *subscription) SendAcks(ctx context.Context, ids []driver.AckID) error {
...
 				req := req
 				g.Go(func() error {
-					return sendBatch(req)
+					defer func() { <-s.acksem }()
+					return sendReq(ctx, req)
 				})
 			}
 			// Start the next batch.
-			req = &sqs.DeleteMessageBatchInput{QueueUrl: aws.String(s.qURL)}
+			req = newReq()
```
jba: Set `n = 0` around here, I would think.

If I'm right, your tests aren't catching this, so fix that too.

pubsub/driver/driver.go
```diff
@@ -38,6 +38,24 @@ type Batcher interface {
 // AckID is the identifier of a message for purposes of acknowledgement.
 type AckID interface{}
 
+// AckAction is an enum of possible actions on an AckID.
+type AckAction int
+
+const (
```
jba:
```
const (
  //  ... 
   ActionAck AckAction = iota
  //
   ActionNack
)
```
or `iota+1` if you really want to start from 1 (would require a justifying comment).

jba: BTW, this is fine if you want, but a simple bool would do too.

pubsub/pubsub.go
```diff
@@ -132,9 +134,25 @@ func (m *Message) Ack() {
...
 	if m.isAcked {
-		panic(fmt.Sprintf("Ack() called twice on message: %+v", m))
+		panic(fmt.Sprintf("Ack/Nack called twice on message: %+v", m))
+	}
+	m.ack(driver.ActionAck)
+	m.isAcked = true
+}
+
+// Nack tells the server that this Message was not processed and should be
+// redelivered. It returns immediately, but the/ actual nack is sent in the
```
jba: fix stray "/"

pubsub/pubsub.go
```diff
@@ -642,13 +670,38 @@ func newSubscription(ds driver.Subscription, fixedBatchSize int, newAckBatcher f
...
-		// Remember a non-retryable error from SendAcks. It will be returned on the
+		var acks, nacks []driver.AckID
+		for _, a := range items.([]*driver.AckInfo) {
+			switch a.Action {
+			case driver.ActionAck:
+				acks = append(acks, a.AckID)
+			case driver.ActionNack:
+				nacks = append(nacks, a.AckID)
+			default:
+				panic("unknown action")
```
jba: A boolean would fix this :)

pubsub/azuresb/azuresb.go
vangent: I filed #1699 for this.

pubsub/awssnssqs/awssnssqs.go
vangent: Good catch. I think this is hard to test today, because any test that does more than 10 acks is going to be flaky due to the ack batcher. I could write an AWS-specific test of this function, but I'd rather write a conformance test; I'll do so as part of #1698.

vangent: (fixed)

pubsub/driver/driver.go
vangent: Done. You are right, bool is simpler. I used an enum because I was thinking of us adding "extend ack deadline" later, but that would probably require another value anyway, so let's worry about that later.

pubsub/pubsub.go

pubsub/pubsub.go

vangent:
PTAL

Issue #1690, opened 2019-03-27
docstore: test unsupported type failing properly by our decoder


shantuo:

jba:
@shantuo can you explain this better, maybe with a repro?

shantuo:
I was talking about documents with types like dynamo's set types that inserted using the native encoder. When we get them and decode them, we should fail if we do not support them. I don't know whether this is already been done.

jba:
We have `TestDecodeErrorOnUnsupported` in dynamodocstore to check for that.

PR #1700, opened 2019-03-28
internal/docstore/dynamodocstore: handle set types

### LINKED ISSUE #1583: docstore/dynamodocstore: support set types

shantuo:
The decoder will fail on stringset, numberset and binaryset attributevalues. The encoder will not encode [][]byte types, instead it throws an error and provide a workaround.

Fixes #1583 
Updates #1690


internal/docstore/dynamodocstore/codec.go
```diff
@@ -332,14 +341,28 @@ func toGoValue(av *dyn.AttributeValue) (interface{}, error) {
...
 	switch v.Type() {
 	case typeOfGoTime:
 		if d.av.S == nil {
 			return false, nil, errors.New("expected string field for time.Time")
 		}
 		t, err := time.Parse(time.RFC3339Nano, *d.av.S)
 		return true, t, err
-	default:
-		return false, nil, nil
+	case typeOfBinarySet:
```
jba: I think what you want to do here is see if `d.av` is one of SS, NS or BS, and return an error. It shouldn't matter what `v` is.

internal/docstore/dynamodocstore/codec.go
```diff
@@ -332,14 +341,28 @@ func toGoValue(av *dyn.AttributeValue) (interface{}, error) {
...
+		return true, nil, fmt.Errorf(unsupportedFmt, `dynamodbav:",binaryset"`)
```
jba: Instead of writing the struct tag, I would just call it "binary set", etc.

shantuo: Done.

internal/docstore/dynamodocstore/codec.go
shantuo: Done.

internal/docstore/dynamodocstore/codec.go
```diff
@@ -65,14 +65,23 @@ func (e *encoder) EncodeMap(n int) driver.Encoder {
...
+)
 
 // EncodeSpecial encodes time.Time specially.
 func (e *encoder) EncodeSpecial(v reflect.Value) (bool, error) {
 	switch v.Type() {
 	case typeOfGoTime:
 		ts := v.Interface().(time.Time).Format(time.RFC3339Nano)
 		e.EncodeString(ts)
+	case typeOfBinarySet:
+		return true, errors.New(`unsupported type: [][]byte, the docstore driver
```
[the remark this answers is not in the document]
shantuo: You're right, there is no data corruption, but we cannot test the native encoder+docstore decoder for `[][]byte` type then.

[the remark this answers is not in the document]
jba: Right, and that's handled by the `UnsupportedType` feature of the test. If you leave this case here, though, then no one can encode a `[][]byte` at all, which is not what we want.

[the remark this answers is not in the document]
shantuo: Done.

PR #1702, opened 2019-03-28
all: add URLMux.XXXSchemes and .ValidScheme

### LINKED ISSUE #1695: all: add a function to URLMux that returns if a scheme is supported
Something like `IsValidScheme(scheme string) bool`.

This would help users identify valid URLs; see https://github.com/pulumi/pulumi/pull/2455/files?file-filters%5B%5D=.go#diff-d335e4296a9a5904243d0f16ea2801a1R55.

vangent:
Fixes #1695.


vangent:
ping @shantuo  ?

blob/blob_test.go
```diff
@@ -296,6 +296,16 @@ func TestURLMux(t *testing.T) {
 	mux.RegisterBucket("foo", fake)
 	mux.RegisterBucket("err", fake)
 
+	if diff := cmp.Diff(mux.BucketSchemes(), []string{"err", "foo"}); diff != "" {
```
shantuo: make `[]string{"err", "foo"}` a local var `schemes` or some such and refer `foo` and `err` using `schemes[i]`.

vangent: IMHO I don't think that makes the test clearer. The schemes are hard-coded in the URLs below too.

shantuo: I was thinking more about not needing to change a bunch of places when we do make change, but it's unlike to change since these are test files so I'm fine with it.

Issue #1469, opened 2019-03-06
internal/docs: move design.MD content somewhere more user-friendly


clausti:
Per @eliben's [comment](https://github.com/google/go-cloud/pull/1464#issuecomment-469902370) on the PR to update design.md, move this sort of documentation to a more user-friendly/accessible place.

eliben:
It probably makes sense to leave some of `design.md` where it currently is but +1 on evaluating what should be moved out. #1448 recently moved out the escape hatches stuff, as an example

zombiezen:
I always hold up [What nobody tells you about documentation](https://www.divio.com/blog/documentation/) as a pretty good article about how to think about documentation audiences. I think the current documents gives pretty good coverage for contributors:

-  **Tutorials:** Our workflow is GitHub, so we gain the benefit of lots of GitHub tutorials.
-  **How-to Guides:** CONTRIBUTING.md
-  **Explanation:** internal/docs/design.md
-  **Reference:** CONTRIBUTING.md + API documentation

On the user side, the situation is more bleak, which I think should be addressed:

-  **Tutorials:** samples/tutorial and samples/guestbook
-  **How-to Guides:** Not many, but the Go package code examples maybe?
-  **Explanation:** Not really present.
-  **Reference:** Go docs

IMO we're missing How-to Guides and Explanation documents aimed at users. These can likely borrow elements from existing contributor-aimed documentation, but probably would focus on different things.

zombiezen:
I'm planning on addressing this by adding new pages to gocloud.dev. I will likely airlift appropriate portions of design.md to form the new content with some light editing.

zombiezen:
I think the user-relevant parts to understand are:

- URLs
- Portable type + driver
- Escape hatches
- Dev+Ops roles

To be clear, the existing sections are fine where they are now, but probably will cross-link with each other.

PR #1704, opened 2019-03-28
internal/website/howto/pubsub: add section

zombiezen:
Like #1630 and #1669, kept as draft for now to batch how-to guides for publication. And like #1669, the reference documentation has TODOs for questions I couldn't answer with reference alone. I will open issues for these, but they should not change the structure of the guides AFAIK.

As part of doing this, I updated the website's archetypes with fields I typically add to these pages' front matter. These get used for `hugo new`.

Updates #1576


internal/website/content/howto/pubsub/publish.md
```diff
@@ -0,0 +1,410 @@
...
+
+The second step is the same across all providers because the first step
+creates a value of the portable [`*pubsub.Topic`][] type. Publishing looks
+like this:
+
+```go
+err := topic.Send(ctx, &pubsub.Message{
+    Body: []byte("Hello, World!\n")
+    Metadata: map[string]string{
+        "Foo": "Bar",
```
jba: Consider using a realistic example of metadata. I sure would like to see one myself.

internal/website/content/howto/pubsub/subscribe.md
```diff
@@ -0,0 +1,504 @@
...
+
+    // ... and then loop on received messages.
+    for {
+        msg, err := subscription.Receive(ctx)
+        if err != nil {
+            // Errors from Receive indicate that Receive will no longer succeed.
+            log.Printf("Receiving message: %v", err)
+            break
+        }
+        // TODO(light): Is this reasonable ack behavior, @jba?
```
jba: Maybe. In the error case, was the error inherent to the message (e.g. it was invalid)? If so, "handle" it by logging, emailing someone, alerting, etc. Then ack it—you handled it, no one else needs to see it. If the error was transient (e.g. couldn't reach a backend), then it's okay to drop it, or even Nack it, so the message will be sent elsewhere.

Acking a message means no one needs to look at it again, ever. That cross-cuts the question of whether processing the message "fails". That is true for all at-least-once systems, so maybe refer people to the wikipedia page before they start this section, so they understand that. (Though it wouldn't hurt to lightly repeat it here.)

internal/website/content/howto/pubsub/subscribe.md
```diff
@@ -0,0 +1,504 @@
...
+
+```go
+func main() {
+    // Parse configuration and open subscription (shutdown deferred).
+    // ...
+
+    // ... and then loop on received messages. We can use a channel as a
+    // semaphore to limit how many goroutines we have active at a time as well
+    // as wait on the goroutines to finish before exiting.
+    const maxHandlers = 10
```
jba: You don't need the constant; it's `cap(sem)`. But maybe keep it for pedagogy? Up to you.

internal/website/content/howto/pubsub/subscribe.md
```diff
@@ -0,0 +1,504 @@
...
+    for {
+        msg, err := subscription.Receive(ctx)
+        if err != nil {
+            // Errors from Receive indicate that Receive will no longer succeed.
+            log.Printf("Receiving message: %v", err)
+            break
+        }
+
+        // Wait if there are too many active handle goroutines and
+        // acquire the semaphore.
```
jba: Add "but if you see that the context is done, stop waiting and start screaming" or similar.

internal/website/content/howto/pubsub/subscribe.md
```diff
@@ -0,0 +1,504 @@
...
+        // Handle the message in a new goroutine.
+        go func() {
+            // TODO(light): Is this reasonable ack behavior, @jba?
+            if err := handle(ctx, msg); err != nil {
+                log.Printf("Handling message: %v", err)
+            } else {
+                msg.Ack()
+            }
+            // Release the semaphore.
+            <-sem
```
jba: Put `defer func() { <-sem }()` at the top of this closure instead. Doesn't matter here but a good thing to encourage.

internal/website/content/howto/pubsub/subscribe.md
```diff
@@ -0,0 +1,504 @@
...
+[`*amqp.Connection`]: https://godoc.org/github.com/streadway/amqp#Connection
+[`rabbitpubsub.OpenSubscription`]: https://godoc.org/gocloud.dev/pubsub/rabbitpubsub#OpenSubscription
+
+## NATS {#nats}
+
+The Go CDK can publish to a [NATS][] subject. A NATS URL only includes the
+subject name. The NATS server is discovered from the `NATS_SERVER_URL`
+environment variable (which is something like `nats://nats.example.com`).
+
+**TODO(light): Something about the `ackfunc` query parameter? Unsure when to use.**
```
jba: NATS is an "at most once" system. The moment you (the subscriber) receive a message, it will never be delivered again. So it's meaningless to ack it.

We talked a lot about what the behavior of `Message.Ack` should be in this case. I argued that if your code calls `Ack`, you think you are running an at-least-once system, and you expect it to be fault tolerant. (If the subscriber crashes before acking, the message will be retransmitted, so every message will eventually be processed.) Therefore, goes this argument, `Ack` should panic, because you have committed the grievous error of linking in at at-most-once provider with your at-least-once system.

@vangent  argued that panicking is going to make some forms of testing very tedious. You may want to run in production with an at-least-once system, but test and do local development with a cheap, on-box system like NATS. To this way of thinking, `Ack` should be a no-op.

Since neither of us could convince the other, we punted to the user. They can (and must) tell us what `Ack` means to them, by providing a function to call when `Message.Ack` is invoked. It is "must," not "can," (i.e. there is no default), as a concession to me. I don't want anyone accidentally getting it wrong (cf. "grievous error" above).

We thought we explained this in [the package doc](https://godoc.org/gocloud.dev/pubsub#hdr-At_most_once_and_At_least_once_Delivery). Clearly we failed. Please suggest alternative wording.

internal/website/content/howto/pubsub/subscribe.md
```diff
@@ -0,0 +1,504 @@
...
+}
+defer subscription.Shutdown(ctx)
+```
+
+### In-Memory Constructor {#mem-ctor}
+
+To create a subscription to an in-memory Pub/Sub topic, pass the [topic you
+created][publish-mem-ctor] into the [`mempubsub.NewSubscription` function][].
+You will also need to pass an acknowledgement deadline. **TODO(light):
+Reference doesn't make clear how this gets used.**
```
jba: See #1709.

internal/website/content/howto/pubsub/publish.md
```diff
@@ -0,0 +1,410 @@
...
+```
+
+The rest of this guide will discuss how to accomplish the first step: opening
+a topic for your chosen Pub/Sub provider.
+
+[`*pubsub.Topic`]: https://godoc.org/gocloud.dev/pubsub#Topic
+
+## Constructors versus URL openers
+
+If you know that your program is always going to use a particular Pub/Sub
```
vangent: There's a longer non-API-specific discussion of URLs here:
https://godoc.org/gocloud.dev#hdr-URLs

Consider linking to it?

internal/website/content/howto/pubsub/publish.md
```diff
@@ -0,0 +1,410 @@
...
+defer pubClient.Close()
+
+// Construct a *pubsub.Topic.
+topic := gcppubsub.OpenTopic(pubClient, projectID, "example-topic", nil)
+defer topic.Shutdown(ctx)
+```
+
+[`gcppubsub.OpenTopic`]: https://godoc.org/gocloud.dev/pubsub/gcppubsub#OpenTopic
+
+## Azure Service Bus {#azure}
```
vangent: I don't want to be annoying, but why are we duplicating all of this provider-specific documentation here? Shouldn't it be in the godoc/examples?

internal/website/content/howto/pubsub/publish.md
zombiezen: I'd like to move that information to a Concepts page as per #1469 and then I'd link to it. I could link to that section in the interrim tho, WDYT?

internal/website/content/howto/pubsub/publish.md
zombiezen: Not sure if it's realistic enough, but Contribute Bot uses metadata to capture headers from the webhook (convenience more than anything else): https://github.com/google/go-cloud/blob/f4bf16298678fbb3b86c91e31a7f3bcaf4019d41/internal/contributebot/webhook/main.go#L93-L103

However, I'm not sure if that example is helpful unless one has the context of how GitHub webhooks work.

internal/website/content/howto/pubsub/subscribe.md
zombiezen: > Maybe. In the error case, was the error inherent to the message (e.g. it was invalid)? If so, "handle" it by logging, emailing someone, alerting, etc. Then ack it—you handled it, no one else needs to see it. If the error was transient (e.g. couldn't reach a backend), then it's okay to drop it, or even Nack it, so the message will be sent elsewhere.

Makes sense. There isn't really a Go standard convention around what constitutes a transient error. The [`Temporary()` method](https://godoc.org/net#Error.Temporary) is probably about as good as it gets, but it's inconvenient to test for and isn't a surefire test, so perhaps prose documentation is good enough, but this does seem difficult for users in general and I'd like to put forward a good pattern.

> Acking a message means no one needs to look at it again, ever. That cross-cuts the question of whether processing the message "fails". That is true for all at-least-once systems, so maybe refer people to the wikipedia page before they start this section, so they understand that. (Though it wouldn't hurt to lightly repeat it here.)

I don't understand this. Can you provide a link to which Wikipedia page you mean, so I can understand it? 🙂

internal/website/content/howto/pubsub/subscribe.md
zombiezen: I think I'll keep it for pedagogical purposes: folks that might not be as familiar with semaphores will at least latch onto the naming and understand that way.

internal/website/content/howto/pubsub/subscribe.md
zombiezen: Done.

internal/website/content/howto/pubsub/subscribe.md
zombiezen: Done.

internal/website/content/howto/pubsub/subscribe.md
zombiezen: Thanks for documenting. Done.

internal/website/content/howto/pubsub/subscribe.md
zombiezen: > NATS is an "at most once" system. The moment you (the subscriber) receive a message, it will never be delivered again. So it's meaningless to ack it.
> 
> We talked a lot about what the behavior of `Message.Ack` should be in this case. I argued that if your code calls `Ack`, you think you are running an at-least-once system, and you expect it to be fault tolerant. (If the subscriber crashes before acking, the message will be retransmitted, so every message will eventually be processed.) Therefore, goes this argument, `Ack` should panic, because you have committed the grievous error of linking in an at-most-once provider with your at-least-once system.
> 
> @vangent argued that panicking is going to make some forms of testing very tedious. You may want to run in production with an at-least-once system, but test and do local development with a cheap, on-box system like NATS. To this way of thinking, `Ack` should be a no-op.

With you so far.

> Since neither of us could convince the other, we punted to the user.

Bluntly, I don't think a user could ever make this decision, at least not with the information provided. (See below.)

> They can (and must) tell us what `Ack` means to them, by providing a function to call when `Message.Ack` is invoked. It is "must," not "can," (i.e. there is no default), as a concession to me. I don't want anyone accidentally getting it wrong (cf. "grievous error" above).

Minor note: the reference doesn't make this clear. I assumed that one of these query parameter values is the default and that the default isn't documented.

> We thought we explained this in [the package doc](https://godoc.org/gocloud.dev/pubsub#hdr-At_most_once_and_At_least_once_Delivery). Clearly we failed. Please suggest alternative wording.

I don't think that I can make such a suggestion because I don't know the information that would help me (or a user) make such a decision.

Concretely, what I am missing is the "why?" not the "what?" or "how?" of each of these options. I understand mechanically what each of these ack function options will do, but I do not understand the tradeoffs associated with picking any one of the individual choices. If I were to come across this cold, I probably would be lazy and pick no-op because that sounds the least dangerous. However, if I were given more information as to "picking no-op is useful in scenarios where X but is dangerous if Y: bad thing Z will happen", that would be tremendously helpful.

internal/website/content/howto/pubsub/publish.md
zombiezen: Because godoc is reference, not a how-to guide. Quoting [What nobody tells you about documentation](https://www.divio.com/blog/documentation/):

> Reference guides are _technical descriptions of the machinery_ and how to operate it.
>
> Reference guides have one job only: to describe. They are code-determined, because ultimately that's what they describe: key classes, functions, APIs, and so they should list things like functions, fields, attributes and methods, and set out how to use them.
>
> Reference material is **information-oriented.**
>
> By all means technical reference can contain examples to illustrate usage, but it should not attempt to explain basic concepts, or how to achieve achieve common tasks.

This is the format I've been doing for the other sections. Rationale:

1.  It's not straight duplication: there is a fair amount of editing that I'm doing as I go to clean up artifacts of godoc examples like predeclaring variables and blank identifier assignments.
2.  Linking would be unsatisfactory: it would force the reader to jump to a number of different packages to understand how to do one task end-to-end.
3.  Developing a way to inline godoc examples would suffer from the aforementioned problem of needing to edit.

I think that the examples in godoc should likely become more focused to individual functions over time, rather than whole-package like they largely are now. Definitely we should link between these things more, but since the how-to guides aren't published, I don't want to drop broken links into production reference.

internal/website/content/howto/pubsub/publish.md
jba: Yeah, too specific. I was hoping you could think of something that would be more universal. Some quick ideas:
```
language: en
importance: high
```

internal/website/content/howto/pubsub/subscribe.md
jba: It seems I was mistaken about the existence of that wikipedia page. @vangent, wasn't there a page that described this well?

Here's the idea in a nutshell. When you publish a message to an at-least-once system, it will keep trying to deliver it until it's acked. If it sends it to subscriber 1 and that subscribers "nacks" it (says "no, I can't accept this message"), then the system delivers it to someone else. If that subscriber doesn't ack it (before some predetermined deadline), the system will try again. It will keep trying (within limits) until it gets an ack. Conversely, once it gets an ack, it will never send the message again.

So the idea is, a subscriber should ack a message exactly when that message has been dealt with. A message has been dealt with when it has been successfully processed. But it also has been dealt with if the message is invalid, and the appropriate party has been notified of same. That is "failure" in one sense, but failing to ack that message is just going to give the same problem to someone else, so it's success in another sense. 

On the other hand, if the message is invalid and the subscribing process tries to send an email to the sender and the email system replies "down for maintenance, come back later," then the message shouldn't be acked, because it hasn't been dealt with. (The right thing to do here is probably to nack it, and also to tell the system to delay redelivery; but you might not want to get into that because some systems don't let you add a delay.)

So when I wrote 

> [Whether to ack] cross-cuts the question of whether processing the message "fails"

I meant that acks are about dealing with the message once and for all, and that may include cases you'd describe as failure.

internal/website/content/howto/pubsub/subscribe.md
jba: > I assumed that one of these query parameter values is the default and that the default isn't documented.

This is a query parameter without a default, and indeed that should be clearly noted. I love URLs!\</sarcasm>

jba: > I do not understand the tradeoffs associated with picking any one of the individual choices.

If you pick no-op, you could be breaking an invariant of your system, causing the system to malfunction (but probably only rarely). Here's the scenario:

1. You design your system to include an at-least-once pubsub provider. You chose at-least-once for reliability: you *must* process every message, and you know that if one of your subscribers crashes or locks up before acking, the message will be redelivered to someone else. You use the `gcloud.dev/pubsub` package and link in one of the at-least-once providers.

2. One year later, a maintainer of your system notices that the company's cloud provider just introduced another pubsub product. The maintainer doesn't read the fine print, which says that this is an at-most-once system, but does notice two numbers: it's a lot faster than the pubsub provider you're using, and it's a lot cheaper too.

3. Seems like a no-brainer! The maintainer swaps out the pubsub provider you chose in step 1 for the cheaper one, which also has a Go CDK driver. When they code up the `OpenSubscription` call of the new provider, they see they  must provide an ack function. Let's say they choose either no-op or panic.

4.
    a. They choose no-op. They build and redeploy. Everything seems fine (but faster and cheaper!). But a week later, they start seeing customer support issues: "I heard that XYZ happened but your app didn't notify me." Not a lot, maybe just one or two out of millions of customers. These support calls trickle in for months. They happen so rarely that they are hard to follow up on, and everyone is busy with adding new features. Eventually, someone notices the correlation between delivery failure and crashed machines. Some time later, someone figures out that if a machine crashes before it handles a pubsub message, the message is never re-delivered. Eventually they track it down to the change in step 3.

   b. They choose panic. They build and redeploy. Almost immediately, their monitoring system goes through the roof: processes are panicking everywhere. Multiple engineers are forced to look at the code causing the panic. Most of them suggest swapping out the panic for a no-op, but someone suggests thinking about what's going on a bit more. Disaster averted.

On the other hand, if you pick no-op, you can test your system locally with a cheaper and simpler at-most-once system, without having to have different code (Ack vs. no Ack) running in testing and production.

(Caveat: There may be some bias in my presentation of the alternatives.)

internal/website/content/howto/pubsub/publish.md
zombiezen: Discussed offline. Resolution:

1. Example code should be embedded in how-to guides, but source-of-truth should be godoc. Keeping as-is for sake of submitting, but I'm going to solve #1827 before moving forward with other guides.
2. Reference docs for particular functions should stay where they are.
3. Larger concept discussion will move to the website, and be linked from the godoc. Tracking in #1469.

internal/website/content/howto/pubsub/subscribe.md
zombiezen: Thanks for the detailed write-up; I understand much better now. I went with recommending `?ackfunc=panic` as the first choice, and detailed under what conditions someone may choose `?ackfunc=noop`, which seems consistent with what you are saying.

zombiezen:
@jba this is ready for re-review. I believe I have addressed all your comments.

PR #1708, opened 2019-03-28
pubsub/test: add burstiness simulations

### LINKED ISSUE #707: pubsub: find out how well batcher handles bursts
Probably the hardest thing the batcher will have to deal with is a short burst of activity following a period of inactivity. I'd like to have some way to characterize how well the batcher deals with this situation. Maybe a benchmark would do it. If we do add a benchmark, we'll need a baseline for comparison. We could compare with bundler for example.

vangent:
Graphs:
* [Google internal](https://docs.google.com/spreadsheets/d/1_JOGCTtaccUjdVvxgwJyU2NDH9ssgiM8HuuF-8KHHsg/edit)
* [External](https://docs.google.com/a/google.com/spreadsheets/d/e/2PACX-1vTzAucjgz3nqKhpzzoZHqPG47iiNr2uXnxkF8mqttvSgwVD1BSpa15d-itliLzKVhA36Jyz5ZIxQjB7/pubhtml?gid=312937778&single=true)

These are not A/B tests, both blue and red are the same. The bottom three rows are the new simulations:

* In the middle third, `ReceiveBatch` returns 0 messages.
* In the middle third, `ReceiveBatch` takes 10x as long.
* In the middle third, simulated message processing time taxes 100x as long.

I also changed the msgs/sec measurement in the benchmark to be more accurate (based on messages it gets via `Receive` rather than reported through a callback hook per `ReceiveBatch` call, which can have a high granularity when simulated RPC times are high), and removed the now-unused `postReceiveBatchHook`.

Fixes #707.


pubsub/benchmark_test.go
```diff
@@ -60,21 +60,28 @@ func (s *fakeSub) ReceiveBatch(ctx context.Context, maxMessages int) ([]*driver.
...
 // TestReceivePerformance enables characterization of Receive under various
 // situations, characterized in "tests" below.
 func TestReceivePerformance(t *testing.T) {
 	t.Skip("Skipped by default")
 
 	const defaultNumGoRoutines = 100
-	defaultReceiveProfile := func(maxMessages int) (int, time.Duration) { return maxMessages, 0 }
-	defaultProcessProfile := func() time.Duration { return 0 }
+	defaultReceiveProfile := func(_ bool, maxMessages int) (int, time.Duration) { return maxMessages, 0 }
+	defaultProcessProfile := func(_ bool) time.Duration { return 0 }
```
jba: You don't need the underscore.

vangent: Done.

PR #1718, opened 2019-03-28
pubsub/natspubsub: clean up and check errors in examples

zombiezen:


zombiezen:
I'm abandoning this change, as I'm doing this and other cleanups as part of #1827.

PR #1721, opened 2019-03-28
pubsub: ensure Message.Metadata is consistently nil if empty

vangent:
Before this PR, the behavior is inconsistent (it depends on driver behavior; some return an empty map, others return nil). Consistency is good.

I noticed this when running conformance tests that used a `nil` Metadata when sending and tried to verify that the received message matched; for some drivers, it did, and for others, it didn't (because the latter drivers returned an empty map instead of nil).


vangent:
(I'll be adding a conformance test for this in a separate PR)

PR #1722, opened 2019-03-28
docstore: test blind decoding

jba:
Test the decoder's AsInterface method, which decodes values
without knowing the intended type.

### DIFF
--- a/internal/docstore/drivertest/drivertest.go
+++ b/internal/docstore/drivertest/drivertest.go
@@ -18,6 +18,7 @@ package drivertest // import "gocloud.dev/internal/docstore/drivertest"
 
 import (
 	"context"
+	"math"
 	"math/rand"
 	"sync"
 	"testing"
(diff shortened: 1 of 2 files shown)


internal/docstore/drivertest/drivertest.go
```diff
@@ -526,6 +499,116 @@ func testCodec(t *testing.T, ct CodecTester) {
...
+	t.Run("NativeEncode", func(t *testing.T) { testBlindDecode1(t, ct.NativeEncode, ct.DocstoreDecode) })
+}
+
+func testBlindDecode1(t *testing.T, encode func(interface{}) (interface{}, error), decode func(_, _ interface{}) error) {
+	// Encode and decode expect a document, so use this struct to hold the values.
+	type S struct{ X interface{} }
+
+	for _, test := range []struct {
+		in    interface{} // the value to be encoded
+		want  interface{} // one possibility
```
shantuo: nit: use a []interface{} to hold possible results?

jba: I tried it that way first, and the tests were ugly because I had to write `[]interface{}` for every one, even the single-valued ones.

PR #1723, opened 2019-03-28
pubsub: replace newBatcher func with batcher.Options; added a conformance test for batching

### LINKED ISSUE #1698: pubsub: get rid of the "newBatcher" args to constructors
I don't think we're ever doing anything other than passing `nil` for the default; although I have seen some flakiness in tests due to batch, it hasn't risen to the level of being a problem.

I think it might be simpler to replace the func arg with something like `fixedBatchSize`, which would get passed through to the batcher and enforce that batch size (with 0 being the current default of dynamic batch sizes).

@jba thoughts? Do you remember why we made this an arg?

vangent:
This PR gets rid of the `driver.Batcher` interface, and the function arguments to `pubsub.newTopic` and `pubsub.newSubscription` that return ` driver.Batcher`. We were never using anything other than the default.

Instead, the constructors now take a `batcher.Options` to set batching options, including max concurrent calls and min/max batch size, and the `batcher.Batcher` type is used explicitly.

Conformance tests set `batcher.Options.MaxBatchSize` to 1 to ensure that record/replay tests are stable.

A new conformance test sets `MaxBatchSize` and `MinBatchSize` to a large value (11) to test larger batch sizes.

I think these new options can be used to simplify provider code in a few cases; for examples, `awssnssqs`'s `SendBatch` only sends a single message at a time; it currently does its own parallelization in `SendBatch`, using a semaphore to limit overall concurrency. I think that this could be replaced by passing `batcher.Options.MaxBatchSize=1` and `batcher.Options.NumHandlers` to the concurrency desired. I left TODOs for these because the cleanup is separable and I want to do benchmarks for them.

The `BREAKING_CHANGE_OK` is for driver implementations; it doesn't affect external users since all driver implementations are currently in our repo.

Fixes #1698.


internal/batcher/batcher.go
```diff
@@ -88,22 +108,41 @@ func (b *Batcher) AddNoWait(item interface{}) <-chan error {
...
+	if len(b.pending) < b.opts.MinBatchSize {
+		return nil
+	}
+	if b.opts.MaxBatchSize == 0 || len(b.pending) <= b.opts.MaxBatchSize {
+		// Send it all!
+		batch := b.pending
+		b.pending = nil
+		return batch
+	}
+	batch := b.pending[0:b.opts.MaxBatchSize]
```
jba: You can omit the initial zero.

pubsub/pubsub.go
```diff
@@ -259,8 +259,8 @@ func (t *Topic) ErrorAs(err error, i interface{}) bool {
 // NewTopic is for use by provider implementations.
 var NewTopic = newTopic
 
-// defaultBatcher creates a batcher for topics, for use with NewTopic.
-func defaultBatcher(ctx context.Context, t *Topic, dt driver.Topic) driver.Batcher {
+// newSendBatcher creates a batcher for topics, for use with NewTopic.
+func newSendBatcher(ctx context.Context, t *Topic, dt driver.Topic, batcherOpts *batcher.Options) *batcher.Batcher {
```
jba: s/batcherOpts/opts/

pubsub/pubsub.go
```diff
@@ -274,26 +274,20 @@ func defaultBatcher(ctx context.Context, t *Topic, dt driver.Topic) driver.Batch
...
 
 // newTopic makes a pubsub.Topic from a driver.Topic.
 //
-// If newBatcher is nil, a default batcher implementation will be used.
-// If non-nil, the driver.Batcher returned by newBatcher should batch
-// *driver.Message, and call the SendBatch func of the driver.Topic in its
-// handler. See defaultBatcher for a sample implementation.
-func newTopic(d driver.Topic, newBatcher func(context.Context, *Topic, driver.Topic) driver.Batcher) *Topic {
+// batcherOpts may be nil to accept defaults.
+func newTopic(d driver.Topic, batcherOpts *batcher.Options) *Topic {
```
jba: ditto. We can see the type name is `batcher.Options`.

vangent:
> Nice.

Yeah! Wait till you get the next PR where I delete a bunch of code and use these options instead.

internal/batcher/batcher.go
vangent: Yup, done.

pubsub/pubsub.go

pubsub/pubsub.go

PR #1724, opened 2019-03-29
pubsub: simplify and improve drivers by using batcher options to manage batch size and concurrency for Send and Ack/Nack

### LINKED ISSUE #1698: pubsub: get rid of the "newBatcher" args to constructors

vangent:
Now that drivers can explicitly control the maximum batch size and the allowed concurrency, they don't have to handle batch sizes bigger than they want. This simplifies their code significantly, while also improving performance in at least one place where we hadn't added concurrency yet (AWS Send).

This also improves retry behavior, since each batch is retried independently; previously, if a huge `SendBatch` was split up into multiple RPCs by the driver, and one of the RPCs failed, we'd potentially end up retrying the whole thing.

Note: We can't easily do the same trick for `ReceiveBatch` because it's not using a batcher, but it's worth thinking about. Maybe the portable type could split a single `ReceiveBatch` call into multiple concurrent `ReceiveBatch` calls based on a similar set of options. I filed #1725.

```
AWS Send 150->1000  /* 8x faster! */
Azure 15->700 /* ~40x faster! */
GCP Send 2200->3000 /* faster, likely due to extra concurrency of 2 */
```

Fixes #1698.


pubsub/awssnssqs/awssnssqs.go
```diff
@@ -537,98 +514,43 @@ func (s *subscription) receiveMessages(ctx context.Context, max int64) ([]*drive
...
-			n = 0
-		}
-	}
-	if err := g.Wait(); err != nil {
+	resp, err := s.client.ChangeMessageVisibilityBatchWithContext(ctx, req)
+	if err != nil {
 		return err
 	}
-	return ctxErr
+	if numFailed := len(resp.Failed); numFailed > 0 {
```
jba: factor out?

vangent: Tricky because the types are different; don't think it's worth it.

PR #1730, opened 2019-03-29
runtimevar/httpvar: add an URLOpener

### LINKED ISSUE #1729: runtimevar: add URL support to new HTTP provider

vangent:
I also modified the mock HTTP server to keep track of distinct variables.

Fixes #1729.


vangent:
ping @shantuo ?

runtimevar/httpvar/httpvar.go
```diff
@@ -36,6 +44,53 @@ import (
...
+		runtimevar.DefaultURLMux().RegisterVariable(scheme, o)
+	}
+}
+
+// Schemes are the URL schemes httpvar registers its URLOpener under on runtimevar.DefaultMux.
+var Schemes = []string{"http", "https"}
+
+// URLOpener opens HTTP URLs like "http://myserver.com/foo.txt".
+//
+// The full URL, including scheme, is used as the endpoint, except that the
```
shantuo: How about just `the URL query values are removed`? Since we'll error out if there is unknown parameter.

runtimevar/httpvar/httpvar.go
```diff
@@ -36,6 +44,53 @@ import (
...
+	u2 := *u
+	q := u2.Query()
+
+	decoderName := q.Get("decoder")
+	q.Del("decoder")
+	decoder, err := runtimevar.DecoderByName(ctx, decoderName, o.Decoder)
+	if err != nil {
+		return nil, fmt.Errorf("open variable %v: invalid decoder: %v", u, err)
+	}
+	// See if we changed the query parameters.
```
shantuo: Should it also check for invalid parameters?

runtimevar/httpvar/httpvar.go
vangent: No, we don't... This one explicitly passes through all unknown query parameters.

runtimevar/httpvar/httpvar.go
vangent: No, for this one I think we should pass them through.

PR #1731, opened 2019-03-29
pubsub/rabbit: give up the lock when we're not receiving anything

### LINKED ISSUE #1728: pubsub/rabbit: TestNack hangs sometimes
I've seen this a couple of times, need to repro.

vangent:
I found this hang exceedingly hard to reproduce, but I think the problem is that sometimes we'd end up looping forever in `ReceiveBatch`, holding the lock the whole time. Meanwhile, someone else is trying to ack/nack, but they can't get the lock. It's safer to return, give up the lock, and let the portable type call us back.

Fixes #1728.

### DIFF
--- a/pubsub/rabbitpubsub/rabbit.go
+++ b/pubsub/rabbitpubsub/rabbit.go
@@ -592,11 +592,9 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*dr
 			}
 
 		case <-maxTime:
-			// Timed out. Return whatever we have, or continue waiting if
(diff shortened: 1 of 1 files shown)


chaserchen:
**Hi, I found this change to consume too many cpu resources.**
When not found message, after max_time(50 * time.Millisecond), we will called again, keep looking until found it.
Is it necessary to optimize again?

vangent:
Hi @chaserchen ! I'm going to move this thread to the [bug you opened](https://github.com/google/go-cloud/issues/2103) (and reopen it), since that's a better place to discuss than a closed PR.

PR #1733, opened 2019-03-29
pubsub: add a conformance test for double acking/nacking

### LINKED ISSUE #1726: pubsub: add a conformance test for sending a double ack/nack
For at-least-once providers, it's possible for a `Subscription` to (n)ack a message that has already been acked (by another recipient). Also, on batch APIs, depending on provider semantics during partial errors, it may be possible for retries to retry some (n)acks that were accepted previously.

I think that we should make sure that sending an (n)ack for a message that has already been acked should be silently ignored. Otherwise, we'll either retry (which presumably won't work), or kill the `Subscription`.

We should be able to add a conformance test for this by using the driver APIs directly (the `Subscription` portable type hides ackIDs, and won't let you double-ack).

@jba for any thoughts.

vangent:
Fixes #1726.


pubsub/azuresb/azuresb.go
```diff
@@ -519,8 +518,38 @@ func (s *subscription) updateMessageDispositions(ctx context.Context, ids []driv
...
+			return err
+		}
+	}
+	return nil
+}
+
+// isNotFoundErr returns true if the error is status code 410, Gone.
+// Azure returns this when trying to ack/nack a message that no longer exists.
+func isNotFoundErr(err error) bool {
+	return strings.Contains(err.Error(), "status code 410")
```
vangent: This is a bit hokey, but the err is a `common.Retryable` which doesn't expose the underlying error.

The alternative would be to change to use `RPC` instead of `RetryableRPC`, and duplicate the code for checking the result status code:
https://github.com/Azure/azure-amqp-common-go/blob/master/rpc/rpc.go#L125

pubsub/rabbitpubsub/rabbit.go
```diff
@@ -580,6 +580,11 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*dr
 				}
 				// s.closec must be closed too. See if it has an error.
 				if err := closeErr(s.closec); err != nil {
+					// PreconditionFailed can happen if we send an Ack or Nack for a
```
vangent: @jba This is also super sucky. It looks like when we send acks/nacks, errors are reported back by closing the channel and we get them the next time we call `Receive`; all we can do is reopen it. Suggestions are welcome.

pubsub/drivertest/drivertest.go
```diff
@@ -464,6 +465,88 @@ func testBatching(t *testing.T, newHarness HarnessMaker) {
...
+	// Ack them again; this should succeed even though we've acked them before.
+	// If providers return an error for this, drivers should drop them.
+	err = ds.SendAcks(ctx, []driver.AckID{dms[0].AckID, dms[1].AckID})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Nack all 3 messages. This should also succeed, and the nack of the third
+	// message should take effect, so we should be able to fetch it again.
+	// Some providers may also re-send the first message.
```
jba: I don't really expect consistent behavior from an ack and a nack. In a significantly distributed system (like Google cloud pubsub), I believe they could happen in either order. So I'm not sure it's worth testing that combination.

pubsub/drivertest/drivertest.go
```diff
@@ -464,6 +465,88 @@ func testBatching(t *testing.T, newHarness HarnessMaker) {
...
+	err = ds.SendNacks(ctx, []driver.AckID{dms[0].AckID, dms[1].AckID, dms[2].AckID})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// The test will hang here if the message isn't redelivered, so give it a
+	// shorter timeout.
+	ctx2, cancel := context.WithTimeout(ctx, 5*time.Second)
```
jba: I'm concerned about the time-sensitive nature of this. I suspect 5s will turn out to be too short, but I guess we'll see.

pubsub/rabbitpubsub/rabbit.go
jba: When we do the ack, could we check for channel closure there by selecting from s.closec?

pubsub/drivertest/drivertest.go
vangent: Agree, and this test doesn't test that (explicitly). I clarified this comment to be more clear that whether the first 2 messages are re-sent depends on provider semantics and timing. The 3rd message **must** be redelivered though.

pubsub/drivertest/drivertest.go
vangent: Yea, it will be flaky if not (but only during `record`). It seemed fine to me, but I raised to 15s.

pubsub/rabbitpubsub/rabbit.go
vangent: No; I tried adding a `select` on `s.closec` at the end of `sendAcksOrNacks`, but there was never any error there, even when one later shows up. There must be some delay in closing the channel/sending the error.

PR #1734, opened 2019-03-30
docstore/driver: make NewDocument stricter

jba:
Documents can only be constructed from non-nil struct pointers and maps.

Also, rewrite `Document.Decode` so it looks like `Document.Encode`. This lets us add a check in the top-level `Decode` method that the first argument is settable. (A `Document` that is a map is not settable, but that's never a problem because it will always have its fields set, not the document itself.)


jba:
@shantuo ping

PR #1738, opened 2019-04-01
pubsub/awssnssqs: OpenTopic&OpenSubscription use ConfigProvider

### LINKED ISSUE #1713: pubsub/awssnssqs: OpenTopic&OpenSubscription require a client instead of a session

The `OpenTopic` and `OpenSubscription` functions take in `*sns.SNS` and `*sqs.SQS`, respectively. Other AWS providers (see [`s3blob.OpenBucket`](https://godoc.org/gocloud.dev/blob/s3blob#OpenBucket) and [`awsparamstore.NewVariable`](https://godoc.org/gocloud.dev/runtimevar/awsparamstore#NewVariable)) take in a `client.ConfigProvider` (commonly a `*session.Session`). This inconsistency across packages adds cognitive overhead when using mostly AWS services.

`OpenTopic` and `OpenSubscription` should take in a `client.ConfigProvider` and create the `*sns.SNS` and `*sqs.SQS` objects under the hood.

One could go the other way and make other providers take in their respective client objects. This is somewhat consistent with GCP, but these "client types" in the AWS SDK are largely wrappers around an instantiated configuration. Making the user create these client types is largely boilerplate and doesn't actually introduce meaningful flexibility. Using the configuration seems like the way to go, and is also the lesser delta of breaking changes.

Zyqsempai:
Migrated OpenTopic&OpenSubscription functions to use ConfigProvider: 

Fixes #1713.

### DIFF
--- a/pubsub/awssnssqs/awssnssqs.go
+++ b/pubsub/awssnssqs/awssnssqs.go
@@ -118,8 +118,6 @@ var Set = wire.NewSet(
 	SubscriptionOptions{},
 	TopicOptions{},
 	URLOpener{},
-	sns.New,
-	sqs.New,
 )
 
 // lazySessionOpener obtains the AWS session from the environment on the first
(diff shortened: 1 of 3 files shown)


pubsub/awssnssqs/awssnssqs.go
```diff
@@ -186,9 +186,9 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
 	if err != nil {
 		return nil, fmt.Errorf("open topic %v: %v", u, err)
 	}
-	client := sns.New(o.ConfigProvider, overrideCfg)
```
vangent: Please also remove the `sns.New` and `sqs.New` from `var Set` above, near line 121.

pubsub/awssnssqs/awssnssqs.go
```diff
@@ -186,9 +186,9 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
...
+	o.TopicOptions.ClientConfig = overrideCfg
```
vangent: See https://github.com/google/go-cloud/blob/master/blob/s3blob/s3blob.go#L139 for an example of how to overlay `overrideCfg` on top of `o.ConfigProvider` without using `Options`.

Zyqsempai:
@vangent Done, thanks for your comments.

PR #1739, opened 2019-04-01
pubsub: add a finalizer that complains if a message isn't acked/nacked

### LINKED ISSUE #1454: pubsub: consider putting a finalizer on Message to complain if a message isn't acked
https://golang.org/pkg/runtime/#SetFinalizer

We could `log.Printf` something if a `Message` is finalized without ever having been acked, as a debugging aid.

Down side is that the user may have chosen not to `Ack` on purpose, so this might work better after we support `Nack`.

vangent:
Fixes #1454.

This is similar to the finalizer we have for blob.Reader/Writer:
https://github.com/google/go-cloud/blob/master/blob/blob.go#L658

### DIFF
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -537,8 +537,10 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
 			if s.ackFunc == nil {
 				// Add a finalizer that complains if the Message we return isn't
 				// acked or nacked.
-				_, file, lineno, ok := runtime.Caller(1)
+				_, file, lineno, ok := runtime.Caller(1) // the caller of Receive
 				runtime.SetFinalizer(m, func(m *Message) {
+					m.mu.Lock()
+					defer m.mu.Unlock()
 					if !m.isAcked {
 						var caller string
 						if ok {
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -87,9 +87,11 @@ package pubsub // import "gocloud.dev/pubsub"
 import (
 	"context"
 	"fmt"
+	"log"
 	"math"
 	"net/url"
 	"reflect"
+	"runtime"
 	"sync"
 	"time"
 	"unicode/utf8"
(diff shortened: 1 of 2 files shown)


vangent:
(also added `Ack` to a test that was missing it, detected via this new finalizer!)

pubsub/pubsub.go
```diff
@@ -532,6 +534,20 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
 			m := s.q[0]
 			s.q = s.q[1:]
 			s.throughputCount++
+			if s.ackFunc == nil {
+				// Add a finalizer that complains if the Message we return isn't
+				// acked or nacked.
+				_, file, lineno, ok := runtime.Caller(1)
```
jba: I assume this refers to Receive's caller? Can you document that?

pubsub/pubsub.go
```diff
@@ -532,6 +534,20 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
...
+				runtime.SetFinalizer(m, func(m *Message) {
+					if !m.isAcked {
```
jba: I'm uncomfortable referencing this without the lock. I know, `m` is dead here, it can't possibly have a concurrent writer. Nevertheless. The rule is to lock for all reads and writes, even if you think something is impossible.

pubsub/pubsub.go
```diff
@@ -532,6 +534,20 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
...
+						var caller string
+						if ok {
+							caller = fmt.Sprintf(" (%s:%d)", file, lineno)
+						}
+						log.Printf("A pubsub.Message was never Acked or Nacked%s", caller)
```
jba: I suspect we'll get complaints about this, and we'll end up turning it into a metric. But we can cross that bridge later.

pubsub/pubsub_test.go
```diff
@@ -167,6 +167,7 @@ func TestConcurrentReceivesGetAllTheMessages(t *testing.T) {
 					}
 					return
 				}
+				m.Ack()
```
jba: Sweet!

pubsub/pubsub.go
vangent: Yes, done.

pubsub/pubsub.go

pubsub/pubsub.go
vangent: Ack.

PR #1740, opened 2019-04-01
pubsub/azuresb: cache amqp connection

### LINKED ISSUE #1699: pubsub/azuresb: consider caching amqp client used for acks
https://github.com/google/go-cloud/blob/master/pubsub/azuresb/azuresb.go#L460

We create a client during `SendAcks` and then throw it away; could we keep the client and re-use it?

@vsaroopchand thoughts?

vangent:
Fixes #1699.

cc/ @vsaroopchand

### DIFF
--- a/pubsub/azuresb/azuresb.go
+++ b/pubsub/azuresb/azuresb.go
@@ -301,12 +301,15 @@ func (*topic) ErrorCode(err error) gcerrors.ErrorCode {
 }
 
 type subscription struct {
-	sbSub           *servicebus.Subscription
-	opts            *SubscriptionOptions
-	topicName       string                // Used in driver.subscription.SendAcks to validate credentials before issuing the message complete bulk operation.
-	sbNs            *servicebus.Namespace // Used in driver.subscription.SendAcks to validate credentials before issuing the message complete bulk operation.
-	verifyExists    sync.Once
-	verifyExistsErr error
+	sbSub *servicebus.Subscription
+	opts  *SubscriptionOptions
+
+	// Used for one-time initialization of, lazily called in ReceiveBatch.
+	// The results of initLinkFn are saved in linkErr and amqpLink.
+	initOnce   sync.Once
+	initLinkFn func(context.Context) (*rpc.Link, error)
+	linkErr    error
+	amqpLink   *rpc.Link
 }
 
 // SubscriptionOptions will contain configuration for subscriptions.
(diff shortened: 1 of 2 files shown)


vangent:
I'm not sure if the error about non-existent subscription with this approach is as clear as the one before. I could put back the explicit check using `SubscriptionManager`.

@vsaroopchand FYI, the error for using too-long subscription names was not great:

`frame larger than peer's max frame size` (from `rpc.NewLink`).

pubsub/azuresb/azuresb_test.go
```diff
@@ -72,7 +72,12 @@ func newHarnessUsingAutodelete(ctx context.Context, t *testing.T) (drivertest.Ha
 }
 
 func (h *harness) CreateTopic(ctx context.Context, testName string) (dt driver.Topic, cleanup func(), err error) {
+	// Keep the topic entity name under 50 characters as per Azure limits.
```
vangent: There's a better fix for this in #1741, which should replace this.

pubsub/azuresb/azuresb.go
```diff
@@ -301,12 +301,15 @@ func (*topic) ErrorCode(err error) gcerrors.ErrorCode {
...
-	sbSub           *servicebus.Subscription
-	opts            *SubscriptionOptions
-	topicName       string                // Used in driver.subscription.SendAcks to validate credentials before issuing the message complete bulk operation.
-	sbNs            *servicebus.Namespace // Used in driver.subscription.SendAcks to validate credentials before issuing the message complete bulk operation.
-	verifyExists    sync.Once
-	verifyExistsErr error
+	sbSub *servicebus.Subscription
+	opts  *SubscriptionOptions
+
+	// Used for one-time initialization of, lazily called in ReceiveBatch.
```
jba: "initialialization of" what?

pubsub/azuresb/azuresb.go
```diff
@@ -342,29 +345,41 @@ func openSubscription(ctx context.Context, sbNs *servicebus.Namespace, sbTop *se
 	if opts == nil {
 		opts = &SubscriptionOptions{}
 	}
+	// One-time initialization of Link to AMQP server, lazily initiated in
```
jba: s/Link/link/

pubsub/azuresb/azuresb.go
```diff
@@ -342,29 +345,41 @@ func openSubscription(ctx context.Context, sbNs *servicebus.Namespace, sbTop *se
...
+	// ReceiveBatch so that we delay "invalid topic/subscription" errors until
```
jba: Why couldn't you just call it here, but save the error and report it in ReceiveBatch?

vsaroopchand:
I actually fixed this in the SB SDK; this custom SendAck can use the newly added Bulk Operation. I'll rework SendAck once this is merged in.

https://github.com/Azure/azure-service-bus-go/pull/99

pubsub/azuresb/azuresb.go
vangent: Reworded.

pubsub/azuresb/azuresb.go
vangent: Done.

pubsub/azuresb/azuresb.go
vangent: Done.

vangent:
PTAL

vangent:
> I actually fixed this in the SB SDK; this custom SendAck can use the newly added Bulk Operation. I'll rework SendAck once this is merged in.
> 
> [Azure/azure-service-bus-go#99](https://github.com/Azure/azure-service-bus-go/pull/99)

Cool, SGTM. Does it support bulk nacks as well?

PR #1741, opened 2019-04-01
pubsub/azuresb: make sure to keep topics/subscriptions unique when trimming them

### LINKED ISSUE #1600: pubsub/azurepubsub: conformance tests don't pass in "autodelete" mode
By "autodelete" mode, I mean creating the Service Bus subscription with `servicebus.SubscriptionWithReceiveAndDelete` and using the `AckFuncForReceiveAndDelete` option set to something non-nil.

The conformance test that fails creates a topic, and two subscriptions for that same topic. It looks to me like using "autodelete" mode means that only one of the subscriptions gets the message. The test expects that both will.

vangent:
Fixes #1600.

### DIFF
--- a/pubsub/azuresb/azuresb_test.go
+++ b/pubsub/azuresb/azuresb_test.go
@@ -78,8 +78,9 @@ func newHarnessUsingAutodelete(ctx context.Context, t *testing.T) (drivertest.Ha
 
 func (h *harness) CreateTopic(ctx context.Context, testName string) (dt driver.Topic, cleanup func(), err error) {
 	topicName := sanitize(fmt.Sprintf("%s-top-%d", testName, atomic.AddUint32(&h.numTopics, 1)))
-
-	createTopic(ctx, topicName, h.ns, nil)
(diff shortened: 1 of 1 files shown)


pubsub/azuresb/azuresb_test.go
```diff
@@ -72,7 +77,7 @@ func newHarnessUsingAutodelete(ctx context.Context, t *testing.T) (drivertest.Ha
 }
 
 func (h *harness) CreateTopic(ctx context.Context, testName string) (dt driver.Topic, cleanup func(), err error) {
-	topicName := fmt.Sprintf("%s-topic-%d", sanitize(testName), atomic.AddUint32(&h.numTopics, 1))
+	topicName := sanitize(fmt.Sprintf("%s-top-%d", testName, atomic.AddUint32(&h.numTopics, 1)))
```
jba: Why doesn't the createTopic call below check for errors?

vangent: I don't know; fixed.

Issue #1660, opened 2019-03-25
docstore: matching error specs with mongodb


jba:
Docstore specifies the following behaviors, which seem impossible to satisfy in MongoDB with a single RPC:
- When writing a document with a revision, if the document doesn't exist a NotFound error is returned, but if it exists with the wrong revision then a FailedPrecondition error is returned.
- When deleting a document with a revision, if the document doesn't exist no error is returned, but if it exists with the wrong revision then a FailedPrecondition error is returned.

Mongo doesn't support revisions natively, so the client will manage them itself, by reading and writing the "DocstoreRevision" field of documents. To write or delete a document with a given ID `id` at revision `rev`, we select for the document using the "filter" `{"_id": id, "DocstoreRevision": rev}`. If there is a matching document, this works as intended: the document is written or deleted. But if no document matches, it could be for two reasons: either the document doesn't exist (no matching ID) or it exists at the wrong revision. Mongo doesn't distinguish between these two: it just tells us that there were no matches to our filter. 

We could distinguish the two cases by first looking up the document just by ID. But making two RPCs for every write and delete (that involves revisions) would be unacceptable for performance.

Trying to do the lookup in parallel won't work in general. If we're deleting a document, the lookup would race with the delete. It might work for writes, though.

We could loosen the semantics of writes, saying that a write with a bad revision might return either FailedPrecondition or NotFound. That doesn't help with deletes, though.

We could change the semantics of Delete to return an error on not found. That coupled with relaxing the expected code to FailedPrecondition or NotFound would enable mongo to conform to the spec.

jba:
@vangent @shantuo thoughts?

jba:
One other piece of the puzzle: the reason Delete doesn't return an error is to make it easier to use in bulk actions. Currently, the spec of `ActionList.Do` says that it returns on first error, with the number of successfully completed actions. A Delete that failed if a document didn't exist would be annoying to use in bulk delete situations where some of the documents might happen to already be deleted. But maybe the spec of `Do` is too strong: as @clausti points out, we can't be sure that providers will execute action lists in any kind of order.

So maybe the right solution is for Delete to return an error for nonexistent documents, coupled with relaxing the semantics of `Do` to attempt all the actions, and return errors for each failed action.

vangent:
Another option would be to optimistically try the write/delete; if it succeeds, great, we're done. If it fails, make another `Get` RPC with just the ID to determine which error to return.

vangent:
[...]

PR #1742, opened 2019-04-02
docstore: distinguish ordered from unordered action lists

jba:
- Clarify the meaning of running an ActionList in order.

- Add the ActionList.Unordered method, so users can request out-of-order execution.
  This also affects errors: in Unordered mode, all actions are run regardless of errors.

- To accommodate unordered-mode errors, change ActionList.Do to return a special error
  type that can represent multiple errors along with the corresponding actions' positions.

Implementing Unordered is left for a subsequent PR.

Updates #1660.


internal/docstore/driver/driver.go
```diff
@@ -71,3 +83,28 @@ type Mod struct {
 	FieldPath []string
 	Value     interface{}
 }
+
+// An ActionListError contains all the errors encountered from a call to RunActions,
+// and the positions of the corresponding actions.
+type ActionListError []struct {
```
shantuo: Maybe `ActionListErrors` or `ActionErrorList`?

internal/docstore/docstore.go
```diff
@@ -179,21 +195,36 @@ func (l *ActionList) Update(doc Document, mods Mods) *ActionList {
...
+// If any action fails, the returned error will be an ActionListError that contains
+// the position in the ActionList of each failed action.
+func (l *ActionList) Do(ctx context.Context) error {
 	var das []*driver.Action
-	for _, a := range l.actions {
+	for i, a := range l.actions {
 		d, err := a.toDriverAction()
 		if err != nil {
-			return 0, wrapError(l.coll.driver, err)
+			return ActionListError{{i, wrapError(l.coll.driver, err)}}
```
shantuo: Should we return all the invalid actions after checking through the list?

internal/docstore/driver/driver.go
jba: I think singular (ActionListError) makes a little more sense. It is one `error`, even though it holds multiple things. For comparison, Hashicorp's version of this is called [multierror](https://github.com/hashicorp/go-multierror).

internal/docstore/docstore.go
jba: Good idea. Done.

jba:
PTAL.

internal/docstore/docstore.go
```diff
@@ -179,21 +195,44 @@ func (l *ActionList) Update(doc Document, mods Mods) *ActionList {
...
-// actions and an error. In general there is no way to know which actions succeeded,
-// but the error will contain as much information as possible about the failures.
-func (l *ActionList) Do(ctx context.Context) (int, error) {
+// An ActionListError is returned by ActionList.Do. It contains all the errors
+// encountered while executing the ActionList, and the positions of the corresponding
+// actions.
+type ActionListError = driver.ActionListError
+
+// Do executes the action list.
+// In ordered mode (when the Unordered method was not called on the list), execution
```
vangent: Nit: consider adding a blank link before "In ordered mode" and "In unordered mode"?

internal/docstore/docstore.go
```diff
@@ -179,21 +195,44 @@ func (l *ActionList) Update(doc Document, mods Mods) *ActionList {
...
+// will stop after the first action that fails.
+// In unordered mode, all the actions will be executed. In either mode, as a special
```
vangent: Consider moving this "none of the actions" thing to be first? E.g., 

```
// Do executes the action list.
//
// If any of the actions are invalid (for example, ...), it returns an error, and no actions are executed.
//
// If ordered mode...
//
// In unordered mode...
```

internal/docstore/docstore.go
```diff
@@ -179,21 +195,44 @@ func (l *ActionList) Update(doc Document, mods Mods) *ActionList {
...
+// case, none of the actions will be executed if any is invalid (for example, a Put
+// whose document is missing its key field).
+// If any action fails, the returned error will be an ActionListError that contains
```
vangent: Make it clear if this comment is only for Unordered mode, or for both (again, if it applies to both, consider moving it earlier?).

internal/docstore/docstore.go
```diff
@@ -179,21 +195,44 @@ func (l *ActionList) Update(doc Document, mods Mods) *ActionList {
...
 
-// Do executes the action list. If all the actions executed successfully, Do returns
-// (number of actions, nil). If any failed, Do returns the number of successful
-// actions and an error. In general there is no way to know which actions succeeded,
-// but the error will contain as much information as possible about the failures.
-func (l *ActionList) Do(ctx context.Context) (int, error) {
+// An ActionListError is returned by ActionList.Do. It contains all the errors
+// encountered while executing the ActionList, and the positions of the corresponding
+// actions.
+type ActionListError = driver.ActionListError
```
vangent: I think you should copy the type here; otherwise, the end user has to look at the driver package to see what an `ActionListError` consists of.

internal/docstore/driver/driver.go
vangent: Consider making a type for the struct so that it doesn't have to be inlined everywhere?

internal/docstore/driver/driver.go
```diff
@@ -71,3 +83,28 @@ type Mod struct {
...
+// TODO(jba): use xerrors formatting.
+func (e ActionListError) Error() string {
+	var s []string
+	for _, x := range e {
+		s = append(s, fmt.Sprintf("at %d: %v", x.Index, x.Err))
+	}
+	return strings.Join(s, "; ")
+}
+
+func (e ActionListError) Unwrap() error {
```
vangent: I don't understand what "Unwrap" is for. I find it surprising that it returns `nil` if there's more than one error. What is this used for?

internal/docstore/docstore.go
```diff
@@ -179,21 +195,44 @@ func (l *ActionList) Update(doc Document, mods Mods) *ActionList {
...
 		}
-		das = append(das, d)
 	}
-	n, err := l.coll.driver.RunActions(ctx, das)
-	return n, wrapError(l.coll.driver, err)
+	if len(alerr) > 0 {
+		return alerr
+	}
+	alerr = l.coll.driver.RunActions(ctx, das, l.unordered)
+	if alerr != nil { // Explicit non-nil check is necessary because alerr is not of type error.
```
vangent: Why is the nil check needed? Isn't it always safe to `range` over the slice?

internal/docstore/driver/driver.go
```diff
@@ -18,23 +18,35 @@ package driver // import "gocloud.dev/internal/docstore/driver"
...
-	// RunActions should return immediately after the first action that fails.
-	// The first return value is the number of actions successfully
-	// executed, and the second is the error that caused the action to fail.
+	// If unordered is false, it must appear as if the actions were executed in the
+	// order they appear in the slice, from the client's point of view. The actions
+	// need not happen atomically, nor does eventual consistency in the provider
+	// system need to be taken into account. For example, after a write returns
+	// successfully, the driver can immediately perform a read on the same document,
+	// even though the provider's semantics does not guarantee that the read will see
+	// the write. RunActions should return immediately after the first action that fails.
```
vangent: But, IIUC, it needs to fill in `ErrorNotExecuted` for all the subsequent actions? Maybe this is something we could do in the portable type (for ordered mode only; for unordered mode I think the driver needs to tell us all the info).

Oh I now see you mentioned this below. Maybe it would be more clear to make less optional; e.g., "in ordered mode, just return the first error and its position" (and the portable type will in the `ErrNotExecuted` for the rest); "in unordered mode, always return n errors"?

internal/docstore/driver/driver.go
vangent: OK, now I see it used above. Maybe "SingleError" or something? At least a docstring.

internal/docstore/drivertest/drivertest.go
```diff
@@ -30,6 +30,9 @@ import (
 	"gocloud.dev/internal/docstore/driver"
 )
 
+// TODO(jba): Test RunActions with unordered=true. We can't actually test the ordering,
+// but we can check that all actions are executed even if some fail.
```
vangent: Do we also have a test that does an ordered list of actions, with an expected error in the middle, and ensures it gets the right error back and that the actions after the error aren't executed?

And one that verifies that a malformed action is returned as an error and none of the actions are executed?

(If no to either please add as TODOs).

internal/docstore/driver/driver.go
jba: This is for conformance with the new errors stuff. See https://tip.golang.org/pkg/errors/#Wrapper.

internal/docstore/docstore.go

internal/docstore/docstore.go

internal/docstore/docstore.go

internal/docstore/docstore.go
jba: It is safe to range over a nil slice. But I need the explicit nil check somewhere. If there is no error, I want to return the interface value `(nil, nil)` from this function, not `(nil, ActionListError)`. See https://golang.org/doc/faq#nil_error.

My comment on this line was supposed to explain this. Let me know how I can improve it.

internal/docstore/driver/driver.go
jba: We inline it once. Giving it a name would pollute the API forever at the cost of a tiny bit of implementation pain.

jba: If we find that implementing the unordered mode means writing that struct out a lot, in every driver, then we can just have an unexported function that returns a struct value.

internal/docstore/driver/driver.go
jba: Sorry you had to puzzle through that. It was obsoleted by `ActionListError`. Originally RunActions just returned a `[]error`, which required all these special cases and conventions.

internal/docstore/drivertest/drivertest.go
jba: Added TODOs.

jba:
PTAL.

internal/docstore/docstore.go
vangent: I still think this (the "special case") should be moved up above, or at least not be part of the "unordered mode" paragraph (since it applies to both modes).

internal/docstore/docstore.go
```diff
@@ -180,21 +197,71 @@ func (l *ActionList) Update(doc Document, mods Mods) *ActionList {
...
+// ActionListError that contains the position in the ActionList of each failed
+// action.
+//
+// In ordered mode (when the Unordered method was not called on the list), execution
+// will stop after the first action that fails.
+//
+// In unordered mode, all the actions will be executed. In either mode, as a special
+// case, none of the actions will be executed if any is invalid (for example, a Put
+// whose document is missing its key field).
+func (l *ActionList) Do(ctx context.Context) error {
```
vangent: If this function always returns `ActionListError`, maybe state that more explicitly?

internal/docstore/docstore.go
vangent: Oh I see. I was misled by the comment being right before the `range`.

Maybe:

````
alerr = ...
if len(alerr) == 0 {
  return nil // Explicitly return nil, because alerr is not of type error
}
for i := range ...{}
return alerr
```

I don't feel that strongly about this, but I didn't get the comment on first read.

internal/docstore/docstore.go

internal/docstore/docstore.go
jba: Clarified.

Issue #1332, opened 2019-02-14
docstore: add queries to the portable API


jba:
Support for queries in the `docstore` and `driver` packages.

Initially, drivers should return an "unimplemented" error.

jba:
Done a while ago.

PR #1745, opened 2019-04-02
internal/docstore: add driver interface and types for query

shantuo:
Updates #1332


internal/docstore/docstore.go
```diff
@@ -31,7 +31,7 @@ import (
 //
 // A Document can be represented as a map[string]int or a pointer to a struct. For
 // structs, the exported fields are the document fields.
-type Document = interface{}
+type Document interface{}
```
jba: Just curious, why did you make this change?

internal/docstore/driver/driver.go
```diff
@@ -71,3 +75,28 @@ type Mod struct {
...
 }
+
+// A Query defines a query operation to find documents within a collection based
+// on a set of requirements.
+type Query struct {
+	FieldPaths []string         // the selected field paths
+	Filters    []Filter         // a list of filter for the query
+	Limit      int              // the number of result in one query request
+	StartAt    interface{}      // define the starting point of the query
+	Itr        DocumentIterator // the iterator for Get action
```
jba: nit: s/Itr/Iter/

internal/docstore/docstore.go
shantuo: I wasn't sure why it was necessary to do an alias vs a type definition? Is that a better practice?

jba: If `Document` is a separate type, then anyone calling `Get`, `Put`, etc. will have to cast their struct or map to it. 

I made it an alias so that it basically serves as documentation, that `Get` etc. take Go types that are valid documents. 

We can discuss this further, but in any case it shouldn't be part of this PR.

shantuo: Reverted.

internal/docstore/driver/driver.go
shantuo: Done.

jba:
The tests are failing because of a change I made. Just comment out the TestConformance/BlindCodec test. I will fix in the next PR.

vangent:
Next time, if you ask for me as a reviewer, please wait until I review :-).

internal/docstore/driver/driver.go
```diff
@@ -71,3 +75,28 @@ type Mod struct {
 	FieldPath []string
 	Value     interface{}
 }
+
+// A Query defines a query operation to find documents within a collection based
+// on a set of requirements.
+type Query struct {
+	FieldPaths []string         // the selected field paths
```
vangent: IMHO these comments need more details.

E.g.,
* for `FieldPath`, add something like: "The returned Documents should only have these fields."
* for `Filters`, add that the filters should be ANDed together
* for `Limit`, can the driver return fewer than Limit even if more are available?
* for `StartAt`, explain where this value will come from; presumably it is from something that the driver previously returned?

internal/docstore/driver/driver.go
```diff
@@ -71,3 +75,28 @@ type Mod struct {
...
+	Filters    []Filter         // a list of filter for the query
+	Limit      int              // the number of result in one query request
+	StartAt    interface{}      // define the starting point of the query
+	Iter       DocumentIterator // the iterator for Get action
+}
+
+// A Filter defines a filter expression used to filter the query result.
+type Filter struct {
+	Field string      // the field path to filter
+	Op    string      // the operation, supports =, >, >=, <, <=
```
vangent: Consider making an enum for this? Not sure it's worth it.

internal/docstore/driver/driver.go
```diff
@@ -71,3 +75,28 @@ type Mod struct {
...
+	Value interface{} // the value to filter
```
vangent: Is there somewhere where we explain what the operators mean for different types? For example, what does it mean for a struct to be `>` another struct?

internal/docstore/driver/driver.go
```diff
@@ -71,3 +75,28 @@ type Mod struct {
...
+}
+
+// A DocumentIterator is an iterator that can iterator through the results (for Get
```
vangent: `A DocumentIterator iterates through...`

internal/docstore/driver/driver.go
```diff
@@ -71,3 +75,28 @@ type Mod struct {
...
+// action), and returns the next page that can be passed into a Query as a
+// starting point.
+type DocumentIterator interface {
```
vangent: Why have an iterator for the driver? For blob, we exposed an iterator on the portable type, but the driver method is "get next page".

internal/docstore/driver/driver.go
```diff
@@ -71,3 +75,28 @@ type Mod struct {
...
 	Value     interface{}
 }
+
+// A Query defines a query operation to find documents within a collection based
+// on a set of requirements.
+type Query struct {
+	FieldPaths []string         // the selected field paths
+	Filters    []Filter         // a list of filter for the query
+	Limit      int              // the number of result in one query request
+	StartAt    interface{}      // define the starting point of the query
```
vangent: I don't see where a caller would get this value, or what it represents?

internal/docstore/driver/driver.go
jba: In Firestore, querying is a streaming RPC.

The Go mongoDB driver exposes an iterator.

internal/docstore/driver/driver.go
jba: We can remove it for now, and figure out how to start a query at a particular place.

internal/docstore/driver/driver.go
jba: I don't think it's worth it.

internal/docstore/driver/driver.go
jba: I doubt we could ever discover the answer to your `Limit` question without reading the server source code. It would be unlikely for that information to be in documentation (for example, the Firestore RPC doc just says that limit is "the maximum number of results to return") and I can't imagine how we could reliably test for it.

vangent: This is driver documentation; it should specify what acceptable behavior for drivers is, and conformance tests should verify that behavior. This is similar to `PageSize` for blob:
https://github.com/google/go-cloud/blob/master/blob/driver/driver.go#L152

```
// PageSize sets the maximum number of objects to be returned.
// 0 means no maximum; driver implementations should choose a reasonable
// max. It is guaranteed to be >= 0.
```

internal/docstore/driver/driver.go
vangent: But, it is easy to implement returning pages on top of an iterator. It is hard to implement an iterator on top of pages. If all drivers support iterators natively, then it's probably fine to have that be the driver interface, but if not, I think it's more straightforward to use pages.

I don't see how this is any different than `List` for blob -- there also, GCS supports an iterator, but other drivers do not.

internal/docstore/driver/driver.go
vangent: Moved to #1750.

internal/docstore/driver/driver.go
vangent: Moved to #1750.

PR #1746, opened 2019-04-02
pubsub/awssnssqs: Changed UrlOpener scheme to awssns and awssqs

### LINKED ISSUE #1707: pubsub/awssnssqs: use different URL schemes for topic and subscription
/cc @jba for thoughts

The topic/publishing side of the `awssnssqs` package uses AWS SNS, and the subscription side uses AWS SQS. However, both types of URLs use the scheme `awssnssqs`. In order to know which service a URL is actually using, you have to consult the documentation and understand it.

To me, it would seem clearer to use `sns` and `sqs` respectively. This would immediately read that one side is connecting to one service or the other.

Zyqsempai:
Changed pubsub/awssnssqs UrlOpener scheme from `awssnssqs` to `sns` and `sqs` respectively.
Fixes #1707

### DIFF
--- a/pubsub/awssnssqs/awssnssqs.go
+++ b/pubsub/awssnssqs/awssnssqs.go
@@ -18,7 +18,7 @@
 // URLs
 //
 // For pubsub.OpenTopic and pubsub.OpenSubscription, awssnssqs registers
-// for the scheme "awssnssqs".
+// for the scheme's "awssns" and "awssqs" respectively.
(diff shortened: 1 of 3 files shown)


vangent:
This is a backwards-incompatible change:

```
Incompatible changes:
- Scheme: removed
Compatible changes:
- SNSScheme: added
- SQSScheme: added
```

so you'll need to add a commit with `BACKWARDS_CHANGE_OK` in the first line of the commmit message.

PR #1747, opened 2019-04-02
mysql: Add Terraform Automation for Azure MySQL 

### LINKED ISSUE #1737: mysql: We are missing terraform file for azure-mysql
### Describe the bug

We are missing terraform file for new azure MySQL implementation.

## Expected behavior

There are should be `main.tf` file in azuremysql folder to run the tests.

vsaroopchand:
Fixes #1737 ; adds to #76

### DIFF
--- a/mysql/azuremysql/main.tf
+++ b/mysql/azuremysql/main.tf
@@ -1,13 +1,12 @@
-
 terraform {
   required_version = "~>0.11"
 }
 
-provider "azurerm" {  
+# See documentation for more info: https://www.terraform.io/docs/providers/azurerm/auth/azure_cli.html
+provider "azurerm" {
   version = "~> 1.22"
-  # See documentation for more info: https://www.terraform.io/docs/providers/azurerm/auth/azure_cli.html
-  subscription_id = "00000000-0000-0000-0000-000000000000"
 }
+
 provider "random" {
   version = "~> 2.1"
 }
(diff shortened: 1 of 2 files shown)


eliben:
Thanks Vishal. I'm assigning to @zombiezen to review when back from vacation (next week)

mysql/azuremysql/main.tf
```diff
@@ -0,0 +1,94 @@
+
+terraform {
+  required_version = "~>0.11"
+}
+
+provider "azurerm" {  
+  version = "~> 1.22"
+  # See documentation for more info: https://www.terraform.io/docs/providers/azurerm/auth/azure_cli.html
+  subscription_id = "00000000-0000-0000-0000-000000000000"
```
zombiezen: IIUC the user of this Terraform setup would have to edit this line to the correct subscription ID. I believe if you remove this, then it will either prompt or use an environment variable.

mysql/azuremysql/main.tf
```diff
@@ -0,0 +1,94 @@
...
+
+  special = false
+  length  = 20
+}
+resource "azurerm_resource_group" "mysqlrg" {
+  name     = "${var.resourcegroup}"
+  location = "${var.location}"
+}
+resource "azurerm_mysql_server" "mysqlserver" {
+  name                = "${var.servername}"
```
zombiezen: If this must be unique, then generating a random suffix for something like "go-cdk-test-" seems easier for the user rather than prompting for it.

mysql/azuremysql/main.tf
```diff
@@ -0,0 +1,94 @@
+
```
zombiezen: nit: remove initial blank line.

(Have you run this through `terraform fmt`?)

mysql/azuremysql/main.tf
vsaroopchand: Correct, it will use the default subscription (if you have more than one). I'll remove it.

mysql/azuremysql/main.tf
vsaroopchand: Ok, servername is auto generated using your suggested prefix.

mysql/azuremysql/main.tf
vsaroopchand: Done.

vsaroopchand:
PTAL

PR #1748, opened 2019-04-02
docstore: support get with field paths

jba:
Support the ability to retrieve a subset of document fields.

- Changed the driver codec framework to re-use existing maps and
  structs rather than creating new ones.

- Wrote conformance tests.

- Added support to all the drivers.

See


internal/docstore/driver/codec.go
```diff
@@ -369,6 +370,9 @@ func Decode(v reflect.Value, d Decoder) error {
 }
 
 func decode(v reflect.Value, d Decoder) error {
+	if !v.CanSet() {
+		panic(fmt.Sprintf("cannot set %+v", v))
```
shantuo: Maybe return an error instead of panicking. If we don't do this it will also panic when we try to set the value right?

internal/docstore/memdocstore/codec.go
```diff
@@ -76,8 +77,25 @@ func (e *mapEncoder) MapKey(k string) { e.m[k] = e.val }
 ////////////////////////////////////////////////////////////////
 
 // decodeDoc decodes m into ddoc.
-func decodeDoc(m map[string]interface{}, ddoc driver.Document) error {
-	return ddoc.Decode(decoder{m})
+func decodeDoc(m map[string]interface{}, ddoc driver.Document, fps [][]string) error {
+	var m2 map[string]interface{}
+	if fps == nil {
```
shantuo: This should be `if len(fps) == 0 {`

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -158,13 +158,27 @@ func (c *collection) get(ctx context.Context, doc driver.Document, fieldpaths []
...
 	in := &dyn.GetItemInput{
 		TableName: &c.table,
 		Key:       av.M,
 	}
+	if len(fieldpaths) > 0 {
+		// Construct a projection expression for the field paths.
+		// See https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ProjectionExpressions.html.
+		nbs := []expression.NameBuilder{expression.Name(docstore.RevisionField)}
+		for _, fp := range fieldpaths {
+			nbs = append(nbs, expression.Name(strings.Join(fp, ".")))
```
shantuo: Forgot about this probably, why do we store a fieldpath as a `[]string`?

internal/docstore/driver/codec.go
jba: Right.

Done.

internal/docstore/memdocstore/codec.go
jba: Done.

internal/docstore/dynamodocstore/dynamo.go
jba: For the convenience of drivers, so they don't have to parse and validate a dot-separated string.

jba:
PTAL.

PR #1750, opened 2019-04-02
internal/docstore: add query types and methods to portable API

shantuo:
Updates #1332 

- Add `Query` and `DocumentIterator` structs
- Add `Get` for query.
- Add context to iterator's `Next` method as we will need to make new query request when a page is finished.


internal/docstore/query.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// Query creates a new Query over the collection.
+func (c *Collection) Query() *Query {
+	return &Query{coll: c, dq: &driver.Query{}}
+}
+
+// Where expresses a condition on the query.
+// Valid ops are: "=", ">", "<", ">=", "<=".
+func (q *Query) Where(field, op string, value interface{}) *Query {
```
jba: s/field/fieldpath/

Call parseFieldPath.


The Query type should have an `err` field that remembers the error; it will be returned immediately by the  `Delete` and `Update` methods, and the `Get` method will copy it into the iterator.

internal/docstore/query.go
```diff
@@ -0,0 +1,89 @@
...
+	q.dq.Limit = n
+	return q
+}
+
+// Get returns an iterator for retrieving the documents specified by the query. If
+// field paths are provided, only those paths are set in the resulting documents.
+//
+// Call Stop on the iterator when finished.
+func (q *Query) Get(ctx context.Context, fieldpaths ...string) *DocumentIterator {
+	_ = q.coll.driver.RunQuery(ctx, q.dq)
```
jba: The field paths should be parsed and stored in the driver.Query.

jba:
I don't know if you've seen #1749 yet. It makes some changes I think you'll want.

internal/docstore/query.go
```diff
@@ -0,0 +1,89 @@
...
+	return &DocumentIterator{iter: q.dq.Iter}
```
jba: Instead of passing the iterator back through a field in the Query, could it be the first return value?

jba: Also, we haven't talked about how to handle delete and update queries in the driver. What about three separate methods?
```
RunGetQuery(ctx, query) (DocumentIterator, error)
RunUpdateQuery(ctx, query) error
RunDeleteQuery(ctx, query) error
```

jba: Sorry, that should be
```
RunUpdateQuery(ctx, query, mods) error
```

internal/docstore/query.go
```diff
@@ -0,0 +1,89 @@
...
+func (q *Query) Limit(n int) *Query {
+	q.dq.Limit = n
+	return q
+}
+
+// Get returns an iterator for retrieving the documents specified by the query. If
+// field paths are provided, only those paths are set in the resulting documents.
+//
+// Call Stop on the iterator when finished.
+func (q *Query) Get(ctx context.Context, fieldpaths ...string) *DocumentIterator {
```
vangent: WDYT about having the fieldpaths specified in a separate method? E.g., `Select` or something, so `Get` always just takes ctx?

internal/docstore/query.go
```diff
@@ -0,0 +1,89 @@
...
+// Limit will limit the results to at most n documents.
+func (q *Query) Limit(n int) *Query {
+	q.dq.Limit = n
+	return q
+}
+
+// Get returns an iterator for retrieving the documents specified by the query. If
+// field paths are provided, only those paths are set in the resulting documents.
+//
+// Call Stop on the iterator when finished.
```
vangent: Why is `Stop` needed? Does it need to be called even if you iterate to the end?

WDYT about `Close` instead?

internal/docstore/query.go
```diff
@@ -0,0 +1,89 @@
...
+type DocumentIterator struct {
+	iter driver.DocumentIterator
+	err  error
+}
+
+// Next stores the next document in dst. It returns io.EOF if there are no more
+// documents.
+// Once Next returns an error, it will always return the same error.
+func (it *DocumentIterator) Next(ctx context.Context, dst Document) error {
```
vangent: Why not `Next(ctx) (*Document, error)` similar to blob?

internal/docstore/query.go
```diff
@@ -0,0 +1,89 @@
...
+	if it.err != nil {
+		return it.err
+	}
+	ddoc, err := driver.NewDocument(dst)
+	if err != nil {
+		it.err = err
+		return err
+	}
+	it.err = it.iter.Next(ctx, ddoc)
```
vangent: See my comments on your driver PR; why not have the driver method be "get next page" (instead of forcing them all to implement an iterator)? Do they all already support iterators?

internal/docstore/query.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Query represents a query over a collection.
+type Query struct {
+	coll *Collection
+	dq   *driver.Query
+}
+
+// Query creates a new Query over the collection.
+func (c *Collection) Query() *Query {
+	return &Query{coll: c, dq: &driver.Query{}}
```
vangent: How does `driver.Query.StartAt` ever get filled in?

internal/docstore/query.go
shantuo: I was originally plan to have the `RunQuery` to be similar to `RunActions`, where the `Action` is a field of `driver.Query`, which is also the reason I pass the iterator back through a field. I feel that seems more consistent and simpler for the driver interface.

Is there any other reason to have three separate methods here?

internal/docstore/query.go
jba: This way `Get` is consistent with `ActionList.Get`. One less thing to remember.

internal/docstore/query.go
jba: Some providers may need to do resource cleanup if the user stops iterating in the middle. Firestore is one.

Naming a method `Close` without an error return is confusing to readers. Brad Fitzpatrick and I had a long discussion about this way back when I was writing the [iterator guidelines](https://github.com/googleapis/google-cloud-go/wiki/Iterator-Guidelines#the-close-and-stop-methods). I now agree with him: readers see a call to `Close` without an error check and get nervous. 

Why not `Close() error`? I don't believe any of the providers need to report an error back. That may change as we implement more.

internal/docstore/query.go
jba: You need to pass in the document to do type-directed decoding. (When the document is a struct.)

internal/docstore/query.go
jba: Answered there.

internal/docstore/query.go
jba: We'll eventually have a `StartAt` method on Query, but we don't know what that's going to look like yet (or even if it's possible to do portably).

shantuo: It gets filled in when the driver needs to fetch a new page when user calls `DocumentIterator.Next`. We don't surface pagination at the portable API level.

internal/docstore/query.go
shantuo: Some driver (firestore) needs to clean up the resource.

My understanding is that `Stop` doesn't return an error, like `Close` does, and doesn't need to be called if iterating to the end? Is that correct? @jba 

I'll add some documentation.

internal/docstore/query.go
shantuo: It needs the type to be passed in for type-driven decoding.

internal/docstore/query.go
shantuo: Let me think about this more and sketch out the interface.

internal/docstore/query.go
jba: No, it was just because of the different signatures.

I'm okay with using an enum but I'd still rather put the iterator in the return values. It's odd to pass something back in an argument. I know that for `Get` we populate the argument doc, but we need the arg to do type-driven decoding, and it's also forced by the fluent API. Neither of those things are at play here.

internal/docstore/query.go
jba: Let's leave it as an iterator. Both firestore and mongo already have iterators. Dynamo does have pagination, but on balance I think the iterator API will be less coding and less clumsiness overall.

internal/docstore/query.go
```diff
@@ -0,0 +1,132 @@
...
+	return q
+}
+
+// Get returns an iterator for retrieving the documents specified by the query. If
+// field paths are provided, only those paths are set in the resulting documents.
+//
+// Call Stop on the iterator when finished.
+func (q *Query) Get(ctx context.Context, fieldpaths ...string) *DocumentIterator {
+	for _, fp := range fieldpaths {
+		fp, err := parseFieldPath(FieldPath(fp))
```
jba: Change the arg type to FieldPath, as in ActionList.Get.

internal/docstore/query.go
```diff
@@ -0,0 +1,132 @@
...
+		if err != nil {
```
jba: We don't need to accumulate the errors. We can stop at the first error.

internal/docstore/query.go
```diff
@@ -0,0 +1,132 @@
...
+// Stop stops the iterator, the state of the iterator will be lost and calling
+// Next will return io.EOF.
+//
+// Stop only needs to be called when you want to terminate the iterator early.
+func (it *DocumentIterator) Stop() {
+	it.err = io.EOF
+	it.iter.Stop()
+}
+
+// QueryError is a list of errors occurred during the assembly of a Query.
```
jba: I don't think this is necessary. Let's just return the first error we find.

Lists of errors are problematical in Go, and that's actually going to get worse with the new `errors.Is/As` proposal. I only did it for action lists because there was no other way to let the user know which of their actions had failed. Here, if we just return the first error that makes the query bad, then that's fine—the user has to fix the query and try again anyway.

internal/docstore/query.go
shantuo: Done.

internal/docstore/query.go
shantuo: Removed.

internal/docstore/query.go
shantuo: Changed to `RunGetQuery`. I think returning an iterator looks good.

internal/docstore/driver/driver.go
```diff
@@ -90,11 +90,10 @@ type ActionListError []struct {
...
 type Query struct {
-	FieldPaths [][]string       // the selected field paths
-	Filters    []Filter         // a list of filter for the query
-	Limit      int              // the number of result in one query request
-	StartAt    interface{}      // define the starting point of the query
-	Iter       DocumentIterator // the iterator for Get action
+	FieldPaths [][]string  // the selected field paths
+	Filters    []Filter    // a list of filter for the query
+	Limit      int         // the number of result in one query request
+	StartAt    interface{} // define the starting point of the query
```
vangent: Let's leave `StartAt` off for now, since we don't know how it will be implemented.

internal/docstore/driver/driver.go
```diff
@@ -90,11 +90,10 @@ type ActionListError []struct {
 // A Query defines a query operation to find documents within a collection based
 // on a set of requirements.
 type Query struct {
-	FieldPaths [][]string       // the selected field paths
-	Filters    []Filter         // a list of filter for the query
-	Limit      int              // the number of result in one query request
-	StartAt    interface{}      // define the starting point of the query
-	Iter       DocumentIterator // the iterator for Get action
+	FieldPaths [][]string  // the selected field paths
```
vangent: IMHO these comments need more details.

E.g.,
* for `FieldPath`, add something like: "The returned Documents should only have these fields."
* for `Filters`, add that the filters should be ANDed together
* for `Limit`, is the driver allowed to return `EOF` before `Limit` is reached even if more are available?

internal/docstore/driver/driver.go
```diff
@@ -104,10 +103,21 @@ type Filter struct {
 	Value     interface{} // the value to compare using the operation
```
vangent: Is there somewhere where we explain what the operators mean for different types? For example, what does it mean for a struct to be > another struct?

internal/docstore/driver/driver.go
```diff
@@ -104,10 +103,21 @@ type Filter struct {
...
 }
 
-// A DocumentIterator is an iterator that can iterator through the results (for Get
-// action), and returns the next page that can be passed into a Query as a
-// starting point.
+// A DocumentIterator iterates through the results (for Get action), and returns
+// the next page that can be passed into a Query as a starting point.
```
vangent: It's not returning pages....?

internal/docstore/driver/driver.go
```diff
@@ -104,10 +103,21 @@ type Filter struct {
...
 type DocumentIterator interface {
-	Next(doc Document) error
+
+	// Next get the next item in the query result and calls decode to decode into
```
vangent: "Next gets...

What is `decode`? It's not defined in this package.

internal/docstore/driver/driver.go
```diff
@@ -104,10 +103,21 @@ type Filter struct {
...
+	// Document.
+	//
+	// When it is called when the iterator is at the end of a page, it should go
```
vangent: Again, not sure why you're talking about pages here. The interface doesn't use pages.

internal/docstore/driver/driver.go
```diff
@@ -104,10 +103,21 @@ type Filter struct {
...
+	// fetch another page if it is possible, though it does not need to guarantee
+	// to have a result.
+	//
+	// When there is no result to return, returns a io.EOF.
```
vangent: Nit: "When there are no more results, to return, it should return io.EOF."

internal/docstore/driver/driver.go
```diff
@@ -104,10 +103,21 @@ type Filter struct {
...
+	Next(context.Context, Document) error
+
+	// Stop cleans up resources of the iterator, if any. It is only expected to be
+	// called when user wants to terminate the iterator early.
```
vangent: How about:

```
// Stop terminates the iterator before Next return io.EOF, allowing any cleanup needed.
// After Stop is called, Next should return an error.
```

?

internal/docstore/query.go
```diff
@@ -0,0 +1,108 @@
...
+			return &DocumentIterator{err: q.err}
+		}
+		q.dq.FieldPaths = append(q.dq.FieldPaths, fp)
+	}
+	it, err := q.coll.driver.RunGetQuery(ctx, q.dq)
+	return &DocumentIterator{iter: it, err: err}
+}
+
+// DocumentIterator iterates over documents.
+// Call Stop on the iterator when finished.
```
vangent: This should be updated to indicate you only need to call it if io.EOF hasn't been returned yet.

The portable type should enforce this too (e.g., if `Stop` is called, but `io.EOF` has been returned, don't call the driver's `Stop`).

internal/docstore/driver/driver.go
shantuo: Clarified, I was thinking that the driver implementation may do a query in a page size smaller than limit.

internal/docstore/driver/driver.go
shantuo: Done.

internal/docstore/driver/driver.go
shantuo: I did the latter part (next returning io.EOF when stop is called) through the portable API.

internal/docstore/query.go
shantuo: Done.

shantuo:
PTAL

internal/docstore/driver/driver.go
```diff
@@ -90,11 +90,18 @@ type ActionListError []struct {
 // A Query defines a query operation to find documents within a collection based
 // on a set of requirements.
 type Query struct {
-	FieldPaths [][]string       // the selected field paths
-	Filters    []Filter         // a list of filter for the query
-	Limit      int              // the number of result in one query request
-	StartAt    interface{}      // define the starting point of the query
-	Iter       DocumentIterator // the iterator for Get action
+	// FieldPaths contain a list of field paths user selects to return in the query
```
vangent: nit: "the user"

internal/docstore/driver/driver.go
```diff
@@ -90,11 +90,18 @@ type ActionListError []struct {
...
+	// results. The returned documents should only have these fields populated.
+	FieldPaths [][]string
+
+	// Filters contain a list of filters for the query. If there are more than one
+	// filter, they should be combined with AND.
+	Filters []Filter
+
+	// Limit sets the maximum number of results returned by one query request. It
+	// is always non-negative. When it <= 0, the driver implementation should
```
vangent: The semantics of this are unclear to me. It feels like the comment is about retrieving a page of result, but the API is not paginated. The `Limit` seems like it should be specifically about capping the maximum number of times you can call `Next` in the return iterator, right? And the driver must always return all of the matching results, but if there are more than `Limit`, it can stop.

It also seems like `0` should mean "no limit", not a reasonable default. If the driver is using a paginated implementation, it can decide how big the pages it gets are.

internal/docstore/driver/driver.go
vangent: Can you document that here?

internal/docstore/driver/driver.go
```diff
@@ -104,10 +111,21 @@ type Filter struct {
...
 }
 
-// A DocumentIterator is an iterator that can iterator through the results (for Get
-// action), and returns the next page that can be passed into a Query as a
-// starting point.
+// A DocumentIterator iterates through the results (for Get action).
 type DocumentIterator interface {
-	Next(doc Document) error
+
+	// Next tries to get the next item in the query result and decode into Document
```
vangent: nit: decodes

internal/docstore/driver/driver.go
```diff
@@ -104,10 +111,21 @@ type Filter struct {
...
+	// with the driver's codec.
+	//
+	// When the driver makes the query request for a page size smaller than the
```
vangent: I would just delete this paragraph; it's talking about pages, when the interface is not paginated.

internal/docstore/driver/driver.go
shantuo: Done.

internal/docstore/driver/driver.go
shantuo: Done.

PR #1751, opened 2019-04-02
server: add an http.Handler argument to the server constructor

### LINKED ISSUE #1104: server: take in http.Handler as constructor argument
In current usage of `*server.Server` with Wire, there is a frequent need to inject both an "application" object and a "server" object in main. This results in awkwardness like placing the `*server.Server` as a field inside the "application" object:

https://github.com/google/go-cloud/blob/2e562e6592295685adfe7376f66fec81f76505c2/samples/guestbook/main.go#L121-L131

Or in the more extreme case, creating a tuple struct just for the purpose of having both:

https://github.com/google/go-cloud/blob/2e562e6592295685adfe7376f66fec81f76505c2/internal/contributebot/setup.go#L38-L69

All that `func main` ought to really care about is having the initialized `*server.Server` object so that it can call `ListenAndServe`. This would provide a better server setup experience.

(Fixing this might simplify some of the semantics of #516.)

clausti:
and updates the `guestbook` sample to use the new constructor, which enables dependency injection of the application into the router so it can register routes before itself being injected into the server, leaving the `main` function to only call `ListenAndServe`

fixes #1104


samples/guestbook/wire_gen.go
```diff
@@ -75,7 +77,7 @@ func setupAWS(ctx context.Context, flags *cliFlags) (*application, func(), error
 		DefaultSamplingPolicy: sampler,
 		Driver:                defaultDriver,
 	}
-	serverServer := server.New(serverOptions)
+	serverServer := server.New(router, serverOptions)
```
clausti: router successfully added to `setup` functions

samples/guestbook/main.go
```diff
@@ -107,14 +107,18 @@ func main() {
 	defer cleanup()
 
 	// Set up URL routes.
-	r := mux.NewRouter()
+	r := app.srv.Handler
+	// r := mux.NewRouter()
+
+	// these all error, saying http.Handler has no method
+	// so is the mux.NewRouter actually getting injected?
 	r.HandleFunc("/", app.index)
```
clausti: These three lines all error with 
```
./main.go:118:3: r.HandleFunc undefined (type http.Handler has no field or method HandleFunc)
./main.go:119:3: r.HandleFunc undefined (type http.Handler has no field or method HandleFunc)
./main.go:120:3: r.HandleFunc undefined (type http.Handler has no field or method HandleFunc)
```

Can anyone help me figure out why the gorrilla/mux router is not actually getting injected? (It is in the wiregen file, see: [other comment](https://github.com/google/go-cloud/pull/1751/files#r271899065) @shantuo @jba @vangent

samples/guestbook/wire_gen.go
```diff
@@ -34,13 +37,12 @@ import (
...
 	"gocloud.dev/server/sdserver"
 	"gocloud.dev/server/xrayserver"
-	"google.golang.org/genproto/googleapis/cloud/runtimeconfig/v1beta1"
-	"net/http"
 )
 
 // Injectors from inject_aws.go:
 
 func setupAWS(ctx context.Context, flags *cliFlags) (*application, func(), error) {
+	router := mux.NewRouter()
```
clausti: `router` is a `mux.NewRouter()`, same as the original code

samples/guestbook/main.go
jba: Sorry, I'm not familiar with wire.

vangent: I don't think this is a wire problem. Before your change, `r` was a `mux.Router`; after this, r is a `http.Handler`. It looks like you are expecting `wire.Bind` to inject the mux.Router into the `src.Handler` field, and maybe that works because mux.Router implements `http.Handler`? But in this code, the type is still `http.Handler`, which doesn't have a `HandleFunc` method.
HTH

clausti: yeah, `mux.Router` implements `http.Handler`. If you look at the other comments, you can see that the `router` is the `mux.NewRouter()` and it's getting passed to the server constructor. Then on [old line 94, new line 96](https://github.com/google/go-cloud/pull/1751/files#diff-f3a898cf69ea8307808637876beeefccR96) it's getting passed to the construction of the application. There are similar lines for all of the 'setup' functions. The `app` struct is getting created with those setup functions, in every case except for the `log.Fatalf` case, [which exits](https://golang.org/pkg/log/#Fatalf). So as far as I can tell, there's no code path that doesn't get injected with the `mux.Router`, so there's no way for it to ever be the zero type of `http.Handler` there, but I am still getting this error.

clausti:
Travis is still failing due to tests in `contributebot`, which is a slightly more complex case than the sample or test apps. @zombiezen let's look tomorrow?

server/server.go
```diff
@@ -44,6 +44,7 @@ var Set = wire.NewSet(
 // The zero value is a server with the default options.
 type Server struct {
 	reqlog        requestlog.Logger
+	Handler       http.Handler
```
eliben: Why is this field exported and others aren't?

samples/guestbook/main.go
```diff
@@ -123,22 +117,30 @@ var applicationSet = wire.NewSet(
 	newApplication,
 	appHealthChecks,
 	trace.AlwaysSample,
+	NewRouter,
+	wire.Bind((*http.Handler)(nil), (*mux.Router)(nil)),
 )
 
+func NewRouter(app *application) *mux.Router {
```
zombiezen: nit: No need to export (it's in `package main`), so let's keep this unexported to match with surrounding code.

(Same comment applies to other NewRouter functions.)

server/server.go
clausti: the intent was to allow the handler to be assigned after construction, since the previous flow was to pass it in to `ListenAndServe`.

clausti: @eliben it's not clear to me if this was a imperative to change this to be unexported or if the question implies the existence of potentially valid reason(s) to export it, can you clarify?

eliben: I was just wondering because it's different from the others. Your comment makes sense to me, but on the other hand it's hard to imagine why handlers would be changed for a server after creation. This is consistent with `http.Server` which is good, but then `http.Server` has all its fields public and here we hide all but `Handler`.
In any case I don't feel strongly about this and will defer to you and @zombiezen for a decision

zombiezen: I was curious about this field being exported because it does introduce potential for it to be mutated, which I don't see a use case for yet. (I tend to lean on the side of keeping as much unexported as possible until someone has a concrete need, and then explicitly design for that.)

clausti: unexported it, PTAL

internal/contributebot/main.go
```diff
@@ -292,8 +292,13 @@ func (w *worker) ghClient(installID int64) *github.Client {
...
 }
 
-// ServeHTTP serves a page explaining that this port is only open for health checks.
-func (w *worker) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
+func (w *worker) CheckHealth() error {
+	return w.auth.CheckHealth()
+}
+
+// frontPage serves a page explaining that this port is only open for health checks.
+//func (w *worker) ServeHTTP(resp http.ResponseWriter, req *http.Request) {
```
zombiezen: Remove this line

Issue #1743, opened 2019-04-02
pubsub/tests: why isn't the lack of ack complained about for As?


vangent:
The conformance test for "As" doesn't call `Ack` on the message it receives.

-- Fix that.
-- Figure out why the new finalizer doesn't complain about this.

jba:
> Figure out why the new finalizer doesn't complain about this.

You might just exit before the first GC.

PR #1755, opened 2019-04-03
pubsub: report ack error at Shutdown if it hasn't been returned earlier

### LINKED ISSUE #1744: pubsub: Shutdown should return error if acking failed and it hasn't already been reported
I noticed recently that we have several tests for which `Ack` failed, but we didn't notice because the error is only reported during the next call to `Receive`, and the test never calls `Receive` again.

Examples:
* `azuresb` was failing for one of the `As` tests because the test name was too long (being fixed in #1741).
* I added `Ack` to the `As` test for #1743 and the replay tests didn't fail even though they are now calling `Ack` and there's nothing in the record file about it.

`pubsub.Subscription.Shutdown` should return the error from `Ack` if it hasn't been reported already, and the fix any tests that are broken.

vangent:
Fixes #1744.

Note: Tests for this PR and #1743 pass independently, but I am pretty sure they will fail when merged together. #1743 adds a call to `Ack` in the record/replay tests; the replay tests pass even though there's a new call that isn't in the record file, because the background `SendAcks` is failing but that error is never being reported. That changes in this PR. So, I'll need to re-record some replay files before submitting the second of the two PRs.


pubsub/pubsub.go
```diff
@@ -327,6 +327,7 @@ type Subscription struct {
 	mu               sync.Mutex    // protects everything below
 	q                []*Message    // local queue of messages downloaded from server
 	err              error         // permanent error
+	unreportedAckErr error         // permanent error from background SendAcks that hasn't been returnd to the user yet
```
jba: "returned"

Issue #1696, opened 2019-03-27
blob: add support for copying an object


vangent:

vangent:
Notes: none of the cloud providers have an atomic move. You copy and delete.

* S3 [copy](https://godoc.org/github.com/aws/aws-sdk-go/service/s3#S3.CopyObject)
* GCS [copy](https://godoc.org/cloud.google.com/go/storage#ObjectHandle.CopierFrom)
* Azure [copy](https://godoc.org/github.com/Azure/azure-storage-blob-go/azblob#BlobURL.StartCopyFromURL)

Zyqsempai:
@vangent are you working on it? can I take a part?

vangent:
@Zyqsempai 

Yes, I am working on it. My plan is to do an initial PR with the driver + portable type APIs and the memory/file implementations, and I'm mostly done with that. Then implementing each Cloud provider can be done independently; you could do one or more of those (Azure, S3, GCS). If that sounds interesting to you, I'll cc you on the first PR, and just let me know which of the providers you want.

Zyqsempai:
Yeah, sounds more than interesting, thanks!

vangent:
Great! I cc'd you on the first PR. After that is submitted I plan on extending the `As` tests for `BeforeCopy`. Please update this bug with which provider you'd like to work on first (links for all 3 are above).

Zyqsempai:
I will start with `aws` and then continue with 'gcs'.

vangent:
I'm going to work on GCS now.

vangent:
Wow, GCS was super easy. @Zyqsempai have you started on AWS yet, or shall I go ahead? We'd like to tag a release soon and it'd be good to have `Copy` implemented across the board.

Zyqsempai:
Yep, i will create a PR asap.

PR #1758, opened 2019-04-04
blob/all: add a Copy method BREAKING_CHANGE_OK

vangent:
cc @Zyqsempai 

This PR adds the portable API method, extends the driver interface, adds conformance tests, and implements `memblob` and `fileblob`.

It is marked as a breaking change due to the new method in the blob driver interface; since all drivers are in this repo this will not be a breaking change for our users.

Left for subsequent PRs (and marked with TODOs):

* Implementation for the 3 cloud provider drivers.
* Extending conformance tests for `As` to include `CopyOptions.BeforeCopy`.
* Verifying and documenting and testing the semantics for what happens if the destination already exists. Both `fileblob` and `memblob` overwrite, but I'm not sure what the cloud providers do yet.

Updates #1696.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -804,10 +804,10 @@ func (b *Bucket) NewWriter(ctx context.Context, key string, opts *WriterOptions)
 // TODO(rvangent): Look into cross-Bucket copies.
 func (b *Bucket) Copy(ctx context.Context, srcKey, dstKey string, opts *CopyOptions) (err error) {
 	if !utf8.ValidString(srcKey) {
-		return fmt.Errorf("blob.Copy: srcKey must be a valid UTF-8 string: %q", srcKey)
+		return gcerr.Newf(gcerr.InvalidArgument, nil, "blob: Copy srcKey must be a valid UTF-8 string: %q", srcKey)
 	}
 	if !utf8.ValidString(dstKey) {
-		return fmt.Errorf("blob.Copy: dstKey must be a valid UTF-8 string: %q", dstKey)
+		return gcerr.Newf(gcerr.InvalidArgument, nil, "blob: Copy dstKey must be a valid UTF-8 string: %q", dstKey)
 	}
 	if opts == nil {
 		opts = &CopyOptions{}
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -799,9 +799,7 @@ func (b *Bucket) NewWriter(ctx context.Context, key string, opts *WriterOptions)
 // If the source blob does not exist, Copy returns an error for which
 // gcerrors.Code will return gcerrors.NotFound.
 //
-// TODO(rvangent): Determine and document semantics if there's already an
-// existing blob at dstKey.
-// TODO(rvangent): Look into cross-Bucket copies.
+// If the destination blob already exists, it is overwritten.
 func (b *Bucket) Copy(ctx context.Context, srcKey, dstKey string, opts *CopyOptions) (err error) {
 	if !utf8.ValidString(srcKey) {
 		return gcerr.Newf(gcerr.InvalidArgument, nil, "blob: Copy srcKey must be a valid UTF-8 string: %q", srcKey)
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -60,6 +60,7 @@
 //
 // This API collects OpenCensus traces and metrics for the following methods:
 //  - Attributes
+//  - Copy
 //  - Delete
 //  - NewRangeReader, from creation until the call to Close. (NewReader and ReadAll
 //    are included because they call NewRangeReader.)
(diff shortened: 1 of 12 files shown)


blob/drivertest/drivertest.go
```diff
@@ -1621,6 +1626,105 @@ func testMD5(t *testing.T, newHarness HarnessMaker) {
...
+		}
+		if !cmp.Equal(got, contents) {
+			t.Errorf("got %q want %q", string(got), string(contents))
+		}
+		// Verify attributes of the copy.
+		gotAttr, err := b.Attributes(ctx, dstKey)
+		if err != nil {
+			t.Error(err)
+		}
+		if diff := cmp.Diff(gotAttr, wantAttr, cmpopts.IgnoreUnexported(blob.Attributes{})); diff != "" {
```
jba: Only compare if err == nil.

blob/fileblob/fileblob.go
```diff
@@ -516,6 +516,48 @@ func (w *writer) Close() error {
...
+	srcPath, _, xa, err := b.forKey(srcKey)
+	if err != nil {
+		return err
+	}
+	f, err := os.Open(srcPath)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	r := io.Reader(f)
```
jba: I don't think this is necessary.

blob/blob.go
```diff
@@ -794,6 +795,37 @@ func (b *Bucket) NewWriter(ctx context.Context, key string, opts *WriterOptions)
...
+		return fmt.Errorf("blob.Copy: dstKey must be a valid UTF-8 string: %q", dstKey)
+	}
+	if opts == nil {
+		opts = &CopyOptions{}
+	}
+	dopts := &driver.CopyOptions{
+		BeforeCopy: opts.BeforeCopy,
+	}
+	b.mu.RLock()
+	defer b.mu.RUnlock()
```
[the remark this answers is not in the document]
vangent: This is following the pattern used in all the other APIs. I was worried about that too, but I think it's OK given that it's a read lock, and the only time we get the full/write lock is during `Close`.

vangent:
> As a general note, I'd expect cross-bucket copying to be somewhat common. Is there a plan for that?

Good point. I haven't looked into that. Added a TODO.

blob/driver/driver.go
```diff
@@ -255,6 +263,16 @@ type Bucket interface {
 	// and do any necessary cleanup in Close. Close should then return ctx.Err().
 	NewTypedWriter(ctx context.Context, key string, contentType string, opts *WriterOptions) (Writer, error)
```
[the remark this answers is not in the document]
vangent: Done.

vangent:
> > As a general note, I'd expect cross-bucket copying to be somewhat common. Is there a plan for that?

My tentative plan would be to add `CopyOptions.SourceBucket *blob.Bucket`. The default (`nil`) means use the same bucket, but you could supply another bucket to copy from.

blob/fileblob/fileblob.go

blob/drivertest/drivertest.go
vangent: Changed `t.Error` to `t.Fatal` above.

blob/fileblob/fileblob.go
```diff
@@ -516,6 +516,47 @@ func (w *writer) Close() error {
 	return nil
 }
 
+// Copy implements driver.Copy.
+func (b *bucket) Copy(ctx context.Context, srcKey, dstKey string, opts *driver.CopyOptions) error {
+	// Note: we could use NewRangedReader here, but since we need to copy all of
+	// the metadata (from xa), it's more efficient to do it directly.
+	srcPath, _, xa, err := b.forKey(srcKey)
```
clausti: is `xa` a conventional abbreviation/var name for metadata?

vangent: I believe it's a shortening of the type (`type xattrs struct`), which I think is short for "extended attributes". `xa` is used in a number of other places in this file though. I'm open to better names, but probably in a different PR.

vangent:
> > > As a general note, I'd expect cross-bucket copying to be somewhat common. Is there a plan for that?
> 
> My tentative plan would be to add `CopyOptions.SourceBucket *blob.Bucket`. The default (`nil`) means use the same bucket, but you could supply another bucket to copy from.

Filed https://github.com/google/go-cloud/issues/1766.

blob/blob.go
```diff
@@ -792,6 +793,36 @@ func (b *Bucket) NewWriter(ctx context.Context, key string, opts *WriterOptions)
...
 }
 
+// Copy the blob stored at srcKey to dstKey.
+// A nil CopyOptions is treated the same as the zero value.
+//
+// If the source blob does not exist, Copy returns an error for which
+// gcerrors.Code will return gcerrors.NotFound.
+//
+// If the destination blob already exists, it is overwritten.
+func (b *Bucket) Copy(ctx context.Context, srcKey, dstKey string, opts *CopyOptions) (err error) {
```
jba: I think you should reverse src and dest, to be compatible with Go (e.g. the `copy` built-in).

vangent: Good point, done.

vangent:
PTAL

PR #1759, opened 2019-04-04
all: update portable types to return errors with gcerr codes

vangent:


vangent:
Apparently this is a breaking change because it changes the type of `runtimevar.ErrClosed` from `error` to `*gcerr.Error`.

blob/blob.go
```diff
@@ -565,7 +564,7 @@ func (b *Bucket) Exists(ctx context.Context, key string) (bool, error) {
 // gcerrors.Code will return gcerrors.NotFound.
 func (b *Bucket) Attributes(ctx context.Context, key string) (_ Attributes, err error) {
 	if !utf8.ValidString(key) {
```
jba: I'd factor this out into a `checkKey` function.

You'd lose the function-specific error string, but gcerrors should have frame information. You'll want to call gcerr.New instead of Newf, and pass in the right caller offset (2, I think).

vangent: I don't think this is worth it. Calling `checkKey` is 3 lines, the same as what's done currently, and the error messages are slightly different (not only the function-specific error, but singular vs plural ("metadata keys" vs "Delete key").

PR #1763, opened 2019-04-05
docstore/firedocstore: queries

jba:
Support for Get queries using Firestore.

Tested omitted until conformance tests are ready.

Updates #1338.


internal/docstore/firedocstore/fs.go
```diff
@@ -35,7 +35,7 @@ import (
 type collection struct {
 	client    *vkit.Client
 	dbPath    string // e.g. "projects/P/databases/(default)"
-	collPath  string // e.g. "projects/P/databases/(default)/documents/MyCollection"
+	collPath  string // e.g. "projects/P/databases/(default)/documents/States/Wisconsin/cities"
```
shantuo: Why make this change here? Does `States/Wisconsin/cities` also represent some type of hierarchy?

internal/docstore/driver/driver.go
```diff
@@ -127,3 +127,7 @@ type DocumentIterator interface {
 	// needed.
 	Stop()
 }
+
+// EqualOp is the name of the equality operator.
+// It is defined here to avoid confusion between "=" and "==".
+const EqualOp = "="
```
shantuo: Should this go into the portable API so that the user uses this as well? I'm not sure whether all drivers have the same standard of using "=".

internal/docstore/firedocstore/fs.go
jba: I wanted to show that `collPath` could be the full path to a nested collection.

internal/docstore/driver/driver.go
jba: We should pick one as the docstore standard, and `=` is a slightly better choice because it matches SQL.

If a user picks the wrong one they'll get an error and fix their code. I added this because I'm worried that a driver will get it wrong, and that would be a real bug.

Let's leave it here and we can always "promote" it later if we think that's a good idea.

PR #1772, opened 2019-04-06
blob/azureblob: implement Copy

vangent:
Updates #1696.


blob/azureblob/azureblob.go
```diff
@@ -318,8 +318,34 @@ var errUnimplemented = errors.New("not implemented")
...
+	dstKey = escapeKey(dstKey, false)
+	dstBlobURL := b.containerURL.NewBlobURL(dstKey)
+	srcKey = escapeKey(srcKey, false)
+	srcURL := b.containerURL.NewBlobURL(srcKey).URL()
+	resp, err := dstBlobURL.StartCopyFromURL(ctx, srcURL, nil /* metadata */, azblob.ModifiedAccessConditions{}, azblob.BlobAccessConditions{})
+	if err != nil {
+		return err
+	}
+	copyStatus := resp.CopyStatus()
+	nErrors := 0
```
shantuo: This can be defined inside the for loop.

blob/azureblob/azureblob.go
```diff
@@ -318,8 +318,34 @@ var errUnimplemented = errors.New("not implemented")
...
+	for copyStatus == azblob.CopyStatusPending {
+		// Poll until the copy is complete.
+		time.Sleep(500 * time.Millisecond)
```
shantuo: I see copy in general may take more time. Is it common enough that we might expose an option for timeouts? Also, maybe we should do exponential waits, or maybe it's too much.

blob/azureblob/testdata/TestConformance/TestAs/verify_As_returns_false_when_passed_nil.yaml
```diff
@@ -8,7 +8,7 @@ interactions:
       Content-Length:
       - "11"
       User-Agent:
-      - go-cloud/blob/0.9.0 Azure-Storage/0.3 (go1.11.2; linux)
+      - go-cloud/blob/0.1.0 Azure-Storage/0.4 (go1.12; linux)
```
shantuo: Why did the version jump back `0.9.0`->`0.1.0`?

blob/azureblob/azureblob.go
vangent: It can't be inside the for loop, as it may be incremented inside the for loop. I guess it could be declared as part of the `for` itself, e.g. `for nErrors := 0; copyStatus == copyStatusPending`, but I find that less readable.

blob/azureblob/azureblob.go
vangent: I've never seen it actually return `CopyStatusPending`, so I suspect it's only for very large files. We can expose an option for the 500ms if we're asked to do so, I don't think it's worth it at this point.

blob/azureblob/testdata/TestConformance/TestAs/verify_As_returns_false_when_passed_nil.yaml
vangent: See #1118.

blob/azureblob/azureblob.go
shantuo: Oops, didn't know what I was reading...

PR #1774, opened 2019-04-08
docstore: document query behavior

jba:


internal/docstore/doc.go
```diff
@@ -48,4 +48,44 @@
 // If a struct doesn't have a DocstoreRevision field, then the logic described above
 // won't apply to documents read and written with that struct. All writes with the
 // struct will succeed even if the document was changed since the last Get.
+//
+//
+// Queries
+//
+// Docstore supports querying within a collection. Call the Query method on
+// Collection to obtain a Query value, Then build your query by calling Query methods
```
vangent: Then -> then

internal/docstore/doc.go
```diff
@@ -48,4 +48,44 @@
...
+// like Where, Limit and so on. Finally, call the Get method on the query to execute it.
+// The result in an iterator, whose use is described below.
```
vangent: in -> is

internal/docstore/doc.go
```diff
@@ -48,4 +48,44 @@
 // If a struct doesn't have a DocstoreRevision field, then the logic described above
 // won't apply to documents read and written with that struct. All writes with the
 // struct will succeed even if the document was changed since the last Get.
+//
+//
+// Queries
+//
+// Docstore supports querying within a collection. Call the Query method on
+// Collection to obtain a Query value, Then build your query by calling Query methods
+// like Where, Limit and so on. Finally, call the Get method on the query to execute it.
```
vangent: Side note: should we even have a `Limit` method? I'm not sure it makes much sense in the context of an API that doesn't supported sorting, only returns results one at a time, and does client-side filtering.

internal/docstore/doc.go
```diff
@@ -48,4 +48,44 @@
...
+// The result in an iterator, whose use is described below.
+//
+//     iter := coll.Query().Where("size", ">", 10).Limit(5).Get(ctx)
+//
+// The Where methods defines a filter condition, much like a WHERE clause in SQL. Conditions
+// are of the form "field op value", where field is any document field path (including dot-separated
+// paths), op is one of "=", ">", "<", ">=" or "<=", and value can be a string or number.
+//
+// Although you can make multiple Where calls, not all provider systems can handle them.
```
vangent: "not all provider systems can handle them" sounds a bit too strong to me, it implies it won't work.

How about something like:

"You can make multiple Where calls. In some cases, parts of a Where clause may be processed on the client rather than natively by the provider, which may have performance implications for large result sets. See the provider-specific package doc for details."

internal/docstore/doc.go
```diff
@@ -48,4 +48,44 @@
...
+// For some providers, Where clauses may be processed on the client.
+//
+// The Limit method specifies the maximum number of results to return.
+//
+// Docstore provides no guarantees about the ordering of results. There is no way to
+// ask for a particular ordering.
+//
+// Use the DocumentIterator returned from Query.Get by repeatedly calling its Next
+// method until it returns io.EOF. Always call Stop when you are finished with an
```
vangent: Do you always have to call `Stop`, or only if you stopped before `io.EOF`? IIRC the `Stop` docstring said the latter. I don't feel strongly either way, but let's make sure we're consistent. 

Also, if it's "always call Stop", consider adding a `SetFinalizer` for it to log if it is GCd without `Stop` being called, similar to `Message` without `Ack` and `blob.Reader/Writer` without `Close`.

shantuo: If the iterator returns EOF, can the driver clean up the resource if needed?

internal/docstore/doc.go
jba: Done.

internal/docstore/doc.go
jba: Done.

internal/docstore/doc.go
jba: Limit is among other things a server-side optimization, even if the client API returns one result at a time.

You're right about sorting, but I'm pretty sure we can't leave the sort order undefined forever. (Working on gathering evidence for that.)

I'm going to leave it for now.

internal/docstore/doc.go
jba: We could make it unnecessary for EOF, but let's say "always." That's easier to remember, matches file Close, and leads to more maintainable code (if you defer the Stop, and then a maintainer changes the iteration to end early, the code still works).

jba: I updated the `DocumentIterator` documentation to match.

jba:
@shantuo PTAL

PR #1776, opened 2019-04-08
natspubsub: replace ugorji/codec with encoding/gob

jba:
The codec package used to encode and decode messages is an uneccesary
dependency for pubsub, and takes up 11M in the binary (according to
https://github.com/jondot/goweight). Replace it with gob.


pubsub/natspubsub/nats.go
```diff
@@ -175,28 +175,11 @@ type topic struct {
...
-	// driver.Message.Metadata type
-	dm := driver.Message{}
-	mh.MapType = reflect.TypeOf(dm.Metadata)
-}
-
 // We define our own version of message here for encoding that
-// only encodes Body and Metadata. Otherwise we would have to
-// add codec decorations to driver.Message.
+// only encodes Body and Metadata.
 type encMsg struct {
```
vangent: Is this used anymore?

vangent:
`FAIL: dependencies changed; compare listdeps.sh output with alldeps`

pubsub/natspubsub/nats.go
jba: No. Removed.

jba:
PTAL.

eliben:
Wow, nice find!

PR #1778, opened 2019-04-09
docstore: validate query filters

jba:


jba:
Good question.

For the op, validating here allows drivers to assume that they are getting a legal op. That slightly simplifies the driver code, because they don't have to return an error. And this way we can make sure the error is identical for all drivers.

Validating the value is more important, because for portability we want to restrict the legal values to a smaller set than some providers may support. So if we didn't check that `[]byte` was not allowed, then that might work for dynamo but maybe fail for Firestore.

PR #1781, opened 2019-04-09
firedocstore: support multiple filters in queries

jba:
Add support for locally evaluating some filters.

TODO: add multi-filter conformance tests (after the dynamo gets basic query support)

Updates #1338.

### DIFF
--- a/internal/docstore/firedocstore/query.go
+++ b/internal/docstore/firedocstore/query.go
@@ -12,13 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// TODO(jba): figure out how to get filters with complex values to work (since they
+// are represented as arrays of floats). Also, uints: since they are represented as
+// int64s, the sign is wrong. Since you can only compare complex numbers for
+// equality, maybe it could work if Firestore arrays can be compared for equality.
+
 package firedocstore
 
 import (
 	"context"
 	"fmt"
 	"math"
+	"math/big"
 	"path"
+	"reflect"
 
 	"github.com/golang/protobuf/ptypes/wrappers"
 	"gocloud.dev/internal/docstore/driver"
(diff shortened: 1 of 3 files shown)


internal/docstore/firedocstore/query.go
```diff
@@ -66,18 +80,112 @@ func (it *docIterator) Next(ctx context.Context, doc driver.Document) error {
...
+	}
+	lhs := reflect.ValueOf(val)
+	rhs := reflect.ValueOf(f.Value)
+	if lhs.Kind() == reflect.String {
+		if rhs.Kind() != reflect.String {
+			return false
+		}
+		return applyComparison(f.Op, strings.Compare(lhs.String(), rhs.String()))
+	}
+	// Compare numbers by using big.Float. This is expensive
```
shantuo: Curious, does the firestore backend handle this the same way?

internal/docstore/firedocstore/query.go
```diff
@@ -114,7 +224,28 @@ func (c *collection) queryToProto(q *driver.Query) (*pb.StructuredQuery, error)
...
+// and those we must evaluate here on the client.
+func splitFilters(fs []driver.Filter) (sendToFirestore, evaluateLocally []driver.Filter) {
+	// Enforce that only one field can have an inequality.
+	var rangeFP []string
+	for _, f := range fs {
+		if f.Op == driver.EqualOp {
+			sendToFirestore = append(sendToFirestore, f)
+		} else {
+			if rangeFP == nil || fpEqual(rangeFP, f.FieldPath) {
+				// Multiple inequality filters on the same field are OK.
```
shantuo: The docstring at the top of the package states only the first where clause with inequality is passed. Should that one be changed?

internal/docstore/firedocstore/query.go
jba: I doubt it; it's expensive. (And I don't think the backend is written in Go.) It may use some larger floating-point type that can accurately represent any int64.

internal/docstore/firedocstore/query.go
jba: Thanks, I clarified.

jba:
PTAL.

PR #1783, opened 2019-04-09
runtimevar: runtimevar.Decode takes its own context. 

### LINKED ISSUE #1667: runtimevar: DecryptDecoder does not use the context of the Variable goroutine
/cc @shantuo @vangent

The Context used for [`DecryptDecoder`](https://godoc.org/gocloud.dev/runtimevar#DecryptDecode) will always be the one passed into `DecryptDecoder` instead of the Context used by the caller of the [`runtimevar.Decode` function](https://godoc.org/gocloud.dev/runtimevar#Decode).

The pressing issue this presents is that a long-running decryption call can block the close of a `*runtimevar.Variable`. `runtimevar.Decode` typically gets called during a `driver.Watcher.WatchVariable` call, which is in turn called by [`Variable.background`](https://github.com/google/go-cloud/blob/6482f3715c170669214c5cf4ba8421c51d772d30/runtimevar/runtimevar.go#L211). `Variable.background` relies on being able to cancel any outstanding long-running work by canceling its context.

Another implication is this can result in incorrect trace information being emitted when using secrets keepers that make HTTP/RPC requests.

I believe the only way we can address this is with a breaking change to make the `runtimevar.Decode` signature take in a Context parameter:

```go
type Decode func(context.Context, []byte, interface{}) error
```

This implies a bunch of breaking changes to support the Context being properly passed, but I don't see another way to do it.

### Additional context

This is a follow-on to #1290. #1580 is the PR that introduced the `DecryptDecode` function, but this issue wasn't caught in review.

eliben:
BREAKING_CHANGE_OK
Fixes #1667

[this is a draft - I'll need to propagate the context further to the Decoder object's Decode too]


runtimevar/runtimevar.go
```diff
@@ -495,16 +497,16 @@ func BytesDecode(b []byte, obj interface{}) error {
 //
 // post defaults to BytesDecode. An optional decoder can be passed in to do
 // further decode operation based on the decrypted message.
-func DecryptDecode(ctx context.Context, k *secrets.Keeper, post Decode) Decode {
-	return func(b []byte, obj interface{}) error {
+func DecryptDecode(unusedCtx context.Context, k *secrets.Keeper, post Decode) Decode {
```
zombiezen: Remove this Context parameter: it was added specifically to curry the context, which is what led me to file #1667.

```suggestion
func DecryptDecode(k *secrets.Keeper, post Decode) Decode {
```

eliben: Thanks, will do!

Issue #1782, opened 2019-04-09
runtimevar: document the different types of decoders in URLs


eliben:
The docstring for `runtimevar/runtimevar.go` has a useful description of the different decoders supported through URLs:

```
// Supported values include:
//   - empty string: Returns the default from the URLOpener.Decoder, or
//       BytesDecoder if URLOpener.Decoder is nil (which is true if you're
//       using the default URLOpener).
//   - "bytes": Returns a BytesDecoder; Snapshot.Value will be of type []byte.
//   - "jsonmap": Returns a JSON decoder for a map[string]interface{};
//       Snapshot.Value will be of type *map[string]interface{}.
//   - "string": Returns StringDecoder; Snapshot.Value will be of type string.
// It also supports using "decrypt+<decoderName>" (or "decrypt" for default
// decoder) to decrypt the data before decoding. It uses the secrets package to
// open a keeper by the URL string stored in a envrionment variable
// "RUNTIMEVAR_KEEPER_URL". See https://godoc.org/gocloud.dev/secrets#OpenKeeper
// for more details.
```

I can't find this elsewhere in the existing docstrings. This should be visible on godoc, probably in the documentation of https://godoc.org/gocloud.dev/runtimevar#OpenVariable ?

eliben:
... this demonstrates a quirk of how godoc generates documentation. We have

```
// DecoderByName returns a *Decoder based on decoderName.
// It is intended to be used by VariableURLOpeners in driver packages.
var DecoderByName = decoderByName
```

So the godoc page for `runtimevar` shows this docstring for the public `DecoderByName`, while the docstring for the private name it aliases is lost. Moreover, the documentation doesn't even specify the arguments `DecoderByName` takes and its return values, because it's just a `var`

PR #1786, opened 2019-04-09
runtimevar: Expose DecoderByName directly

eliben:
For #1782

This should show the actually useful documentation of this function in the godoc, and list its arguments

### DIFF
--- a/runtimevar/runtimevar.go
+++ b/runtimevar/runtimevar.go
@@ -509,9 +509,6 @@ func DecryptDecode(ctx context.Context, k *secrets.Keeper, post Decode) Decode {
 }
 
 // DecoderByName returns a *Decoder based on decoderName.
-// It is intended to be used by VariableURLOpeners in driver packages.
-var DecoderByName = decoderByName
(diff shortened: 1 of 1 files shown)


runtimevar/runtimevar.go
```diff
@@ -509,9 +509,6 @@ func DecryptDecode(ctx context.Context, k *secrets.Keeper, post Decode) Decode {
 }
 
 // DecoderByName returns a *Decoder based on decoderName.
-// It is intended to be used by VariableURLOpeners in driver packages.
```
vangent: Let's keep this line in the docstring, to let end users know they shouldn't be using this directly.

vangent:
```
Incompatible changes:
- DecoderByName: changed from var to func
```

runtimevar/runtimevar.go
eliben: Done

eliben:
Thanks @shantuo , @vangent and I chatted about this

PR #1787, opened 2019-04-10
pubsub/test: check for NotFound error code, and fail fast in conformance tests

### LINKED ISSUE #1791: pubsub: NotFound errors are not being returned correctly
The conformance tests check that an error is returned, but not that its code is `NotFound`. Several drivers are getting this wrong.

Update the tests and fix the drivers.

vangent:
* Check for `NotFound` error code in a couple of tests, and fix `mempubsub` to return it correctly.
* Update a couple of tests that `Receive` messages to time out faster if it isn't there (15s vs default 10m).

Fixes #1791.


vangent:
Doh, it looks like a few other providers are not returning `NotFound` correctly... fixing.

vangent:
PTAL

pubsub/awssnssqs/awssnssqs_test.go
```diff
@@ -33,33 +33,10 @@ import (
 	"gocloud.dev/pubsub/drivertest"
 )
 
-const region = "us-east-2"
-
-func TestOpenTopic(t *testing.T) {
```
eliben: So this went into drivertest?

vangent: Yes, these two tests are already covered by conformance tests.

PR #1795, opened 2019-04-11
blob/s3blob: implement Copy

Zyqsempai:
Updates #1696.

/cc @vangent The only question is, how to generate all that testData?


blob/s3blob/s3blob.go
```diff
@@ -670,12 +667,17 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
...
-
 // Copy implements driver.Copy.
 func (b *bucket) Copy(ctx context.Context, dstKey, srcKey string, opts *driver.CopyOptions) error {
-	// TODO(rvangent): Implement this and delete errUnimplemented.
-	return errUnimplemented
+	dstKey = escapeKey(dstKey)
+	srcKey = escapeKey(srcKey)
+	input := &s3.CopyObjectInput{
+		Bucket:     aws.String(b.name),
+		CopySource: aws.String(srcKey),
```
vangent: According to the godoc, you need the bucket name as a prefix here, with a `/` separator.

`CopySource: aws.String(b.name + "/" + srcKey)`

vangent:
Looks good, one comment inline about `CopySource`. I was able to download your PR, and after that fix, the tests pass with updated goldens.

> /cc @vangent The only question is, how to generate all that testData?

There are 3 ways to do this:

1. You can modify this PR to not delete the old code ("not implemented", etc.), and only add the new/correct code, but commented out. Once it is submitted, I'll send another PR uncommenting the new code, deleting the "not implemented" stuff, and adding the test golden files.

2. You can create an AWS test project + bucket, and update the constants at the top of `s3blob_test.go` to refer to your bucket name and region. You will then have to re-record all of the golden files (`cd s3blob; go test --record`). After that is submitted, I'll change the constants back to the Go CDK team project and re-record all of the files.

3. You can give me access to your fork, and I can submit the updated/added test golden files to it directly, updating this PR. I haven't done this before but I think it should work. https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork.

I'm fine with any of these, just let me know.

blob/s3blob/s3blob.go
```diff
@@ -670,12 +667,17 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
...
+		Key:        aws.String(dstKey),
+	}
+	_, err := b.client.CopyObject(input)
```
vangent: Can you change this to `CopyObjectWithContext(ctx, input)` so it respect the ctx? I notice that there are other requests in this file that don't do this, I'll send a separate PR to fix.

vangent: I did send a separate PR already for this, so updating this line should make it consistent with the rest.

Zyqsempai:
@vangent Fixed, and I allowed you to edit my fork branch, so please add the tests!

PR #1803, opened 2019-04-11
pubsub/tests: clean up some leaked goroutines in tests

vangent:
I noticed this when running tests with `--count=100` -- there were tons of goroutines on the stack. I then went through and looked for other places where `Shutdown` wasn't being called.


pubsub/awssnssqs/awssnssqs_test.go
```diff
@@ -297,10 +297,13 @@ func TestOpenTopicFromURL(t *testing.T) {
 
 	ctx := context.Background()
 	for _, test := range tests {
-		_, err := pubsub.OpenTopic(ctx, test.URL)
+		top, err := pubsub.OpenTopic(ctx, test.URL)
```
eliben: Nit: is "top" a short name we use frequently? Seems a bit mysterious, and saves only 2 chars.

pubsub/azuresb/azuresb_test.go
```diff
@@ -313,6 +313,7 @@ func BenchmarkAzureServiceBusPubSub(b *testing.B) {
 	if err != nil {
 		b.Fatal(err)
 	}
+	defer topic.Shutdown(ctx)
```
eliben: topic here...

pubsub/awssnssqs/awssnssqs_test.go
vangent: Done.

PR #1805, opened 2019-04-11
internal/cmd/gocdk: apply

### LINKED ISSUE #1816: internal/cmd/gocdk: implement minimal apply command
This command should simply be a shim over `terraform apply` run in a biome directory. This blocks #1812.

See #1805 for the current PR, which has a planned follow-on of #1809.

clausti:
implements a minimal `apply` command, with some initial tests

fixes #1816

### DIFF
--- a/internal/cmd/gocdk/apply.go
+++ b/internal/cmd/gocdk/apply.go
@@ -16,10 +16,41 @@ package main
 
 import (
 	"context"
+	"flag"
+	"os/exec"
+	"path/filepath"
 
 	"golang.org/x/xerrors"
 )
 
 func apply(ctx context.Context, pctx *processContext, args []string) error {
-	return xerrors.New("not implemented")
+	f := newFlagSet(pctx, "apply")
+	if err := f.Parse(args); xerrors.Is(err, flag.ErrHelp) {
+		return nil
+	} else if err != nil {
+		return usagef("gocdk apply: %w", err)
+	}
+
+	if f.NArg() != 1 {
+		return usagef("gocdk apply BIOME")
+	}
+	biome := f.Arg(0)
+	moduleRoot, err := findModuleRoot(ctx, pctx.workdir)
+	if err != nil {
+		return xerrors.Errorf("apply %s: %w", biome, err)
+	}
+	// TODO(cla): extract into helper method
+	biomePath := filepath.Join(moduleRoot, "biomes", biome)
+
+	// TODO(cla): take over steps (plan, confirm, apply) so we can
+	// dictate the messaging and errors. We should visually differentiate
+	// when we insert verbiage on top of terraform.
+	c := exec.CommandContext(ctx, "terraform", "apply")
+	c.Dir = biomePath
+	c.Stdout = pctx.stdout
+	c.Stderr = pctx.stderr
+	if err := c.Run(); err != nil {
+		return xerrors.Errorf("apply %s: %w", biome, err)
+	}
+	return nil
 }
--- a/internal/cmd/gocdk/apply.go
+++ b/internal/cmd/gocdk/apply.go
@@ -46,7 +46,7 @@ func apply(ctx context.Context, pctx *processContext, args []string) error {
 		return xerrors.Errorf("apply %s: %w", biome, err)
 	}
 
-	// TODO(cla): take over steps (plan, confirm, apply) so we can
+	// TODO(#1821): take over steps (plan, confirm, apply) so we can
 	// dictate the messaging and errors. We should visually differentiate
 	// when we insert verbiage on top of terraform.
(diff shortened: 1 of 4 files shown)


internal/cmd/gocdk/apply_test.go
```diff
@@ -0,0 +1,112 @@
...
+			stdout:  ioutil.Discard,
+			stderr:  ioutil.Discard,
+		}
+		ctx := context.Background()
+
+		if err := run(ctx, pctx, []string{"apply", biomeName}, new(bool)); err != nil {
+			t.Errorf("run error: %+v", err)
+		}
+
+		outputs, err := tfReadOutput(findBiomeDir(dir, biomeName))
```
zombiezen: Let's add a comment above this line explaining the assertion we're making: after a successful `terraform apply`, `terraform output` should return the greeting we configured. One non-obvious property (we ended up manually testing) was that `terraform output` fails if `terraform init` was not called, so this is actually verifying quite a bit of state with a seemingly simple assertion.

internal/cmd/gocdk/apply.go
```diff
@@ -16,10 +16,68 @@ package main
...
+	if err := ensureTerraformInit(ctx, pctx, biomePath); err != nil {
+		return xerrors.Errorf("apply %s: %w", biome, err)
+	}
+
+	// TODO(cla): take over steps (plan, confirm, apply) so we can
+	// dictate the messaging and errors. We should visually differentiate
+	// when we insert verbiage on top of terraform.
+	c := exec.CommandContext(ctx, "terraform", "apply")
+	c.Dir = biomePath
+	c.Stdout = pctx.stdout
```
zombiezen: If the `terraform apply` needs to prompt the user, stdin is not hooked up. Let's plumb stdin through the `processContext` from `osProcessContext` to here.

internal/cmd/gocdk/main_test.go
```diff
@@ -73,14 +69,28 @@ func TestFindModuleRoot(t *testing.T) {
...
-		}
 
 		got, err := findModuleRoot(context.Background(), subdir)
 		if got != dir || err != nil {
 			t.Errorf("findModuleRoot(ctx, %q) = %q, %v; want %q, <nil>", dir, got, err, dir)
 		}
 	})
 }
+
+func newTestModule() (dir string, cleanup func(), _ error) {
```
eliben: Why do you need the return types to be named here?

[I thought these are mainly useful for `defer`s, though I may be wrong]

internal/cmd/gocdk/main_test.go
```diff
@@ -73,14 +69,28 @@ func TestFindModuleRoot(t *testing.T) {
...
-			t.Fatal(err)
-		}
 
 		got, err := findModuleRoot(context.Background(), subdir)
 		if got != dir || err != nil {
 			t.Errorf("findModuleRoot(ctx, %q) = %q, %v; want %q, <nil>", dir, got, err, dir)
 		}
 	})
 }
+
```
eliben: Please document this function (esp. semantics of return values)

internal/cmd/gocdk/apply_test.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+	})
+}
+
+func newTestBiome(dir, name, tfGreeting string) error {
+	biomeDir := filepath.Join(dir, "biomes", name)
+	if err := os.MkdirAll(biomeDir, 0777); err != nil {
+		return err
+	}
+	terraformSource := `output "greeting" {
```
eliben: Nit: this would be slightly easier on the eyes if it used a sprintf or template

internal/cmd/gocdk/main_test.go
clausti: we don't *need* them to be named, I guess. I thought naming the cleanup function made it clearer what it was for.

eliben: How about naming only cleanup? (the `_` name for the error doesn't seem useful anyway).  i looked through the style guide, it's ambivalent w.r.t. named parameters. If you feel naming cleanup will improve clarity, that sounds good

clausti: It compile errors if I only name the function, but I can give the `string` a blank identifier?

eliben: I see.

Don't feel strongly about this. We don't hame return values in similar cases elsewhere in the CDK (for example see `xrayserver.NewExporter`), but if you feel it helps clarify the function that's fine. Up to you.

In any case the function needs a documentation comment explaining the semantic meaning of parameters and return values.

clausti: Ended up removing the names on the return values; seemed clear enough once the function was documented.

PR #1814, opened 2019-04-12
health: add CheckerFunc type

zombiezen:
Convenience type similar to `http.HandlerFunc`. I would like to get this into the release if possible.

### DIFF
--- a/health/health.go
+++ b/health/health.go
@@ -90,10 +90,12 @@ type Checker interface {
 	CheckHealth() error
 }
 
-// CheckerFunc implements Checker by calling a function.
+// CheckerFunc is an adapter type to allow the use of ordinary functions as
+// health checks. If f is a function with the appropriate signature,
+// CheckerFunc(f) is a Checker that calls f.
 type CheckerFunc func() error
 
-// CheckHealth calls the function.
+// CheckHealth calls f().
 func (f CheckerFunc) CheckHealth() error {
 	return f()
 }
--- a/health/health.go
+++ b/health/health.go
@@ -59,7 +59,7 @@ func writeUnhealthy(w http.ResponseWriter) {
 	io.WriteString(w, status)
 }
 
-// HandleLive is an http.HandleFunc that handles liveness checks by
+// HandleLive is an http.HandlerFunc that handles liveness checks by
 // immediately responding with an HTTP 200 status.
 func HandleLive(w http.ResponseWriter, _ *http.Request) {
 	writeHealthy(w)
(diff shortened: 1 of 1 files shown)


health/health.go
```diff
@@ -31,6 +31,11 @@ func (h *Handler) Add(c Checker) {
 	h.checkers = append(h.checkers, c)
 }
 
+// AddFunc adds a new check function to the handler.
+func (h *Handler) AddFunc(f CheckerFunc) {
```
vangent: Why is this necessary, won't `Add` work directly?

health/health.go
```diff
@@ -84,3 +89,11 @@ func writeHealthy(w http.ResponseWriter) {
 type Checker interface {
 	CheckHealth() error
 }
+
+// CheckerFunc implements Checker by calling a function.
+type CheckerFunc func() error
```
vangent: Should it take a context?

health/health.go
```diff
@@ -84,3 +89,11 @@ func writeHealthy(w http.ResponseWriter) {
 type Checker interface {
 	CheckHealth() error
 }
+
+// CheckerFunc implements Checker by calling a function.
```
vangent: I'd suggest a comment similar to `http.HandlerFunc` to make it more clear what the use case is for this type.

```
The CheckerFunc type is an adapter to allow the use of ordinary functions as Checkers. If f is a function with the appropriate signature, CheckerFunc(f) is a Checker that calls f.
```

health/health.go
zombiezen: It's convenient for writing:

```go
healthHandler.AddFunc(func() error {
  // ...
})
```

instead of:


```go
healthHandler.Add(health.CheckerFunc(func() error {
  // ...
}))
```

([`net/http.ServeMux.HandleFunc`](https://godoc.org/net/http#ServeMux.HandleFunc) exists for similar reasons.)

It's a minor thing, so I'd be fine pulling this out.

health/health.go
zombiezen: Done.

health/health.go
zombiezen: The `Checker` interface intentionally does not take in a `Context` because a `Checker` should not perform anything expensive like an RPC: it should purely report on the health at a snapshot in time.

vangent: SGTM.

health/health.go
vangent: I'd suggest either leaving it out, or making it clear in the docstring that this is just a shorthand. I think I'd lean toward the former, but either way is fine.

PR #1817, opened 2019-04-12
pubsub: add a Close method to both driver.Topic and driver.Subscription

### LINKED ISSUE #1813: pubsub: consider adding Close/Shutdown methods for drivers
We currently have `Shutdown` at the portable type, but not in the driver interface, for both `Topic` and `Subscription`.

I'm working on Kafka; it's looking like I'm going to need to start a goroutine in `OpenSubscription`. It's not obvious how to clean it up.

* We could add `Shutdown` or `Close` methods to the driver interface and have the portable type call them.

* It could use the `ctx` passed to `OpenSubscription`, but that's a bit subtle IMHO; many users will pass `context.Background()` and leak the goroutine, or pass something else and not realize the context needs to be closed.

* I could make `OpenSubscription` return a `cleanup` function. This seems a bit weird given that it also returns a `pubsub.Subscription` that has a `Shutdown` method.

@jba thoughts?

vangent:
Fixes #1813.

### DIFF
--- a/pubsub/driver/driver.go
+++ b/pubsub/driver/driver.go
@@ -97,6 +97,11 @@ type Topic interface {
 	// ErrorCode should return a code that describes the error, which was returned by
 	// one of the other methods in this interface.
 	ErrorCode(error) gcerrors.ErrorCode
+
+	// Close cleans up any resources used by the Topic. Once Close is called,
+	// there will be no method calls to the Topic other than As, ErrorAs, and
+	// ErrorCode.
+	Close() error
 }
 
 // Subscription receives published messages.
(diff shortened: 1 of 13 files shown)


pubsub/driver/driver.go
```diff
@@ -55,6 +55,8 @@ type Message struct {
 }
 
 // Topic publishes messages.
+// Drivers may optionally also implement io.Closer; Close will be called
```
vangent: I could define a new type instead of `io.Closer` if you prefer.

jba:
As I commented on the issue, I think it's simpler to just add `Close() error` to `Topic` and `Subscription`. I think it is worth the small implementation burden to have a fully compile-time interface.

vangent:
OK, done.

PR #1822, opened 2019-04-12
internal/testing: don't break if upstream changes during test execution

### LINKED ISSUE #1802: internal/testing: check_api_change.sh raises an error if master made a breaking change during test run
As per this comment:

https://github.com/google/go-cloud/blob/4a4b28c6ec7c46a7b4a945e427298862efb65958/internal/testing/check_api_change.sh#L39-L44

The script currently fails in the case where a breaking change has been committed to master since the PR work started. I haven't drilled into whether merging resolves this issue, but I don't think this should be an error case, as this will affect any PRs outstanding after a breaking change commits. (Thundering herd merge.)

Travis CI job with context: https://travis-ci.com/google/go-cloud/jobs/192256983#L1445-L1453

zombiezen:
I also made changes suggested by [shellcheck](https://www.shellcheck.net/) to avoid escaping issues. They're made as separate commits if you want to isolate the changes for review.

Fixes #1802

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -885,6 +885,7 @@ golang.org/x/tools v0.0.0-20181221154417-3ad2d988d5e2/go.mod h1:n7NCudcB/nEzxVGm
 golang.org/x/tools v0.0.0-20190114222345-bf090417da8b/go.mod h1:n7NCudcB/nEzxVGmLbDWY5pfWTLqBcC2KZ6jyYvM4mQ=
 golang.org/x/tools v0.0.0-20190226205152-f727befe758c h1:vamGzbGri8IKo20MQncCuljcQ5uAO6kaCeawQPVblAI=
 golang.org/x/tools v0.0.0-20190226205152-f727befe758c/go.mod h1:9Yl7xja0Znq3iFh3HoIrodX9oNMXvdceNzlUR8zjMvY=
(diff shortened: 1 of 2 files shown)


zombiezen:
Forgot to cc: @vangent @jba 

@jadekler, you may want to make the corresponding change to google-cloud-go @ https://github.com/googleapis/google-cloud-go/blob/ce66726a686431294383c2afcbc583f71f1cb630/internal/kokoro/check_incompat_changes.sh#L17-L21

jadekler:
> Forgot to cc: @vangent @jba
> 
> @jadekler, you may want to make the corresponding change to google-cloud-go @ https://github.com/googleapis/google-cloud-go/blob/ce66726a686431294383c2afcbc583f71f1cb630/internal/kokoro/check_incompat_changes.sh#L17-L21

I'm not entirely sure what's going on in this PR but it seems travis specific - is that right? (we don't use travis)

zombiezen:
It is not Travis-specific; the problem is [a rare race condition](https://github.com/google/go-cloud/issues/1802#issuecomment-482703618) that can occur especially if presubmit checks are long. (This change has also has the side-benefit of making this check run faster and hermetically, but that's not the primary reason for this change.)

The critical part is changing `git clone https://example.com/my/repo.git "$MASTER_CLONE_DIR"` to `git clone -b master . "$MASTER_CLONE_DIR"`. This way, instead of operating on the state of the live repository (which may have changed since the start of the run), the script operates on a coherent snapshot of the repository.

jadekler:
Ah!! Thanks for the writeup, will try to find some time to put in these changes!

PR #1824, opened 2019-04-13
internal/docstore: process partition key in dynamo query

shantuo:


jba:
Let's hold off on making any more modifications to drivertest to support DynamoDB. That's not a direction that makes sense in the long run, because then we won't have a portable API. When I get back to work next week we can discuss alternative ideas.

shantuo:
Drop this for now, see #1832

PR #1831, opened 2019-04-16
dynamodocstore: use Scan if we can't use Query

jba:
Fall back to using the Scan RPC if the conditions for a Query RPC
aren't met.

### DIFF
--- a/internal/docstore/dynamodocstore/query.go
+++ b/internal/docstore/dynamodocstore/query.go
@@ -26,75 +26,126 @@ import (
 	"gocloud.dev/internal/docstore/driver"
 )
 
+// TODO: support parallel scans (http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Scan.html#Scan.ParallelScan)
+
+type avmap = map[string]*dynamodb.AttributeValue
+
 func (c *collection) RunGetQuery(ctx context.Context, q *driver.Query) (driver.DocumentIterator, error) {
-	out, err := c.runGetQuery(ctx, q, nil)
+	items, last, err := c.runQueryOrScan(ctx, q, nil)
 	if err != nil {
 		return nil, err
 	}
 	return &documentIterator{
 		c:     c,
 		q:     q,
-		items: out.Items,
-		count: *out.Count,
+		items: items,
 		limit: q.Limit, // manually count limit since dynamodb uses "limit" as scan limit before filtering
-		last:  out.LastEvaluatedKey,
+		last:  last,
 	}, nil
 }
 
-func (c *collection) runGetQuery(ctx context.Context, q *driver.Query, startAfter map[string]*dynamodb.AttributeValue) (*dynamodb.QueryOutput, error) {
-	pb := expression.NamesList(expression.Name(docstore.RevisionField))
-	for _, fp := range q.FieldPaths {
-		pb = pb.AddNames(expression.Name(strings.Join(fp, ".")))
+func (c *collection) runQueryOrScan(ctx context.Context, q *driver.Query, startAfter avmap) (items []avmap, last avmap, err error) {
+	var cb expression.Builder
+	cbUsed := false // It's an error to build an empty Builder.
+	if len(q.FieldPaths) > 0 {
+		pb := expression.NamesList(expression.Name(docstore.RevisionField))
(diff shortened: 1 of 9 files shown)


internal/docstore/dynamodocstore/query.go
```diff
@@ -26,75 +26,126 @@ import (
...
+	if len(cfs) > 0 {
+		cb = cb.WithFilter(filtersToConditionBuilder(cfs))
+	}
+	return cb
+}
+
+func filtersToConditionBuilder(fs []driver.Filter) expression.ConditionBuilder {
+	var cb expression.ConditionBuilder
+	if len(fs) == 0 {
+		return cb
```
shantuo: An empty builder will later on result in a build error on the expression builder. We should make sure the fs is not empty, otherwise we might just let it panic since it is unexported.

internal/docstore/dynamodocstore/query.go
```diff
@@ -26,75 +26,126 @@ import (
...
-	cb := expression.NewBuilder().WithProjection(pb)
-	cb = processFilters(cb, q.Filters, c.partitionKey, c.sortKey)
-	ce, err := cb.Build()
-	if err != nil {
-		return nil, err
+	// If there is an equality filter on the partition key, do a query.
+	// Otherwise, do a scan.
+	doQuery := false
+	for _, f := range q.Filters {
+		if len(f.FieldPath) == 1 && f.FieldPath[0] == c.partitionKey && f.Op == driver.EqualOp {
```
shantuo: partition key and sort key could contain '/'.

internal/docstore/dynamodocstore/query.go
```diff
@@ -26,75 +26,126 @@ import (
...
 	}, nil
 }
 
-func (c *collection) runGetQuery(ctx context.Context, q *driver.Query, startAfter map[string]*dynamodb.AttributeValue) (*dynamodb.QueryOutput, error) {
-	pb := expression.NamesList(expression.Name(docstore.RevisionField))
-	for _, fp := range q.FieldPaths {
-		pb = pb.AddNames(expression.Name(strings.Join(fp, ".")))
+func (c *collection) runQueryOrScan(ctx context.Context, q *driver.Query, startAfter avmap) (items []avmap, last avmap, err error) {
+	var cb expression.Builder
+	cbUsed := false // It's an error to build an empty Builder.
```
shantuo: Is this necessary? The `Build` will return an error if it is empty.

internal/docstore/dynamodocstore/query.go
```diff
@@ -26,75 +26,126 @@ import (
...
+			KeyConditionExpression:    ce.KeyCondition(),
+			FilterExpression:          ce.Filter(),
+			ProjectionExpression:      ce.Projection(),
+			ExclusiveStartKey:         startAfter,
+		})
+		if err != nil {
+			return nil, nil, err
+		}
+		return out.Items, out.LastEvaluatedKey, nil
+	} else {
```
shantuo: nit: the else can be removed.

internal/docstore/dynamodocstore/query.go
jba: Do you mean '.'?

It turns out you're right, but it doesn't matter because it would still be a top-level field. See https://godoc.org/github.com/aws/aws-sdk-go/service/dynamodb#KeySchemaElement: "A KeySchemaElement must be a scalar, top-level attribute (not a nested attribute)." I actually created a table with a partition key called `name.first`, and I had to use that name (as a top-level key).

Currently there would be no way to query such a table because the `Where` method always parses its first argument as a field path, splitting at dots. I think that's probably fine, since keys with dots in them are probably unlikely.

internal/docstore/dynamodocstore/query.go
jba: Done.

internal/docstore/dynamodocstore/query.go
jba: I agree, panic makes sense here. Done.

internal/docstore/dynamodocstore/query.go
shantuo: Thanks for the explanation!

PR #1841, opened 2019-04-17
blob: implement BeforeCopy As hook for all drivers; cleanup Unimplemented checks

### LINKED ISSUE #1696: blob: add support for copying an object

vangent:
Fixes #1696.

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -49,6 +49,7 @@
 //  - ListOptions.BeforeList: *azblob.ListBlobsSegmentOptions
 //  - Reader: azblob.DownloadResponse
 //  - Attributes: azblob.BlobGetPropertiesResponse
+//  - CopyOptions.BeforeCopy: azblob.Metadata, *azblob.ModifiedAccessConditions, *azblob.BlobAccessConditions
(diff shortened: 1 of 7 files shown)


blob/azureblob/azureblob.go
```diff
@@ -322,7 +323,29 @@ func (b *bucket) Copy(ctx context.Context, dstKey, srcKey string, opts *driver.C
...
 	srcURL := b.containerURL.NewBlobURL(srcKey).URL()
-	resp, err := dstBlobURL.StartCopyFromURL(ctx, srcURL, nil /* metadata */, azblob.ModifiedAccessConditions{}, azblob.BlobAccessConditions{})
+	md := azblob.Metadata{}
+	mac := azblob.ModifiedAccessConditions{}
+	bac := azblob.BlobAccessConditions{}
+	if opts.BeforeCopy != nil {
+		asFunc := func(i interface{}) bool {
+			switch v := i.(type) {
+			case *azblob.Metadata:
+				*v = md
```
shantuo: Why isn't `md` a pointer type?

vangent:
The underlying type is a map, so we don't need a pointer to it to make it
modifiable.

Issue #1651, opened 2019-03-22
docstore: handle URLs


jba:
Add URL support:
```
docstore.OpenCollection(ctx, urlstr) (*Collection, error)
```
Also the URLMux type. (But see #1650).

shantuo:
Done.

PR #1843, opened 2019-04-17
internal/docstore/dynamodocstore: add URL opener

shantuo:
Updates #1651

### DIFF
--- a/internal/docstore/dynamodocstore/dynamo.go
+++ b/internal/docstore/dynamodocstore/dynamo.go
@@ -12,21 +12,133 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package dynamodocstore provides a docstore implementation backed by AWS
+// DynamoDB.
+// Use OpenCollection to construct a *docstore.Collection.
+//
+// URLs
+//
+// For docstore.OpenCollection, dynamodocstore registers for the scheme
+// "dynamodocstore". The default URL opener will use an AWS session with the
+// default credentials and configuration; see
+// https://docs.aws.amazon.com/sdk-for-go/api/aws/session/ for more details.
+// To customize the URL opener, or for more details on the URL format, see
+// URLOpener.
+// See https://godoc.org/gocloud.dev#hdr-URLs for background information.
 package dynamodocstore
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net/url"
 	"strings"
+	"sync"
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/client"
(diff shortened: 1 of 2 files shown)


internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -12,22 +12,134 @@
...
 // limitations under the License.
 
+// Package dynamodocstore provides a docstore implementation backed by AWS
+// DynamoDB.
+// Use OpenCollection to construct a *docstore.Collection.
+//
+// URLs
+//
+// For docstore.OpenCollection, dynamodocstore registers for the scheme
+// "dynamodocstore". The default URL opener will use an AWS session with the
```
vangent: "dynamo" or "dynamodb" should be sufficient, here and elsewhere.

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -12,22 +12,134 @@
...
+}
+
+// Scheme is the URL scheme dynamodocstore registers its URLOpener under on
+// docstore.DefaultMux.
+const Scheme = "dynamodocstore"
+
+// URLOpener opens dynamodocstore URLs like
+// "dynamodocstore://mytable?partition_key=partkey&sort_key=sortkey".
+//
+// The URL Host is used as the table name. See
```
vangent: Same question here as for Mongo: Is "foo/bar" a valid table name?

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -12,22 +12,134 @@
...
+	overrideCfg, err := gcaws.ConfigFromURLParams(q)
+	if err != nil {
+		return nil, fmt.Errorf("open collection %s: %v", u, err)
+	}
+	configProvider.Configs = append(configProvider.Configs, overrideCfg)
+	db, err := Dial(configProvider)
+	if err != nil {
+		return nil, fmt.Errorf("open collection %s: %v", u, err)
+	}
+	return OpenCollection(db, tableName, partitionKey, sortKey), nil
```
vangent: OK for separate PR, but `OpenCollection` needs an `Options`, which should then be added here and to `URLOpener`.

internal/docstore/dynamodocstore/dynamo.go
shantuo: Done.

internal/docstore/dynamodocstore/dynamo.go
shantuo: No, I updated the link below directly to the naming rules.

internal/docstore/dynamodocstore/dynamo.go
shantuo: Ack, filed #1877

PR #1844, opened 2019-04-17
internal/docstore/mongodocstore: add URL opener

shantuo:
Updates #1651


internal/docstore/mongodocstore/mongo.go
```diff
@@ -14,6 +14,16 @@
 
 // Package mongodocstore provides an implementation of the docstore API for MongoDB.
 //
+// URLs
+//
+// For docstore.OpenCollection, mongodocstore registers for the schemes
```
vangent: Nit: schemes -> scheme.

internal/docstore/mongodocstore/mongo.go
```diff
@@ -14,6 +14,16 @@
...
+// "mongodocstore".
```
vangent: "mongo" should be sufficient.

internal/docstore/mongodocstore/mongo.go
```diff
@@ -40,6 +55,89 @@ import (
...
+		return nil, err
+	}
+	client, err := mongo.NewClient(opts)
+	if err != nil {
+		return nil, err
+	}
+	if err := client.Connect(ctx); err != nil {
+		return nil, err
+	}
+	// Connect doesn't seem to actually make a connection, so do an RPC.
```
vangent: Why? E.g., I don't think creating a gRPC connection fails if the remote is not available (as that may be transient).

internal/docstore/mongodocstore/mongo.go
```diff
@@ -40,6 +55,89 @@ import (
...
+	if err := client.Ping(ctx, nil); err != nil {
+		return nil, err
+	}
+	return client, nil
+}
+
+// Scheme is the URL scheme mongodocstore registers its URLOpener under on
+// docstore.DefaultMux.
+const Scheme = "mongodocstore"
```
vangent: As above: `mongo`.

internal/docstore/mongodocstore/mongo.go
```diff
@@ -40,6 +55,89 @@ import (
...
+
+// URLOpener opens URLs like "mongodocstore://mydb/mycollection".
```
vangent: and here

internal/docstore/mongodocstore/mongo.go
```diff
@@ -40,6 +55,89 @@ import (
...
+//
+// The URL Host is used as the database name.
+// The URL Path is used as the colleciton name.
```
vangent: Nit: colleciton -> collection

internal/docstore/mongodocstore/mongo.go
```diff
@@ -40,6 +55,89 @@ import (
...
+	return client, nil
+}
+
+// Scheme is the URL scheme mongodocstore registers its URLOpener under on
+// docstore.DefaultMux.
+const Scheme = "mongodocstore"
+
+// URLOpener opens URLs like "mongodocstore://mydb/mycollection".
+//
+// The URL Host is used as the database name.
```
vangent: What do database names look like? E.g., is "foo/bar" a valid database name? If so, this won't work.

internal/docstore/mongodocstore/mongo.go
```diff
@@ -40,6 +55,89 @@ import (
...
+// The URL Path is used as the colleciton name.
+//
+// No query parameters are supported.
+type URLOpener struct {
+	Client *mongo.Client
```
vangent: Nit: Add comment for `Client`.

internal/docstore/mongodocstore/mongo.go
```diff
@@ -40,6 +55,89 @@ import (
...
+
+	dbName := u.Host
+	if dbName == "" {
+		return nil, fmt.Errorf("open collection %s: URL must have a non-empty Host (database name)", u)
+	}
+	collName := strings.TrimPrefix(u.Path, "/")
+	if collName == "" {
+		return nil, fmt.Errorf("open collection %s: URL must have a non-empty Path (collection name)", u)
+	}
+	return OpenCollection(o.Client.Database(dbName).Collection(collName), nil), nil
```
vangent: Should pass a real `Options` instead of `nil`. Just add it to `URLOpener`.

internal/docstore/mongodocstore/mongo.go
shantuo: Done.

internal/docstore/mongodocstore/mongo.go
shantuo: Done.

internal/docstore/mongodocstore/mongo.go
shantuo: Done.

internal/docstore/mongodocstore/mongo.go
shantuo: Done.

internal/docstore/mongodocstore/mongo.go
shantuo: Done.

internal/docstore/mongodocstore/mongo.go
shantuo: Removed.

internal/docstore/mongodocstore/mongo.go
shantuo: No, only the collection name can contain `/`s. Added a link to the naming rules.

internal/docstore/mongodocstore/mongo.go
shantuo: Done.

internal/docstore/mongodocstore/mongo.go
shantuo: Done.

PR #1845, opened 2019-04-17
dynamodocstore: separate query planning from execution

jba:
Before the first RPC of a query, create the ScanInput or QueryInput
for it.  Reuse that input for subsequent RPCs.

Also, redo the limit logic since the previous version had a couple of
problems.


internal/docstore/dynamodocstore/query.go
```diff
@@ -68,41 +71,55 @@ func (c *collection) runQueryOrScan(ctx context.Context, q *driver.Query, startA
...
+
+type queryRunner struct {
+	c       *collection
+	scanIn  *dynamodb.ScanInput
+	queryIn *dynamodb.QueryInput
+}
+
+func (qr *queryRunner) run(ctx context.Context, startAfter avmap) (items []avmap, last avmap, err error) {
+	if qr.scanIn != nil {
+		qr.scanIn.ExclusiveStartKey = last
```
shantuo: Should `last` be `startAfter`?

internal/docstore/dynamodocstore/query.go
```diff
@@ -68,41 +71,55 @@ func (c *collection) runQueryOrScan(ctx context.Context, q *driver.Query, startA
...
+		out, err := qr.c.db.ScanWithContext(ctx, qr.scanIn)
+		if err != nil {
+			return nil, nil, err
+		}
+		return out.Items, out.LastEvaluatedKey, nil
+	}
+	qr.queryIn.ExclusiveStartKey = last
```
shantuo: Same here.

internal/docstore/dynamodocstore/query.go
jba: Good catch. Fixed.

internal/docstore/dynamodocstore/query.go
jba: ditto

jba:
PTAL

PR #1846, opened 2019-04-17
internal/testing: include gocdk CLI module

zombiezen:
Forgot to do this in earlier PRs.

### DIFF
--- a/internal/testing/alldeps
+++ b/internal/testing/alldeps
@@ -30,6 +30,7 @@ github.com/google/subcommands
 github.com/google/uuid
 github.com/google/wire
 github.com/googleapis/gax-go
+github.com/googleapis/gax-go/v2
 github.com/gorilla/mux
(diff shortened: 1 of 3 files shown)


vangent:
Test is failing: FAIL: dependencies changed; compare listdeps.sh output with alldeps

but basic idea LGTM.

zombiezen:
Curious, since I did already run listdeps.sh: it seems like this discrepancy is coming from the fact that Travis is configured to run Go 1.11, but I thought we all agreed that we should start running Go 1.12. Should I update the Travis config?

vangent:
I would check with jba@. There's #1557 in flight, but it hasn't been touched in a while.

zombiezen:
Ah, that's probably where my confusion's coming from. I'll leave it as-is, hopefully tests should pass now.

PR #1848, opened 2019-04-18
internal/docstore/dynamodocstore: support running actions in group

shantuo:
This uses dynamo's transaction API. It doesn't use the `driver.SplitAction` since each transaction can only have up to 10 items, and no two write's can target the same item. I had to write its own split function for simplicity.


internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -61,45 +62,159 @@ func (c *collection) RunActions(ctx context.Context, actions []*driver.Action, u
...
 		if err != nil {
-			return driver.ActionListError{{i, err}}
+			return err
 		}
+		tws[i] = tw
 	}
-	return nil
+
+	_, err := c.db.TransactWriteItemsWithContext(ctx, &dyn.TransactWriteItemsInput{
+		ClientRequestToken: aws.String(driver.UniqueString()),
```
jba: I don't think we need this if we aren't rerunning the same RPC.

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -61,45 +62,159 @@ func (c *collection) RunActions(ctx context.Context, actions []*driver.Action, u
...
+// the composite key to be guaranteed unique.
+func (c *collection) primaryKey(a *driver.Action) string {
+	pkey, err := a.Doc.GetField(c.partitionKey)
+	if err != nil {
+		if c.ErrorCode(err) == gcerr.NotFound && a.Kind == driver.Create {
+			newPartitionKey := driver.UniqueString()
+			a.Doc.SetField(c.partitionKey, newPartitionKey)
+			return newPartitionKey
+		}
+		panic(err) // shouldn't happen
```
jba: What if someone forgets to set the key fields in a Put action? Where is the logic with `missingKeyField`?

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -61,45 +62,159 @@ func (c *collection) RunActions(ctx context.Context, actions []*driver.Action, u
...
+	}
+	if c.sortKey != "" {
+		skey, err := a.Doc.GetField(c.sortKey)
+		if err != nil {
+			panic(err) // shouldn't happen
```
jba: ditto

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -61,45 +62,159 @@ func (c *collection) RunActions(ctx context.Context, actions []*driver.Action, u
...
+		}
+		return fmt.Sprintf("%s, %s", pkey, skey)
+	}
+	return pkey.(string)
```
jba: It doesn't have to be a string.

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -61,45 +62,159 @@ func (c *collection) RunActions(ctx context.Context, actions []*driver.Action, u
...
+			return newPartitionKey
+		}
+		panic(err) // shouldn't happen
+	}
+	if c.sortKey != "" {
+		skey, err := a.Doc.GetField(c.sortKey)
+		if err != nil {
+			panic(err) // shouldn't happen
+		}
+		return fmt.Sprintf("%s, %s", pkey, skey)
```
jba: This could give the same string for two different items: `a, b, c` could be pkey=`a, b` and skey=`c`, or pkey=`a` and skey=`b, c`.

Also, the keys need not be strings.

I think this function should return a `[2]interface{}`.

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -61,45 +62,159 @@ func (c *collection) RunActions(ctx context.Context, actions []*driver.Action, u
...
+// primaryKey tries to get the primary key from the doc, which is the partition
+// key if there is no sort key, or the combination of both keys. If there is not
+// a key for Create action, it generates a partition key of the doc. It returns
+// the composite key to be guaranteed unique.
+func (c *collection) primaryKey(a *driver.Action) string {
+	pkey, err := a.Doc.GetField(c.partitionKey)
+	if err != nil {
+		if c.ErrorCode(err) == gcerr.NotFound && a.Kind == driver.Create {
+			newPartitionKey := driver.UniqueString()
+			a.Doc.SetField(c.partitionKey, newPartitionKey)
```
jba: This should only happen on success.

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -214,7 +212,17 @@ func (c *collection) runWrites(ctx context.Context, actions []*driver.Action) er
...
 		TransactItems:      tws,
 	})
-	return err
+	if err != nil {
+		return err
+	}
+	for i, a := range actions {
+		if a.Kind == driver.Create {
+			if _, err := a.Doc.GetField(c.partitionKey); err != nil && gcerrors.Code(err) == gcerrors.NotFound {
+				actions[i].Doc.SetField(c.partitionKey, tws[i].Put.Item[c.partitionKey])
```
jba: `tws[i].Put.Item[c.partitionKey]` is an `AttributeValue`, I think. You have to extract the `S` field.

jba: Can you also add something to the driver test for this case that checks that the generated partition key is a string, so this bug would have been caught? Either in this PR, or add a TODO.

shantuo: Good catch, fixed.

There is already a TODO in `testCreate` that I currently comment out the `unnamed` test case. Filed #1857

PR #1850, opened 2019-04-18
blob/gcsblob: return *storage.Reader from gcsblob escape hatch

zombiezen:
**BREAKING_CHANGE**
    
For consistency with `WriterOptions.BeforeWrite` and to preserve reader identity (don't want to duplicate state).

### DIFF
--- a/blob/gcsblob/gcsblob.go
+++ b/blob/gcsblob/gcsblob.go
@@ -40,7 +40,7 @@
 //  - Error: *googleapi.Error
 //  - ListObject: storage.ObjectAttrs
 //  - ListOptions.BeforeList: *storage.Query
-//  - Reader: storage.Reader
+//  - Reader: *storage.Reader
 //  - Attributes: storage.ObjectAttrs
 //  - CopyOptions.BeforeCopy: *storage.Copier
 //  - WriterOptions.BeforeWrite: *storage.Writer
@@ -241,11 +241,11 @@ func (r *reader) Attributes() driver.ReaderAttributes {
 }
 
 func (r *reader) As(i interface{}) bool {
-	p, ok := i.(*storage.Reader)
+	p, ok := i.(**storage.Reader)
 	if !ok {
 		return false
 	}
-	*p = *r.raw
+	*p = r.raw
 	return true
 }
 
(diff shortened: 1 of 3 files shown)


zombiezen:
Err, wait this isn't quite right. Hold off for a moment...

zombiezen:
Better. Ready for review.

vangent:
What is the problem you are trying to solve? IIRC, I made this a non-pointer on purpose to make it read-only from the user perspective. It not uncommon for `As` types to be struct for reads and pointers for writes (because only for the latter do we expect the caller to modify the value).

It seems fine to me either way.

You will have to update the test:
https://github.com/google/go-cloud/blob/master/blob/gcsblob/gcsblob_test.go#L182

zombiezen:
As a reader, I was confused by why some were pointers and others weren't. Especially since Writer is a pointer type, it looked like a copy-paste error. I understand now that the others are copying information, and I think that makes a lot of sense.

However, Reader looks like it should be treated the same way as Writer: it is fundamentally a mutable thing (reading consumes the response), and we shouldn't depend on the underlying `storage.Writer` type being safe to copy (it is passed around as a pointer).

blob/gcsblob/gcsblob.go
```diff
@@ -241,11 +241,11 @@ func (r *reader) Attributes() driver.ReaderAttributes {
 }
 
 func (r *reader) As(i interface{}) bool {
-	p, ok := i.(*storage.Reader)
+	p, ok := i.(**storage.Reader)
```
vangent: We talked about this in person, looks fine.

You need to update the test though.

zombiezen: Done.

blob/gcsblob/gcsblob_test.go
```diff
@@ -180,7 +180,7 @@ func (verifyContentLanguage) AttributesCheck(attrs *blob.Attributes) error {
 }
 
 func (verifyContentLanguage) ReaderCheck(r *blob.Reader) error {
-	var rr storage.Reader
```
vangent: Actually, you also need to update this test:
https://github.com/google/go-cloud/blob/master/blob/example_test.go#L451

zombiezen: Done. I would not have found that.

PR #1852, opened 2019-04-18
blob: set ContentMD5 in WriteAll if not given

zombiezen:
Extra safety for little cost and no end-user code change.


blob/blob.go
```diff
@@ -669,7 +669,15 @@ func (b *Bucket) newRangeReader(ctx context.Context, key string, offset, length
 
 // WriteAll is a shortcut for creating a Writer via NewWriter and writing p.
```
Rolinh: Shouldn't the docstring mention that the MD5 checksum of `p` is computed each time and verified?
Also, could there be use cases where a caller might not want such a check to happen because, eg, a blob storage solution doesn't provide MD5 verification or uses another hash algorithm such as SHA256?

zombiezen: > Shouldn't the docstring mention that the MD5 checksum of `p` is computed each time and verified?

Added a doc comment.

> Also, could there be use cases where a caller might not want such a check to happen because, eg, a blob storage solution doesn't provide MD5 verification or uses another hash algorithm such as SHA256?

No. Drivers that don't use MD5 verification already safely ignore this field as the portable Bucket type will already do the verification anyway. And if even if there were, it would be out of scope for this PR.

zombiezen:
I did... do we usually do partial records based on which tests failed? Is there a better way than just running once to gather failures and then setting `-run`? (I have not had to update the replay logs for any test in a long while, so I'm a bit rusty.)

vangent:
> I did... do we usually do partial records based on which tests failed? Is there a better way than just running once to gather failures and then setting `-run`? (I have not had to update the replay logs for any test in a long while, so I'm a bit rusty.)

Yes, usually we only update goldens for the tests that failed due to the change. It makes it more obvious what the impact of the change is (e.g., I might actually be able to review a couple of the changed goldens, whereas with 67 of them there's no way).

Separately, it's probably a good idea to re-record everything to make sure they are up-to-date with reality (e.g., new Go versions, new client library versions, etc.), but that could be done as a separate PR with no semantic change.

Might be worth updating https://github.com/google/go-cloud/blob/master/CONTRIBUTING.md#a-warning-about-tests about this.

zombiezen:
This change broke 54 tests (out of 67) without re-recording.

```
$ cd blob/gcsblob
$ git checkout 45d76e1f4203ee7e136ba9e416490de39612a706 -- testdata
$ go test |& grep 'FAIL: Test' | sed -e 's/^.*\(TestConformance.*\) (.*/\1/' | wc -l
54
```

The reviewer difference between 54 and 67 doesn't seem enough to justify reverting to me. WDYT?

vangent:
> This change broke 54 tests (out of 67) without re-recording.

Wow! I guess we use `WriteAll` a lot in the conformance tests...

> The reviewer difference between 54 and 67 doesn't seem enough to justify reverting to me. WDYT?

Agreed.

Issue #1832, opened 2019-04-16
dynamodocstore: support any query


jba:
We can and should make the dynamo driver support any query. 

The first step is falling back to using Scan. This is done in  #1831. But we should do more.

The `DescribeTable` RPC will tell us all the indexes on the table. We can use that information to construct an efficient query. The algorithm is something like:
1. If the query does not contain an equality filter, do a Scan.
2. Find all the queryables (index or table) whose partition key matches the field of some equality filter in the query.
3. Narrow the list further to those queryables whose sort key matches the field of another filter in the query, if any.
4. Of the remaining list, prefer the table to any index, and prefer local indexes to global ones. Both the table and local indices can retrieve any item attribute without additional RPCs.
5. If the resulting list contains only global indexes, select one whose list of fields includes the query's list of selected fields. 
6. If there is no such global index, do a Scan.

Step 6 is there because if a global index's fields do not include a selected field, then the only way to retrieve that field is to read the item directly from the table. We don't want to do an additional RPC per item.

One more thing: we should provide a way for the user to discover what we did. At a minimum, we should report whether we did a scan, so the user can find which queries would benefit from an index. We could also report the index we used, if any. I haven't thought much about how to surface this information in a portable way, but a simple human-readable string might suffice, since the information is probably only useful to the developer.

shantuo:
7. Firestore produces error messages when a certain query fails to suggest the user to create certain query. We might do something similar here that when we fall back to use Scan, we can produce a warning and guide them to make their query more efficient.

jba:
>  Firestore produces error messages when a certain query fails to suggest the user to create certain query. We might do something similar here that when we fall back to use Scan, we can produce a warning and guide them to make their query more efficient.

I made this #1873.

PR #1854, opened 2019-04-19
dynamodocstore: query plan with indexes

jba:
Take table indexes into account when deciding how to run a query.

Add tests for query planning.

Updates #1832. This PR builds and runs the RPCs, but we also need to
report the plan to the user.

### DIFF
--- a/internal/docstore/dynamodocstore/query.go
+++ b/internal/docstore/dynamodocstore/query.go
@@ -58,60 +58,199 @@ func (c *collection) RunGetQuery(ctx context.Context, q *driver.Query) (driver.D
 	return it, nil
 }
 
-func (c *collection) planQueryOld(q *driver.Query) (*queryRunner, error) {
+func (c *collection) planQuery(q *driver.Query) (*queryRunner, error) {
 	var cb expression.Builder
 	cbUsed := false // It's an error to build an empty Builder.
+	// Set up the projection expression.
 	if len(q.FieldPaths) > 0 {
-		pb := expression.NamesList(expression.Name(docstore.RevisionField))
+		var pb expression.ProjectionBuilder
+		hasRevisionField := false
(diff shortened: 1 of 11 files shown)


internal/docstore/dynamodocstore/query.go
```diff
@@ -47,7 +58,7 @@ func (c *collection) RunGetQuery(ctx context.Context, q *driver.Query) (driver.D
 	return it, nil
 }
 
-func (c *collection) planQuery(q *driver.Query) (*queryRunner, error) {
+func (c *collection) planQueryOld(q *driver.Query) (*queryRunner, error) {
```
shantuo: Why keeps it?

internal/docstore/dynamodocstore/query_test.go
```diff
@@ -0,0 +1,372 @@
...
+			},
+		},
+		{
+			desc: "equality filter on table partition, filter on global index sort, bad projection",
+			// Although there is a global index that matches the filters best, it doesn't
+			// have the necessary fields. So we query against the table.
+			globalIndexPartitionKey: "tableP",
+			globalIndexSortKey:      "globalS",
+			globalIndexFields:       []string{"other"},
+			query: &driver.Query{Filters: []driver.Filter{
```
shantuo: Consider adding a case that the query asks for some paths that the global index doesn't contain.

internal/docstore/dynamodocstore/query.go
jba: Fixed.

internal/docstore/dynamodocstore/query_test.go
jba: Thanks, this case actually does that but not in a clear way. I fixed it.

PR #1856, opened 2019-04-19
mongodocstore: extract the ID field with a function

jba:
Previously, we assumed the key field of every collection was "_id",
the same as MongoDB's key field. That forces every collection that
wants to be portable to use that field name.

Now, have the user provide a function to extract the ID from the
document. This allows any field (or even a combination of fields)
to be used as the MongoDB _id field.

### DIFF
--- a/internal/docstore/mongodocstore/mongo.go
+++ b/internal/docstore/mongodocstore/mongo.go
@@ -42,7 +42,7 @@ import (
 
 type collection struct {
 	coll   *mongo.Collection
-	idFunc func(interface{}) interface{}
+	idFunc func(docstore.Document) interface{}
 }
 
 type Options struct {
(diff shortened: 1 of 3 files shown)


jba:
I realized that if the user can only supply a function to extract the ID, then we can't support the feature where Create makes a unique ID, because it wouldn't know how to populate the doc with the new ID. So I also provided a way for the user to name a field as the ID field.

I think the URL should take a field name as a query parameter, and use that for `Options.IDField`.

internal/docstore/mongodocstore/mongo.go
```diff
@@ -253,6 +289,21 @@ func makeFilter(doc driver.Document) (filter bson.D, id, rev interface{}, err er
...
 
+func (c *collection) docID(doc driver.Document) (interface{}, error) {
+	if c.idField != "" {
+		id, err := doc.GetField(c.idField)
+		if err != nil {
+			return nil, gcerr.New(gcerr.InvalidArgument, err, 2, "document missing ID")
+		}
+		return id, nil
+	}
+	id := c.idFunc(doc.Origin)
```
shantuo: Should it check `idFunc != nil`?

jba:
> The user needs to provide a function to extract the ID field and I don't know why that's necessary.

Just to be clear, the user can provide an ID field or a function. The function is not needed if they provide a field. 

The function is needed in this scenario:
1. The user develops a collection in DynamoDB that has a partition key and a sort key.
2. They start to use docstore.
3. They want to migrate their data to MongoDB, or use the two providers at the same time. They will want to write code like
   ```
   dynamocoll.Get(doc)
   mongocoll.Put(doc)
   ```
For that to work, the Mongo collection must know how to create a single ID from the two key fields of doc.

internal/docstore/mongodocstore/mongo.go
jba: That's not possible because of the logic on lines 72 and 73.

PR #1862, opened 2019-04-19
pubsub: document intended use of Nack

zombiezen:
Follow-up from an offline discussion about Pub/Sub where I learned how to use Nack properly from @vangent.

### DIFF
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -163,11 +163,11 @@ func (m *Message) Ack() {
 	m.isAcked = true
 }
 
-// Nack tells the server that this Message was not processed and should be
-// redelivered (negative acknowledgement). It returns immediately, but the
+// Nack (short for negative acknowledgment) tells the server that this Message
+// was not processed and should be redelivered. It returns immediately, but the
 // actual nack is sent in the background, and is not guaranteed to succeed.
 //
-// Nack is a performance optimization for retrying transient failures: it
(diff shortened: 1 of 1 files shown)


pubsub/pubsub.go
```diff
@@ -164,8 +164,14 @@ func (m *Message) Ack() {
 }
 
 // Nack tells the server that this Message was not processed and should be
-// redelivered. It returns immediately, but the actual nack is sent in the
-// background, and is not guaranteed to succeed.
+// redelivered (negative acknowledgement). It returns immediately, but the
+// actual nack is sent in the background, and is not guaranteed to succeed.
+//
+// Nack is a performance optimization for retrying transient failures: it
```
jba: "transient failures. It"

pubsub/pubsub.go
```diff
@@ -164,8 +164,14 @@ func (m *Message) Ack() {
 }
 
 // Nack tells the server that this Message was not processed and should be
```
jba: How about

   Nack (short for negative acknowledgment) is a ...

Also note the omitted "e" in acknowledgment; that is the preferred American spelling. https://grammarist.com/spelling/acknowledgment-acknowledgement

zombiezen: Done.

pubsub/pubsub.go
zombiezen: Not digging the multiple colon-delimited phrases? 😆 Done.

PR #1865, opened 2019-04-22
internal/website: publish blob and pub/sub how-to guides

zombiezen:
The secrets guide is still under development, so not publishing yet.

Updates #1576

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -13,7 +13,8 @@
 // limitations under the License.
 
 // Package blob provides an easy and portable way to interact with blobs
-// within a storage location, hereafter called a "bucket".
+// within a storage location, hereafter called a "bucket". See
+// https://gocloud.dev/howto/blob/ for getting started guides.
 //
 // It supports operations like reading and writing blobs (using standard
 // interfaces from the io package), deleting blobs, and listing blobs in a
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -14,7 +14,7 @@
 
 // Package blob provides an easy and portable way to interact with blobs
 // within a storage location, hereafter called a "bucket". See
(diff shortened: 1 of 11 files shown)


blob/blob.go
```diff
@@ -13,7 +13,8 @@
 // limitations under the License.
 
 // Package blob provides an easy and portable way to interact with blobs
-// within a storage location, hereafter called a "bucket".
+// within a storage location, hereafter called a "bucket". See
+// https://gocloud.dev/howto/blob/ for getting started guides.
```
eliben: "for getting started guides" reads a bit weird, like it's missing punctuation. Maybe just "for a tutorial"?

pubsub/pubsub.go
```diff
@@ -13,7 +13,8 @@
 // limitations under the License.
 
 // Package pubsub provides an easy and portable way to interact with publish/
-// subscribe systems.
+// subscribe systems. See https://gocloud.dev/howto/pubsub/ for getting started
+// guides.
```
eliben: Ditto

blob/blob.go
zombiezen: Would "how-to guides" read better? The tutorials are a different section.

eliben: It would read better, though I'm not sure I would understand the distinction between "tutorials" and "how-to guides". It's a bit too fine-grained in my mind. I think of documentation in terms of "tutorial" vs "reference". Not feeling strongly about this though - wording suggestion LGTM

zombiezen: Done.

pubsub/pubsub.go
zombiezen: Done.

PR #1866, opened 2019-04-22
blob: return pointer to Attributes structs instead of copying

### LINKED ISSUE #1849: blob: Bucket.Attributes returns a value instead of a pointer
While recently porting some old code over to using the Go CDK, I was surprised to discover that `blob.Bucket.Attributes` returns `blob.Attributes` as opposed to `*blob.Attributes`:

https://github.com/google/go-cloud/blob/0c1c45556d4bd0186aa4ecffa7afb72607067c97/blob/blob.go#L562-L566

As of 0c1c45556d4bd0186aa4ecffa7afb72607067c97, `blob.Attributes` contains 10 fields (totaling 152 bytes on linux/amd64) and may increase over time: https://github.com/google/go-cloud/blob/0c1c45556d4bd0186aa4ecffa7afb72607067c97/blob/blob.go#L165-L198

IMO `blob.Attributes` should be passed around by pointer in accordance with guidance from [Go Code Review Comments](https://golang.org/wiki/CodeReviewComments#pass-values):

> Don't pass pointers as function arguments just to save a few bytes. If a function refers to its argument `x` only as `*x` throughout, then the argument shouldn't be a pointer. Common instances of this include passing a pointer to a `string` (`*string`) or a pointer to an interface value (`*io.Reader`). In both cases the value itself is a fixed size and can be passed directly. **This advice does not apply to large structs, or even small structs that might grow. [emphasis mine]**

vangent:
Fixes #1849.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -563,22 +563,22 @@ func (b *Bucket) Exists(ctx context.Context, key string) (bool, error) {
 //
 // If the blob does not exist, Attributes returns an error for which
 // gcerrors.Code will return gcerrors.NotFound.
-func (b *Bucket) Attributes(ctx context.Context, key string) (_ Attributes, err error) {
+func (b *Bucket) Attributes(ctx context.Context, key string) (_ *Attributes, err error) {
 	if !utf8.ValidString(key) {
-		return Attributes{}, gcerr.Newf(gcerr.InvalidArgument, nil, "blob: Attributes key must be a valid UTF-8 string: %q", key)
+		return nil, gcerr.Newf(gcerr.InvalidArgument, nil, "blob: Attributes key must be a valid UTF-8 string: %q", key)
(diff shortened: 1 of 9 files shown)


blob/blob.go
```diff
@@ -563,22 +563,22 @@ func (b *Bucket) Exists(ctx context.Context, key string) (bool, error) {
 //
 // If the blob does not exist, Attributes returns an error for which
 // gcerrors.Code will return gcerrors.NotFound.
```
zombiezen: Should we perhaps remark that the caller should not modify the returned `Attributes` struct?

vangent: I don't think it matters if they do, as we create a new `blob.Attributes` to return to them, so they have the only reference. It's worth adding to the driver methods though; done.

PR #1868, opened 2019-04-22
secrets/all: add a Close function to Keeper and driver.Keeper

### LINKED ISSUE #1653: secrets: add Keeper.Close method
/cc @shantuo @clausti

In the future, we may have secret keeper drivers that require resource cleanup or other such behavior. At the moment, there is no hook that a driver can expect to receive to signal such cleanup.

Similar to #1492, I think that the `Keeper` portable type and driver interface should include a `Close` method.

vangent:
Fixes #1653.

### DIFF
--- a/secrets/secrets.go
+++ b/secrets/secrets.go
@@ -61,6 +61,7 @@ package secrets // import "gocloud.dev/secrets"
 import (
 	"context"
 	"net/url"
+	"sync"
 
 	"gocloud.dev/internal/gcerr"
 	"gocloud.dev/internal/oc"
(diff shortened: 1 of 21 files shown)


secrets/secrets.go
```diff
@@ -73,6 +74,11 @@ import (
 type Keeper struct {
 	k      driver.Keeper
 	tracer *oc.Tracer
+	// mu protects the closed variable.
```
zombiezen: nit: blank line before separate mu-protected block

vangent: Done.

PR #1874, opened 2019-04-22
docstore: provide the user with query plan information

### LINKED ISSUE #1832: dynamodocstore: support any query

jba:
Let the user ask whether a query used an index, or did a table scan.

This can help users optimize queries.

Fixes #1832.


internal/docstore/dynamodocstore/query.go
```diff
@@ -385,3 +385,26 @@ func (it *documentIterator) Stop() {
 	it.items = nil
 	it.last = nil
 }
+
+func (c *collection) QueryPlan(q *driver.Query) (*driver.QueryPlan, error) {
+	if c.description == nil {
```
vangent: Does c need locking?

internal/docstore/query.go
```diff
@@ -144,3 +153,26 @@ func (it *DocumentIterator) Stop() {
 	it.err = io.EOF
 	it.iter.Stop()
 }
+
+// A QueryPlan describes how a query will be executed.
```
vangent: I'm a bit nervous about putting too much structure here given that we have an n of only 1. WDYT about just returning a string instead? Drivers can put whatever description they want, but it's intended for a human to read it.

internal/docstore/dynamodocstore/query.go
jba: I just moved it to the constructor instead. Much simpler.

internal/docstore/query.go
jba: Done.

jba:
PTAL.

internal/docstore/dynamodocstore/query.go
```diff
@@ -386,25 +379,20 @@ func (it *documentIterator) Stop() {
...
 	return qr.queryPlan(), nil
 }
 
-func (qr *queryRunner) queryPlan() *driver.QueryPlan {
-	qp := &driver.QueryPlan{Scan: qr.scanIn != nil}
-	if !qp.Scan && qr.queryIn.IndexName != nil {
-		qp.Index = *qr.queryIn.IndexName
+func (qr *queryRunner) queryPlan() string {
+	if qr.scanIn != nil {
+		return "Scan"
```
vangent: Nit: maybe "Full Scan"?

internal/docstore/firedocstore/query.go
```diff
@@ -324,7 +324,7 @@ func fieldRef(fp []string) *pb.StructuredQuery_FieldReference {
 	return &pb.StructuredQuery_FieldReference{FieldPath: toServiceFieldPath(fp)}
 }
 
-func (c *collection) QueryPlan(q *driver.Query) (*driver.QueryPlan, error) {
-	// We don't know how Firestore executes the query, so we return an empty struct.
-	return &driver.QueryPlan{}, nil
+func (c *collection) QueryPlan(q *driver.Query) (string, error) {
+	// We don't know how Firestore executes the query, so we return an empty string.
+	return "", nil
```
vangent: Maybe "unknown" instead of empty string for the ones where we don't know?

internal/docstore/dynamodocstore/query.go
jba: I don't think that adds anything. The standard term is "table scan."

internal/docstore/firedocstore/query.go

internal/docstore/query.go
```diff
@@ -144,3 +153,13 @@ func (it *DocumentIterator) Stop() {
 	it.err = io.EOF
 	it.iter.Stop()
 }
+
+// Plan describes how the query would be executed if its Get method were called with
+// the given field paths. Plan uses only information available to the client, so it
+// cannot know whether whether a service uses indexes or scans internally.
```
shantuo: nit: s/whether whether/whether

jba: Done.

PR #1876, opened 2019-04-22
firedocstore: support a function that returns the document name

### LINKED ISSUE #1858: docstore: support multi-field primary keys
Currently, docstore assumes that a single field in the document is the unique key. Many document stores work this way, but not all: DynamoDB allows two fields to act as the primary key together (that is, each one individually may not be unique, but the pair is unique). If we want to portably support all document stores, we have to account for this case in the other drivers. 

As an example, consider a collection of high scores for games. The fields are: game name, user name, and score. The primary key is the pair (game name, user name). The collection could use the struct
```
type HighScore struct {
    Game  string
    User  string
    Score int
}
```

The most general way to support this would be for drivers to accept an option that is a function from documents to primary key values. In the above example, it could be
```
func(doc docstore.Document) interface{} {
    hs := doc.(*HighScore)
    return hs.Game + "," + hs.User
}
```
(assuming that no `Game` has a comma).

I propose we support that function. But in doing so we lose the ability for Create to make up a primary key and store it back in the document: it wouldn't know how to construct the key or where to store it. So we should also support a single field name to use as the key, as we do now in most drivers. 

#1856 is the PR that does this for mongodocstore.

jba:
Fixes #1858.


internal/docstore/firedocstore/fs.go
```diff
@@ -48,18 +63,28 @@ type collection struct {
 // firedocstore requires that a single string field, nameField, be designated the
 // primary key. Its values must be unique over all documents in the collection, and
 // the primary key must be provided to retrieve a document.
-func OpenCollection(client *vkit.Client, projectID, collPath, nameField string) *docstore.Collection {
-	return docstore.NewCollection(newCollection(client, projectID, collPath, nameField))
+func OpenCollection(client *vkit.Client, projectID, collPath string, opts Options) (*docstore.Collection, error) {
```
vangent: Per your comment on the issue, I think this should be `*Options`, not `Options`. I suspect you made a non-pointer because it contains required options, but see above for an alternative.

internal/docstore/firedocstore/fs.go
```diff
@@ -33,10 +33,25 @@ import (
...
+	dbPath   string // e.g. "projects/P/databases/(default)"
+	collPath string // e.g. "projects/P/databases/(default)/documents/States/Wisconsin/cities"
+
+}
+
+// Options sets options for constructing a *docstore.Collection backed by Google Firestore.
+type Options struct {
+	// The document field to use for the document name.
+	// Exactly one of NameField or NameFunc must be set.
+	NameField string
```
vangent: It's a bit unfortunate that one of these is required, but they are hidden away in `Options`.

Also, IIUC, having a dual key now results in (likely) storing them twice (once, combined, as the name, and once separately in the doc), unlike a single key. Presumably this is because you can't recover the two fields from the name?

WDYT about something like:

```
type Namer interface {
  // Name returns a unique name for the document, along with a slice of
  // fields to delete from the document itself (because they are captured by
  // the name).
  Name(doc docstore.Document /* or should this be pdoc? */) (string, []string)
  
  // FromName processes a document name into the fields it came from.
  FromName(name string) map[string]string
}
```

We could have simple helpers to create single field and dual field implementations of `Namer`. For example, the simple case of a name based on a single field would return the value of the field plus the field name from `Name`, and a `map` from the field name to the passed-in name for `FromName`. A two-field `Namer` would concatenate the two fields somehow in `Name`, and return that plus the two field names; in `FromName`, it would reverse the concatenation operation on the passed-in name and use it to create a map from the two field names to the two split values.

The `OpenDocument` constructor would take an instance of `Namer`:

```
OpenCollection(client *vkit.Client, projectID, collPath string, namer Namer, opts *Options)
```

Alternatively, we could have separate constructors, e.g. `OpenCollection` would take a single field name and create a `Namer` internally, and have a separate `OpenCollectionWithNamer` or whatever.

jba: Regarding Options: I split into two separate OpenXXX functions.

I don't think we can remove the primary-key fields from the document because they're needed for queries.

jba:
PTAL

internal/docstore/firedocstore/fs.go
jba: Obsolete.

internal/docstore/firedocstore/fs.go
vangent: Why is it OK to remove the primary key field when there's a single one? Are we clever enough to redirect all query filters on that field to be on the document name?

internal/docstore/firedocstore/fs.go
```diff
@@ -158,17 +160,44 @@ type collection struct {
...
-	return docstore.NewCollection(newCollection(client, projectID, collPath, nameField))
+func OpenCollection(client *vkit.Client, projectID, collPath, nameField string) (*docstore.Collection, error) {
+	c, err := newCollection(client, projectID, collPath, nameField, nil)
+	if err != nil {
+		return nil, err
+	}
+	return docstore.NewCollection(c), nil
+}
+
+// OpenCollectionFunc creates a *docstore.Collection representing a Firestore collection.
```
vangent: Nit: I don't find this name descriptive. How about "OpenCollectionCompoundName" or "OpenCollection[With]NameFunc"?

jba: Went with OpenCollectionNameFunc.

internal/docstore/firedocstore/fs.go
```diff
@@ -557,24 +590,34 @@ func (c *collection) commit(ctx context.Context, ws []*pb.Write) ([]*pb.WriteRes
...
-		// Return missing field as empty string.
-		return "", nil
+// docName returns the name of the document. This is either the value of the field
+// called c.nameField, or the result of calling c.nameFunc. It must be a
+// string.
+// docName returns an error if NameField isn't present, or NameFunc returns "".
+// The second return value reports whether c.nameField is set and the field is
+// missing. This is to support the Create action, which can create new document
+// names.
+
```
shantuo: Remove the blank line.

internal/docstore/firedocstore/fs.go
```diff
@@ -557,24 +590,34 @@ func (c *collection) commit(ctx context.Context, ws []*pb.Write) ([]*pb.WriteRes
...
-// create new document names.
-func (c *collection) docName(doc driver.Document) (string, error) {
-	n, err := doc.GetField(c.nameField)
-	if err != nil {
-		// Return missing field as empty string.
-		return "", nil
+// docName returns the name of the document. This is either the value of the field
+// called c.nameField, or the result of calling c.nameFunc. It must be a
+// string.
+// docName returns an error if NameField isn't present, or NameFunc returns "".
```
shantuo: Should this be an "and" in between? It returns error when neither name field nor name func works right?

internal/docstore/firedocstore/fs.go
jba: That's a good point. I forgot about handling queries on the name. I'll have to see how that works in firestore. I filed #1886.

The problem with multiple primary key fields is that the driver couldn't know how they're combined into the key. A collection with `Game` and `User` fields may construct the key as `Game,User` or `Game|User` or `User/Game`. How could the driver translate the filter `User = "bob"` into a query on the key?

internal/docstore/firedocstore/fs.go

internal/docstore/firedocstore/fs.go
jba: I reworded to clarify.

jba:
PTAL

PR #1880, opened 2019-04-23
all: add cleanup functions to Cloud SQL Open function returns

### LINKED ISSUE #1798: postgres/cloudpostgres: Open should return a cleanup function

The [`cloudpostgres.Open`](https://godoc.org/gocloud.dev/postgres/cloudpostgres#Open) function can't be used directly in Wire without leaking resources because it does not include a [cleanup function](https://github.com/google/wire/blob/master/docs/guide.md#cleanup-functions)

The signature of `Open` should change from:

```go
func Open(ctx context.Context, certSource proxy.CertSource, params *Params) (*sql.DB, error)
```

to:

```go
func Open(ctx context.Context, certSource proxy.CertSource, params *Params) (*sql.DB, func(), error)
```

The cleanup function should close the returned database connection.

zombiezen:
Fixes #1798

### DIFF
--- a/mysql/cloudmysql/cloudmysql.go
+++ b/mysql/cloudmysql/cloudmysql.go
@@ -75,8 +75,9 @@ type Params struct {
 	TraceOpts []ocsql.TraceOption
 }
 
-// Open opens a Cloud SQL database.
-func Open(ctx context.Context, certSource proxy.CertSource, params *Params) (*sql.DB, error) {
+// Open opens a Cloud SQL database. The second return value is a Wire cleanup
+// function that calls Close on the returned database.
+func Open(ctx context.Context, certSource proxy.CertSource, params *Params) (*sql.DB, func(), error) {
 	// TODO(light): Avoid global registry once https://github.com/go-sql-driver/mysql/issues/771 is fixed.
 	dialerCounter.mu.Lock()
 	dialerNum := dialerCounter.n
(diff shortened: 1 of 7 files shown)


mysql/cloudmysql/cloudmysql.go
```diff
@@ -75,8 +75,9 @@ type Params struct {
 	TraceOpts []ocsql.TraceOption
 }
 
-// Open opens a Cloud SQL database.
-func Open(ctx context.Context, certSource proxy.CertSource, params *Params) (*sql.DB, error) {
+// Open opens a Cloud SQL database. The second return value is a Wire cleanup
```
vangent: Nit: I would leave `Wire` out of the description; if you use this without wire you can still use it.

postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -155,13 +156,14 @@ type Params struct {
 	TraceOpts []ocsql.TraceOption
 }
 
-// Open opens a Cloud SQL database.
-func Open(ctx context.Context, certSource proxy.CertSource, params *Params) (*sql.DB, error) {
+// Open opens a Cloud SQL database. The second return value is a Wire cleanup
```
vangent: Same.

mysql/cloudmysql/cloudmysql.go
zombiezen: Done.

PR #1894, opened 2019-04-23
secrets: tests for a closed Keeper; make sure to wrap Close err

vangent:


secrets/secrets.go
```diff
@@ -155,7 +155,7 @@ func (k *Keeper) Close() error {
 	if prev {
 		return errClosed
 	}
-	return k.k.Close()
+	return wrapError(k, k.k.Close())
```
shantuo: wrapError needs to check `err != nil`

vangent: I think it ended up doing so via gcerrs.DoNotWrap, but better to be explicit. Done.

PR #1896, opened 2019-04-23
internal/website: update how-to docs for recent changes

vangent:


vangent:
We are going to have to be very careful to keep the documentation up to date. This set of changes from the last month or so illustrates how easy it is to get out of sync. The live examples will help.

zombiezen:
> We are going to have to be very careful to keep the documentation up to date. This set of changes from the last month or so illustrates how easy it is to get out of sync. The live examples will help.

I think that's a good incentive to make us think about breaking changes. It affects our users, too. The fact that good documentation requires maintenance doesn't justify not writing documentation.

Issue #1675, opened 2019-03-26
docstore: support As for action lists and queries


jba:
Docstore users will want an escape hatch before running actions or queries. For example, before executing an action, a user might want to modify the actual provider request to use a provider-specific feature not supported by docstore.

Docstore supports action lists, written as
```
coll.Actions().Get(doc1).Put(doc2).Do(ctx)
```
as well as convenience function for single actions:
```
coll.Get(ctx, doc)
```
Queries will have a fluid API like action lists:
```
coll.Where("field", "==", value).Limit(10).Get(ctx)
```

I propose that we support `As` for action lists and queries by exploiting their fluid API. We can just add an `As` method:
```
type AsFunc func(interface{}) bool

coll.Actions().Get(doc1).Put(doc2).As(func(AsFunc)).Do(ctx)

coll.Where("field", "==", value).Limit(10).As(func(AsFunc)).Get(ctx)
```
The "asFunc" pattern is the same that we use elsewhere (e.g. [blob ListOptions](https://godoc.org/gocloud.dev/blob#ListOptions)).

We don't need to support `As` for individual action methods, because users can always use the action list form. That means we don't need to clutter the individual methods with an additional argument.

vangent:
SGTM.

shantuo:
For query, there are two things we could support:

1. `BeforeQuery` for query to let the user modify the request before query runs
1. `As` for iterator to let the user directly access the query output

For action list, it is a bit complicated, as we are doing group actions and the user shouldn't know about the grouping. For example, if the user does `coll.Actions().Get(doc1)...Get(doc100).BeforDo(..).Do(ctx)`, those actions will run in several groups. There a several things we might do:

1. `BeforeDo` for running the action list and let the user modify some universal options that applies to all of the actions.
1. Per action `As` functions that only affect the action last added to the list, which give the user access to the individual actions and modify them. For example, `coll.Actions().Get(doc1).BeforeGet(..).Put(doc2).BeforePut(..).Do(ctx)`.This seems a bit tedious but should be usable since the user would likely do this in a loop and write once for each type of action.
1. Result of the action, probably not very useful even for get, since it is the raw format before the decoder.

jba:
In conversation, we decided to go with a single `BeforeDo` per action list. Its argument, a callback function, will be invoked multiple times, one for each RPC input. For example, if an action list is divided into three separate batch RPCs, the `BeforeDo` callback will be invoked three times.

jba:
[...]

PR #1897, opened 2019-04-24
internal/docstore: portable API as support for query

shantuo:
Updates #1675

### DIFF
--- a/internal/docstore/query.go
+++ b/internal/docstore/query.go
@@ -91,6 +91,15 @@ func (q *Query) Limit(n int) *Query {
 	return q
 }
 
+// BeforeQuery takes a callback function that will be called before the Query is
+// executed to the underlying provider's query functionality. The callback takes
+// a parameter, asFunc, that converts its argument to provider-specific types.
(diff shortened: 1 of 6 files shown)


shantuo:
This PR contains two `As` support:

- `BeforeQuery` for query to let the user modify the request before query runs
- `As` for iterator to let the user directly access the query output

internal/docstore/query.go
```diff
@@ -91,6 +91,15 @@ func (q *Query) Limit(n int) *Query {
 	return q
 }
 
+// BeforeQuery takes a callback function that will be called before the Query is
+// executed to the underlying provider's query functionality. The callback takes
+// a parameter, asFunc, that converts its argument to provider-specific types.
+// See https://godoc.org/gocloud.dev#hdr-As for background information.
+func (q *Query) BeforeQuery(beforeQuery func(asFunc func(interface{}) bool) error) *Query {
```
jba: nit: call the arg something simple like `f` to avoid repeating "beforeQuery".

shantuo: Done.

Issue #1881, opened 2019-04-23
internal/cmd/gocdk: serve should rebuild server when source changes


zombiezen:
When `gocdk serve` runs, it should also watch for filesystem changes and use that to trigger a rebuild. To keep the transition graceful, `gocdk serve` should reverse proxy the HTTP traffic, waiting until the newly built server reports healthy to start serving traffic to the other server.

PR #1903, opened 2019-04-24
internal/cmd/gocdk: add reverse proxy and live rebuilding

zombiezen:
I'm using SIGUSR1 to simulate filesystem changes for now, but I think this useful enough behavior (manual rebuild trigger) that I think we should keep it around.

No tests, unfortunately, because the end-to-end would require two complete programs as input. Do-able, but trickier than manual testing at the moment.

Updates #1881


internal/cmd/gocdk/serve.go
```diff
@@ -45,39 +51,70 @@ func serve(ctx context.Context, pctx *processContext, args []string) error {
...
+	biome      string
+	address    string
+	basePort   int
+}
+
+// serveBuildLoop builds and runs the user's server and sets the proxy's
+// backend whenever a new built server becomes healthy. This loop will continue
+// until ctx's Done channel is closed. serveBuildLoop returns an error only
+// if it was unable to start the main build loop.
+func serveBuildLoop(ctx context.Context, pctx *processContext, logger *log.Logger, myProxy *serveProxy, proxyAddr *net.TCPAddr, opts *serveOptions) error {
```
vangent: It looks like `proxyAddr` is only used as a string for output, maybe just pass it as a string?

vangent: (avoiding the kind-of-weird cast to `*net.TCPAddr`)

internal/cmd/gocdk/serve.go
```diff
@@ -91,47 +128,73 @@ func serveLoop(ctx context.Context, pctx *processContext, logger *log.Logger, ad
...
-	if err := buildForServe(ctx, pctx, moduleRoot, alloc.exePath); err != nil {
-		return err
+	allocB := &serverAlloc{
+		exePath: filepath.Join(buildDir, "serverB"),
+		port:    opts.basePort + 1,
 	}
-	process, err := alloc.start(ctx, pctx, logger)
-	if err != nil {
-		return err
+	currAlloc, nextAlloc := allocA, allocB
```
vangent: Nit: I found the name of these variables to be backwards (you build/run a new server into `curAlloc`, while `nextAlloc` has the currently running server most of the time).

internal/cmd/gocdk/serve.go
```diff
@@ -22,19 +22,25 @@ import (
...
 )
 
 func serve(ctx context.Context, pctx *processContext, args []string) error {
 	f := newFlagSet(pctx, "serve")
-	address := f.String("address", "localhost:8080", "address to serve on")
-	biome := f.String("biome", "dev", "biome to apply and use configuration from")
+	opts := new(serveOptions)
+	f.StringVar(&opts.address, "address", "localhost:8080", "`host:port` to address to serve on")
+	f.StringVar(&opts.biome, "biome", "dev", "`name` of biome to apply and use configuration from")
+	f.IntVar(&opts.basePort, "base-port", 9090, "use `port` and port+1 for the built servers to listen on")
```
vangent: Consider a flag to disable the reverse proxy and just run the server directly?

internal/cmd/gocdk/serve.go
```diff
@@ -91,47 +128,73 @@ func serveLoop(ctx context.Context, pctx *processContext, logger *log.Logger, ad
...
+			logger.Printf("Starting server: %v", err)
+			if process == nil {
+				myProxy.setBuildError(err)
+			}
+			continue
+		}
+
+		// Server started successfully. Cut traffic over.
+		myProxy.setBackend(currAlloc.url(""))
+		if process == nil {
```
clausti: could we also use `first` for this?

internal/cmd/gocdk/main.go
```diff
@@ -111,6 +111,11 @@ func osProcessContext() (*processContext, error) {
 	}, nil
 }
 
+// overrideEnv returns a copy of pctx.env that has vars appended to the end.
+func (pctx *processContext) overrideEnv(vars ...string) []string {
+	return append(pctx.env[:len(pctx.env):len(pctx.env)], vars...)
```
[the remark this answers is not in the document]
zombiezen: My intent with this line is to always create a new slice, since I don't want to mutate `pctx.env` or its backing array. Thus, this makes `overrideEnv` safe to call from multiple goroutines. By setting the capacity to the length, `append` will always allocate unless `len(vars) == 0`.

internal/cmd/gocdk/serve.go
```diff
@@ -22,19 +22,25 @@ import (
...
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/xerrors"
 )
 
 func serve(ctx context.Context, pctx *processContext, args []string) error {
 	f := newFlagSet(pctx, "serve")
-	address := f.String("address", "localhost:8080", "address to serve on")
-	biome := f.String("biome", "dev", "biome to apply and use configuration from")
+	opts := new(serveOptions)
+	f.StringVar(&opts.address, "address", "localhost:8080", "`host:port` to address to serve on")
```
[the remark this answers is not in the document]
zombiezen: Done.

internal/cmd/gocdk/serve.go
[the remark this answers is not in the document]
zombiezen: `-address` specifies the port the user should connect to that has the reverse proxy and `-base-port` specifies the two ports that the proxied servers should run under. Open to better names for this.

zombiezen: The user can directly connect to the server at `localhost:base-port` or run the server with `go run`. Configurability of whether to reverse-proxy or not seems like it add more branching to the logic for a relatively uncommon case.

internal/cmd/gocdk/serve.go
```diff
@@ -45,39 +51,70 @@ func serve(ctx context.Context, pctx *processContext, args []string) error {
...
+	proxyListener, err := net.Listen("tcp", opts.address)
+	if err != nil {
+		return xerrors.Errorf("gocdk serve: %w", err)
+	}
+	myProxy := new(serveProxy)
+	group, groupCtx := errgroup.WithContext(ctx)
+	group.Go(func() error {
+		return runHTTPServer(groupCtx, proxyListener, myProxy)
+	})
+	proxyListener.Addr().String()
```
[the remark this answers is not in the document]
zombiezen: Nothing. Gone.

internal/cmd/gocdk/serve.go
zombiezen: Done. (Albeit still with the type conversion.)

internal/cmd/gocdk/serve.go
zombiezen: Ack, thanks for pointing that out. What about `buildAlloc` and `otherAlloc`?

internal/cmd/gocdk/serve.go
zombiezen: No: the loop could iterate a few times before the server starts successfully in the case of failure to build or failure to start. `process != nil` indicates that a process has built and started successfully in the past.

internal/cmd/gocdk/main.go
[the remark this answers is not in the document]
vangent: That is super subtle. I would suggest making it more explicit (e.g., just copy the source slice explicitly, then append. Or at the very least add a comment.

internal/cmd/gocdk/serve.go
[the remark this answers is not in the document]
vangent: Just use constants for the latter? Or use the `--address` port +1 and +2?

Otherwise, I think you need to explain the architecture (that there is a reverse proxy, and 1-2 additional servers being proxied) in the flag. I don't think there's any flag name that will make any sense if the user doesn't know that.

internal/cmd/gocdk/serve.go
vangent: How about `liveAlloc` and `spareAlloc`? (where you build into `spareAlloc`, and then swap it to `liveAlloc` once it's up).

zombiezen: I like that naming, but I'm concerned about a slight imprecision. At the start, `liveAlloc` will not represent a live server, but in steady state it behaves as you say.

I'd be fine using the naming if you and @clausti are okay with it. It does seem strictly better than `currAlloc` and `nextAlloc`.

internal/cmd/gocdk/serve.go
vangent: Will `go run` actually work? IIUC, `gocdk serve` is doing some work other than just launching the server, no? E.g., generating the config file, adding a command-line flag with the config.

I can think of a number of reasons why this "run 2 servers proxying 1 of them" thing might not work for a given application:

* It binds to a specific port other than the HTTP one (e.g., gRPC server).
* It connects to some kind of global resource (e.g., subscribes to a topic).
* It writes to a specific file on the filesystem.

It's fine with me to add a TODO for this, but I think we need a reasonable escape hatch. If `go run` works, that would be fine, but I'm not sure it will. Connecting to the base-port directly doesn't help.

internal/cmd/gocdk/main.go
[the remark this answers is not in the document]
zombiezen: Comment added.

internal/cmd/gocdk/serve.go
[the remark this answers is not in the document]
zombiezen: Used `-address` port +1 and +2, thanks for the idea! (I slightly modified for the case where the `-address` port is zero.)

zombiezen: Opened #1914 to track.

internal/cmd/gocdk/serve.go
clausti: I agree that `liveAlloc` and `spareAlloc` are easier to make sense of than `current` and `next`

zombiezen: Done.

internal/cmd/gocdk/serve.go
clausti: @vangent if a user doesn't desire the live reloading, or it doesn't work for their application for any reason, in terms of escape hatch, we've said that they will also be able to deploy to local (and directly launch to local, as long as they've built the docker image at least once).

PR #1915, opened 2019-04-25
pubsub/all: add an Message.BeforeSend As hook

### LINKED ISSUE #1907: pubsub: add an As hook for sending a message
We have a `Message.As` for accessing the provider's type for a received message, but no way for a user to set provider-specific things as a message is being sent.

vangent:
Fixes #1907.

### DIFF
--- a/pubsub/azuresb/azuresb.go
+++ b/pubsub/azuresb/azuresb.go
@@ -269,6 +269,18 @@ func (t *topic) SendBatch(ctx context.Context, dms []*driver.Message) error {
 	for k, v := range dm.Metadata {
 		sbms.Set(k, v)
 	}
+	if dm.BeforeSend != nil {
+		asFunc := func(i interface{}) bool {
+			if p, ok := i.(**servicebus.Message); ok {
+				*p = sbms
+				return true
+			}
+			return false
+		}
+		if err := dm.BeforeSend(asFunc); err != nil {
+			return err
+		}
+	}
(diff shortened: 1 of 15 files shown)


pubsub/driver/driver.go
```diff
@@ -52,6 +52,15 @@ type Message struct {
 	// see Topic.As for more details.
 	// AsFunc must be populated on messages returned from ReceiveBatch.
 	AsFunc func(interface{}) bool
+
+	// BeforeSend is a callback used when sending a message. It should remain
+	// nil on messages returned from ReceiveBatch.
+	//
+	// The callback must called exactly once, before the message is sent.
```
jba: must => must be

pubsub/azuresb/azuresb.go
```diff
@@ -269,6 +269,18 @@ func (t *topic) SendBatch(ctx context.Context, dms []*driver.Message) error {
 	for k, v := range dm.Metadata {
 		sbms.Set(k, v)
 	}
+	if dm.BeforeSend != nil {
```
jba: Missing "As" doc comment at top of file?

pubsub/driver/driver.go
vangent: Done.

pubsub/azuresb/azuresb.go
vangent: Done.

PR #1917, opened 2019-04-25
blob/gcsblob: expose storage.ObjectHandle via As

### LINKED ISSUE #1916: blob/gcsblob: permit setting of GCS preconditions for write
See #1851 for context. This issue is specifically to give `As` access to the `ObjectHandle` during writes.

vangent:
Fixes #1916.

### DIFF
--- a/blob/gcsblob/gcsblob.go
+++ b/blob/gcsblob/gcsblob.go
@@ -253,8 +253,13 @@ func (b *bucket) ErrorCode(err error) gcerrors.ErrorCode {
 	if err == storage.ErrObjectNotExist {
 		return gcerrors.NotFound
 	}
-	if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == 404 {
-		return gcerrors.NotFound
+	if gerr, ok := err.(*googleapi.Error); ok {
+		switch gerr.Code {
+		case 404:
+			return gcerrors.NotFound
+		case 412:
+			return gcerrors.FailedPrecondition
+		}
 	}
 	return gcerrors.Unknown
 }
@@ -451,7 +456,7 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
 	var w *storage.Writer
 	if opts.BeforeWrite != nil {
 		asFunc := func(i interface{}) bool {
-			if p, ok := i.(***storage.ObjectHandle); ok {
+			if p, ok := i.(***storage.ObjectHandle); ok && w == nil {
 				*p = objp
 				return true
 			}
--- a/blob/gcsblob/gcsblob.go
+++ b/blob/gcsblob/gcsblob.go
@@ -43,7 +43,7 @@
 //  - Reader: *storage.Reader
 //  - Attributes: storage.ObjectAttrs
 //  - CopyOptions.BeforeCopy: *storage.Copier
-//  - WriterOptions.BeforeWrite: *storage.Writer
+//  - WriterOptions.BeforeWrite: **storage.ObjectHandle, *storage.Writer
 package gcsblob // import "gocloud.dev/blob/gcsblob"
 
 import (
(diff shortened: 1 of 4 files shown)


blob/gcsblob/gcsblob.go
```diff
@@ -430,28 +430,47 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
...
+		return w
+	}
+
+	var w *storage.Writer
 	if opts.BeforeWrite != nil {
 		asFunc := func(i interface{}) bool {
-			p, ok := i.(**storage.Writer)
-			if !ok {
-				return false
+			if p, ok := i.(***storage.ObjectHandle); ok {
```
zombiezen: Should we perhaps disallow reading the `ObjectHandle` once the `Writer` has been made?

blob/gcsblob/gcsblob.go
```diff
@@ -43,7 +43,7 @@
 //  - Reader: *storage.Reader
 //  - Attributes: storage.ObjectAttrs
 //  - CopyOptions.BeforeCopy: *storage.Copier
-//  - WriterOptions.BeforeWrite: *storage.Writer
+//  - WriterOptions.BeforeWrite: **storage.ObjectHandle, *storage.Writer
```
zombiezen: The double (and thus triple as seen below) pointer gives me pause. I could imagine providing some larger API surface to smooth it out, but it's likely not worth it.

blob/gcsblob/gcsblob_test.go
```diff
@@ -143,9 +143,17 @@ func (verifyContentLanguage) ErrorCheck(b *blob.Bucket, err error) error {
 }
 
 func (verifyContentLanguage) BeforeWrite(as func(interface{}) bool) error {
+	var objp **storage.ObjectHandle
+	if !as(&objp) {
+		return errors.New("Writer.As failed to get ObjectHandle")
+	}
+	// Replace the ObjectHandle with a new one that adds Conditions
+	// (this condition will always be true).
+	*objp = (*objp).If(storage.Conditions{DoesNotExist: true})
```
zombiezen: I think we should make this a separate test case: it seems like making the precondition fail is the test case that would  ensure that this escape hatch worked. Otherwise, if it's always true, then it would be the same as if the escape hatch didn't modify the outgoing request.

blob/gcsblob/gcsblob.go
vangent: Yes, good idea; done.

blob/gcsblob/gcsblob.go
vangent: Agree.

blob/gcsblob/gcsblob_test.go
vangent: Done.

vangent:
PTAL.

blob/gcsblob/gcsblob.go
```diff
@@ -253,8 +253,13 @@ func (b *bucket) ErrorCode(err error) gcerrors.ErrorCode {
 	if err == storage.ErrObjectNotExist {
 		return gcerrors.NotFound
 	}
-	if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == 404 {
-		return gcerrors.NotFound
+	if gerr, ok := err.(*googleapi.Error); ok {
+		switch gerr.Code {
+		case 404:
```
zombiezen: nit: You can use the `http.Status*` constants, like `http.StatusNotFound`, to make this mapping a little easier to follow.

blob/gcsblob/testdata/TestPreconditions.yaml
```diff
@@ -0,0 +1,219 @@
...
+      Content-Type:
+      - multipart/related; boundary=d438767e09cb038bb48032e680a4b45b57396989db1a085ea8b2d908869a
+      User-Agent:
+      - google-api-go-client/0.5 go-cloud/blob/0.1.0
+      X-Goog-Api-Client:
+      - gl-go/1.12.0 gccl/20180226
+    url: https://www.googleapis.com/upload/storage/v1/b/go-cloud-blob-test-bucket/o?alt=json&ifGenerationMatch=-999&prettyPrint=false&projection=full&uploadType=multipart
+    method: POST
+  response:
+    body: '{"error":{"errors":[{"domain":"global","reason":"conditionNotMet","message":"Precondition
```
zombiezen: Nice stack trace.

blob/gcsblob/gcsblob.go
vangent: Done.

Issue #1827, opened 2019-04-15
internal/website: how-to example code can drift from godoc examples


zombiezen:
### Is your feature request related to a problem? Please describe.

The how-to guides under [internal/website/content/howto](https://github.com/google/go-cloud/tree/master/internal/website/content/howto) often duplicate example code from the godoc reference, but sometimes with slight differences.

Long-term, this can cause maintenance burden as improvements in one rendering likely should carry over into the other rendering. Further, the Hugo-pasted example code does not ever get compiled, so it is more likely that the code may not compile correctly.

### Describe the solution you'd like

The Hugo site should somehow reference/copy example code from the godoc examples. The hackiest way of implementing this would be a presubmit check that fails if skew occurs, but ideally the Hugo site could reference examples from godoc, preferably with a way of doing some minimal rewrites (e.g. `log.Fatal(err)` -> `return err` and removing blank identifier assignments).

### Describe alternatives you've considered

I've considered making the how-to guides in Hugo be the canonical source of truth, as we definitely need them in the how-to guides. However, I rejected this because:

-  many people already consume documentation in godoc.org directly
-  examples in godoc get compiled and optionally run, improving confidence in their correctness
-  making the user navigate back and forth just to show examples seems unreasonable

zombiezen:
I did some research on this today. Hugo doesn't allow running subprocesses (see gohugoio/hugo#796), so such a mechanism would have to run out-of-band, thus complicating the site build step. Not terrible, but it makes `hugo serve` not directly usable.

I looked at the [GCP Go samples](https://github.com/GoogleCloudPlatform/golang-samples/) and the approach there seems to be to use magic comments like `// [START/END name_of_example]` to gather snippets for rendering in GCP docs ([example with GCS](https://github.com/GoogleCloudPlatform/golang-samples/blob/master/storage/buckets/main.go)). Importantly, they have standalone sample files that contain the code: these are separate from the godoc examples. What this buys is that the godoc examples won't have these magic comments rendered in them. On the other hand, these samples live far away (a wholly different repository) from the functions they exemplify. Doing this with simple text parsing also loses the import statements.

What I'm thinking is this:

1.  We annotate our godoc examples with more naturalistic magic comments. For example `// Variables set up elsewhere` for the intro block of `ctx := context.Background()` and `// Unused variables` for assignments to the blank identifier at the end.
[...]

PR #1920, opened 2019-04-25
internal/website: rewrite blob how-to guide to use blob godoc examples

zombiezen:
Add Hugo shortcode to read from examples.json.

I rewrote the blob godoc examples to match what was in the how-to guides. The page rendering is slightly different, but no substantive content change.

Updates #1827

### DIFF
--- a/blob/azureblob/example_test.go
+++ b/blob/azureblob/example_test.go
@@ -24,12 +24,17 @@ import (
 )
 
 func Example() {
+	// This example is used in https://gocloud.dev/howto/blob/open-bucket/#azure-ctor
+
+	// Variables set up elsewhere:
+	ctx := context.Background()
+
 	const (
-		// Your Azure Storage Acount and Access Key.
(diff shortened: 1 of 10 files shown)


blob/azureblob/example_test.go
```diff
@@ -93,14 +91,21 @@ func Example_sasToken() {
 }
 
 func Example_openBucket() {
+	// This example is used in https://gocloud.dev/howto/blob/open-bucket/#azure
+
+	// import _ "gocloud.dev/blob/azureblob"
```
eliben: Why is this needed?

zombiezen: The embedded examples will also show used imports, but it won't show imports not referenced in the example. This is one of the lines gatherexamples processes to add it into the imports shown in the example.

It's useful to show this in documentation because without this blank import, the `blob.OpenBucket` line will return an error (the URL opener won't be linked in).

PR #1921, opened 2019-04-25
blob/all: add a ReaderOptions.BeforeRead As hook; fill in more GCS As hook

### LINKED ISSUE #1851: blob/all: add a BeforeRead As hook (e.g., to set GCS preconditions for a read)

GCS supports a feature called [preconditions](https://cloud.google.com/storage/docs/generations-preconditions#_Preconditions) which informs the server to abort the read or write if conditions about the object's version are not met. This is useful for avoiding races in read-modify-write semantics. The way this is exposed in the GCS client library is by creating a new [`ObjectHandle`](https://godoc.org/cloud.google.com/go/storage#ObjectHandle) using [`ObjectHandle.If`](https://godoc.org/cloud.google.com/go/storage#ObjectHandle.If) or [`ObjectHandle.Generation`](https://godoc.org/cloud.google.com/go/storage#ObjectHandle.Generation).

Unfortunately, the current set of escape hatches we provide don't allow a user to get at the `ObjectHandle`: the reader and writer escape hatches use `*storage.Reader` and `*storage.Writer`, respectively, which are created from the `ObjectHandle`.

I'd like the `As` functions for both reading and writing to permit mutating the `ObjectHandle` used to create the `storage.Reader` and `storage.Writer`.

On the writing side, this would mostly mean deferring the creation of the writer until the `BeforeWrite` function explicitly requests the `*storage.Writer`. This is likely minimal impact.

The reading side is more complicated. The current implementation of `gcsblob.bucket.NewRangeReader` calls `storage.ObjectHandle.NewRangeReader`: https://github.com/google/go-cloud/blob/0c1c45556d4bd0186aa4ecffa7afb72607067c97/blob/gcsblob/gcsblob.go#L388-L395

Under the hood, this already sends the HTTP request and reads the status code and headers. This is necessary because `NewRangeReader` is required to return an error if the object does not exist: https://github.com/google/go-cloud/blob/0c1c45556d4bd0186aa4ecffa7afb72607067c97/blob/driver/driver.go#L242-L248

The current escape hatch is on the `blob.Reader` type, which can't possibly meaningfully mutate the `ObjectHandle` at this point, as the request necessarily must have started. Therefore, we likely need to introduce a `ReaderOption` escape hatch that is a `BeforeRead` or some such callback that mirrors the semantics of `BeforeWrite`.

### Additional context

AFAIK none of our other supported blob provider supports functionality similar to preconditions, but in any case, this is fine to stay as a non-portable operation requiring `As()` methods.

vangent:
Fixes #1851.

I filled in GCS BeforeCopy's As hook as well.


vangent:
Ping?

PR #1923, opened 2019-04-25
dynamodocstore: reject a partial local index when selecting all fields

jba:
Unless a local index specifies ALL as its projection, it cannot in
general satisfy a query that wants all the fields of the
document. (The DynamoDB documentation claims that fields missing in a
local index are retrieved from the table, but apparently that only
applies to fields that are explicitly named in the query.)

### DIFF
--- a/internal/docstore/dynamodocstore/query.go
+++ b/internal/docstore/dynamodocstore/query.go
@@ -171,6 +171,7 @@ func (c *collection) bestQueryable(q *driver.Query) (indexName *string, pkey, sk
 // of a query. Since DynamoDB will read explicitly provided fields from the table if
 // they are not projected into the index, the only case where a local index cannot
 // be used is when the query wants all the fields, and the index projection is not ALL.
+// See https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/LSI.html#LSI.Projections.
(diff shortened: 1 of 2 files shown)


internal/docstore/dynamodocstore/query.go
```diff
@@ -166,12 +167,20 @@ func (c *collection) bestQueryable(q *driver.Query) (indexName *string, pkey, sk
...
-// Reports whether the fields selected by the query are projected into (that is,
-// contained directly in) the global index. We need this check before using the
-// index, because if a global index doesn't have all the desired fields, then a
-// separate RPC for each returned item would be necessary to retrieve those fields,
-// and we'd rather scan than do that.
-func (c *collection) fieldsIncluded(q *driver.Query, gi *dynamodb.GlobalSecondaryIndexDescription) bool {
+// localFieldsIncluded reports whether a local index supports all the selected fields
+// of a query. Since DynamoDB will read explicitly provided fields from the table if
+// they are not projected into the index, the only case where a local index cannot
+// be used is when the query wants all the fields, and the index projection is not ALL.
```
shantuo: Could you add a link of where this is documented, I'm talking about the last sentence.

jba: Done.

PR #1924, opened 2019-04-25
blob: migrate gcsblob tests to use httpreplay instead of go-vcr

eliben:


internal/testing/setup/setup.go
```diff
@@ -151,44 +151,51 @@ func NewAWSSession2(ctx context.Context, t *testing.T, region string) (sess *ses
...
 		if err != nil {
 			t.Fatal(err)
 		}
+		cleanup := func() {
+			if err := rec.Close(); err != nil {
+				t.Fatal(err)
+			}
+		}
+		return &gcp.HTTPClient{Client: *client}, client.Transport, cleanup
 	} else {
```
vangent: Nit: You can drop the `else` here because of the return above.

vangent:
> setup.go:188: open testdata/TestPreconditions.replay: no such file or directory

I think that's a merge issue, it's a new test I just added.

internal/testing/setup/setup.go
eliben: Done

eliben:
> > setup.go:188: open testdata/TestPreconditions.replay: no such file or directory
> 
> I think that's a merge issue, it's a new test I just added.

Ugh... and this new test now fails when recording :-) Will investigate why

eliben:
> > > setup.go:188: open testdata/TestPreconditions.replay: no such file or directory
> > 
> > 
> > I think that's a merge issue, it's a new test I just added.
> 
> Ugh... and this new test now fails when recording :-) Will investigate why

Aaaand it fails when recording in a clean client too

eliben:
> > > > setup.go:188: open testdata/TestPreconditions.replay: no such file or directory
> > > 
> > > 
> > > I think that's a merge issue, it's a new test I just added.
> > 
> > 
> > Ugh... and this new test now fails when recording :-) Will investigate why
> 
> Aaaand it fails when recording in a clean client too

Merged @vangent 's fix in #1933. Travis rerunning tests now

Issue #732, opened 2018-11-16
pubsub: write a Kafka driver


jba:
As per @spf13.

jba:
I'm going to remove this from Sprint 20 because the vanity import path has higher priority.

garukun:
Not sure if this is right issue to start a discussion, the original [design on pubsub](https://github.com/google/go-cloud/blob/master/internal/docs/pubsub/design.md) mentions Confluent Kafka only. @jba, @zombiezen, I'm wondering if there are plans to support a "homebrew" Kafka deployment? Also, are there plans to integrate the Shopify/sarama (more defacto library in the Kafka world) to this package.

zombiezen:
@ijt, what's the current state here?

jba:
We haven't started exploring Kafka yet. We certainly intend that our integration will work with any Kafka deployment, not just hosted ones like Confluent. That's what we've done for RabbitMQ.

When we dig deeper, we'll evaluate whether Shopify/sarama is the right package for us to use.

ijt:
+1 to what @jba wrote.

jasongerard:
I'm doing a lot of work in go with kafka right now. I'd love to lend a hand to this.

jba:
@jasongerard That would be great! You should start by browsing around the pubsub tree. The Rabbit driver will probably be the closest match, in that we'll want it to be tested against a local Kafka instance rather than a cloud service. 

Our current plan is to have contributed drivers live outside this repo, but we are considering giving them `gocloud.dev` import paths. So in your case, you'd create a repo that contained your driver package at top level, and we would make "gocloud.dev/pubsub/kafkapubsub" point to it.

Although your driver won't live in this repo, we'd still like to help you as much as possible, and we would like to review the code.

jasongerard:
@jba sounds good. I'll start digging through the other implementations and hit back with questions as needed.

vangent:
Hi @jasongerard have you started this? If not, I'm going to take a look.

vangent:
I've started on this.

h7kanna:
What go library is being considered for this ? Sarama or https://github.com/segmentio/kafka-go ?

vangent:
I started with `segmentio/kafka-go` because the API seemed nicer, but I think it's one level of abstraction too high for us; in a way, it's trying to do something similar to what we're doing, building a nice Go API on top of the provider API. I also had issues making it performant. So, I'm currently using Sarama.

[...]

PR #1926, opened 2019-04-26
pubsub/kafkapubsub: don't do any queuing of messages from Kafka claims

vangent:
Updates #732.


pubsub/kafkapubsub/kafka.go
```diff
@@ -407,75 +407,109 @@ func (s *subscription) Cleanup(sarama.ConsumerGroupSession) error {
...
+		})
+		// A case per claim.
+		for _, claim := range s.claims {
+			cases = append(cases, reflect.SelectCase{
+				Dir:  reflect.SelectRecv,
+				Chan: reflect.ValueOf(claim.Messages()),
+			})
+		}
+		s.mu.Unlock()
+		i, v, ok := reflect.Select(cases)
```
jba: `reflect.Select` is rarely used, so sprinkle a few comments in.

pubsub/kafkapubsub/kafka.go
```diff
@@ -407,75 +407,109 @@ func (s *subscription) Cleanup(sarama.ConsumerGroupSession) error {
...
+		if !ok {
```
jba: For instance, on this line:
```
// i'th channel was closed
```
or something like that.

pubsub/kafkapubsub/kafka.go
```diff
@@ -407,75 +407,109 @@ func (s *subscription) Cleanup(sarama.ConsumerGroupSession) error {
...
-			dms = append(dms, dm)
-
-			s.mu.Lock()
-			s.unacked = append(s.unacked, ack)
-			s.mu.Unlock()
+			// Otherwise, if one of the claim channels closed, we're probably ending
+			// a session. Just keep trying.
+			continue
+		}
+		if v.IsNil() {
```
jba: These next two cases should be handled somehow. If sarama's doc says or implies the channel values will never be nil, then this is a bug in their package, and you should either panic or return an Internal error.

pubsub/kafkapubsub/kafka.go
```diff
@@ -407,75 +407,109 @@ func (s *subscription) Cleanup(sarama.ConsumerGroupSession) error {
...
+			// Don't think this should happen, but avoid panicking.
+			continue
+		}
+		msg, ok := v.Interface().(*sarama.ConsumerMessage)
```
jba: Presumably this is the type of the channel, so this case is actually impossible (an artifact of your use of reflection) and definitely warrants a panic.

pubsub/kafkapubsub/kafka.go
```diff
@@ -407,75 +407,109 @@ func (s *subscription) Cleanup(sarama.ConsumerGroupSession) error {
...
-				AckID:    ack,
-				AsFunc: func(i interface{}) bool {
-					// TODO(rvangent): Implement.
-					return false
-				},
+		// We'll give up after maxWaitCtx is Done, or if s.closeCh is closed.
+		// Otherwise, we want to pull a message from one of the channels in the
+		// claim(s) we've been given.
+		//
+		// Note: we could multiplex this by ranging over each claim.Messages(),
```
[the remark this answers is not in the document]
jba: Actually, one objection: this is a lot of work for one message. I guess your benchmarks will tell us how much work, but if it is significant, we should consider trying to get more than one message at a time.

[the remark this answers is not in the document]
vangent: It is definitely slower, but the benchmark is still getting > 10k msgs/sec, so it should be OK.

If we get more than one message at a time, then we are back in the place where we are buffering messages and likely dropping them; I'd prefer not to do that unless we need to for performance (10k/sec seems sufficient to me, but we can revisit if needed).

You are right that the Sarama library may be doing its own buffering, but I'm not sure how exactly they handle that (e.g., they may be doing something fancy to make sure those messages are redelivered), and I'd rather not add our own buffering on top of that.

pubsub/kafkapubsub/kafka.go
```diff
@@ -407,75 +407,109 @@ func (s *subscription) Cleanup(sarama.ConsumerGroupSession) error {
...
-			// that's OK.
-			// TODO(rvangent): See if we can avoid that.
-			// TODO(rvangent): Also consider setting ReceiveBatch to maxMessages=1
-			//                 and maxHandlers=1 similar to NATS.
-		}
-	}
+	s.mu.Lock()
+	s.claims = append(s.claims, claim)
+	s.mu.Unlock()
+	<-sess.Context().Done()
```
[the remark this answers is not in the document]
vangent: Yes. ConsumeClaim is supposed to `range` over the `claim.Messages` channel, which is closed when the session is closed.

pubsub/kafkapubsub/kafka.go
vangent: Can you suggest some? I did sprinkle a few comments in already, and if they're not helping it would help me to know what would help. I agree though, this was new to me.

pubsub/kafkapubsub/kafka.go
vangent: Done.

pubsub/kafkapubsub/kafka.go
vangent: I tried adding some.

pubsub/kafkapubsub/kafka.go
vangent: OK, I removed these two cases, will just let it panic. I think it should be fine :-).

pubsub/kafkapubsub/kafka.go
vangent: Done.

vangent:
PTAL.

PR #1928, opened 2019-04-26
internal/docstore/firedocstore: add as support for queries

shantuo:
Updates #1675


internal/docstore/firedocstore/fs_test.go
```diff
@@ -106,6 +107,30 @@ func (c *codecTester) DocstoreDecode(value, dest interface{}) error {
...
+func (verifyAs) Name() string {
+	return "verify As"
+}
+
+func (verifyAs) BeforeQuery(as func(i interface{}) bool) error {
+	var req *pb.RunQueryRequest
+	if !as(&req) {
+		return errors.New("Query.BeforeQuery failed")
+	}
+	_ = req.GetStructuredQuery()
```
jba: What's the purpose of this line?

shantuo: I was thinking the `StructuredQuery` would be something the user wants to change and this is just to show it gets the underlying query struct.

PR #1935, opened 2019-04-26
docstore: wrap query errors

jba:
Wrap errors returned from query RPCs so that they have error codes from gcerrors.


internal/docstore/drivertest/drivertest.go
```diff
@@ -976,6 +984,16 @@ func testMultipleActions(t *testing.T, coll *ds.Collection) {
 		{KeyField: "testMultipleActions11", "s": "k"},
 		{KeyField: "testMultipleActions12", "s": "l"},
 	}
+
+	actions := coll.Actions()
```
shantuo: Let's do the `forEach` + `Delete` way inside `defer` to clean up after the test docs like in the query and as tests.

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -540,4 +540,5 @@ var errorCodeMap = map[string]gcerrors.ErrorCode{
 	dyn.ErrCodeTransactionCanceledException:             gcerr.FailedPrecondition,
 	dyn.ErrCodeTransactionInProgressException:           gcerr.InvalidArgument,
 	dyn.ErrCodeIdempotentParameterMismatchException:     gcerr.InvalidArgument,
+	"ValidationException":                               gcerr.InvalidArgument,
```
shantuo: Weird they don't expose the type for dynamo and this seems a pretty common error.

internal/docstore/dynamodocstore/dynamo_test.go
```diff
@@ -99,12 +100,37 @@ func (verifyAs) QueryCheck(it *docstore.DocumentIterator) error {
 func TestConformance(t *testing.T) {
 	// Note: when running -record repeatedly in a short time period, change the argument
 	// in the call below to generate unique transaction tokens.
-	drivertest.MakeUniqueStringDeterministicForTesting(2)
+	drivertest.MakeUniqueStringDeterministicForTesting(4)
```
shantuo: Revert this change so that you don't need to include the unnecessary record files below as the tokens will be the same.

internal/docstore/query.go
```diff
@@ -134,7 +135,8 @@ func (q *Query) init(fps []FieldPath) error {
 // Always call Stop on the iterator.
 type DocumentIterator struct {
 	iter driver.DocumentIterator
-	err  error
+	coll driver.Collection
```
shantuo: I think we can add `ErrorCode` method to `driver.DocumentIterator`, or create a `driver.ErrorCoder`, then both the  iterator and the collection would be able to directly call the underlying `errorCode` method and iterator doesn't need to include the collection while it doesn't use any information of it.

internal/docstore/drivertest/drivertest.go

internal/docstore/dynamodocstore/dynamo.go
jba: Agreed. I grepped the package tree and didn't see it.

internal/docstore/query.go
jba: I think I did what you're suggesting. Please confirm.

internal/docstore/dynamodocstore/dynamo_test.go
jba: Done.

jba:
PTAL.

PR #1936, opened 2019-04-26
internal/website/content/howto/pubsub: use godoc examples

### LINKED ISSUE #1827: internal/website: how-to example code can drift from godoc examples

zombiezen:
Required some restructuring to existing Pub/Sub examples, but should make them more closely associated with the API they are exemplifying.

Also mark examples.json and replay YAML files as generated, see https://thoughtbot.com/blog/github-diff-supression.

Fixes #1827


.gitattributes
```diff
@@ -0,0 +1,2 @@
+/internal/website/data/examples.json  linguist-generated=true
+**/testdata/**/*.yaml  linguist-generated=true
```
vangent: Please add `**/testdata/**/*.replay` as well. Some of them are binary, so are probably collapsed by default, but others are text (recorded HTTP requests/responses).

internal/website/content/howto/pubsub/subscribe.md
```diff
@@ -555,19 +225,7 @@ You will also need to pass an acknowledgement deadline: once a message is
...
-    "gocloud.dev/pubsub/mempubsub"
-)
-
-// ...
-
-topic := mempubsub.NewTopic()
-defer topic.Shutdown(ctx)
-subscription := mempubsub.NewSubscription(topic, 1 * time.Minute)
-```
+{{< goexample "gocloud.dev/pubsub/mempubsub.ExampleNewSubscription" >}}
```
vangent: Question: if you typo this and `ExampleNewSubscription` does not exist, will Travis catch that?

.gitattributes
zombiezen: Done.

internal/website/content/howto/pubsub/subscribe.md
zombiezen: Good point. Added logic to the shortcode to make it do exactly that.

PR #1943, opened 2019-04-29
mongodocstore: provide an option to lowercase field names

### LINKED ISSUE #1899: mongodocstore: fields encode as lower-case
The default encoder for the MongoDB driver lower-cases struct fields before encoding them. This is incompatible with the encoders of the other providers.

We should override the default encoder, which shouldn't be hard to do: we need to construct our own [RegistryBuilder](https://godoc.org/go.mongodb.org/mongo-driver/bson/bsoncodec#RegistryBuilder).

jba:
Add an option to OpenCollection that causes field names to be lowercased
on encoding, and also when mentioned in actions and queries.

See #1899 for rationale.

Fixes #1899.

### DIFF
--- a/internal/docstore/mongodocstore/codec.go
+++ b/internal/docstore/mongodocstore/codec.go
@@ -31,8 +31,8 @@ import (
 // This code is copied from memdocstore/codec.go, with some changes:
 // - special treatment for primitive.Binary
 
-func encodeDoc(doc driver.Document) (map[string]interface{}, error) {
-	var e encoder
+func encodeDoc(doc driver.Document, lowercaseFields bool) (map[string]interface{}, error) {
+	e := encoder{lowercaseFields: lowercaseFields}
 	if err := doc.Encode(&e); err != nil {
 		return nil, err
 	}
@@ -48,7 +48,8 @@ func encodeValue(x interface{}) (interface{}, error) {
 }
 
 type encoder struct {
-	val interface{}
+	val             interface{}
+	lowercaseFields bool
 }
 
 func (e *encoder) EncodeNil()                 { e.val = nil }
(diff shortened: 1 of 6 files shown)


internal/docstore/mongodocstore/codec.go
```diff
@@ -31,8 +31,8 @@ import (
 // This code is copied from memdocstore/codec.go, with some changes:
 // - special treatment for primitive.Binary
 
-func encodeDoc(doc driver.Document) (map[string]interface{}, error) {
-	var e encoder
+func encodeDoc(doc driver.Document, lowercaseFields bool) (map[string]interface{}, error) {
```
shantuo: Consider make the second argument a `encoderOptions` which includes the `lowercaseFields` just like opening a collection.

internal/docstore/mongodocstore/codec.go
```diff
@@ -93,15 +94,14 @@ type mapEncoder struct {
 	encoder
 }
 
-func (e *encoder) EncodeMap(n int, isStruct bool) driver.Encoder {
+func (e *encoder) EncodeMap(n int, _ bool) driver.Encoder {
```
shantuo: Can the second argument be removed then?

internal/docstore/mongodocstore/codec.go
jba: Can we do that if we ever need more than one or two of these? 

We do it for opening a collection to preserve backwards-compatibility. Here we don't have to worry about that, so I'd rather keep things as simple as possible.

internal/docstore/mongodocstore/codec.go
jba: Yes, I added #1969. (I didn't do it in this PR. I have many PRs out and I want to avoid merge conflicts as much as possible.)

jba:
PTAL.

internal/docstore/mongodocstore/codec.go
shantuo: Yeah that makes sense.

Issue #1918, opened 2019-04-25
internal/gocdk: permit init in existing directory


zombiezen:
### Is your feature request related to a problem? Please describe.

As a user of the `gocdk` CLI, I may want to run `gocdk init` on an existing directory, perhaps because I created the directory through my IDE or I want to start using the CLI for my local development or deployment.

### Describe a solution you'd like

(Treat this as a possible solution in the space, not as a final design.)

When presented with an existing directory, `gocdk init` would create biomes, a Go module, and/or a Dockerfile depending on what is missing from the directory.

zombiezen:
One thing that might be interesting to try is to run `gocdk biome add` in an existing Go project.

PR #1945, opened 2019-04-29
internal/cmd/gocdk: minimal init

### LINKED ISSUE #1806: internal/cmd/gocdk: implement minimal init command
It's useful to create projects with the biome structure for the purpose of manual testing. Regardless of the scaffolding code we want to use, it seems fairly important to have an init command sooner rather than later.

clausti:
fixes #1806


internal/cmd/gocdk/init.go
```diff
@@ -15,11 +15,101 @@
...
+	f.StringVar(&modpath, "m", "", "alias for --module-path")
+
+	if err := f.Parse(args); xerrors.Is(err, flag.ErrHelp) {
+		return nil
+	} else if err != nil {
+		return usagef("gocdk init: %w", err)
+	}
+
+	if f.NArg() != 1 {
+		return usagef("gocdk init [options] PATH")
```
vangent: Can you clarify for me (and the user) what the difference between `PATH` here and `module-path` is? I'll probably figure it out as I read the code below but it's not clear from the help.

vangent: Did we consider dropping the `PATH` argument and having `gocdk init` initialize the current directory? I.e., you'd do:

```
$ mkdir myproject
$ cd myproject
$ gocdk init
```

internal/cmd/gocdk/init.go
```diff
@@ -15,11 +15,101 @@
...
+
+	tmplValues := struct {
+		ProjectName string
+		ModulePath  string
+	}{
+		ProjectName: filepath.Base(path),
+		ModulePath:  modpath,
+	}
+	for name, templateSource := range templates.InitTemplates {
+		tmpl := template.Must(template.New("").Parse(templateSource))
```
vangent: Should the arg to `New` here be `name`?

internal/cmd/gocdk/internal/templates/init.go
```diff
@@ -0,0 +1,81 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package templates contains string literals for use by the gocdk
+// command when creating user project files.
+package templates
+
+var InitTemplates = map[string]string{
```
vangent: IMHO we should modify this to use the `go generate` / `static.go` approach from my other PR, as it is more flexible. Can you add a TODO and/or file an issue?

internal/cmd/gocdk/init.go
zombiezen: Yes: #1918 tracks this sort of usage.

clausti: `module-path` is the import path for the module we'll be creating, and `path` is the path to the directory we will be creating to put the project in. 

Do you think it would be clearer if `module-path` was named `import-path`?

vangent: Where is it going to be used? In the `.mod` file, or ...?

vangent: I meant instead of this, not in addition...

clausti: We talked about that, but then there was a fair amount of back and forth about if it needed to be in an empty directory and then what constituted "empty", and the decision was made for simplicity sake to create a new directory, at least at first.

internal/cmd/gocdk/init.go
```diff
@@ -15,11 +15,101 @@
...
+// import path for the project.
+func inferModulePath(ctx context.Context, pctx *processContext, path string) (string, error) {
+	cmd := exec.CommandContext(ctx, "go", "env", "GOPATH")
+	cmd.Dir = pctx.workdir
+	cmd.Env = pctx.env
+	gopath, err := cmd.Output()
+	if err != nil {
+		return "", xerrors.Errorf("infer module path: %w", err)
+	}
+	//check if path is relative to gopath
```
vangent: Nit: comments should be full sentences.

internal/cmd/gocdk/init.go
vangent: OK.

vangent: I get it now, it's the module path for the go.mod file. Maybe just update the help to indicate what the default is. Although not sure it will be possible to explain that succinctly.

Overall this is fine for now but I'm concerned that users will have a hard time understanding it.

internal/cmd/gocdk/init.go
```diff
@@ -15,11 +15,101 @@
...
+		if err := ioutil.WriteFile(fullPath, buf.Bytes(), 0666); err != nil {
+			return xerrors.Errorf("gocdk init: %w", err)
+		}
+	}
+	return nil
+}
+
+// inferModulePath will check the default GOPATH to attempt to infer the module
+// import path for the project.
+func inferModulePath(ctx context.Context, pctx *processContext, path string) (string, error) {
```
vangent: I'm having trouble understanding how this works. Maybe it's not important. Tests might provide some examples to make it more clear (maybe factor out the "infer" part from the "exec go" part).

For now, maybe just add a TODO to add tests for this.

internal/cmd/gocdk/internal/templates/init.go
clausti: https://github.com/google/go-cloud/issues/2015

internal/cmd/gocdk/init.go
clausti: https://github.com/google/go-cloud/issues/2016

clausti:
@vangent, PTAL?

internal/cmd/gocdk/internal/templates/init.go
```diff
@@ -0,0 +1,81 @@
+// Copyright 2019 The Go Cloud Authors
```
vangent: This should be "The Go Cloud Development Kit Authors". Please fix anywhere you are copy/pasting this from, it keeps coming back like a zombie.

internal/cmd/gocdk/internal/templates/init.go
vangent: Travis wants you to add a declaration for this new package. Since we're going to delete it soon as part of adding them to `_static`, I suggest just deleting this file and moving its contents into `init.go`.

Issue #626, opened 2018-11-06
blob/tests: add some concurrency tests


vangent:
Add some conformance tests that use the `blob` API concurrently and make sure things work.

eliben:
The record/replay infrastructure we currently use for blob testing (`go-vcr`) doesn't deal well with non-determinism, and concurrent tests will produce non-deterministic orders of requests/responses.

We'll be looking into using `httpreplay` for blob conformance tests

eliben:
@vangent would you like to see additional concurrent tests beyond #1952?

vangent:
Nope, LGTM!

PR #1952, opened 2019-04-29
blob: add concurrent blob write/read testing

eliben:
Updates #626


blob/drivertest/drivertest.go
```diff
@@ -1821,6 +1825,75 @@ func testDelete(t *testing.T, newHarness HarnessMaker) {
 	})
 }
 
+// testConcurrentWriteAndRead tests that concurrent writing and reading of
```
vangent: Nit:

```
// testConcurrentWriteAndRead tests that concurrent writing to multiple blob keys and concurrent reading from multiple blob keys works.
```

blob/drivertest/drivertest.go
```diff
@@ -1821,6 +1825,75 @@ func testDelete(t *testing.T, newHarness HarnessMaker) {
...
+	}
+	defer h.Close()
+	drv, err := h.MakeDriver(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	b := blob.NewBucket(drv)
+	defer b.Close()
+
+	// Prepare data. Each blob dataSize bytes equal to its numeric key.
```
vangent: Nit, something like:

```
// Prepare data. Each of the numKeys blobs has dataSize bytes, with each byte set to the
// numeric key index. For example, the blob at "key0" consists of all dataSize bytes set to 0.

File: blob/drivertest/drivertest.go [hunk 3]
```diff
@@ -1821,6 +1825,75 @@ func testDelete(t *testing.T, newHarness HarnessMaker) {
...
+		wg.Add(1)
+		go func(key int) {
+			buf, err := b.ReadAll(ctx, blobName(key))
+			if err != nil {
+				t.Fatal(err)
+			}
+			if len(buf) != dataSize {
+				t.Errorf("data size for key %d, got %d want %d", key, len(buf), dataSize)
+			}
+			for i := 0; i < len(buf); i++ {
```
vangent: Could also use `bytes.Equal(buf, keyData[k])`.

blob/drivertest/drivertest.go
```diff
@@ -1821,6 +1825,75 @@ func testDelete(t *testing.T, newHarness HarnessMaker) {
...
+
+	// Write all blobs concurrently.
+	for k := 0; k < numKeys; k++ {
+		wg.Add(1)
+		go func(key int) {
+			if err := b.WriteAll(ctx, blobName(key), keyData[key], nil); err != nil {
+				t.Fatal(err)
+			}
+			wg.Done()
+		}(k)
```
vangent: Add a `defer b.Delete(ctx, blobName(key))` here to clean up the blobs....

blob/drivertest/drivertest.go
eliben: Done

blob/drivertest/drivertest.go
eliben: Done

blob/drivertest/drivertest.go
eliben: Yes, but it would make error reporting much harder. Emitting huge byte slices isn't too helpful, whereas here you get a helpful error message. It actually saved my bacon once when I got the sync wrong and one key got data from another, was very quick to see what the problem is. Seems worth it for the cost of 4 lines of trivial code?

blob/drivertest/drivertest.go
eliben: Done

eliben:
Thanks for the quick review! PTAL

blob/drivertest/drivertest.go
vangent: Up to you, but IMHO the error could just emit the key number, not the index. It seems unlikely that some of the bytes are going to be right and others wrong.

eliben: OK, I checked and we use `bytes.Equal` a bunch in tests, so will align to that. Maybe we should write some of our own testing utils to makes this nicer (or wait for generics)

PR #1955, opened 2019-04-30
firedocstore: unordered action lists

jba:
Implement unordered action lists in a simple way, using the same grouping
as for ordered lists. The difference is that we run the groups concurrently.

TODO: add conformance tests.


internal/docstore/firedocstore/fs.go
```diff
@@ -487,6 +506,38 @@ func (c *collection) updateWrites(doc driver.Document, docName string, mods []dr
...
+			if g[0].Kind == driver.Get {
+				errs := c.runGets(ctx, g)
+				for i, err := range errs {
+					errs[base+i] = err
+				}
+			} else {
+				err := c.runWrites(ctx, g)
+				// Writes run in a transaction, so there is a single error for the group.
+				for i := 0; i < len(g); i++ {
+					errs[base+i] = err
```
shantuo: Why do we need to repeat the same errors and return them? It doesn't seem to help the user this way, instead, we could just return all different ones.

internal/docstore/firedocstore/fs.go
```diff
@@ -487,6 +506,38 @@ func (c *collection) updateWrites(doc driver.Document, docName string, mods []dr
 	return []*pb.Write{w}, nil
 }
 
+func (c *collection) runActionsUnordered(ctx context.Context, actions []*driver.Action) driver.ActionListError {
+	// Split into groups the same way, but run them concurrently.
+	// TODO(jba): group without considering order.
+	groups := driver.SplitActions(actions, shouldSplit)
+	errs := make([]error, len(actions))
```
shantuo: `errs` doesn't need to be the length of the actions, see comment below.

internal/docstore/firedocstore/fs.go
```diff
@@ -255,55 +259,70 @@ func shouldSplit(cur, new *driver.Action) bool {
...
+	errs := c.runGets(ctx, gets)
+	for i, err := range errs {
+		if err != nil {
+			return i, err
+		}
+	}
+	return len(gets), nil
+}
+
+// runGets executs a group of Get actions by calling the BatchGetDocuments RPC.
```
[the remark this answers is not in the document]

internal/docstore/firedocstore/fs.go
jba: Unordered lists have to return an error for each failed action. In this case, all the actions in the write failed. We have to a return an error for all of them.

How would we make the errors different? The only thing that's different is the action itself, but the index lets the user look up the action, so we are already encoding that information.

internal/docstore/firedocstore/fs.go
jba: This is just a convenience. Only the non-nil errors are returned, see `driver.driver.NewActionListError`.

jba:
PTAL.

PR #1957, opened 2019-04-30
pubsub/rabbitpubsub: remove unnecessary and flaky tests

### LINKED ISSUE #1950: pubsub/rabbit: TestCancelSendAndReceive seems to be flaky
Example where it failed on Travis: https://travis-ci.com/google/go-cloud/jobs/196471448

@vangent says it failed once when running with `-count 100`

vangent:
`TestCancelSendAndReceive` is flaky, and heavily overlaps with a conformance test of the same name.

`TestPublishConcurrently` overlaps with at least a conformance benchmark, and current configuration means `SendBatch` isn't called concurrently anyway.

Fixes #1950.

### DIFF
--- a/pubsub/rabbitpubsub/rabbit_test.go
+++ b/pubsub/rabbitpubsub/rabbit_test.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


pubsub/rabbitpubsub/rabbit_test.go
```diff
@@ -149,48 +148,6 @@ func (h *harness) Close() {
 
 func (h *harness) MaxBatchSizes() (int, int) { return 0, 0 }
 
-// This test is important for the RabbitMQ driver because the underlying client is
```
vangent: I don't this test is needed anymore; there are conformance tests and benchmarks that do concurrent sends. We also configure the `Send` batcher with `MaxHandlers=1` so that `SendBatch` isn't called concurrently anyway.

pubsub/rabbitpubsub/rabbit_test.go
```diff
@@ -207,8 +164,8 @@ func TestUnroutable(t *testing.T) {
 		{Body: []byte("")},
 	}
 	err := topic.SendBatch(ctx, msgs)
-	merr, ok := err.(MultiError)
-	if !ok {
+	var merr MultiError
```
vangent: Might as well use `As` here....

PR #1960, opened 2019-04-30
secrets/localsecrets: error if secret material not 32 bytes

### LINKED ISSUE #1664: secrets/localsecrets: improve key input for URLs
/cc @vangent @clausti 

The current implementation of `localsecrets.URLOpener.OpenKeeperURL` will use `Base64Key` or 
`ByteKey`:

https://github.com/google/go-cloud/blob/6482f3715c170669214c5cf4ba8421c51d772d30/secrets/localsecrets/localsecrets.go#L72-L85

If the user provides secret material of exactly the correct size, then this URL opener works exactly as intended. However, if the secret material is of an incorrect size, then the URL opener will not return an error (as I would expect), but instead silently proceed doing one of two things:

1.  If the decoded argument is longer than 32 bytes, then the bytes after 32 are silently truncated.
2.  If the decoded argument is shorter than 32 bytes, then the value is extended to 32 bytes by adding NUL bytes to the end of the input.

The further issue with the `stringkey` scheme is that it will use the bytes as-is, which will likely be ASCII-clean and thus reduce the entropy of the key by a factor of 2^8 (1 bit per character).

From a usability standpoint, the danger here is that someone will see "secret material" and assume they need to create a "password", which will not be cryptographically strong key material. Since they will not see an error on the wrong length of key material. Quoting the paper [Cryptographic Extraction and Key Derivation: The HKDF Scheme](https://eprint.iacr.org/2010/264):

> We associate the notion of “cryptographically strong” keys with that of pseudorandom keys, namely, indistinguishable by feasible computation from a random uniform string of the same length.

(While this is strictly speaking an enhancement, the implications of improper usage can introduce security issues for the user of `localsecrets` so I am classifying this as a bug.)

I believe the URL opener should only permit base-64 encoded input of the correct size and return errors for anything that doesn't exactly match. Even if the user interprets such an error as requiring a password of 32 characters, that's strictly more entropy than a zero-padded password.

I'd also like to reconsider the `ByteKey` and `Base64Key` functions as they exist now. The silent truncation/zero-extension behavior has the problems I list above and I think they don't provide the right set of anti-affordances to encourage secure usage. At the very least, they should validate the length and return an error if it does not match. An even more accessible design could be to use a key-derivation function to build a cryptographically strong key from a weak password.

(There may be other solutions than the ones I've thought of here, but I'll list the first few ideas that I considered while reporting this bug.)

Using a key-derivation function in all cases could make insecure usage harder, but at the cost of introducing a performance hit for users that correctly generated a 32-byte random key.

One could imagine addressing this problem only through documentation and extolling the virtues of generating keys correctly, but I feel this likely wouldn't scale: it relies on the user reading the documentation. Such a documentation-only approach also doesn't give the user feedback when they're doing something that is likely a security concern when the library could detect it.

### Additional context

The first few paragraphs in the Introduction section of [Cryptographic Extraction and Key Derivation: The HKDF Scheme](https://eprint.iacr.org/2010/264) provide a fairly decent academic introduction to Key-Derivation Functions and the problems around passwords and key material.

The Wikipedia articles on [Key stretching](https://en.wikipedia.org/wiki/Key_stretching) and [Key Derivation Function](https://en.wikipedia.org/wiki/Key_derivation_function) fill in a few more details, but don't give a very good overview of the problem.

clausti:
This PR updates the `localsecrets` methods `ByteKey` and `Base64Key` to error if not given 32 bytes of secret material (measured after decoding, in the case of `Base64Key`).

It also adds a convenience method for generating 32 random bytes to use. 

fixes #1664

### DIFF
--- a/secrets/localsecrets/localsecrets.go
+++ b/secrets/localsecrets/localsecrets.go
@@ -81,7 +81,11 @@ func (o *URLOpener) OpenKeeperURL(ctx context.Context, u *url.URL) (*secrets.Kee
 		}
 		return NewKeeper(sk), nil
 	}
-	return NewKeeper(ByteKey(u.Host)), nil
+	sk, err := ByteKey(u.Host)
+	if err != nil {
+		return nil, err
+	}
+	return NewKeeper(sk), nil
 }
 
 // keeper holds a secret for use in symmetric encryption,
@@ -106,16 +110,39 @@ func Base64Key(base64str string) ([32]byte, error) {
 	if err != nil {
 		return sk32, err
 	}
+	keySize := len([]byte(key))
+	if keySize != 32 {
+		return sk32, fmt.Errorf("Base64Key: secret material is %v bytes, want 32 bytes", keySize)
+	}
 	copy(sk32[:], key)
 	return sk32, nil
 }
 
 // ByteKey takes a secret key as a string and converts it
-// to a [32]byte, cropping it if necessary.
-func ByteKey(sk string) [32]byte {
+// to a [32]byte, erroring if the string is not 32 bytes of data.
+func ByteKey(sk string) ([32]byte, error) {
 	var sk32 [32]byte
-	copy(sk32[:], []byte(sk))
-	return sk32
+	skb := []byte(sk)
+	keySize := len(skb)
+	if keySize != 32 {
+		return sk32, fmt.Errorf("ByteKey: secret material is %v bytes, want 32 bytes", keySize)
+	}
+	copy(sk32[:], skb)
+	return sk32, nil
+}
+
+// ThirtyTwoByteSecret will generate random secret material suitable to be
+// used as the secret key argument to NewKeeper.
+func ThirtyTwoByteSecret() ([32]byte, error) {
+	var sk32 [32]byte
+	randomBytes := make([]byte, 32)
+	// Read random numbers into the passed slice until it's full
+	_, err := rand.Read(randomBytes)
+	if err != nil {
+		return sk32, err
+	}
+	copy(sk32[:], randomBytes)
+	return sk32, nil
 }
 
 const nonceSize = 24
--- a/secrets/localsecrets/localsecrets.go
+++ b/secrets/localsecrets/localsecrets.go
@@ -103,7 +103,7 @@ func NewKeeper(sk [32]byte) *secrets.Keeper {
 }
 
 // Base64Key takes a secret key as a base64 string and converts it
-// to a [32]byte, cropping it if necessary.
+// to a [32]byte, erroring if the decoded data is not 32 bytes.
 func Base64Key(base64str string) ([32]byte, error) {
 	var sk32 [32]byte
 	key, err := base64.StdEncoding.DecodeString(base64str)
@@ -135,13 +135,11 @@ func ByteKey(sk string) ([32]byte, error) {
 // used as the secret key argument to NewKeeper.
 func ThirtyTwoByteSecret() ([32]byte, error) {
 	var sk32 [32]byte
-	randomBytes := make([]byte, 32)
-	// Read random numbers into the passed slice until it's full
-	_, err := rand.Read(randomBytes)
+	// Read random numbers into the passed slice until it's full.
+	_, err := rand.Read(sk32[:])
 	if err != nil {
 		return sk32, err
 	}
-	copy(sk32[:], randomBytes)
 	return sk32, nil
 }
 
--- a/secrets/localsecrets/localsecrets.go
+++ b/secrets/localsecrets/localsecrets.go
@@ -112,7 +112,7 @@ func Base64Key(base64str string) ([32]byte, error) {
 	}
 	keySize := len([]byte(key))
 	if keySize != 32 {
-		return sk32, fmt.Errorf("Base64Key: secret material is %v bytes, want 32 bytes", keySize)
+		return sk32, fmt.Errorf("Base64Key: secret key material is %v bytes, want 32 bytes", keySize)
 	}
 	copy(sk32[:], key)
 	return sk32, nil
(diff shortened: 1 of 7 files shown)


secrets/localsecrets/localsecrets.go
```diff
@@ -106,16 +110,39 @@ func Base64Key(base64str string) ([32]byte, error) {
 	if err != nil {
 		return sk32, err
 	}
+	keySize := len([]byte(key))
```
vangent: I think the docstring for this function is stale; it no longer crops.

secrets/localsecrets/localsecrets.go
```diff
@@ -106,16 +110,39 @@ func Base64Key(base64str string) ([32]byte, error) {
...
+	}
 	copy(sk32[:], key)
 	return sk32, nil
 }
 
 // ByteKey takes a secret key as a string and converts it
-// to a [32]byte, cropping it if necessary.
-func ByteKey(sk string) [32]byte {
+// to a [32]byte, erroring if the string is not 32 bytes of data.
+func ByteKey(sk string) ([32]byte, error) {
```
vangent: IMHO we should drop this function (and the `stringkey` scheme) entirely. Let's make people use the `[]byte` form (or base64 of `[]byte`).

secrets/localsecrets/localsecrets.go
```diff
@@ -106,16 +110,39 @@ func Base64Key(base64str string) ([32]byte, error) {
...
+	if keySize != 32 {
+		return sk32, fmt.Errorf("ByteKey: secret material is %v bytes, want 32 bytes", keySize)
+	}
+	copy(sk32[:], skb)
+	return sk32, nil
+}
+
+// ThirtyTwoByteSecret will generate random secret material suitable to be
+// used as the secret key argument to NewKeeper.
+func ThirtyTwoByteSecret() ([32]byte, error) {
```
vangent: Elsewhere we call it a "Key" (e.g., `Base64Key`), so maybe just `NewKey`?

secrets/localsecrets/localsecrets.go
```diff
@@ -106,16 +110,39 @@ func Base64Key(base64str string) ([32]byte, error) {
...
 	var sk32 [32]byte
-	copy(sk32[:], []byte(sk))
-	return sk32
+	randomBytes := make([]byte, 32)
+	// Read random numbers into the passed slice until it's full
```
vangent: Nit: trailing ".".

secrets/localsecrets/localsecrets.go
```diff
@@ -106,16 +110,39 @@ func Base64Key(base64str string) ([32]byte, error) {
...
+	_, err := rand.Read(randomBytes)
+	if err != nil {
+		return sk32, err
+	}
+	copy(sk32[:], randomBytes)
```
vangent: Do we need to copy? I think you can just pass `sk32[:]` directly to `rand.Read` and avoid `randomBytes`.

secrets/localsecrets/localsecrets.go
```diff
@@ -106,16 +110,39 @@ func Base64Key(base64str string) ([32]byte, error) {
...
+// used as the secret key argument to NewKeeper.
+func ThirtyTwoByteSecret() ([32]byte, error) {
 	var sk32 [32]byte
-	copy(sk32[:], []byte(sk))
-	return sk32
+	randomBytes := make([]byte, 32)
+	// Read random numbers into the passed slice until it's full
+	_, err := rand.Read(randomBytes)
+	if err != nil {
+		return sk32, err
```
vangent: https://golang.org/pkg/math/rand/#Rand.Read says it always returns a `nil` error, so I think you can ignore the error here (or panic), and drop the `error` return value from this function.

vangent:
Since this PR is a breaking change, you'll need to add a commit with `BREAKING_CHANGE_OK` in the commit message.

From Travis:
```
Found breaking API change(s) in: gocloud.dev/secrets/localsecrets.
FAIL. If this is expected and OK, you can pass this check by adding a commit with BREAKING_CHANGE_OK in the first line of the message.
```

secrets/localsecrets/localsecrets.go
clausti: ah, this is from [crypto/rand](https://godoc.org/crypto/rand#Read), rather than `math/rand`, which seems to indicate it sometimes returns a real error.

secrets/localsecrets/localsecrets.go
clausti: Yeah, we get a type error if we try and pass `sk32` directly, because `Read` takes `[]bye`, not [32]byte`

secrets/localsecrets/localsecrets.go
clausti: Getting rid of the `stringkey` scheme entirely seems like it might warrant its own issue/PR, WDYT?

secrets/localsecrets/localsecrets.go
clausti: slightly related: in the error messages, I've made it say "secret material [...]", should it instead say "key material"?

secrets/localsecrets/localsecrets.go
vangent: If you prefer, but it seems like part and parcel of this change to me (i.e., you'd need to update/remove some of the same `stringkey` examples you're modifying in this PR).

secrets/localsecrets/localsecrets.go
vangent: Or "secret key material"? I don't feel strongly, but we should try to be consistent.

I don't think `ThirtyTwoByte` is needed in the name for this. `NewRandomKey` or `NewRandomSecret` are fine with me.

secrets/localsecrets/localsecrets.go
vangent: Yup, you're right, never mind.

secrets/localsecrets/localsecrets.go
clausti: Opened https://github.com/google/go-cloud/issues/1980 for removing the `ByteKey` function and `stringkey` scheme.

PR #1961, opened 2019-04-30
internal/cmd/gocdk: listen for filesystem changes

### LINKED ISSUE #1881: internal/cmd/gocdk: serve should rebuild server when source changes

zombiezen:
Fixes #1881

### DIFF
--- a/internal/cmd/gocdk/fswatch.go
+++ b/internal/cmd/gocdk/fswatch.go
@@ -30,14 +30,14 @@ import (
 func waitForDirChange(ctx context.Context, path string, pollInterval time.Duration) error {
 	tick := time.NewTicker(pollInterval)
 	defer tick.Stop()
-	init, err := snapshotTree(path)
+	initSnapshot, err := snapshotTree(path)
 	if err != nil {
 		return xerrors.Errorf("wait for %s to change: %w", path, err)
 	}
 	for {
 		select {
 		case <-tick.C:
-			same, err := init.sameAs(path)
+			same, err := initSnapshot.sameAs(path)
 			if err != nil {
 				// TODO(light): Log error.
 				continue
@@ -103,6 +103,7 @@ func snapshotTree(path string) (*dirSnapshot, error) {
 			return nil, xerrors.Errorf("snapshot tree %s: %w", path, err)
 		}
 		snap := newDirSnapshot(contents)
+		// Save snapshot in requested variable.
 		*curr.dst = snap
 
 		// Push subdirectories onto stack in reverse order, so they are popped in
--- a/internal/cmd/gocdk/fswatch.go
+++ b/internal/cmd/gocdk/fswatch.go
@@ -16,17 +16,16 @@ package main
 
 import (
 	"context"
-	"io/ioutil"
 	"os"
 	"path/filepath"
-	"sort"
+	"strings"
 	"time"
 
 	"golang.org/x/xerrors"
 )
 
 // waitForDirChange blocks until the directory changes or the context's Done
-// channel is closed.
+// channel is closed. waitForDirChange returns nil only if it detected a change.
 func waitForDirChange(ctx context.Context, path string, pollInterval time.Duration) error {
 	tick := time.NewTicker(pollInterval)
 	defer tick.Stop()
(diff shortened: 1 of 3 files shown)


internal/cmd/gocdk/fswatch.go
```diff
@@ -0,0 +1,205 @@
...
+
+	"golang.org/x/xerrors"
+)
+
+// waitForDirChange blocks until the directory changes or the context's Done
+// channel is closed.
+func waitForDirChange(ctx context.Context, path string, pollInterval time.Duration) error {
+	tick := time.NewTicker(pollInterval)
+	defer tick.Stop()
+	init, err := snapshotTree(path)
```
vangent: Nit: consider calling this `snapshot` instead of `init`? Or `initSnapshot`?

internal/cmd/gocdk/fswatch.go
```diff
@@ -0,0 +1,205 @@
...
+	if err != nil {
+		return xerrors.Errorf("wait for %s to change: %w", path, err)
```
vangent: How does the user distinguish this error from "there was a change"?

internal/cmd/gocdk/fswatch.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	for {
+		select {
+		case <-tick.C:
+			same, err := init.sameAs(path)
```
vangent: WDYT about doing an explicit snapshot here and calling `recursiveEqual` (and dropping `sameAs`)?

internal/cmd/gocdk/fswatch.go
```diff
@@ -0,0 +1,205 @@
...
+	if ent == nil && ent2 == nil {
+		return true
+	}
+	if ent == nil || ent2 == nil {
+		return false
+	}
+	return ent.name == ent2.name && ent.mode == ent2.mode && ent.modTime.Equal(ent2.modTime)
+}
+
+// isIgnoredForWatch reports whether a given filename is ignored for the
```
vangent: Do we care about anything other than `.go` files?

internal/cmd/gocdk/fswatch.go
```diff
@@ -0,0 +1,205 @@
...
+	sort.Slice(snap.entries, func(i, j int) bool {
+		return snap.entries[i].name < snap.entries[j].name
+	})
+	return snap
+}
+
+// snapshotTree recursively snapshots a directory.
+func snapshotTree(path string) (*dirSnapshot, error) {
+	// Unfortunately filepath.Walk does not work well for this because
+	// we need to keep track of the parent snapshot variable to write to.
```
vangent: I did not understand this comment, can you clarify? Aren't we always appending to the same `dirSnapshot.entries`?

internal/cmd/gocdk/fswatch.go
```diff
@@ -0,0 +1,205 @@
...
+		curr := stack[len(stack)-1]
+		stack = stack[:len(stack)-1]
+
+		// Read directory and create snapshot.
+		contents, err := ioutil.ReadDir(curr.path)
+		if err != nil {
+			return nil, xerrors.Errorf("snapshot tree %s: %w", path, err)
+		}
+		snap := newDirSnapshot(contents)
+		*curr.dst = snap
```
vangent: What is this used for?

internal/cmd/gocdk/fswatch.go
```diff
@@ -0,0 +1,205 @@
...
+		// Pop next node off stack.
+		curr := stack[len(stack)-1]
+		stack = stack[:len(stack)-1]
+
+		// Read directory and create snapshot.
+		contents, err := ioutil.ReadDir(curr.path)
+		if err != nil {
+			return nil, xerrors.Errorf("snapshot tree %s: %w", path, err)
+		}
+		snap := newDirSnapshot(contents)
```
vangent: So this includes directories as well as files? Couldn't the directories be ignored (other than recursing into them)?

internal/cmd/gocdk/fswatch.go
vangent: I see, no. Could it be done that way? I.e., flatten the recursive directory structure into `entries` (where the entry name is the full path from the root). Then the equality comparisons are much easier.

internal/cmd/gocdk/fswatch.go
zombiezen: Done.

internal/cmd/gocdk/fswatch.go
zombiezen: `waitForDirChange` returns `nil` if there was a change.

internal/cmd/gocdk/fswatch.go
zombiezen: I left a TODO in `sameAs`: in the case where there has been a change, `sameAs` could use less stat calls than a full snapshot because it could find the difference and return early before descending on the walk. However, I recognize that this only optimizes for the edge event, not the steady-state. WDYT?

internal/cmd/gocdk/fswatch.go
zombiezen: Yes, because there could be template files or other resources that the binary depends on. (Also Terraform files would need to be included.)

internal/cmd/gocdk/fswatch.go
zombiezen: A removal of an empty directory would be caught by this watcher, but not using the scheme you outline. I suppose the case doesn't matter too much.

internal/cmd/gocdk/fswatch.go
zombiezen: Added comment to clarify.

internal/cmd/gocdk/fswatch.go
zombiezen: I didn't do it that way because I was worried about path normalization issues, but I think the two approaches are roughly the same. Do you think that gives a significant benefit for readability?

internal/cmd/gocdk/fswatch.go
vangent: Agree, either way is fine.

internal/cmd/gocdk/fswatch.go
vangent: Still not following what `.dst` is used for. If it's filling it in here, maybe it should be initialized to `nil` somewhere else? Maybe this will go away if you go with `os.Walk`.

internal/cmd/gocdk/fswatch.go
vangent: Right.

vangent: I wonder if we should make this configurable somewhere... e.g., if the user's app writes some files, it could go wrong. Maybe add a TODO?

internal/cmd/gocdk/fswatch.go
zombiezen: Done.

internal/cmd/gocdk/fswatch.go
zombiezen: Done.

internal/cmd/gocdk/fswatch.go
zombiezen: As you say, obsoleted by using `Walk`.

internal/cmd/gocdk/fswatch.go
zombiezen: To-Done.

Issue #1948, opened 2019-04-29
internal/cmd/gocdk: add minimal add-portable-type command


vangent:
* Add a portable type like `blob.Bucket`.
* Add code to the user's project demonstrating use of the type.

Issue #1810, opened 2019-04-11
internal/cmd/gocdk: add help command


zombiezen:
Currently, the CLI only surfaces help information if you run `gocdk subcommand --help`. It would be nice to support the common alternative spelling `gocdk help subcommand`. In gg, I implemented this by rewriting the second spelling into the first spelling during subcommand dispatch and allowing the subcommand to render its own help documentation.

zombiezen:
Link to gg source: https://github.com/zombiezen/gg/blob/c2ce6c1b011cd5cdfaf92b1f1fd42df508c0dc3d/cmd/gg/main.go#L263-L271

vangent:
This will be handled in https://github.com/google/go-cloud/issues/2147.

PR #1962, opened 2019-04-30
internal/gocdk: a very partial implementation of "add"ing a portable type

vangent:
Lots of TODOs, but seems good to go incrementally. Feedback on the TODOs is welcome.

* I added the `static/` directory using `github.com/shurcooL/vfsgen`, overriding `ModTime` for each entry to be `time.Time{}`.
* Only `blob.Bucket` is supported; adding it just copies the `.go` file for the demo into the user's project.
* The demo supports listing and viewing blobs in the Bucket.

Updates #1948.


internal/cmd/gocdk/static/demo/blob.bucket/demo_blob_bucket.go
```diff
@@ -0,0 +1,186 @@
+// Copyright 2019 The Go Cloud Authors
```
zombiezen: Won't this add a copyright header to code generated for the user? I don't believe we want that.

Also, on the location of this file:

1. It's not under an inner internal directory, so when we eventually move internal/cmd/ -> cmd/, then anyone could import this package.
2. This will get built when `./...` is matched, which is not what we want. The easiest hack is to add an underscore to the beginning of the static directory (see very last sentence of [Package lists and patterns](https://golang.org/cmd/go/#hdr-Package_lists_and_patterns) for guarantee that this works).

internal/cmd/gocdk/static/demo/blob.bucket/demo_blob_bucket.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	ctx := req.Context()
+	reader, err := bucket.NewReader(ctx, key, nil)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("Failed to create Reader: %v\n", err), http.StatusNotFound)
+		return
+	}
+	defer reader.Close()
+	io.Copy(w, reader)
```
zombiezen: Also set `Content-Type` and `Content-Length` headers

internal/cmd/gocdk/add.go
```diff
@@ -16,10 +16,100 @@ package main
...
+
 func add(ctx context.Context, pctx *processContext, args []string) error {
-	return xerrors.New("not implemented")
+	// Compute a sorted slice of available portable types for usage.
+	var avail []string
+	for _, pt := range portableTypes {
+		avail = append(avail, pt.name)
+	}
+	sort.Strings(avail)
+	usageMsg := fmt.Sprintf("gocdk add <%s>", strings.Join(avail, "|"))
```
zombiezen: We already know this list is going to be 5 or more, which seems like it would be unwieldy for the places this message would be used.

internal/cmd/gocdk/add.go
```diff
@@ -16,10 +16,100 @@ package main
...
+	force := f.Bool("force", false, "re-add even the portable type even if it has already been added, overwriting previous files")
+	if err := f.Parse(args); xerrors.Is(err, flag.ErrHelp) {
+		return nil
+	} else if err != nil {
+		return usagef("%s: %w", usageMsg, err)
+	}
+
+	args = f.Args()
+	if len(args) != 1 {
+		return usagef("%s: expected 1 argument, got %d", usageMsg, len(args))
```
zombiezen: Pointing out difference from original design doc: this command should take the form `gocdk add pt(ype?) TYPE`. I do wonder whether for simplicity it could be made into a single argument `add-ptype`, but I don't have thoughts on the ergonomics of this. @clausti for comment.

internal/cmd/gocdk/generate_static.go
```diff
@@ -0,0 +1,68 @@
...
+type zeroModTimeFileInfo struct{ os.FileInfo }
+
+func (zeroModTimeFileInfo) ModTime() time.Time {
+	return time.Time{}
+}
+
+func main() {
+	var fs http.FileSystem = zeroModTimeFS{http.Dir("./static")}
+	err := vfsgen.Generate(fs, vfsgen.Options{
+		Filename:     "static.go", // output file name
```
zombiezen: This makes me sad that we can't just directly pipe to stdout, since it makes the check a little bit trickier.

One simple approach is to use `os.Args[1]` for the file name.

internal/cmd/gocdk/add.go
```diff
@@ -16,10 +16,100 @@ package main
...
+	}
+
+	srcFile, err := static.Open(pt.goDemoPath)
+	if err != nil {
+		return err
+	}
+	defer srcFile.Close()
+
+	// TODO(rvangent): This creates the demo file at the top level of the user's
+	// project. Should we create a subdirectory/sub-package?
```
zombiezen: -1 for subpackage from me.

internal/cmd/gocdk/main.go
```diff
@@ -27,6 +27,13 @@ import (
 	"golang.org/x/xerrors"
 )
 
+// generate_static converts the files in static/ into constants in a new
+// file,
+//go:generate go run generate_static.go
+
+// TODO(rvangent): Add a presubmit to Travis to ensure that the generated file
```
zombiezen: I feel somewhat strongly that this check should be added before submitting this PR.

internal/website/content/internal/cmd/gocdk/static/demo/blob.bucket/_index.md
```diff
@@ -0,0 +1,4 @@
+---
```
zombiezen: I believe once you've addressed the directory naming issue you can remove this file. Otherwise, we should amend the presubmit check.

internal/cmd/gocdk/generate_static.go
vangent: Done.

internal/cmd/gocdk/static/demo/blob.bucket/demo_blob_bucket.go
vangent: Removed copyright.

Renamed directory to `_static`.

internal/cmd/gocdk/main.go

internal/website/content/internal/cmd/gocdk/static/demo/blob.bucket/_index.md
vangent: Done.

internal/cmd/gocdk/add.go
vangent: What other things would `add` add?

internal/cmd/gocdk/static/demo/blob.bucket/demo_blob_bucket.go
vangent: Add a TODO for this. It doesn't play nice with `fileblob`, as none of the blobs on the filesystem have `ContentType`, so it ends up being "application/octet-stream", so the browser downloads them instead of trying to display them.

internal/cmd/gocdk/add.go
vangent: What do you suggest for help then? The list needs to viewable somehow. On separate lines? Or a separate command?

internal/cmd/gocdk/add.go
zombiezen: Biomes.

vangent:
PTAL

internal/cmd/gocdk/add.go
clausti: I'm fine with changing `add ptype` and `add biome` to `add-ptype` and `add-biome` if that makes more sense to most of us.

vangent: Done. I renamed the command (`add` to `add-ptype`) and the function (`add` to `addPortableType`), but didn't rename the file in this PR because it will make rebasing follow-on PRs hard. I added a TODOs to add `add-biome` and to rename the file.

internal/cmd/gocdk/add.go
zombiezen: I mean, I think the final print you have is a fine way of showing it. The information seems appropriate if the type is misspelled or missing.

The problem with this usage message is that the list will be spilled out when passing a misspelled flag or passing too many arguments, which seems unfortunate to me.

vangent: IMHO we need to take an overall look at how flag parsing and help works; we can fix/improve this as part of that effort.

For example, `--help` kind of sucks now; it shows the available flags but not the actual usage message (including non-flag arguments):

```
$ gocdk add-ptype --help
Usage of add:
  -force
    	re-add even the portable type even if it has already been added, overwriting previous files
  -help
    	displays this help message
```

while using a wrong flag shows the full usage, but not the available flags:


```
$ gocdk add-ptype --unknownflag
usage: gocdk add <blob.Bucket>: flag provided but not defined: -unknownflag
`

vangent: Filed https://github.com/google/go-cloud/issues/1997 to follow u.

zombiezen: I agree. I've opened #1991 to look at the flag parsing one and #1810 tracks having a more comprehensive help command. My tendency is to try to address these sort of things in a cross-cutting way once we know what we need. I've added details to both issues with links to code on how I addressed these for gg, my Git wrapper that bumped into most of these same problems.

vangent:
Friendly ping @clausti ...?

PR #1968, opened 2019-05-01
tests: refactor creation of httpreplay.Recorder into common code

eliben:
Updates #770

### DIFF
--- a/internal/testing/setup/setup.go
+++ b/internal/testing/setup/setup.go
@@ -95,14 +95,12 @@ func awsSession(region string, client *http.Client) (*session.Session, error) {
 	})
 }
 
-type RecorderModifier func(r *httpreplay.Recorder)
-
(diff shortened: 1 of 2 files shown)


internal/testing/setup/setup.go
```diff
@@ -95,55 +95,65 @@ func awsSession(region string, client *http.Client) (*session.Session, error) {
 	})
 }
 
-// NewAWSSession2 is like NewAWSSession, but it uses a different record/replay proxy.
-func NewAWSSession2(ctx context.Context, t *testing.T, region string) (sess *session.Session, rt http.RoundTripper, cleanup func()) {
+type RecorderModifier func(r *httpreplay.Recorder)
```
jba: nit: you don't need this type.

eliben: Done

Issue #1972, opened 2019-05-01
gcppubsub: conformance tests -record fails in clean client


eliben:
### Describe the bug

Running conformance tests recording fails in `pubsub/gcppubsub`

### To Reproduce

In a clean master client (with proper credentials):

```
cd pubsub/gcppubsub
go test -v --record
```

Getting failure:

```
    --- FAIL: TestConformance/TestNack (0.36s)
        replay.go:187: Recording into golden file testdata/TestConformance/TestNack.replay
        drivertest.go:342: creating topic: rpc error: code = AlreadyExists desc = Resource already exists in the project (resource=TestConformance_TestNack-topic-1).
 ```

jba:
As a general rule, I don't think our tests should create resources. That should be done once per project/account and out of band, by terraform or a shell script.

eliben:
> As a general rule, I don't think our tests should create resources. That should be done once per project/account and out of band, by terraform or a shell script.

Agreed

vangent:
You should be able to unblock this by going to the UI and deleting that
topic (and likely subscription as well). It's likely that the tests were
interrupted and didn't get a chance to clean up after themselves.

A one time provisioning using terraform might work, but would likely lead
to other forms of flakiness (e.g. undelivered messages remaining in a topic
between runs).

eliben:
Thanks, @vangent this fixed the problem for this test.

I think it's ok for tests to create topics/subscriptions as long as they clean up after themselves. It could help if we modified the test to remove these topics/subscriptions if they already exist before creating. WDYT?

vangent:
Sure, that would make them more stable, but it might also make them slower. Probably fine since `-record` doesn't happen often. They might still be flaky since some providers don't let you delete and then recreate a resource quickly (you get an error that says try again in 1 minute or whatever).

Another approach would be to create a new *different* topic/subscription every time (e.g., put timestamp in the name), but then we'd likely strand them.

eliben:
> Sure, that would make them more stable, but it might also make them slower. Probably fine since `-record` doesn't happen often. They might still be flaky since some providers don't let you delete and then recreate a resource quickly (you get an error that says try again in 1 minute or whatever).
> 

I was thinking keep the code as it is now, but also add a check to delete topics if they already exist before the test starts. This would be rare in practice (only when these leftovers happen)

> Another approach would be to create a new _different_ topic/subscription every time (e.g., put timestamp in the name), but then we'd likely strand them.

[...]

PR #1973, opened 2019-05-01
pubsub/gcppubsub: refactor usage of recorder flag

eliben:
This is refactoring one of the last remaining uses of the `recorder` package - here used just for a flag, so I'm replacing it with a boolean.

This is currently stuck on #1972 since I can't test the recording properly (replay passes).

Updates #770


eliben:
Immediate issue in #1972, submitting

eliben:
This is refactoring one of the last remaining uses of the recorder package - here used just for a flag, so I'm replacing it with a boolean.

Updates #770

PR #1977, opened 2019-05-02
server: fix to start span before logging

### LINKED ISSUE #1976: server: logged traces are always empty
### Describe the bug

Server always logs empty traces (i.e., `"logging.googleapis.com/trace":"00000000000000000000000000000000"`).

### To Reproduce

Run server with logging on.

```
server.New(handler, &server.Options{
	RequestLogger: requestlog.NewStackdriverLogger(os.Stdout, nil),
})
```

## Expected behavior

It should output non-empty traces (e.g., `"logging.googleapis.com/trace":"ed9828b4ed67b6600c1e0c2ba1782f01"`).

### Version

`v0.13.0`.

### Additional context

None.

shouichi:
Fixes #1976.

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -24,7 +24,7 @@ import (
 	"gocloud.dev/requestlog"
 )
 
-func TestDriver(t *testing.T) {
+func TestListenAndServe(t *testing.T) {
 	td := new(testDriver)
 	s := New(http.NotFoundHandler(), &Options{Driver: td})
 	err := s.ListenAndServe(":8080")
@@ -34,11 +34,17 @@ func TestDriver(t *testing.T) {
 	if !td.listenAndServeCalled {
 		t.Error("ListenAndServe was not called from the supplied driver")
 	}
+	if td.handler == nil {
+		t.Error("testDriver must set handler, got nil")
+	}
 }
 
 func TestMiddleware(t *testing.T) {
+	onLogCalled := false
+
 	tl := &testLogger{
 		onLog: func(ent *requestlog.Entry) {
+			onLogCalled = true
 			if ent.TraceID.String() == "" {
 				t.Error("TraceID is empty")
 			}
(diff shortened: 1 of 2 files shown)


shouichi:
> Would you be willing to write a unit test for this change?

Sure 👍

shouichi:
@zombiezen Added a test!

server/server_test.go
```diff
@@ -18,10 +18,13 @@ import (
...
 	"errors"
 	"net/http"
+	"net/http/httptest"
 	"testing"
+
+	"gocloud.dev/requestlog"
 )
 
-func TestListenAndServe(t *testing.T) {
+func TestDriver(t *testing.T) {
```
zombiezen: Let's keep this `ListenAndServe`, because that is what this test is verifying.

server/server_test.go
```diff
@@ -33,15 +36,53 @@ func TestListenAndServe(t *testing.T) {
 	}
```
zombiezen: Let's also check that `td.handler != nil`, since we depend on that in the new test.

server/server_test.go
```diff
@@ -33,15 +36,53 @@ func TestListenAndServe(t *testing.T) {
...
 }
 
+func TestLogging(t *testing.T) {
```
zombiezen:
```suggestion
func TestMiddleware(t *testing.T) {
```

(We may want to verify other properties of the middleware later.)

server/server_test.go
```diff
@@ -33,15 +36,53 @@ func TestListenAndServe(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	req, err := http.NewRequest("GET", "/", nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	rr := httptest.NewRecorder()
+	td.handler.ServeHTTP(rr, req)
```
zombiezen: We should also verify that `onLog` ends up being called.

shouichi:
@zombiezen Thanks for the feedback, fixed the test 🚀

PR #1981, opened 2019-05-02
secrets/localsecrets: fix Decrypt for ciphertext < nonceSize

vangent:
The test would crash before the fix with a slice index out of bounds panic.


vangent:
Friendly ping @clausti ...?

secrets/localsecrets/localsecrets_test.go
```diff
@@ -63,6 +63,35 @@ func (v verifyAs) ErrorCheck(k *secrets.Keeper, err error) error {
 	return nil
 }
 
+func TestSmallData(t *testing.T) {
+	tests := [][]byte{
+		nil,
+		{},
+		{0},
```
eliben: What's the difference between `{0}` and `{65}` as far as this function is concerned? You may want to test a larger value like nonceSize-1 instead?

vangent: Good idea; done.

PR #1983, opened 2019-05-02
secrets/vault: run tests against a vault docker container

### LINKED ISSUE #1958: secrets/vault: stop depending on Vault root module in test
### Describe the bug

Our integration test spins up a Vault instance by linking in the code and running it in-process AFAICT. Vault recently split their repository into submodules: `github.com/hashicorp/vault`, `github.com/hashicorp/vault/api`, and `github.com/hashicorp/vault/sdk`. However, they have not released a version of the `vault` top-level module that excludes the `api` packages. This causes ambiguous import errors during builds.

### To Reproduce

1. Run `go mod tidy` in `gocloud.dev/`
2. Run `internal/testing/listdeps.sh`

## Expected behavior

A list of dependencies is printed to stdout.

## Actual behavior

The following error message occurs:

```
build gocloud.dev/samples/gocdk-secrets: cannot load github.com/hashicorp/vault/api: ambiguous import: found github.com/hashicorp/vault/api in multiple modules:
        github.com/hashicorp/vault v1.1.2 (/home/light/pkg/mod/github.com/hashicorp/vault@v1.1.2/api)
        github.com/hashicorp/vault/api v1.0.1 (/home/light/pkg/mod/github.com/hashicorp/vault/api@v1.0.1)
```

### Version

gocloud.dev master (a5a281f3fa4925245d3f76614677ebb24fb0de08)

### Additional context

We have been near-missing this because up to this point, we haven't imported anything that specified a module requirement on `github.com/hashicorp/vault/api`. Running `go mod tidy` appears to be grabbing everything from Vault tests, which then includes [`github.com/hashicorp/vault/helper/builtinplugins`](https://github.com/hashicorp/vault/blob/master/helper/builtinplugins/registry.go), which seems like a big offender of bringing in more stuff.

This is a known issue reported (and closed as WAI) upstream: hashicorp/vault#6634

shantuo:
Fixes #1958


eliben:
Also related to #886

secrets/vault/vault_test.go
```diff
@@ -45,44 +45,28 @@ func (h *harness) MakeDriver(ctx context.Context) (driver.Keeper, driver.Keeper,
...
-	// Start a new test server.
-	c, cleanup := testVaultServer(t)
+	c, err := Dial(ctx, &Config{
+		Token: testToken,
+		APIConfig: api.Config{
+			Address: apiAddress,
+		},
+	})
+	if err != nil {
+		return nil, err
```
vangent: If the local Vault server is not available, this should `Skip` the test instead of failing....

vangent: See https://github.com/google/go-cloud/pull/1987 for an example.

shantuo: Dial doesn't make request, added `Skip` to the next call.

secrets/vault/vault_test.go
```diff
@@ -21,19 +21,19 @@ import (
...
-	vhttp "github.com/hashicorp/vault/http"
-	"github.com/hashicorp/vault/logical"
-	"github.com/hashicorp/vault/vault"
 	"gocloud.dev/secrets"
 	"gocloud.dev/secrets/driver"
 	"gocloud.dev/secrets/drivertest"
 )
 
+// To run these tests against a real Vault server, first run ./localvault.sh.
+// Then wait a few seconds for the server to be ready.
```
zombiezen: Is there a way we can health check it for readiness? Or perhaps dial-looping for some period of time? This seems like a source of flakiness.

zombiezen: Gotcha. I do wonder whether start_local_deps.sh or localvault.sh should wait for a healthy response before continuing with the tests.

shantuo: Yeah I thought about doing some checks on the server health and retries. But I think it is hard to  know whether we should restart the container or wait longer for the health probe. It would become as flaky. That's why I did more retries when dialing the test server and skip the test after a few attempts. WDYT?

zombiezen: What I'm saying is: the test is using presence of server to determine whether to run the test, so that should fail fast. The localvault.sh script should block on health checks before coming up. This is likely a problem with all current local dep scripts, so it would be fine to file an issue and handle it separately. In the meantime, I think these tests should fail fast.

PR #1987, opened 2019-05-02
runtimevar/etcd: run local etcd from a docker container instead embedding it

vangent:


runtimevar/etcdvar/etcdvar_test.go
```diff
@@ -23,51 +23,45 @@ import (
...
-type harness struct {
-	client *clientv3.Client
-}
+type harness struct{}
 
 func newHarness(t *testing.T) (drivertest.Harness, error) {
-	if etcdErr != nil {
-		return nil, etcdErr
+	if etcdError != nil {
+		t.Skip("No local Kafka running, see pubsub/kafkapubsub/localkafka.sh")
```
eliben: etcd?

Also, why not fail? what's the point of running the test if there is no server?

vangent: Oops, fixed.

We don't run the test if there is no server, we skip the tests.

If we just fail, then the tests on MacOS/Windows would always fail, because the `docker` stuff only works on Linux.

PR #1989, opened 2019-05-02
pubsub/gcppubsub: delete and retry when testing with topic/subscription creation

eliben:
This is a straw-man idea for #1972. Seems to work nicely except I can't figure out a way to check for the error type returned from pubsub RPC calls. @jba suggested:

```
import google.golang.org/grpc/codes
import google.golang.org/grpc/status
if status.Code(err) == codes.AlreadyExists
```

But this doesn't work (`status.Code` returns `Unknown`). So I just put in a workaround for now (see TODO in code).

The idea is: if we find an existing topic we delete it (clean up from previous run), and also advance the topic number so we don't run into cool-off issues (re-create a topic with the same name just after deleting).


vangent:
How about `grpc.Code(err)` ?

pubsub/gcppubsub/gcppubsub_test.go
```diff
@@ -58,9 +58,22 @@ func newHarness(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
...
-	return createTopic(ctx, h.pubClient, topicName, topicPath)
+	// We may encounter topics that were created by a previous test run and were
+	// not properly cleaned up. In such a case delete the existing token and create
+	// a new topic with a higher topic number (to avoid cool-off issues between
+	// deletion and re-creation of the same topic).
+	for {
+		topicName := fmt.Sprintf("%s-topic-%d", sanitize(testName), atomic.AddUint32(&h.numTopics, 1))
+		topicPath := fmt.Sprintf("projects/%s/topics/%s", projectID, topicName)
+		dt, cleanup, err := createTopic(ctx, h.pubClient, topicName, topicPath)
+		// TODO: not for submit; how to test this error correctly?
```
vangent: Would this be simpler with recursion? E.g.,

```
topicName := ...
topicPath := ...
dt, cleanup, err := createTopic(...)
if err == nil {
  return dt, cleanup, err
}
h.pubClient.DeleteTopic(...)
return h.CreateTopic(...)
```

pubsub/gcppubsub/gcppubsub_test.go
```diff
@@ -58,9 +58,22 @@ func newHarness(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
...
+		if err != nil && strings.Contains(err.Error(), "AlreadyExists") {
```
vangent: The problem might be that we're wrapping the error inside `createTopic` below rather than returning exactly what `pubClient.CreateTopic` returns. Try removing that `fmt.Printf` and seeing if the error code thing @jba suggested works.

eliben: Nice spot! It works now with @jba's check

Just removing the `fmt.Errorf` wrapping for now. Is this something xerrors could help with?

pubsub/gcppubsub/gcppubsub_test.go
eliben: I love recursion as much as the next person, but the iterative version seems much more readable to me here

pubsub/gcppubsub/gcppubsub_test.go
```diff
@@ -58,15 +58,27 @@ func newHarness(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
...
+		topicPath := fmt.Sprintf("projects/%s/topics/%s", projectID, topicName)
+		dt, cleanup, err := createTopic(ctx, h.pubClient, topicName, topicPath)
+		if err != nil && status.Code(err) == codes.AlreadyExists {
+			// Delete the topic and retry.
+			h.pubClient.DeleteTopic(ctx, &pubsubpb.DeleteTopicRequest{Topic: topicPath})
+			continue
+		}
+		return dt, cleanup, err
+	}
 }
```
vangent: Do the same thing for `CreateSubscription` below?

eliben: Done

eliben:
Added retry to subscription too

pubsub/gcppubsub/gcppubsub_test.go
jba: Yes, in the new error world you can add text to an error but still preserve the error value itself.

PR #1992, opened 2019-05-03
testing: remove go-vcr dependency: we no longer use it

### LINKED ISSUE #770: all: use a better replay tool

eliben:
Fixes #770


eliben:
Reran listdeps to fix travis

internal/testing/replay/replay.go
```diff
@@ -16,168 +16,13 @@
...
-				}
-			}
-		}
-		keep = append(keep, action)
-	}
-	c.Interactions = keep
-	c.Mu.Unlock()
-	return c.Save()
-}
-
```
vangent: Can we move `NewGCPDialOptions` to `setup.go` and just delete this `replay` package entirely?

vangent: (feel free to do in a separate PR)

vangent:
\o/ thanks @eliben !!!

shantuo:
Also remove the old recorded files?

internal/testing/replay/replay.go
eliben: Yep, have separate PR in progress for that

eliben:
> Also remove the old recorded files?

Oops, yes. will do in a separate PR

PR #1993, opened 2019-05-03
internal/docstore/firedocstore: add BeforeDo for actions

shantuo:
Updates #1675 

For writes, `As` supports the slice type `[]*pb.Write` since we return that in the update call. It has only one element but my understanding is that there could be more than one and the slice type allows the user to append more `Write`s before commit.


internal/docstore/firedocstore/fs.go
```diff
@@ -357,6 +373,19 @@ func (c *collection) runWrites(ctx context.Context, actions []*driver.Action) er
 		}
 		newNames[i] = nn
 		pws = append(pws, ws...)
+		if opts.BeforeDo != nil {
+			asFunc := func(i interface{}) bool {
+				p, ok := i.(*[]*pb.Write)
```
jba: I think this should be the CommitRequest instead.

shantuo: That's right, I somehow missed that type.

PTAL

PR #1995, opened 2019-05-03
test: inline remaining replay.go function and remove stale .yaml replay files

eliben:
Followup on #1992


eliben:
> I think you should also delete `internal/website/content/internal/testing/replay/_index.md`.

Done

Issue #1994, opened 2019-05-03
website: hugo seems to keep generating "changing" sitemap.xml files


eliben:
See https://github.com/google/go-cloud/commit/332c1bde650b09ca2589871809e0ae6 -- we have many commits every day to gh-pages, which essentially reshuffle the `sitemap.xml` file order.

Robert suggests trying a newer Hugo version

Or we could open an issue for Hugo

vangent:
I filed https://github.com/gohugoio/hugo/issues/5934.

zombiezen:
I agree with Robert's assessment here: this seems like an issue with upstream Hugo.

eliben:
The default Hugo `sitemap.xml` template uses:

```
{{ range .Data.pages }}
...
{{ end }}
```

We could create our own template that uses a sorted range? (`range sort`)?

eliben:
#1996 should address this for now (I hope). Will keep a watch on the commits to gh-pages to see if the spurious sitemap updates stop. Leaving open for now to see if this gets fixed and what bep says on the Hugo issue

eliben:
I don't see spurious commits to gh-pages any more, so closing as fixed.

PR #1996, opened 2019-05-03
website: add custom sitemap.xml template

eliben:
Use a sorted range loop of data pages to prevent undefined order which
causes superfluous diffs

Updates #1994


eliben:
Compare to https://github.com/gohugoio/hugo/blob/master/tpl/tplimpl/embedded/templates/_default/sitemap.xml

I added `sort "Permalink"` to sort by a field; dunno if it's the best field to sort by, but then again the sorting order shouldn't matter

PR #2001, opened 2019-05-06
pubsub/doc: delete obsolete sentence about Message.CanNack

Rolinh:
CanNack() is a method of Subscription, not Message.

### DIFF
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -76,8 +76,7 @@
 //
 // Since Message.Nack never makes sense for some providers (for example, for
 // at-most-once providers, the provider can't redeliver the message), Nack will
-// panic if called for some providers. You can call Message.CanNack to see if
-// it is available.
+// panic if called for some providers.
 //
 // OpenCensus Integration
 //


pubsub/pubsub.go
```diff
@@ -76,8 +76,8 @@
 //
 // Since Message.Nack never makes sense for some providers (for example, for
 // at-most-once providers, the provider can't redeliver the message), Nack will
-// panic if called for some providers. You can call Message.CanNack to see if
-// it is available.
+// panic if called for some providers. You can call Subscription.CanNack to see
```
vangent: Thanks for catching this! Actually, this sentence should be deleted entirely. `Subscription.CanNack` is on the driver interface type (`driver.Subscription`, not `pubsub.Subscription`), so it is not exposed to end users. We are considering exposing it to end users via `Message.Nackable` (currently commented out below).

So, do you mind just deleting the sentence starting with "You can call"?

Rolinh: Sure. Done now.

pubsub/pubsub.go
```diff
@@ -74,11 +74,6 @@
 //    ever call Message.Ack.
 //  - func() { log.Info("ack called!") }: log. Softer than panicking.
 //
```
vangent: Sorry, please only delete the last sentence from the paragraph, not the whole paragraph.

Rolinh: Oh yes. It was indeed clear, sorry about that...

PR #2008, opened 2019-05-06
blob: add support for creating a blob.Bucket based on a prefix

### LINKED ISSUE #909: blob/all: consider supporting a base path during construction
`fileblob` already has this since you give it a root dir, but other providers are assumed to be rooted at the top-level of the Bucket. The user can always add a prefix everywhere, but it might be useful to allow them to specify a prefix at construction time.

For example, I'm uploading stuff to my Bucket in a virtual folder `foo/bar`; I should be able to specify `foo/bar` at construction time and be confident that everything is scoped to `foo/bar` and below.

vangent:
Fixes #909.

I went ahead with this because Hugo needs it (according to bep@, it is common to target subfolders like this). It is of course possible to implement client side, but it's quite tricky to intercept the key in all the right places; see https://github.com/gohugoio/hugo/pull/5937 for an example.

I used your [suggestion](https://github.com/google/go-cloud/issues/909#issuecomment-445393080) of wrapping the Bucket rather than trying to do via per-driver `Options`. I think it worked out pretty well. I wrapped the `driver.Bucket` rather than the `blob.Bucket` itself.

I implemented it using a prefix rather than a more flexible `func(string) string` mapper because I couldn't really think of a use case other than targeting a subfolder, and it's somewhat tricky to implement an arbitrary function since the mapping needs to be reversible (so, we'd likely need to take 2 functions, not 1).


blob/blob.go
```diff
@@ -1073,3 +1093,9 @@ func wrapError(b driver.Bucket, err error) error {
 }
 
 var errClosed = gcerr.Newf(gcerr.FailedPrecondition, nil, "blob: Bucket has been closed")
+
+// PrefixedBucket returns a *Bucket based on b with all keys modified to have
+// prefix.
+func PrefixedBucket(b *Bucket, prefix string) *Bucket {
+	return NewBucket(driver.NewPrefixedBucket(b.b, prefix))
```
vangent: Open question: should we allow/support providing a prefix that doesn't end in a "/"? I'm not sure what the use case is for a non-directory prefix, and it's easy to mess up (e.g,. `?prefix=mysubdir` instead of `prefix=mysubdir/`).

We could:
a) Document that `prefix` should likely end with a `/`.
b) Add a `/` to prefix if it doesn't have one (probably in `driver.NewPrefixedBucket`).
c) Change this and `driver.NewPrefixedBucket` to return an error if `prefix` doesn't end with a `/`.

I'm mildly leaning toward c), but let me know what you think.

blob/driver/driver.go
```diff
@@ -304,3 +306,56 @@ type SignedURLOptions struct {
...
+	if opts != nil {
+		myopts = *opts
+	}
+	myopts.Prefix = b.prefix + myopts.Prefix
+	page, err := b.base.ListPaged(ctx, &myopts)
+	if err != nil {
+		return nil, err
+	}
+	for _, p := range page.Objects {
+		p.Key = strings.TrimPrefix(p.Key, b.prefix)
```
vangent: IIUC, `p.Key` **should** have a `b.prefix` prefix here. We could `panic` or return an error if it doesn't instead of silently continuing.

blob/fileblob/fileblob_test.go
```diff
@@ -35,17 +35,23 @@ import (
...
 type harness struct {
 	dir       string
+	prefix    string
 	server    *httptest.Server
 	urlSigner URLSigner
 	closer    func()
 }
 
-func newHarness(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
+func newHarness(ctx context.Context, t *testing.T, prefix string) (drivertest.Harness, error) {
```
vangent: I test the new functionality in `fileblob_test` and `memblob_test`; I think that should be sufficient, no?

`fileblob_test` tests the URL part too (`memblob_test` just checks that it doesn't fail).

vangent:
An alternative to this that just occurred to me is to add a `WithPrefix` function to `blob.Bucket` itself, which would just modify a struct field `prefix`, and apply the prefix in the `blob.Bucket` code as needed. I think we could get the `URLOpener` to work still, by grabbing the `prefix` query param, using the regular `URLOpener`, and then calling `WithPrefix` on the result.

Not sure if that's preferable.

blob/blob.go
zombiezen: My preference is a), since I could imagine some other delimiter being used (like dashes)

blob/blob.go
```diff
@@ -1042,7 +1042,27 @@ func (mux *URLMux) OpenBucketURL(ctx context.Context, u *url.URL) (*Bucket, erro
 	if err != nil {
 		return nil, err
 	}
-	return opener.(BucketURLOpener).OpenBucketURL(ctx, u)
+	return applyPrefixParam(ctx, opener.(BucketURLOpener), u)
```
zombiezen: We should definitely document this query parameter.

blob/driver/driver.go
```diff
@@ -304,3 +306,56 @@ type SignedURLOptions struct {
...
+
+// prefixedBucket implements Bucket by prepending prefix to all keys.
+type prefixedBucket struct {
+	base   Bucket
+	prefix string
+}
+
+// NewPrefixedBucket returns a Bucket based on b with all keys modified to have
+// prefix.
+func NewPrefixedBucket(b Bucket, prefix string) Bucket {
```
zombiezen: I don't understand why we're exporting this function in the driver. Can you help me understand the use case?

blob/blob.go
zombiezen: Is it necessary to implement this in terms of a `driver.Bucket`? I am concerned that this implies that our portable type is not good at composing, since folks may want to create similar layers and wouldn't have access to the driver of an arbitrary portable type instance.

blob/fileblob/fileblob_test.go
zombiezen: SGTM

blob/driver/driver.go
vangent: It is exported from the driver so that `blob.go` can call it.

We could alternatively move the `prefixedBucket` type into `blob.go` entirely, but
a) I felt like it was more natural here.
b) The conformance tests are run given a `driver.Bucket`, so we need to be able to wrap a `driver.Bucket` in order to run the conformance test suite on a prefixed bucket.

An alternative might be to add a conformance test that starts with a `driver.Bucket`, makes a `blob.Bucket` out of it, and then turns it into a prefixed `Bucket`, and then runs a test. This would make it so that every driver was tested with a prefixed bucket instead of just `mem` + `file`, but would mean that we only run a specific conformance test (not yet written), as opposed to what happens now, which is that the entire conformance test suite is run against a prefixed bucket.

blob/blob.go
vangent: Let's chat in person.

-- No, it is not necessary, but I think you necessarily end up with a different type. I.e., `type PrefixedBucket` for the wrapped type, not `blob.Bucket`.
-- That makes it hard to use a possibly-prefixed bucket.
-- In particular, I'm not sure if it composes well with an `URLOpener` that returnes a prefixed bucket.

vangent:
I filed #2053 to discuss a more general purpose solution. As discussed, I think it's independent from this PR.

blob/blob.go
vangent: Updated docstring.

blob/blob.go

vangent: (by adding a docstring to `blob.OpenBucket`, and by adding an example).

I'll update the HOWTO on `gocloud.dev` separately.

vangent:
PTAL

PR #2009, opened 2019-05-06
docstore: implement Query.Delete

jba:


internal/docstore/drivertest/drivertest.go
```diff
@@ -914,6 +919,90 @@ func testQuery(t *testing.T, coll *ds.Collection) {
 	}
 }
 
+func testDeleteQuery(t *testing.T, coll *ds.Collection) {
+	cleanUpTable(t, newHighScore, coll)
```
shantuo: Could `cleanUpTable` also use `RunDeleteQuery`?

internal/docstore/firedocstore/query.go
```diff
@@ -349,3 +356,44 @@ func fieldRef(fp []string) *pb.StructuredQuery_FieldReference {
...
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return err
+		}
+		pws = append(pws, &pb.Write{
+			Operation: &pb.Write_Delete{Delete: res.Document.Name},
+		})
+		if len(pws) >= maxWritesPerRPC {
```
shantuo: Is there a limit on the client library? Why do we want to limit the number of writes here? More efficient?

internal/docstore/mongodocstore/query.go
```diff
@@ -115,3 +137,54 @@ func (it *docIterator) As(i interface{}) bool {
...
+	// If there is no limit, use the DeleteMany call.
+	if q.Limit <= 0 {
+		filter, err := c.filtersToBSON(q.Filters)
+		if err != nil {
+			return err
+		}
+		_, err = c.coll.DeleteMany(ctx, filter, nil)
+		return err
+	} else {
+		// If there is a limit, we have to run a query, because DeleteMany doesn't
```
shantuo: If the user provides a limit when they want to "blind" delete on a query, how would they know whether it deletes the documents they wanted? I suggest we might want to not support limit for query-and-writes at the first place.

internal/docstore/firedocstore/query.go
jba: To limit the memory on the client. We don't want to read a million items and then delete them, because the client would have to store all million.

internal/docstore/mongodocstore/query.go
jba: Good point, it does seem weird to do that. I'll make it an error.

jba:
PTAL.

internal/docstore/firedocstore/query.go
shantuo: I just got a `rpc error: code = InvalidArgument desc = maximum 500 writes allowed per request` when running some benchmarks, I guess there is a limit on the writes.

Issue #2013, opened 2019-05-07
internal/gocdk: add automated tests for launch


zombiezen:
I'm sending out the launch command for review, but I haven't found a satisfactory way to test the command. The process is by definition non-hermetic: it is taking a binary and making it available in a different location. HTTP/RPC replay may help a little, but it won't make the application available at the address claimed by the launcher, which is an important property to test.

The local launcher is a little bit better since all we have to communicate with is the local Docker daemon, but that doesn't tell us much other than we correctly issued a `docker run` command.

WDYT @clausti and @vangent? Should we keep these tests manual for now with some test plan, or do you have ideas on how we can test this functionality in an automated way?

vangent:
Can we test the local launcher using HTTP requests to the local port? It seems like we're doing that already to some extent using `waitForHealthy`. I'm doing something similar in the `ptype-add` tests (PR coming as soon as the base one is approved).

I agree that the remote launchers are going to be tough to test. HTTP replay might be good if `-record` works for it.

clausti:
>HTTP/RPC replay may help a little, but it won't make the application available at the address claimed by the launcher, which is an important property to test.

Is it possible to have a test run only in record mode (skip it if you're not recording)? if so, would it be sufficient to do the "is it actually live at the address" testing only every so often? Instead of skipping the test entirely, we could have it check how old the recording is and fail if it's over the threshold?

zombiezen:
> Is it possible to have a test run only in record mode (skip it if you're not recording)?

Yes.

> if so, would it be sufficient to do the "is it actually live at the address" testing only every so often?

Yes, we could skip the check for server liveness when the test is run in replay mode. But replay mode is the default and is run pervasively: when we run `go test ./...`, on Travis, etc.

My observation here is that at that point the test is largely then verifying whether the RPC is made with the correct parameters, which is usually easier to maintain as a unit test. I think that would be my preference, but it does not give as much confidence as an end-to-end test. On the other hand, I think this more limited unit test would make the limited utility of the test more apparent.

> Instead of skipping the test entirely, we could have it check how old the recording is and fail if it's over the threshold?

-1 from me on that idea. The failure mode I see there is that the test would start failing on an unrelated PR and it would block any submissions until the test is fixed.

On a previous team, logs would get regenerated weekly by the build cop, but it was much more automated than our setup. Trying to implement such a plan without having that in place would likely generate a fair amount of toil (noise in recording and flaky backends) with not enough benefit (actual regressions discovered).

[...]

Issue #1861, opened 2019-04-19
internal/cmd/gocdk: implement minimal launch command


zombiezen:
The most minimal thing I can think of for this to do is to take a Docker image and get it to run with [Cloud Run](https://cloud.google.com/run/docs/). While the steps I outline below are specific to Cloud Run, the logic should be able to cleanly branch for other deployment environments.

The Docker image should already exist by the time `gocdk launch` runs (see #1807) and will be passed as a flag to `gocdk launch`, but the command will need to:

1. Gather the selected launcher from the biome's biome.json and the launch specifier from the biome's Terraform outputs (`output "launcher_specifier"`).
2. (Maybe) Push the Docker image to Google Container Registry.
3. Make the correct Cloud Run API calls to start up the service from the Docker image.

Later on, we can worry about propagating environment variables/configuration from Terraform outputs to the Cloud Run startup, but that can be additive.

zombiezen:
Taking this from @vangent as discussed offline. I've prototyped this last week: I'll send out PRs that implement the Cloud Run functionality described above.

Issue #1890, opened 2019-04-23
internal/cmd/gocdk: specify configuration mechanism


zombiezen:
In initial prototyping/design, @clausti and I proposed using a JSON configuration file whose path would be passed in via command-line flag and the server's port would be specified via command-line flag.

However, in examining some of our potential launch targets, I think it might make more sense to use environment variables for configuration. Especially for server port, there's precedence of using the `PORT` environment variable in:

- [Cloud Run](https://cloud.google.com/run/docs/reference/container-contract#port)
- [Heroku](https://devcenter.heroku.com/articles/runtime-principles#web-servers), and derivative systems like [Foreman](https://ddollar.github.io/foreman/#PROCFILE)
- [Cloud Foundry](https://docs.run.pivotal.io/devguide/deploy-apps/environment-variable.html#PORT)

Further, it looks like there are some potential launch targets (like Cloud Run) where it is difficult to set command-line arguments. Instead of inventing a new scheme, it seems like it might just make sense to follow the general consensus on this one and just use environment variables.

Once we decide, we should update `gocdk serve` and `gocdk launch` to use this configuration mechanism.

/cc @clausti @vangent

zombiezen:
I haven't heard any objections to environment variables, so I'm going to implement this.

PR #2014, opened 2019-05-07
internal/cmd/gocdk: implement local launcher

zombiezen:
Adds full launch command.

Unsure what the testing strategy here should be, #2013 tracks that.

Updates #1861


internal/cmd/gocdk/launch.go
```diff
@@ -15,11 +15,170 @@
...
+	// Read the launch specifier from the biome's Terraform output.
+	tfOutput, err := tfReadOutput(ctx,
+		findBiomeDir(moduleRoot, biome),
+		pctx.overrideEnv("TF_IN_AUTOMATION=1"))
+	if err != nil {
+		return xerrors.Errorf("gocdk launch: %w", err)
+	}
+
+	// Launch the application.
+	launchURL, err := launcher.Launch(ctx, &launchInput{
```
vangent: How are you choosing which arguments to pass to `Launch` vs which ones are passed to `newLauncher`? It might be simpler to put them all in once place or the other...?

internal/cmd/gocdk/launch.go
```diff
@@ -15,11 +15,170 @@
...
+	c.Dir = local.dockerDir
+	out, err := c.CombinedOutput()
+	if err != nil {
+		if len(out) == 0 {
+			return nil, xerrors.Errorf("local launch: docker run: %w", err)
+		}
+		return nil, xerrors.Errorf("local launch: docker run:\n%s", out)
+	}
+
+	local.logger.Printf("Docker container %s started, waiting for healthy...", bytes.TrimSuffix(out, []byte("\n")))
```
vangent: How do they stop it?

internal/cmd/gocdk/launch.go
```diff
@@ -15,11 +15,170 @@
...
+	serveURL := &url.URL{
+		Scheme: "http",
+		Host:   "localhost:8080",
+		Path:   "/",
+	}
+	healthCheckURL := &url.URL{
```
vangent: Nit: Or, slightly shorter:

```
healthCheckURL:= *serveURL
healthCheckURL.Path = "/healthz/readiness"
```
?

internal/cmd/gocdk/launch.go
```diff
@@ -15,11 +15,170 @@
...
+		Scheme: serveURL.Scheme,
+		Host:   serveURL.Host,
+		Path:   "/healthz/readiness",
+	}
+	if err := waitForHealthy(ctx, healthCheckURL); err != nil {
```
vangent: Nit: would it be easier to use if `waitForHealthy` took an url string instead of a `url.URL`?

internal/cmd/gocdk/launch.go
```diff
@@ -15,11 +15,170 @@
...
+		return nil, xerrors.Errorf("local launch: %w", err)
+	}
+	return serveURL, nil
+}
+
+// launchSpecifier is the set of arguments passed from a biome's Terraform
+// module to the biome's launcher.
+type launchSpecifier map[string]interface{}
```
vangent: I am not sure that this type is adding much value. Consider leaving it as a `map[string]interface{}`. You can have still have a `stringValue` helper that takes one as an argument.

internal/cmd/gocdk/launch.go
```diff
@@ -15,11 +15,170 @@
...
+	dockerDir string
+}
+
+// Launch implements Launcher.Launch.
+func (local *localLauncher) Launch(ctx context.Context, input *launchInput) (*url.URL, error) {
+	dockerArgs := []string{
+		"run",
+		"--rm",
+		"--detach",
+		"--publish", "8080:8080",
```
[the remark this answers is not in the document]
clausti: it might be nice for it to be settable, but I think it's ok for running `serve` and `deploy` in the same biome simultaneously to be mutually exclusive.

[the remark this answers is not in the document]
zombiezen: Added a launch specifier option.

internal/cmd/gocdk/launch.go
zombiezen: That which is specific to a launcher type and could be reused between launches goes into the launcher constructor. \
That which is generic to any launcher goes into the `launchInput`.

internal/cmd/gocdk/launch.go
```diff
@@ -15,11 +15,170 @@
...
+	return nil
+}
+
+// Launcher is the interface for any type that can launch a Docker image.
+type Launcher interface {
+	Launch(ctx context.Context, input *launchInput) (*url.URL, error)
+}
+
+// launchInput is the input to a launcher.
+type launchInput struct {
```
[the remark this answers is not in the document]
zombiezen: Done.

internal/cmd/gocdk/launch.go
```diff
@@ -15,11 +15,170 @@
...
+	})
+	if err != nil {
+		return xerrors.Errorf("gocdk launch: %w", err)
+	}
+	fmt.Fprintf(pctx.stdout, "Serving at %s\n", launchURL)
+	return nil
+}
+
+// Launcher is the interface for any type that can launch a Docker image.
+type Launcher interface {
```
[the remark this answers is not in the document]
zombiezen: TODO added.

internal/cmd/gocdk/launch.go
```diff
@@ -15,11 +15,170 @@
...
+}
+
+// newLauncher creates the launcher for the given name.
+func newLauncher(ctx context.Context, pctx *processContext, launcherName string) (Launcher, error) {
+	logger := log.New(pctx.stderr, "gocdk: ", log.Ldate|log.Ltime)
+	switch launcherName {
+	case "local":
+		return &localLauncher{
+			logger:    logger,
+			dockerEnv: pctx.env,
```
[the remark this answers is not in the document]
zombiezen: > It's confusing to me that `localLauncher` has a `dockerEnv` from `pctx.env`, but `launchInput` also has an `env`. What goes where? Can we just have one set of environment variables?

Based on discussions around #1890, using environment variables to pass configuration to the running binary seems right, so these are two distinct sets of environment variables. We can't join them.

> IIUC, one is the set used to exec `docker`, while the other is the set passed to the docker image being run. I guess that's reasonable. Maybe `localLauncher` can just have a `processContext` field instead of copying its subfields?

I was trying to be as specific as possible with what the `localLauncher` depends on from the `processContext`.

internal/cmd/gocdk/launch.go
```diff
@@ -15,11 +15,198 @@
...
+	"path/filepath"
+	"strconv"
 
 	"golang.org/x/xerrors"
 )
 
 func launch(ctx context.Context, pctx *processContext, args []string) error {
-	return xerrors.New("not implemented")
+	f := newFlagSet(pctx, "launch")
+	dockerImage := f.String("image", "", "Docker image to launch in the form `image[:tag]` (defaults to project latest)")
```
clausti: The dockermentation uses `name:tag`, and I think we should too. (Not objecting to the optional brackets, just to calling it "image" instead of "name".)

internal/cmd/gocdk/launch.go
```diff
@@ -15,11 +15,170 @@
...
+	"os/exec"
+	"path/filepath"
 
 	"golang.org/x/xerrors"
 )
 
 func launch(ctx context.Context, pctx *processContext, args []string) error {
-	return xerrors.New("not implemented")
+	f := newFlagSet(pctx, "launch")
+	imageName := f.String("image", "", "Docker image name to launch (defaults to project latest)")
```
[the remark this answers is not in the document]
clausti: Definitely, I think the biome/launcher concepts will be useful with non-gocdk projects. I'm cool with having users specify both the name and the tag, just not just the name.

internal/cmd/gocdk/launch.go
```diff
@@ -15,11 +15,170 @@
...
+	}
+	launcher, err := newLauncher(ctx, pctx, *cfg.Launcher)
+	if err != nil {
+		return xerrors.Errorf("gocdk launch: %w", err)
+	}
+
+	// Read the launch specifier from the biome's Terraform output.
+	tfOutput, err := tfReadOutput(ctx,
+		findBiomeDir(moduleRoot, biome),
+		pctx.overrideEnv("TF_IN_AUTOMATION=1"))
```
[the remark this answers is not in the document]
vangent: Shouldn't we set it in `tfReadOutput` then?

internal/cmd/gocdk/launch.go
vangent: Is `Launch` going to support an `Unlaunch` or `Stop` command? That would seem reasonable to me.

vangent: (maybe file an issue for that).

And consider just extending this `Printf` with a note on how to stop the container.

internal/cmd/gocdk/launch.go
vangent: Sorry, I don't follow your comment. The extra type made this harder to follow for me, not easier. We'll probably be reading config variables from the json config in various places, I think it will be easier to follow if we just have a set of utilities for doing so rather than adding a new type for each use case.

internal/cmd/gocdk/launch.go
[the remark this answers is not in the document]
vangent: I guess so, but it makes it less readable since you're basically aliasing a bunch of things -- I have to keep track of the fact that `dockerEnv` is really `pctx.env`. There are so many different envs and dirs floating around that extra names for existing ones just adds to the confusion.

internal/cmd/gocdk/launch.go
vangent: But isn't `newLauncher` used for all launchers as well? I think your comment would make sense to me if there were a separate function being called to construct each launcher, but there's only one (with a switch inside it). So the args to `newLauncher` aren't really specific to a launcher type are they?

internal/cmd/gocdk/launch.go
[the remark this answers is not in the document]
zombiezen: Now I understand. I'm worried about the breaking of convention: when specifying Docker image name + tag, the `name[:tag]` form is prevailing. I've never seen a system that accepts `[name:]tag` or `[name]:tag` as input. Even `docker build -t` (the "tag" flag) takes in the `name[:tag]` form. We could separate this out as two different flags, but that can have surprising effects for the user.

It may be nice for `gocdk deploy` to have such a tag flag, since that would affect both `build` and `launch`, but I don't know if that's in scope for this PR. Can we open an issue for this and proceed from there?

internal/cmd/gocdk/launch.go
zombiezen: It is broadly speaking helpful for the interface split, especially if we move forward with #2046 as we discussed offline.

internal/cmd/gocdk/launch.go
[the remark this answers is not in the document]
zombiezen: Ack. I'd like to address this as part of #2046.

internal/cmd/gocdk/launch.go
zombiezen: Added log statement. Unlaunch/stop is likely going to be too provider-specific since you need to give a handle to a previous deployment. Honestly, I think the more common situation will be that folks just run `gocdk launch -image=MYIMAGE:OLDTAG` to do a rollback.

internal/cmd/gocdk/launch.go
zombiezen: Done.

internal/cmd/gocdk/launch.go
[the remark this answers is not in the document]
zombiezen: Did a little more digging on this, since I want to consider the effect on `gocdk build`. The definitive grammar for a Docker image reference is here: https://github.com/docker/distribution/blob/master/reference/reference.go

So in examining that, `:tag` would never be an acceptable image reference, so we could permit that short form to take the name from the project Dockerfile. Implemented that, WDYT?

internal/cmd/gocdk/launch.go
vangent: I'm fine with delaying this until then. Can you add an explicit TODO about figuring out which arguments should be passed to create the launcher, vs which arguments should be passed to `launcher.Launch`? It's not obvious, especially since `Launch` is likely only going to be called once.

internal/cmd/gocdk/launch.go
vangent: Do you mind removing it for now? If you feel like it needs to be added to make the Cloud Run launcher work well, you can reintroduce it in that PR, but at this point I think it just adds a level of indirection that's hard to follow and doesn't provide much value.

zombiezen: Removed.

PR #2019, opened 2019-05-08
gocdk: add tests for add-ptype and the portable type demos

vangent:
Updates #1948.


internal/cmd/gocdk/add_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+import (
+  "log"
+	"net/http"
+  "os"
+)
+
+func main() {
+	port := os.Getenv("PORT")
+  log.Fatal(http.ListenAndServe(":" + port, nil))
```
zombiezen: Indentation differs on these lines

internal/cmd/gocdk/add_test.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	defer cleanup()
+
+	pctx := &processContext{
+		workdir: dir,
+		env:     os.Environ(),
+		stdout:  ioutil.Discard,
+		stderr:  ioutil.Discard,
+	}
+	// Uncomment when debugging.
```
zombiezen: Delete?

internal/cmd/gocdk/add_test.go
```diff
@@ -0,0 +1,195 @@
...
+	// pctx.stdout = os.Stdout
+	// pctx.stderr = os.Stderr
+	ctx := context.Background()
+
+	// TODO(rvangent): Run gocdk init instead of this.
+	if err := ioutil.WriteFile(filepath.Join(dir, "main.go"), []byte(mainContent), 0666); err != nil {
```
zombiezen: `newTestModule` should do what you want.

internal/cmd/gocdk/add_test.go
```diff
@@ -0,0 +1,195 @@
...
+				}
+				if logBody {
+					t.Error("Full HTTP response body:\n\n", body)
+				}
+			}
+		})
+	}
+}
+
+func freePort() int {
```
zombiezen: nit: add "find" because "free" can be ambiguous as a verb:

```suggestion
func findFreePort() int {
```

internal/cmd/gocdk/add_test.go
```diff
@@ -0,0 +1,195 @@
...
+	l, _ := net.Listen("tcp", ":0")
```
zombiezen: Should probably still check the error so that the ensuing panic has the proper error message instead of a type assertion panic. Also, maybe restrict to loopback interface (`localhost`) to avoid exposing ports to the world.

internal/cmd/gocdk/add_test.go
vangent: Done.

internal/cmd/gocdk/add_test.go
vangent: Done.

internal/cmd/gocdk/add_test.go
vangent: I call `newTestModule` up above. But it currently doesn't include a `main.go` file, which is what this line is adding.

internal/cmd/gocdk/add_test.go
vangent: Done.

internal/cmd/gocdk/add_test.go
vangent: Done and done.

PR #2020, opened 2019-05-08
internal/gocdk: add a demo for runtimevar

vangent:
Updates #1948.


internal/cmd/gocdk/_static/demo/runtimevar.variable/demo_runtimevar_variable.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+var variableURL string
+var variable *runtimevar.Variable
+var variableErr error
+
+func init() {
+	variableURL = os.Getenv("RUNTIMEVAR_VARIABLE_URL")
+	if variableURL == "" {
+		variableURL = "constant://?val=my-variable&decoder=string"
```
clausti: does this result in something functional for the demo? (It's not clear to me if this is only intended as an example of the url form or if it results in a usable thing based on `constantvar`)

vangent: Yes, it is usable -- it ends up with a variable that always has this value.

vangent: https://screenshot.googleplex.com/phmAJSZuikt

PR #2022, opened 2019-05-08
travis: tell Travis that static.go is generated

vangent:


zombiezen:
I don't understand the problem this PR is solving. If it is to hide the diff from GitHub PRs, then the static.go file already contains the necessary pattern described at https://golang.org/s/generatedcode to trigger the Linguist generated handling.

However, the PR description suggests this has something to do with Travis. Can you elaborate?

vangent:
Oh OK, I did not realize that Github would already recognize it. Never mind then....

PR #2023, opened 2019-05-08
gocdk: flesh out the demo for blob.Bucket

vangent:
Updates #1948.

You can now write to a blob and select a specific blob for reading.


internal/cmd/gocdk/_static/demo/blob.bucket/demo_blob_bucket.go
```diff
@@ -71,101 +90,193 @@ const blobBucketBaseTemplate = `
...
+  {{if .WriteSuccess}}
+    Wrote it!
   {{else}}
-    <div>no blobs in bucket</div>
-  {{end}}
-</body>
-</html>
-`
+    <form>
+    Blob key to write to (any previous blob will be overwritten):
```
clausti: The explanations for the input fields should ideally be labels, but at least grouped by semantic html for accessibility.

internal/cmd/gocdk/_static/demo/blob.bucket/demo_blob_bucket.go
```diff
@@ -60,7 +79,7 @@ const blobBucketBaseTemplate = `
     This page demonstrates the use of a Go CDK blob.Bucket.
   </p>
   <p>
-    It is currently using a blob.Bucket based on the URL "{{ . }}", which
+    It is currently using a blob.Bucket based on the URL "{{ .URL }}", which
```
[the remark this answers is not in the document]
vangent: IIUC, both are true. The variable should eventually be configured in the biome config, and the launch/serve process should translate those config entries into environment variables, which will be read in the code.

At this point, there's no biome config yet, so the statement here is correct. Once the `TODO` you linked to is implemented, we'll need to update this text to  be more clear about how it's working.

internal/cmd/gocdk/_static/demo/blob.bucket/demo_blob_bucket.go
vangent: Added labels.

vangent:
PTAL.

internal/cmd/gocdk/_static/demo/blob.bucket/demo_blob_bucket.go
```diff
@@ -71,101 +90,193 @@ const blobBucketBaseTemplate = `
...
       {{end}}
     </div>
+  {{end}}` + blobBucketTemplateSuffix
+
+	// Input: *blobBucketViewData.
+	blobBucketViewTemplate = blobBucketTemplatePrefix + `
+  {{if .ListObjects}}
+    <form>
+    Choose a blob to view:
+    <select name="key" label="choose a blob to view">
```
clausti: Unfortunately, I don't think this will work, as `label` is an element but not a valid attribute. 

You can either give the `input` element an `id` attribute and have a `label` element with a `for` attribute with a value that is the same as the `id` value, or you can nest the input inside of the label. [MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label)

vangent: Ah, done, here and in `secrets` demo.

PR #2025, opened 2019-05-08
mongodocstore: implement unordered action lists

jba:
We use the BulkWrite call when possible, for performance. This affects
our ability to report errors precisely.


internal/docstore/mongodocstore/mongo.go
```diff
@@ -289,49 +290,59 @@ func (c *collection) toMongoFieldPath(fp []string) string {
...
-		return err
+func sliceToLower(s []string) {
+	for i, e := range s {
+		s[i] = strings.ToLower(e)
 	}
-	mdoc, err := c.encodeDoc(a.Doc, id)
+}
+
+func (c *collection) create(ctx context.Context, a *driver.Action) error {
+	// See https://docs.mongodb.com/manual/reference/method/db.collection.insertOne.
```
shantuo: This and L309 might better be summarized and put on top of `create`.

internal/docstore/mongodocstore/mongo.go
```diff
@@ -289,49 +290,59 @@ func (c *collection) toMongoFieldPath(fp []string) string {
...
+	return nil
 }
 
-func (c *collection) replace(ctx context.Context, a *driver.Action, upsert bool) error {
-	// See https://docs.mongodb.com/manual/reference/method/db.collection.replaceOne
+func (c *collection) prepareCreate(a *driver.Action) (mdoc, createdID interface{}, err error) {
 	id, err := c.docID(a.Doc)
+	if id == "" {
+		if c.idField == "" {
+			// If the user provides a function to get the ID rather than a field, then Create
```
shantuo: Where is `idFunc` used now?

internal/docstore/mongodocstore/mongo.go
```diff
@@ -454,6 +488,166 @@ func (c *collection) docID(doc driver.Document) (interface{}, error) {
...
+		for i, w := range bwe.WriteErrors {
+			alerr = append(alerr, indexedError{-i - 1, gcerr.Newf(translateMongoCode(w.Code), nil, "%s", w.Message)})
+		}
+		return alerr
+	}
+	if res.DeletedCount != nDeletes {
+		alerr = append(alerr, indexedError{-1, gcerr.Newf(gcerr.NotFound, nil, "some delete failed")})
+	}
+	if res.MatchedCount != nMatches {
+		alerr = append(alerr, indexedError{-2, gcerr.Newf(gcerr.NotFound, nil, "some replace failed")})
```
shantuo: Should this `-2` be `-1`?

internal/docstore/mongodocstore/mongo.go
```diff
@@ -480,8 +670,15 @@ func (c *collection) ErrorCode(err error) gcerrors.ErrorCode {
...
 }
 
-func sliceToLower(s []string) {
-	for i, e := range s {
-		s[i] = strings.ToLower(e)
+// Error code for a write error when no documents match a filter.
+// (The Go mongo driver doesn't define an exported constant for this.)
+const mongoNotFoundCode = 11000
+
+func translateMongoCode(code int) gcerrors.ErrorCode {
```
shantuo: Can this be merged into `ErrorCode`?

internal/docstore/mongodocstore/mongo.go
```diff
@@ -454,6 +488,166 @@ func (c *collection) docID(doc driver.Document) (interface{}, error) {
...
+		go func() {
+			err := c.get(ctx, g.action)
+			getc <- indexedError{g.index, err}
+		}()
+	}
+
+	// Run all writes in a single BulkWrite RPC.
+	alerrBulk := c.unorderedBulkWrite(ctx, writes)
+	if alerrBulk != nil {
+		alerr = append(alerr, alerrBulk...)
```
shantuo: How about doing something similar to the firestore one, make a slice here and give a proper length to the error list?

internal/docstore/mongodocstore/mongo.go

internal/docstore/mongodocstore/mongo.go
jba: I guess it doesn't matter, they could all be -1.
Done.

internal/docstore/mongodocstore/mongo.go
jba: The problem is how to handle the errors that aren't associated with any action (the ones with -1).

internal/docstore/mongodocstore/mongo.go
jba: It's called in `docID`.

internal/docstore/mongodocstore/mongo.go
```diff
@@ -454,6 +488,166 @@ func (c *collection) docID(doc driver.Document) (interface{}, error) {
...
+		}
+		// The returned indexes of the WriteErrors are wrong. See https://jira.mongodb.org/browse/GODRIVER-1028.
+		// Until it's fixed, use negative values for the indexes in the errors we return.
+		for _, w := range bwe.WriteErrors {
+			alerr = append(alerr, indexedError{-1, gcerr.Newf(translateMongoCode(w.Code), nil, "%s", w.Message)})
+		}
+		return alerr
+	}
+	if res.DeletedCount != nDeletes {
+		alerr = append(alerr, indexedError{-1, gcerr.Newf(gcerr.NotFound, nil, "some delete failed")})
```
shantuo: If user redo the same deletes when this happen, will Mongo succeed for the remaining ones and silently ignore those that are already deleted? If so we can suggest the user to retry it.

internal/docstore/mongodocstore/mongo.go
jba: Moved this line above the function declaration.

internal/docstore/mongodocstore/mongo.go
```diff
@@ -289,49 +290,59 @@ func (c *collection) toMongoFieldPath(fp []string) string {
...
-func (c *collection) replace(ctx context.Context, a *driver.Action, upsert bool) error {
-	// See https://docs.mongodb.com/manual/reference/method/db.collection.replaceOne
+func (c *collection) prepareCreate(a *driver.Action) (mdoc, createdID interface{}, err error) {
 	id, err := c.docID(a.Doc)
+	if id == "" {
+		if c.idField == "" {
+			// If the user provides a function to get the ID rather than a field, then Create
+			// can't make a unique ID for a document that's missing one, because it doesn't
+			// know how to create it or where to store it in the original document.
+			return nil, nil, err
```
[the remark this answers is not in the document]

internal/docstore/mongodocstore/mongo.go
jba: It's much more likely that the documents just don't exist. If there were a real error, like a network issue, then I'm pretty sure the call would return an error.

PR #2027, opened 2019-05-08
all: update go.mod

eliben:
Following the fix of #1958, we can remove the vault requirement in go.mod. For this PR:

* Removed vault from go.mod
* Reran `go test ./...`, and the go tool updated go.mod


vangent:
Tests failed:
```
FAIL: dependencies changed; run: internal/testing/listdeps.sh > internal/testing/alldeps
```

eliben:
> Tests failed:
> 
> ```
> FAIL: dependencies changed; run: internal/testing/listdeps.sh > internal/testing/alldeps
> ```

Yes, rerunning tests now. `vault` masked a bunch of packages it seems, so now they explicitly appear in `go list`, not sure why... but they all are used by `vault/api`. Maybe a general refactoring of their packages

PR #2029, opened 2019-05-08
all: run Docker dependent tests whenever possible

### LINKED ISSUE #2005: testing: explore options to check local container status before running tests
We discussed in the vault [PR](https://github.com/google/go-cloud/pull/1983#discussion_r281254551) that we want to get a sync point after localXXX.sh boots a container of a certain server for testing.

There are a few options:
1. Do nothing more, we set a timeout on the constructors and fail if the server is not up after a few seconds.
1. Wait a few seconds after the container starts and check its health.
1. If the previous one is still flaky, write a Go program to do more proper waiting.

shantuo:
Fixes #2005 

All tests that require Docker container to run will run when either:
 1) Not on Travis.
 2) On Travis Linux environment, where Docker is available.

### DIFF
--- a/internal/testing/setup/setup.go
+++ b/internal/testing/setup/setup.go
@@ -334,3 +334,11 @@ func NewGCPDialOptions(t *testing.T, recording bool, filename string) (opts []gr
 	}
 	return opts, done
 }
+
+// RunTestsDependingOnDocker returns true when either:
+// 1) Not on Travis.
+// 2) On Travis Linux environment, where Docker is available.
+func RunTestsDependingOnDocker() bool {
+	s := os.Getenv("TRAVIS_OS_NAME")
+	return s == "" || s == "linux"
+}
--- a/internal/testing/setup/setup.go
+++ b/internal/testing/setup/setup.go
@@ -335,10 +335,10 @@ func NewGCPDialOptions(t *testing.T, recording bool, filename string) (opts []gr
 	return opts, done
 }
 
-// RunTestsDependingOnDocker returns true when either:
+// HasDockerTestEnvironment returns true when either:
 // 1) Not on Travis.
 // 2) On Travis Linux environment, where Docker is available.
-func RunTestsDependingOnDocker() bool {
+func HasDockerTestEnvironment() bool {
 	s := os.Getenv("TRAVIS_OS_NAME")
 	return s == "" || s == "linux"
 }
--- a/pubsub/kafkapubsub/kafka_test.go
+++ b/pubsub/kafkapubsub/kafka_test.go
@@ -33,6 +33,7 @@ import (
 	"github.com/Shopify/sarama"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"gocloud.dev/internal/testing/setup"
 	"gocloud.dev/pubsub"
 	"gocloud.dev/pubsub/driver"
 	"gocloud.dev/pubsub/drivertest"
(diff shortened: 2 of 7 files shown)


secrets/vault/vault_test.go
```diff
@@ -37,6 +39,8 @@ const (
 	testToken  = "faketoken"
 )
 
+var enableTransit sync.Once
```
eliben: Document this var

internal/testing/setup/setup.go
```diff
@@ -334,3 +334,11 @@ func NewGCPDialOptions(t *testing.T, recording bool, filename string) (opts []gr
 	}
 	return opts, done
 }
+
+// RunTestsDependingOnDocker returns true when either:
+// 1) Not on Travis.
+// 2) On Travis Linux environment, where Docker is available.
+func RunTestsDependingOnDocker() bool {
```
eliben: I feel the name of this function could be better, conveying it's a predicate. Something like `HasDockerTestEnvironment` or `CanRunLocalServerTests`, etc.

eliben:
OOC, how do you verify these tests actually ran on Travis, rather than being skipped?

secrets/vault/vault_test.go

internal/testing/setup/setup.go
shantuo: Went with `HasDockerTestEnvironment`

pubsub/kafkapubsub/kafka_test.go
```diff
@@ -53,21 +54,21 @@ type harness struct {
 }
 
 var checkOnce sync.Once
-var kafkaRunning bool
+var kafkaError error
```
vangent: I think the `sync.Once` and `kafkaError` variables and `localKafkaRunning` can all be dropped. The tests will just fail naturally if Kafka isn't running.

pubsub/kafkapubsub/kafka_test.go
```diff
@@ -196,8 +197,11 @@ func (asTest) BeforeSend(as func(interface{}) bool) error {
 
 // TestKafkaKey tests sending/receiving a message with the Kafka message key set.
 func TestKafkaKey(t *testing.T) {
-	if !localKafkaRunning() {
-		t.Skip("No local Kafka running, see pubsub/kafkapubsub/localkafka.sh")
+	if !setup.HasDockerTestEnvironment() {
+		t.Skip("Skipping Kafka tests since the Kafka server is not available")
+	}
+	if err := localKafkaRunning(); err != nil {
```
vangent: Same here, not needed.

runtimevar/etcdvar/etcdvar_test.go
```diff
@@ -38,20 +39,25 @@ var (
...
 func init() {
-	etcdClient, etcdError = clientv3.NewFromURL("http://localhost:2379")
-	if etcdError == nil {
-		// Check to see if the local etcd is actually running.
-		ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
-		defer cancel()
-		_, etcdError = etcdClient.Put(ctx, "unused", "unused")
+	if setup.HasDockerTestEnvironment() {
+		etcdClient, etcdError = clientv3.NewFromURL("http://localhost:2379")
+		if etcdError == nil {
```
vangent: Lines 44-48 can be dropped; same as above, tests will just fail naturally if etcd isn't running.

secrets/vault/vault_test.go
```diff
@@ -37,6 +39,9 @@ const (
 	testToken  = "faketoken"
 )
 
+// enableTransit cheks and makes sure the Transit API is enable only once.
```
vangent: nit: cheks -> checks, enable -> enabled

eliben:
> I ran locally with the travis env, our Travis script doesn't have `-v` for `go test` so can't tell from there.

I think it's worthwhile to test this at least once (like add `-v` in your PR and remove it later)?

secrets/vault/vault_test.go

Issue #1883, opened 2019-04-23
internal/testing: consolidate list of project modules


zombiezen:
### Is your feature request related to a problem? Please describe.

When a new module is added, it must be added to:

-  gocloud.code-workspace (see golang/go#30841)
-  internal/testing/listdeps.sh
-  internal/testing/runchecks.sh

Failure to do so means that tests silently don't run for the module. Especially if we start adding more submodules, this problem is going to get worse.

### Describe the solution you'd like

I would like to have either a consolidated list of the modules or have the scripts automatically detect the go.mod files and declare boundaries there, if feasible.

### Additional context

#1846 is an example of a time that I did forget, #1847 is an example of a time that I almost forgot to update one of the files (runchecks.sh) but not the other (listdeps.sh).

eliben:
I can see us having some single place where all the modules in the repo are listed, and the shell scripts reading from there. Not sure how this applies to the vscode config file though - can it actively read from a separate file? Or would it need to be some sort of codegen (config-gen) solution for that case?

zombiezen:
VSCode would need code-gen, yeah.

eliben:
@zombiezen - I've been thinking about this and I'm going back and forth on the helpfulness of code-gen-ing it vs. once-in-a-rare-while updating the VScode file. Not sure what workflow would be helpful to you, so assigning it to you

zombiezen:
Manual workspace edits are probably fine, as long as we don't change modules too often.

zombiezen:
Closing since I don't think I have anything to contribute here.

PR #2036, opened 2019-05-09
testing: consolidate list of modules into a single place

eliben:
Strawman that only updates listdeps.sh for now, just to test the idea. WDYT?

The reason I'm placing it in a .sh file that I `source` is that we can add comments to it explaining what it does and also makes the "user" scripts cleaner.

A similar script can be added later to generate the vscode stuff

Updates #1883


zombiezen:
This looks nice. One idea: you could enter the directories line by line and still have comments by reading it using:

```bash
mapfile -t ALLMODULES < <(sed -e '/^#/d' path/to/file.txt)
```

zombiezen:
Amending for blank lines:

```bash
mapfile -t ALLMODULES < <(sed -e '/^#/d' -e '/^$/d' path/to/file.txt)
```

eliben:
Thanks. Changed to data file and using `sed | while read` to read now. PTAL

allmodules
```diff
@@ -0,0 +1,11 @@
+# This file lists all the modules in our repository; it's used by scripts that
+# have to perform operations for each module.
+#
+# Any line that doesn't begin with a '#' character and isn't empty is treated
+# as a path that has a module in it.
```
zombiezen:
```suggestion
# as a path relative to the top of the repository that has a module in it.
```

allmodules
```diff
@@ -0,0 +1,11 @@
...
+
+.
+./internal/cmd/gocdk
```
zombiezen: nit: these don't need the `./` prefix.

eliben: Done

PR #2037, opened 2019-05-09
gocdk: use gocdk init in tests instead of custom main.go content

vangent:
Updates #1948.

We can rip the flag out if we have a better solution for "init into an empty directory" later.


internal/cmd/gocdk/init.go
```diff
@@ -31,9 +31,11 @@ import (
 func init_(ctx context.Context, pctx *processContext, args []string) error {
 	f := newFlagSet(pctx, "init")
 	var modpath string
+	var allowExistingDir bool
 	f.StringVar(&modpath, "module-path", "", "the module import path for your "+
 		"project's go.mod file (required if project is outside of GOPATH)")
 	f.StringVar(&modpath, "m", "", "alias for --module-path")
+	f.BoolVar(&allowExistingDir, "allow-existing-dir", false, "true to allow initializing an existing directory (contents may be overwritten!)")
```
zombiezen: Please add a TODO before this line referencing #1918.

vangent: Done. I also updated the existing TODO down below to refer to it as well.

PR #2039, opened 2019-05-09
testing: go program to summarize test output

jba:
I tried using jq, but gave up. This is pretty simple.


eliben:
https://gist.github.com/eliben/f9321b97b18a71328eca6ad21ba6d592 isn't much longer :-)

jba:
Agreed, changed to Go. 

PTAL

internal/testing/test-summary/test-summary.go
```diff
@@ -0,0 +1,67 @@
...
+	Elapsed float64 // seconds
+	Output  string
+}
+
+func main() {
+	s, err := run(os.Stdin)
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Println(s)
```
eliben: Should it return a nonzero value if there were failed tests?

internal/testing/test-summary/test-summary.go
```diff
@@ -0,0 +1,67 @@
...
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"time"
+)
+
+// From runing "go doc test2json".
```
eliben: running

internal/testing/test-summary/test-summary.go
```diff
@@ -0,0 +1,67 @@
...
+
+func run(r io.Reader) (string, error) {
+	counts := map[string]int{}
+	line := 0
+	scanner := bufio.NewScanner(bufio.NewReader(r))
+	for scanner.Scan() {
+		line++
+		var event TestEvent
+		if err := json.Unmarshal(scanner.Bytes(), &event); err != nil {
+			return "", fmt.Errorf("line %d: %v", line, err)
```
eliben: Why not just show the line itself instead of the line number?

internal/testing/test-summary/test-summary.go
jba: Done.

internal/testing/test-summary/test-summary.go
jba: Done.

jba:
> Can we just replace `go test ./...` with this?

As currently designed, you'd pipe `go test -json` into it.

jba:
PTAL

internal/testing/test-summary/test-summary.go
eliben: Nit: fix this before merging

jba:
Enhancements sound good to me. I'll work on it later in the week.

PR #2043, opened 2019-05-09
gocdk: add more tests for the blob demo

vangent:
Updates #1948.


vangent:
FYI: The diff is cleaner if you hide whitespace changes....

PR #2044, opened 2019-05-10
internal/docstore/dynamodocstore: RunDeleteQuery uses unordered actions

shantuo:


internal/docstore/dynamodocstore/query.go
```diff
@@ -456,33 +456,39 @@ func (qr *queryRunner) queryPlan() string {
...
-		if err != nil {
-			return err
+		if last == nil {
+			break
 		}
-		actions = append(actions, &driver.Action{Kind: driver.Delete, Doc: doc})
+		startAfter = last
 	}
-	alerr := c.RunActions(ctx, actions, &driver.RunActionsOptions{})
+	alerr := c.runActionsUnordered(ctx, actions, &driver.RunActionsOptions{Unordered: true})
```
jba: Something seems wrong here. Why do we need to set `Unordered: true` if the function is called `runActionsUnordered`?

shantuo: Yeah this is going to be discarded inside `runActionsUnordered`, I was thinking it might use some other API like batch API's but no. I removed it.

PR #2045, opened 2019-05-10
docstore: implement update queries

jba:


jba:
@shantuo ping

internal/docstore/firedocstore/fs.go
```diff
@@ -634,23 +648,37 @@ func toServiceFieldPathComponent(key string) string {
 // revisionPrecondition returns a Firestore precondition that asserts that the stored document's
 // revision matches the revision of doc.
 func revisionPrecondition(doc driver.Document) (*pb.Precondition, error) {
+	rev, err := revisionTimestamp(doc)
```
shantuo: Why create these two functions? The 3 names look similar and may cause some confusion and it seems you can use the `revisionPrecondition` in the above scenarios. At least one of the three can be saved.

internal/docstore/firedocstore/query.go
```diff
@@ -357,12 +357,32 @@ func (c *collection) QueryPlan(q *driver.Query) (string, error) {
...
+}
+
 // For delete and update queries, limit the number of write actions per RPC, to bound
 // client memory.
 // This is a variable so it can be modified for tests.
 var maxWritesPerRPC = 500
 
-func (c *collection) RunDeleteQuery(ctx context.Context, q *driver.Query) error {
+// runActionQuery runs the query, calls writes for each returned document, and then commits those writes.
+func (c *collection) runActionQuery(ctx context.Context, q *driver.Query, writes func(*pb.Document) ([]*pb.Write, error)) error {
```
shantuo: nit: s/runActionQuery/runWriteQuery, `RunGetQuery` doesn't use it.

internal/docstore/firedocstore/fs.go
jba: I use preconditionFromTimestamp in newUpdateWrites and revisionTimestamp in updateWrites.

internal/docstore/firedocstore/query.go

jba:
> Also revert the change of dynamo record tests since they are not changed.

I tried to do this. On my branch the files in master are the same. I don't know why they show up as diffs here.

Issue #2041, opened 2019-05-09
testing: reported failure with go test -json


eliben:
In a clean clone, running `go test ./...` doesn't report failures. 

However, running `go test -json ./...` reports one failing test:

```
{"Time":"2019-05-09T15:26:51.338409101-07:00","Action":"fail","Package":"gocloud.dev/tests/aws/app","Elapsed":0.101}
```

Even though the return code from `go test` is still 0.

eliben:
Opened https://github.com/golang/go/issues/31969

eliben:
For us this is fixed by #2049 
Upstream bug tracking underlying fix

PR #2049, opened 2019-05-10
testing: properly skip tests/aws/app tests when no environment is set up

eliben:
Returning from TestMain doesn't play well with `go test -json` (see #2041). Instead we'll properly skip these tests when no test environment is set up

### DIFF
--- a/tests/aws/app/app_test.go
+++ b/tests/aws/app/app_test.go
@@ -63,10 +63,10 @@ func TestMain(m *testing.M) {
 	if hostIP == "" || gcpProjectID == "" {
 		log.Println("Test environments need to be setup to run server tests.")
 		flag.PrintDefaults()
-		return
-	}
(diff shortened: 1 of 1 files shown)


tests/aws/app/app_test.go
```diff
@@ -63,10 +63,10 @@ func TestMain(m *testing.M) {
...
-		return
-	}
-	if err := startApp(); err != nil {
-		log.Fatal("startApp: ", err)
+	} else {
+		if err := startApp(); err != nil {
+			log.Fatal("startApp: ", err)
+		}
 	}
 	os.Exit(m.Run())
```
shantuo: So `os.Exit` must be called to make the test pass? Add a comment here as a reminder.

eliben: Done

PR #2054, opened 2019-05-11
pubsub/awssnssqs: add a pubsub.Topic implementation for sending directly to an SQS queue

vangent:
This ended up being a pretty big PR, sorry.

It involved a couple of changes to the conformance tests:
1. The ability to skip the test that verifies fanout; if you're using an SQS topic for sending & receiving, there's no fanout.
2. AWS SQS only allows 10 metadata keys; our test was using 18. It was working before because we weren't testing `Raw` mode; by default, messages delivered to SQS from SNS come with everything (body + metadata) JSON-encoded in the `Body` field. Once I tried `Raw` mode (including sending directly to SQS), I hit this limitation. I modified the test to split the weird metadata keys into sorted groups of 10.

For awssnssqs:
1. There are now two constructors returning a `pubsub.Topic`, one for publshing to SNS, and one for publishing to SQS.
2. I did a bunch of renaming and refactoring to share code (mostly the encoding/decoding) stuff as needed.
3. `ReceiveMessages` now handles both the "raw" and default-SNS cases.
4. We now run conformance tests with topics sending to SNS, SNS+raw mode, and SQS.

I will update the examples and the website in a separate PR.


pubsub/awssnssqs/awssnssqs.go
```diff
@@ -448,45 +582,74 @@ func openSubscription(ctx context.Context, sess client.ConfigProvider, qURL stri
...
+		// MessageAttributes. Same deal if you send directly to the SQS queue.
+		//
+		// If it looks like the body might be JSON, we try decoding it. If it
+		// doesn't look like JSON, or if the JSON decode fails, we use the raw
+		// data from m.
+		bodyStr := aws.StringValue(m.Body)
+		rawAttrs := map[string]string{}
+		useJSON := len(m.MessageAttributes) == 0 // if we got attributes, it's raw
+		if useJSON {
+			type MsgBody struct {
```
jba: You don't need the type name. Just
```
var bodyJSON struct { ... }
```

pubsub/drivertest/drivertest.go
```diff
@@ -65,6 +65,10 @@ type Harness interface {
 	// MaxBatchSizes returns the maximum size of SendBatch/Send(Na|A)cks, or 0
 	// if there's no max.
 	MaxBatchSizes() (int, int)
+
+	// SupportsMultipleSubscriptions returns true iff the provider supports
```
jba: s/returns true iff/reports whether/

pubsub/drivertest/drivertest.go
```diff
@@ -291,6 +295,9 @@ func testSendReceiveTwo(t *testing.T, newHarness HarnessMaker) {
 		t.Fatal(err)
 	}
 	defer h.Close()
+	if !h.SupportsMultipleSubscriptions() {
+		t.Skip("multiple subscriptions to a topic not supported")
```
jba: What will happen if someone actually tries to create a second subscription to an SQS queue? Do we return an informative error?

pubsub/drivertest/drivertest.go
```diff
@@ -705,9 +712,24 @@ func testMetadata(t *testing.T, newHarness HarnessMaker) {
 	}
 	defer h.Close()
 
-	weirdMetadata := map[string]string{}
-	for _, k := range escape.WeirdStrings {
-		weirdMetadata[k] = k
+	// Some provider limit the number of metadata per message.
```
jba: We should document this non-portability somewhere. Probably on `Message.Metadata`. The current doc comment reads
```
// Metadata has key/value metadata for the message. It will be nil if the
// message has no associated metadata.
```
which only describes the receive side. It should be changed to describe the send side (e.g. "Before sending, set this ....") and to say that some providers have a max.

pubsub/awssnssqs/awssnssqs.go
```diff
@@ -448,45 +582,74 @@ func openSubscription(ctx context.Context, sess client.ConfigProvider, qURL stri
...
+			for k, v := range m.MessageAttributes {
+				rawAttrs[k] = aws.StringValue(v.StringValue)
+			}
 		}
-		// See BodyBase64Encoding for details on when we base64 decode message bodies.
+
 		decodeIt := false
 		attrs := map[string]string{}
-		for k, v := range body.MessageAttributes {
+		for k, v := range rawAttrs {
```
jba: (Re the line above.) The doc comment for `Message.Metadata` says it will be nil if there are no attributes, but in fact it will be an empty map.

pubsub/awssnssqs/awssnssqs.go
vangent: Done.

pubsub/drivertest/drivertest.go
vangent: Done.

pubsub/drivertest/drivertest.go
vangent: No; I don't think we can tell. The behavior would be the same as if they created a subscription in one server and another subscription in another server -- each subscription would get ~half of the messages.

pubsub/drivertest/drivertest.go
vangent: Updated in `pubsub.go`.

pubsub/awssnssqs/awssnssqs.go
vangent: The portable type sets it to `nil` if empty:
https://github.com/google/go-cloud/blob/master/pubsub/pubsub.go#L606-L608

vangent:
Thanks for the quick review! PTAL.

Issue #2048, opened 2019-05-10
samples: make samples its own module


eliben:
We currently have `samples/appengine` as a separate module.

We should instead make `samples/` a module, with all the samples just commands within that module. This will be needed as part of #886, since samples import all our providers and thus can't be part of the core module (if we want to not force some providers as dependencies on users of core)

eliben:
Fixed by #2057

PR #2057, opened 2019-05-13
samples: move go.mod into samples root, making samples its own module

eliben:
Updates #2048 - moves the go.mod file from samples/appengine to samples.

Note the `replace gocloud.dev => ../` added to the go.mod. Without this, running `go test ./...` in `samples/` results in many errors like:

```
can't load package: package gocloud.dev/samples/server: unknown import path "gocloud.dev/samples/server": ambiguous import: found gocloud.dev/samples/server in multiple modules:
	gocloud.dev/samples (/home/eliben/eli/go-cloud/samples/server)
	gocloud.dev v0.13.0 (/home/eliben/eli/go/pkg/mod/gocloud.dev@v0.13.0/samples/server)
```

The symptom and solution is explained by bcmills in https://github.com/ugorji/go/issues/279 (which refers also to https://github.com/golang/go/issues/27899). The new go.mod points to gocloud.dev v0.13, which also provides these packages - so the go command is confused - it sees the same package(s) provided by two different modules.

The ugorji/go solution was to use a pseudo-version pointing at an existing commit in the core module which removes the packages - this removes the ambiguity. In our case, there is no existing commit yet - so I'm using a replace line. The replace line should be unnecessary when we release a new CDK version.

This has interesting implications for #886 - we'll likely have to do the same when we split out providers to their own modules and retain `replace` lines until a new release.


zombiezen:
Can you also update the VSCode workspace? Otherwise IntelliSense breaks.

eliben:
@zombiezen, sure thing #2076

PR #2059, opened 2019-05-13
all: update cloud.google.com/go to 0.39.0

jba:
This will bring faster httpreplay and improved rpcreplay.


jba:
This is going to fail just like the PR that updated to master. For some reason, either my changes to httpreplay or some downstream change, the replays fail and I have to re-record.

eliben:
> This is going to fail just like the PR that updated to master. For some reason, either my changes to httpreplay or some downstream change, the replays fail and I have to re-record.

I'm OK with re-recording, if this is what it takes. But it would be nice to at least have some idea what the difference is and whether it's "safe" from our POV. Do you have any clue to explain what's going on?

jba:
I think the issue was this change, which was done in response to a problem you encountered with Content-Type:
```
-               // No content-type header. There should not be a body.
-               if len(body) != 0 {
-                       return "", nil, errors.New("no Content-Type, but body")
-               }
-               return "", nil, nil
+               // No content-type header. Treat the body as a single part.
+               return "", [][]byte{body}, nil
```
The match was failing because previously there was no body at all, and now there is an empty body.

In other words, this is a totally harmless, trivial thing, probably not worth fixing in the code, but it does require a re-record.

eliben:
> I think the issue was this change, which was done in response to a problem you encountered with Content-Type:
> 
> ```
> -               // No content-type header. There should not be a body.
> -               if len(body) != 0 {
> -                       return "", nil, errors.New("no Content-Type, but body")
> -               }
> -               return "", nil, nil
> +               // No content-type header. Treat the body as a single part.
> +               return "", [][]byte{body}, nil
> ```
> 
> The match was failing because previously there was no body at all, and now there is an empty body.
> 
> In other words, this is a totally harmless, trivial thing, probably not worth fixing in the code, but it does require a re-record.

Ah, OK. Is it only failing for Azure (where this was an issue), or for all the providers?

jba:
It failed for GCP too. Looks like it's the same reason. I think the empty-body case worked before, but there must have been another case only with azure.

jba:
All tests pass for me locally. PTAL.

jba:
ping @vangent

PR #2068, opened 2019-05-13
docs: update examples and how-tos for AWS pubsub changes

### LINKED ISSUE #2032: pubsub/awssnssqs: add an implementation for Topic that sends directly to SQS
It seems it is not possible to send Messages on subscriptions, only receive messages or send acks.

Maybe I missed something, but it thus does not seem possible to write the publisher part that generate messages received by subscribers.

So, it would be nice to be able to implement both roles with go-cloud.

vangent:
Fixes #2032.


internal/website/content/howto/pubsub/publish.md
```diff
@@ -45,7 +45,7 @@ application connects to the correct region, but otherwise `pubsub.OpenTopic`
 will use the region found in the environment variables or your AWS CLI
 configuration.
 
-{{< goexample "gocloud.dev/pubsub/awssnssqs.Example_openTopic" >}}
+{{< goexample "gocloud.dev/pubsub/awssnssqs.Example_openSNSTopicFromURL" >}}
```
vangent: Note: I renamed the URL-based examples to have a "FromURL" suffix, not sure if that's good. I find the current examples a bit confusing:
https://godoc.org/gocloud.dev/pubsub/awssnssqs#pkg-examples

(there's no way to distinguish the URL-based examples from the other ones).

If you like this, I'll do a separate PR to update the other URL examples similarly.

zombiezen: Seems fine to me: I don't have a strong preference.

vangent: https://github.com/google/go-cloud/issues/2091

Issue #2067, opened 2019-05-13
docstore: support increment/decrement as an update operation


jba:
Currently you can use the Update action to set a field to a different value, or delete a field. We should also provide a way to increment a numeric field.

jba:
Provider support:
- MongoDB: the [`$inc` update operator](https://docs.mongodb.com/manual/reference/operator/update/inc/#up._S_inc)
- Firestore: the [`increment` FieldTransform](https://cloud.google.com/firestore/docs/reference/rpc/google.firestore.v1#google.firestore.v1.DocumentTransform.FieldTransform)
- DynamoDB: the [ADD update action](https://godoc.org/github.com/aws/aws-sdk-go/service/dynamodb#UpdateItemInput)

PR #2069, opened 2019-05-13
docstore: API for increment modifications

jba:
This PR contains only the new API for an increment modification.

Implementations and tests to follow.

See #2067.


internal/docstore/docstore.go
```diff
@@ -331,11 +342,30 @@ func toDriverMods(mods Mods) ([]driver.Mod, error) {
 		if err != nil {
 			return nil, err
 		}
+
+		if inc, ok := v.(driver.IncOp); ok && !isIncNumber(inc.Amount) {
+			return nil, gcerr.Newf(gcerr.InvalidArgument, nil,
+				"Increment amount %v must be an integer or floating-point number", inc.Amount)
```
shantuo: nit: use verb `%T`.

internal/docstore/memdocstore/mem.go
```diff
@@ -275,6 +283,39 @@ func (c *collection) update(doc map[string]interface{}, mods []driver.Mod) error
...
 
+// Add two encoded numbers.
+// Since they're encoded, they are either int64 or float64.
+// Allow adding a float to an int, producing a float.
+// TODO(jba): see how other providers handle that.
+func add(x, y interface{}) (interface{}, error) {
+	if x == nil {
+		return y, nil
+	}
+	switch x := x.(type) {
```
shantuo: Is it safe to do
1. convert both to `float64` and calculate.
1. convert back if both are `int64`?

jba: Usually, but not always.
https://play.golang.org/p/WAv_3ZENgnx

internal/docstore/docstore.go

PR #2070, opened 2019-05-13
internal/gocdk: add a demo for pubsub

vangent:
Updates #1948.


internal/cmd/gocdk/_static/demo/pubsub/demo_pubsub.go
```diff
@@ -0,0 +1,171 @@
...
+var topicURL string
+var topic *pubsub.Topic
+var topicErr error
+
+var subscriptionURL string
+var subscription *pubsub.Subscription
+var subscriptionErr error
+
+func init() {
+	topicURL = os.Getenv("PUBSUB_TOPIC_URL")
```
clausti: This is not specific to this PR, but I do wonder if we want to preface/namespace our demo env vars to make them easier for the user to clean up? or should that be another cli operation? (remove demo)

internal/cmd/gocdk/_static/demo/pubsub/demo_pubsub.go
```diff
@@ -0,0 +1,171 @@
...
+
+	pubsubSendTemplate = pubsubTemplatePrefix + `
+  {{if .Success }}
+    <p>Message sent!</p>
+  {{end}}
+  <p>
+  <form>
+    Enter a message to send to the topic:
+    <br/>
+    <textarea rows="4" cols="50" name="msg">{{ .Msg }}</textarea>
```
clausti: Can we put a `<label>` on this?

internal/cmd/gocdk/_static/demo/pubsub/demo_pubsub.go
vangent: IIUC, the environment variables will be set based on the `outputs.tf`, so they won't have to clean them up. We might still want to namespace them though. I suspect this will come up with we start setting them based on `outputs.tf`.

internal/cmd/gocdk/_static/demo/pubsub/demo_pubsub.go
vangent: Oops, done.

vangent:
PTAL

PR #2073, opened 2019-05-14
docstore: check for field path prefixes in Mods

jba:


internal/docstore/docstore.go
```diff
@@ -314,28 +314,50 @@ func toDriverMods(mods Mods) ([]driver.Mod, error) {
...
+	var dmods []driver.Mod
+	for _, k := range keys {
 		k := FieldPath(k)
 		v := mods[k]
 		fp, err := parseFieldPath(k)
 		if err != nil {
 			return nil, err
 		}
-		dmods[i] = driver.Mod{FieldPath: fp, Value: v}
+		for _, d := range dmods {
```
shantuo: One suggestion, you can do the whole prefix check process in the beginning of this function, I think that way looks more clear as you are just looping through the keys of mods and trying to find all possible prefix and then append them to `dmods` later.

jba: It's hard to find the prefixes before I parse the keys into field paths. For example, `a` is a prefix of `ab` and `a.b`, but only the second is wrong.

PR #2075, opened 2019-05-14
all: go mod tidy on all modules, go install apidiff in tempdir

vangent:
This PR is mostly just running `go mod tidy` on all of our modules. I believe some dependencies in the core module (e.g., `subcommands`) were removable because of turning `samples` into a module.

I am not sure why so many dependencies were removed from the other `go.mod` files.

I am also not sure why the `apidiff` install worked before this without trying to modify the `go.mod` file to add the `golang.org/x/exp/cmd/apidiff` dependencies the way it does in the Wire repo. Maybe there's an indirect dependency on it somewhere? But `go mod why` says it's not used... At any rate, this change makes it safer.


internal/testing/check_api_change.sh
```diff
@@ -31,19 +31,25 @@ set -euo pipefail
...
 function cleanup() {
+  rm -rf "$INSTALL_DIR"
   rm -rf "$MASTER_CLONE_DIR"
   rm -f "$PKGINFO_BRANCH"
   rm -f "$PKGINFO_MASTER"
 }
 trap cleanup EXIT
 
+# Move to a temporary directory while installing apidiff to avoid changing
+# the local .mod file.
```
eliben: Nit: point to most relevan golang issue here, so we can track if this ever gets resolved?

internal/testing/check_api_change.sh
```diff
@@ -31,19 +31,25 @@ set -euo pipefail
...
+pushd /tmp &> /dev/null
```
eliben: You should be able to use a subcommand, same as we do for runnign tests in submodules, without push/pop

internal/testing/check_api_change.sh
vangent: I can't find one that is open :-(.

internal/testing/check_api_change.sh
vangent: Done.

PR #2078, opened 2019-05-14
blob: update docs/examples for PrefixedBucket

vangent:
Updates #909.


internal/website/content/howto/blob/open-bucket.md
```diff
@@ -26,6 +26,18 @@ both forms for each storage provider.
 ["blank import"]: https://golang.org/doc/effective_go.html#blank_import
 [documentation on URLs]: {{< ref "/concepts/urls.md" >}}
 
+## Prefixed Buckets {#prefix}
```
zombiezen: Optional: This seems like more of an advanced topic, consider moving to the end.

vangent: I'll make it a h3. I don't think anyone will notice it at the end after all of the provider-specific stuff.

PR #2100, opened 2019-05-15
internal/docstore/mongodocstore: add BeforeQuery and BeforeDo

### LINKED ISSUE #1675: docstore: support As for action lists and queries

shantuo:
mongo is a little bit different, it doesn't have special types for the inputs, but one can use the corresponding options types to do similar things like other drivers.

Fixes #1675

### DIFF
--- a/internal/docstore/mongodocstore/query.go
+++ b/internal/docstore/mongodocstore/query.go
@@ -42,6 +42,19 @@ func (c *collection) RunGetQuery(ctx context.Context, q *driver.Query) (driver.D
 		}
 		filter = append(filter, bf)
 	}
+	if q.BeforeQuery != nil {
+		asFunc := func(i interface{}) bool {
+			p, ok := i.(**options.FindOptions)
+			if !ok {
+				return false
+			}
+			*p = opts
+			return true
+		}
+		if err := q.BeforeQuery(asFunc); err != nil {
+			return nil, fmt.Errorf("Find: %v", err)
+		}
+	}
 	cursor, err := c.coll.Find(ctx, filter, opts)
 	if err != nil {
 		return nil, fmt.Errorf("Find: %v", err)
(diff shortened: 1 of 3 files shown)


internal/docstore/mongodocstore/query.go
```diff
@@ -42,6 +42,19 @@ func (c *collection) RunGetQuery(ctx context.Context, q *driver.Query) (driver.D
...
+		asFunc := func(i interface{}) bool {
+			p, ok := i.(**options.FindOptions)
+			if !ok {
+				return false
+			}
+			*p = opts
+			return true
+		}
+		if err := q.BeforeQuery(asFunc); err != nil {
+			return nil, fmt.Errorf("Find: %v", err)
```
jba: We shouldn't be using fmt.Errorf here. It hides the error. Just return `err`.

Same on line 60.

shantuo: Done.

PR #2102, opened 2019-05-15
internal/docstore: benchmark action list

### LINKED ISSUE #1882: docstore: write conformance benchmarks
We are doing some level of optimization work now while we are trying to get all drivers work properly under the hood. It will be useful to add some benchmarks to give us some insights. Places like query, actionlist (order and unorder) are worth considering.

shantuo:
Fixes #1882


internal/docstore/drivertest/drivertest.go
```diff
@@ -1067,10 +1075,8 @@ func filterHighScores(hs []*HighScore, f func(*HighScore) bool) []*HighScore {
 }
 
 // cleanUpTable delete all documents from this collection after test.
-func cleanUpTable(t *testing.T, coll *docstore.Collection) {
-	if err := coll.Query().Delete(context.Background()); err != nil {
-		t.Fatalf("%+v", err)
-	}
+func cleanUpTable(coll *docstore.Collection) error {
```
jba: You can put back the first argument as 
```
interface { Fatalf(string, ...interface{}) }
```

jba: Another idea I had is to put a call in with `withCollection` and `withTwoKeyCollection`, before we call the function. That would mean most of the tests wouldn't need to call it.

shantuo: Done.

PR #2105, opened 2019-05-15
travis: add a check to see if go mod tidy has been run

### LINKED ISSUE #2097: travis: add a presubmit with "go mod tidy"

vangent:
Fixes #2097.


internal/testing/runchecks.sh
```diff
@@ -174,15 +181,18 @@ echo "Running wire diff..."
...
 while read -r path || [[ -n "$path" ]]; do
   echo
   echo "Testing sub-module at '$path'..."
-  echo "  Go tests:"
+  echo "  go test:"
   ( cd "$path" && exec go test -mod=readonly ./... ) || result=1
+  echo "  go mod tidy:"
+  ( cd "$path" && check_mod_tidy.sh && echo "    OK" ) || { echo "FAIL: run go mod tidy on sub-module at '$path'" && result=1; }
   echo "  wire check:"
-  ( cd "$path" && exec wire check ./... && echo "  OK" ) || result=1
```
vangent: This was broken before; the `echo OK` never ran because of the exec.

internal/testing/runchecks.sh
```diff
@@ -174,15 +181,18 @@ echo "Running wire diff..."
 
 echo
 echo "Running Go tests for sub-modules..."
+PATH="$PATH":"${PWD}"/internal/testing
```
eliben: I wouldn't do this just to avoid prepending the dir to check_mod_tidy. Prefer explicit - also we do this elsewhere - starting local deps, listdeps, etc.

check_mod_tidy can take a dir and cd into it?

internal/testing/runchecks.sh
eliben: But why is the "exec" used for "go test" then? Comment on the difference

vangent: @zombiezen ?

IIUC, the `exec` is just for efficiency, to avoid forking another process.

internal/testing/check_mod_tidy.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# Make copies of the current files.
+cp ./go.mod "$TMP_GOMOD"
+cp ./go.sum "$TMP_GOSUM"
+
+# Modifies the files in-place.
+go mod tidy
+
+# Check for diffs.
+diff "$TMP_GOMOD" ./go.mod
```
zombiezen: nit: I prefer `-u` (unified) diff output

internal/testing/runchecks.sh
zombiezen: Yes. It's probably safer to just remove them at this point.

internal/testing/runchecks.sh
vangent: Done.

internal/testing/check_mod_tidy.sh
vangent: Done, for all of our diffs.

internal/testing/runchecks.sh
vangent: Done.

vangent:
> LGTM. Be sure to run your changes against shellcheck.

Yup, done. I fixed one thing it turned up in another shell script.

eliben:
It's strange it fails on the change I made to contributebot's go.mod, I'm pretty sure I did run tidy. Mysterious modules...

vangent:
> It's strange it fails on the change I made to contributebot's go.mod, I'm pretty sure I did run tidy. Mysterious modules...

It is failing because we call `go list` earlier, and `go list` changes the `go.mod` file with something that `go mod tidy` will remove. Sigh.

vangent:
PTAL both of you.

Due to the fact that `go list` and `wire` can both modify `go.mod` and/or `go.sum`, I've changed this to use the result of `go mod tidy && go list -deps ./...` as the canonical form for what we expect to be in `go.mod`/`go.sum`.

The `check_mod_tidy.sh` script runs pretty early on (after `go test` but before anything else, both for the root module and sub-modules), and `go test` has a `-mod=readonly` flag, so we can be sure that there haven't been changes before we run `check_mod_tidy.sh`. It makes a backup copy, then runs `go mod tidy && go list -deps ./...` and makes sure nothing changed.

I also added a `gomodcleanup.sh` that can be run from the root of the repo, and does the same process (`go mod tidy && go list -deps ./...`) on all modules.

internal/testing/check_mod_tidy.sh
```diff
@@ -0,0 +1,62 @@
...
+}
+trap cleanup EXIT
+
+# Make copies of the current files.
+cp ./go.mod "$TMP_GOMOD"
+cp ./go.sum "$TMP_GOSUM"
+
+# Modifies the files in-place.
+go mod tidy
+go list -deps ./... &> /dev/null || :
```
eliben: Nit: can you add a comment to explain the `|| :`? This is super cryptic

internal/testing/runchecks.sh
```diff
@@ -106,6 +107,13 @@ else
 fi;
 
 
+echo
+echo "Ensuring that go mod tidy has been run for the root module..."
```
eliben: Sidenote: it's annoying that runcheks does $THINGS for the root module and then repeats them for submodules. The inherent difference seems to be that we're running some extra stuff for root (like coverage). Perhaps there's a cleaner way to do that all in one loop. Just thinking aloud for the future

internal/testing/check_mod_tidy.sh
vangent: Actually, it's not needed anymore. `go list -deps` fails for some modules, but adding the `./...` makes it always work, and this `|| :` is to make it not fail.

Removed.

internal/testing/runchecks.sh
vangent: Ack.

vangent: We could probably move the "loop over modules" stuff to the top, and do `go test` + `wire` etc. in that loop, with a conditional for coverage. Although, we probably want coverage for more than the root module... e.g., we'll want it for the driver submodules, and for `gocdk`.

vangent: Filed https://github.com/google/go-cloud/issues/2117.

PR #2106, opened 2019-05-15
internal/cmd/gocdk: add Cloud Run launcher

### LINKED ISSUE #1861: internal/cmd/gocdk: implement minimal launch command

zombiezen:
Fixes #1861

### DIFF
--- a/internal/cmd/gocdk/launch.go
+++ b/internal/cmd/gocdk/launch.go
@@ -21,13 +21,19 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"net/http"
 	"net/url"
 	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
+	"gocloud.dev/gcp"
 	"golang.org/x/xerrors"
+	"google.golang.org/api/googleapi"
+	"google.golang.org/api/option"
+	cloudrun "google.golang.org/api/run/v1alpha1"
 )
 
 func launch(ctx context.Context, pctx *processContext, args []string) error {
@@ -124,6 +130,22 @@ func newLauncher(ctx context.Context, pctx *processContext, launcherName string)
 			dockerEnv: pctx.env,
 			dockerDir: pctx.workdir,
 		}, nil
+	case "cloudrun":
+		creds, err := pctx.gcpCredentials(ctx)
+		if err != nil {
+			return nil, xerrors.Errorf("prepare cloudrun launcher: %w", err)
+		}
+		httpClient, _ := gcp.NewHTTPClient(http.DefaultTransport, creds.TokenSource)
+		runService, err := cloudrun.NewService(ctx, option.WithHTTPClient(&httpClient.Client))
+		if err != nil {
+			return nil, xerrors.Errorf("prepare cloudrun launcher: %w", err)
+		}
+		return &cloudRunLauncher{
+			logger:    logger,
+			client:    runService,
+			dockerEnv: pctx.env,
+			dockerDir: pctx.workdir,
+		}, nil
 	default:
 		return nil, xerrors.Errorf("prepare launcher: unknown launcher %q", launcherName)
 	}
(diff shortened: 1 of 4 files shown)


internal/cmd/gocdk/launch.go
```diff
@@ -124,6 +130,22 @@ func newLauncher(ctx context.Context, pctx *processContext, launcherName string)
...
+		if err != nil {
+			return nil, xerrors.Errorf("prepare cloudrun launcher: %w", err)
+		}
+		httpClient, _ := gcp.NewHTTPClient(http.DefaultTransport, creds.TokenSource)
+		runService, err := cloudrun.NewService(ctx, option.WithHTTPClient(&httpClient.Client))
+		if err != nil {
+			return nil, xerrors.Errorf("prepare cloudrun launcher: %w", err)
+		}
+		return &cloudRunLauncher{
+			logger:    logger,
```
vangent: Kind of same question as before: if `logger`, `dockerEnv`, and `dockerDir` are going to be needed by every launcher, why not pass them in `LaunchInput`?

internal/cmd/gocdk/launch.go
```diff
@@ -187,6 +209,172 @@ func (local *localLauncher) Launch(ctx context.Context, input *LaunchInput) (*ur
...
+	dockerDir string
+}
+
+// Launch implements Launcher.Launch.
+func (crl *cloudRunLauncher) Launch(ctx context.Context, input *LaunchInput) (*url.URL, error) {
+	projectID := specifierStringValue(input.Specifier, "project_id")
+	location := specifierStringValue(input.Specifier, "location")
+	serviceName := specifierStringValue(input.Specifier, "service_name")
+	if projectID == "" || location == "" || serviceName == "" {
+		return nil, xerrors.New("cloud run launch: specifier does include project_id, location, and/or service_name")
```
vangent: does -> missing ?

vangent: And maybe be more specific than `specifier`; IIUC, this is the `launch_specifier` section in `outputs.tf`?

internal/cmd/gocdk/launch.go
```diff
@@ -187,6 +209,172 @@ func (local *localLauncher) Launch(ctx context.Context, input *LaunchInput) (*ur
...
+	}
+
+	// Push to GCR if needed.
+	// TODO(light): Check for presence in Docker daemon and country-specific
+	// prefixes.
+	if strings.HasPrefix(input.DockerImage, "gcr.io/") {
```
vangent: If it doesn't start with `gcr.io`, what does that mean? Is the rest of this going to work?

internal/cmd/gocdk/launch.go
```diff
@@ -187,6 +209,172 @@ func (local *localLauncher) Launch(ctx context.Context, input *LaunchInput) (*ur
...
+			return nil, xerrors.Errorf("cloud run launch: %w", err)
+		}
+	}
+
+	// Launch on Cloud Run.
+	var env []*cloudrun.EnvVar
+	for i, inputVar := range input.Env {
+		eqIdx := strings.IndexByte(inputVar, '=')
+		if eqIdx == -1 {
+			return nil, xerrors.Errorf("cloud run launch: bad format for specifier env[%d] (got %q)", i, inputVar)
```
vangent: Same here about being more specific about where this is coming from. And tell them what the format should be, not just that it's bad.

internal/cmd/gocdk/launch.go
```diff
@@ -187,6 +209,172 @@ func (local *localLauncher) Launch(ctx context.Context, input *LaunchInput) (*ur
...
+		}
+		env = append(env, &cloudrun.EnvVar{
+			Name:  inputVar[:eqIdx],
```
vangent: Nit: you could use `strings.Split` for this instead.

internal/cmd/gocdk/launch.go
```diff
@@ -187,6 +209,172 @@ func (local *localLauncher) Launch(ctx context.Context, input *LaunchInput) (*ur
...
+			Value: inputVar[eqIdx+1:],
+		})
+	}
+	serviceMeta := &cloudrun.ObjectMeta{
+		Name:      serviceName,
+		Namespace: projectID,
+	}
+	serviceSpec := &cloudrun.ServiceSpec{
```
vangent: Consider add a link here to where this is documented....

internal/cmd/gocdk/launch.go
```diff
@@ -187,6 +209,172 @@ func (local *localLauncher) Launch(ctx context.Context, input *LaunchInput) (*ur
...
+		// Handle error below in the same way a create call is handled.
+		_, err = replaceCall.Context(ctx).Do()
+	} else if err == nil {
+		// Make publicly accessible.
+		policyCall := crl.client.Projects.Locations.Services.SetIamPolicy(serviceString, &cloudrun.SetIamPolicyRequest{
+			Policy: &cloudrun.Policy{
+				Bindings: []*cloudrun.Binding{
+					{
+						Role:    "roles/run.invoker",
+						Members: []string{"allUsers"},
```
vangent: Should we make this be configured in `launch_specifier`?

internal/cmd/gocdk/launch.go
```diff
@@ -187,6 +209,172 @@ func (local *localLauncher) Launch(ctx context.Context, input *LaunchInput) (*ur
...
+		// Already exists, add revision.
+		replaceCall := crl.client.Projects.Locations.Services.ReplaceService(serviceString, &cloudrun.Service{
+			ApiVersion: "serving.knative.dev/v1alpha1",
+			Kind:       "Service",
+			Metadata:   serviceMeta,
+			Spec:       serviceSpec,
+		})
+		// Handle error below in the same way a create call is handled.
+		_, err = replaceCall.Context(ctx).Do()
+	} else if err == nil {
```
vangent: Nit: I found this error flow a bit hard to follow. How about flipping the if/else?

```
err := createCall.Do()
if err == nil {
  // set IAM
  err = policyCall.Do()
} else if err == StatusConflict {
  // Add revision
  err = replaceCall.Do()
}

if err != nil {
  // either the create failed, or the policy call failed, or the replace failed
  return err
}

...
```

internal/cmd/gocdk/launch.go
```diff
@@ -187,6 +209,172 @@ func (local *localLauncher) Launch(ctx context.Context, input *LaunchInput) (*ur
...
+		// Make publicly accessible.
+		policyCall := crl.client.Projects.Locations.Services.SetIamPolicy(serviceString, &cloudrun.SetIamPolicyRequest{
```
vangent: IIUC, we only do this the first time the service is launched? So if the user changes it, we won't clobber their change?

internal/cmd/gocdk/launch.go
```diff
@@ -187,6 +209,172 @@ func (local *localLauncher) Launch(ctx context.Context, input *LaunchInput) (*ur
...
+	tick := time.NewTicker(5 * time.Second)
+	defer tick.Stop()
+	for {
+		select {
+		case <-tick.C:
+		case <-ctx.Done():
+			return nil, xerrors.Errorf("cloud run launch: wait for ready: %w", ctx.Err())
+		}
+
+		currService, err := crl.client.Projects.Locations.Services.Get("projects/" + projectID + "/locations/" + location + "/services/" + serviceName).Context(ctx).Do()
```
vangent: Don't we have this as `serviceString` already?

internal/cmd/gocdk/launch.go
```diff
@@ -187,6 +209,172 @@ func (local *localLauncher) Launch(ctx context.Context, input *LaunchInput) (*ur
...
+		if len(out) == 0 {
+			return xerrors.Errorf("docker push: %w", err)
+		}
+		return xerrors.Errorf("docker push:\n%s", out)
+	}
+	return nil
+}
+
+// conditionStatus finds the Cloud Run condition with the given name and returns
+// its status string (one of "True", "False", or "Unknown").
```
vangent: ... or the empty string, right?

internal/cmd/gocdk/launch.go
```diff
@@ -187,6 +209,172 @@ func (local *localLauncher) Launch(ctx context.Context, input *LaunchInput) (*ur
...
+			if err != nil {
+				return nil, xerrors.Errorf("cloud run launch: parse service URL: %w", err)
+			}
+			return u, nil
+		}
+		crl.logger.Println("Not ready yet, will poll again...")
+	}
+}
+
+func (crl *cloudRunLauncher) push(ctx context.Context, imageName string) error {
```
vangent: Nit: maybe dockerPush?

internal/cmd/gocdk/launch.go
```diff
@@ -187,6 +209,172 @@ func (local *localLauncher) Launch(ctx context.Context, input *LaunchInput) (*ur
...
+func conditionStatus(conds []*cloudrun.ServiceCondition, name string) string {
+	for _, c := range conds {
+		if c.Type == name {
+			return c.Status
+		}
+	}
+	return ""
+}
+
+// launchSpecifier is the set of arguments passed from a biome's Terraform
```
vangent: zombie type :-)

internal/cmd/gocdk/launch.go
zombiezen: Done.

internal/cmd/gocdk/launch.go
zombiezen: Done.

internal/cmd/gocdk/launch.go
zombiezen: Done.

internal/cmd/gocdk/launch.go
zombiezen: Yes. Fixed.

internal/cmd/gocdk/launch.go
zombiezen: Done.

internal/cmd/gocdk/launch.go
zombiezen: Done.

internal/cmd/gocdk/launch.go
vangent: SGTM.

internal/cmd/gocdk/launch.go
vangent: SGTM.

internal/cmd/gocdk/launch.go
vangent: OK.

PR #2109, opened 2019-05-15
internal/docstore/mongodocstore: count bulkWrite results respectively

### LINKED ISSUE #2107: docstore/mongodocstore: matched number doesn't seem reliable
### Describe the bug

I'm getting errors like `some replace failed (replaced 25 out of 0)` when doing unordered put actions.

### To Reproduce

See #2102 

From some local debugging, I see items are correctly put into the collection, but I still get errors on those situations.

## Expected behavior

There shouldn't be error if the items are correctly put into the collection.

### Additional context

I also examined the `UpsertedCount` when doing 100 put action lists, and got results like

```
    driverbenchmark.go:109: at -1: some replace failed (replaced 13 out of 0), Upserted 87 (code=NotFound)
    driverbenchmark.go:109: at -1: some replace failed (replaced 47 out of 0), Upserted 53 (code=NotFound)
    driverbenchmark.go:109: at -1: some replace failed (replaced 1 out of 0), Upserted 99 (code=NotFound)
    driverbenchmark.go:109: at -1: some replace failed (replaced 20 out of 0), Upserted 80 (code=NotFound)
    driverbenchmark.go:109: at -1: some replace failed (replaced 17 out of 0), Upserted 83 (code=NotFound)
    driverbenchmark.go:109: at -1: some replace failed (replaced 1 out of 0), Upserted 99 (code=NotFound)
    driverbenchmark.go:109: at -1: some replace failed (replaced 1 out of 0), Upserted 99 (code=NotFound)
```

shantuo:
Fixes #2107


internal/docstore/mongodocstore/mongo.go
```diff
@@ -649,22 +652,31 @@ func (c *collection) unorderedBulkWrite(ctx context.Context, iactions []indexedA
 			if newID != nil {
 				newIDs[i] = newID
 			}
+			if err == nil {
```
jba: I don't think we need this, because I think a create will return an error if the key already exists.

internal/docstore/mongodocstore/mongo.go
```diff
@@ -687,13 +699,21 @@ func (c *collection) unorderedBulkWrite(ctx context.Context, iactions []indexedA
 		}
 		return alerr
 	}
+	if res.InsertedCount != nCreate {
```
jba: See above. I believe that if InsertedCount != nCreate, we would have gotten an error back from the RPC. Do you have a test case that shows otherwise?

internal/docstore/mongodocstore/mongo.go
```diff
@@ -649,22 +652,31 @@ func (c *collection) unorderedBulkWrite(ctx context.Context, iactions []indexedA
...
+		case driver.Replace:
+			m, err = c.newReplaceModel(a, false)
+			if err == nil {
+				nNonCreateWrite++
+				nModified++
+			}
+		case driver.Put:
+			m, err = c.newReplaceModel(a, true)
+			if err == nil {
+				nNonCreateWrite++
```
jba: Are you sure that a Put to an existing document doesn't show up as a modification? (Is there a test for that?)

internal/docstore/mongodocstore/mongo.go
shantuo: Added a test.

jba:
Remember to write PTAL or I may miss changes.

PR #2112, opened 2019-05-15
all: add link to go reportcard, fix various things it flagged

vangent:
https://goreportcard.com/report/github.com/google/go-cloud

I fixed a bunch of the non-controversial (I think) things it flagged like misspellings and ineffective assignments (it found a few bugs...).

Its `go vet` complaints seem wrong to me.


internal/docstore/firedocstore/fs.go
```diff
@@ -427,9 +427,9 @@ func (c *collection) actionToWrites(a *driver.Action) ([]*pb.Write, string, erro
 
 	case driver.Replace:
 		// If the given document has a revision, use it as the precondition (it implies existence).
-		pc, err := revisionPrecondition(a.Doc)
```
vangent: This was definitely a subtle bug right here.

Issue #2111, opened 2019-05-15
allmodules: mark modules into categories


eliben:
Right now we have an `allmodules` file that just lists all the modules (directories where `go.mod` resides):

```
.
internal/cmd/gocdk
internal/contributebot
internal/website
samples
```

A couple of scripts use these lists of modules to perform tasks "for all modules in the repo".

As part of the work on #886, we'll need tooling to help with the release process. This tooling will need to distinguish between several categories of modules:

1. The root module (admittedly, it's easy as it's the only one that is `.`)
2. Submodules that we plan to release/track, such as separate providers split out from root as part of #886 
3. Samples modules that import all the other released modules
4. Modules we use only internally and don't release. These don't need tagging

For example, prior to a release, a tool will have to add `replace` lines to all released modules that depend on root so they can be tested with the newest version. Same for `samples` depending on root and other released modules. Et cetera

PR #2114, opened 2019-05-15
testing: initial code for release helper tool

eliben:
Initial release helper tool for #2111. Doesn't help with subsequent releases yet - will add that functionality as we go.


internal/testing/releasehelper.go
```diff
@@ -0,0 +1,104 @@
...
+)
+
+var doReplace = flag.Bool("replace", false, "add replace directives to root")
+var doDropReplace = flag.Bool("dropreplace", false, "drop replace directives to root")
+
+func cmdCheck(s string) []byte {
+	fmt.Printf(" -> %s\n", s)
+	fields := strings.Fields(s)
+	if len(fields) < 1 {
+		log.Fatal("Expected \"command <arguments>\"")
```
jba: `Expected "command <arguments>"`

internal/testing/releasehelper.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+	b, err := exec.Command(fields[0], fields[1:]...).Output()
+	if exiterr, ok := err.(*exec.ExitError); ok {
+		log.Fatalf("%s; stderr: %s\n", err, string(exiterr.Stderr))
+	}
```
jba: Handle other non-nil `err`s.

internal/testing/releasehelper.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, r := range modInfo.Require {
+		if r.Path == "gocloud.dev" {
+			// Find relative path needed for the replace directive.
+			// TODO(eliben): handle paths to submodules here too, not only root
+			fmt.Println("Found reference to gocloud.dev")
+			rel, err := filepath.Rel(filepath.Dir(path), ".")
+			if err != nil {
+				log.Fatal(err)
+			}
+			dir, _ := filepath.Split(rel)
```
jba: `filepath.Dir`?

internal/testing/releasehelper.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	input := bufio.NewScanner(f)
+	input.Split(bufio.ScanLines)
+	for input.Scan() {
+		if len(input.Text()) > 0 && !strings.HasPrefix(input.Text(), "#") {
+			runOnGomod(path.Join(input.Text(), "go.mod"))
+		}
+	}
+}
```
jba: after the loop: 
```
if input.Err() != nil ...
```

internal/testing/releasehelper.go

internal/testing/releasehelper.go
eliben: Not sure what you mean?

internal/testing/releasehelper.go

internal/testing/releasehelper.go

eliben:
There's enough functionality now to help with the initial split release. Will add more flags and goodies later as we go along and learn the ropes of juggling multiple modules

internal/testing/releasehelper.go
jba: Ah, I see what happened here!

I suggested that you use the Go backtick string quotes, so you wouldn't have to escape the double quotes. But of course Markdown interpreted my backticks. Let me try again:
```
`Expected "command <arguments>"`
```

internal/testing/releasehelper.go
```diff
@@ -0,0 +1,143 @@
...
+	"os"
+	"os/exec"
+	"path"
+	"path/filepath"
+	"strings"
+)
+
+// TODO(eliben): add bumpversion flag that bumps version by 1, or to a version
+// given on the command line
+var doReplace = flag.Bool("replace", false, "add replace directives to root")
```
vangent: These file like they should be commands, not flags. In particular, it doesn't make sense to set both flags to `true`.

internal/testing/releasehelper.go
```diff
@@ -0,0 +1,143 @@
...
+	for _, r := range modInfo.Require {
+		// Find requirements on modules within the gocloud.dev tree.
+		if strings.HasPrefix(r.Path, "gocloud.dev") {
+			// Find the relative path from 'path' (the module file we're processing)
+			// and the module required here.
+			var reqPath string
+			if r.Path == "gocloud.dev" {
+				reqPath = "."
+			} else {
+				reqPath = r.Path[12:]
```
vangent: Alternatively, something like this:

```
if reqPath := strings.TrimPrefix(r.Path, "gocloud.dev"); reqPath != r.Path {
  if reqPath == "" {
    reqPath = "."
  }
  ...
}
```

internal/testing/releasehelper.go
```diff
@@ -0,0 +1,143 @@
...
+	rawJson := cmdCheck("go mod edit -json " + path)
+	var modInfo GoMod
+	err := json.Unmarshal(rawJson, &modInfo)
+	if err != nil {
+		log.Fatal(err)
+	}
+	return modInfo
+}
+
+func runOnGomod(path string) {
```
vangent: Consider passing the module path instead of the full path to the `go.mod` file? It's easy to construct the path to the `go.mod` here instead, and I think it's clearer to do it that way than to call `filepath.Dir(path)` down below.

E.g.,

```
func runOnModule(modulepath string) {
  gomodpath = filepath.Join(modulepath, "go.mod")
  ...
```

internal/testing/releasehelper.go
```diff
@@ -0,0 +1,143 @@
...
+			if r.Path == "gocloud.dev" {
+				reqPath = "."
+			} else {
+				reqPath = r.Path[12:]
+			}
+			rel, err := filepath.Rel(filepath.Dir(path), reqPath)
+			if err != nil {
+				log.Fatal(err)
+			}
+			if strings.HasSuffix(rel, "/.") {
```
vangent: Can you add a comment for this, it's not obvious what it's doing.

internal/testing/releasehelper.go
```diff
@@ -0,0 +1,143 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Helper tool for creating new releases of the Go CDK.
+// Run from the repository root:
+// $ go run internal/testing/releasehelper.go <flags>
```
vangent: Can you expand this help a bit (or, probably better, add `help` to the command itself)? IIUC, `replace` is going to add `replace` directives to all submodules, pointing `gocloud.dev` to `../..` or whatever. "Add replace directives to root" makes it sound like it will modify root, instead of the inverse.

internal/testing/releasehelper.go
```diff
@@ -0,0 +1,143 @@
...
+package main
```
vangent: Add a simple test?

internal/testing/releasehelper.go
eliben: Done, nice catch

internal/testing/releasehelper.go
eliben: There could be other flags that can be shared with `-replace`, such as bumping dep version (will be needed for future releases). Currently replace wins over remove, and I can just add an error if both are given. Flags are much simpler to add/tweak than getting into commands

eliben: Added a check

internal/testing/releasehelper.go
eliben: Ack. Mine feels a bit simpler, so I'll stick with it if you don't mind

internal/testing/releasehelper.go

internal/testing/releasehelper.go

internal/testing/releasehelper.go

internal/testing/releasehelper.go
eliben: I'm not sure it's worth it. To really test it I'll need to recreate a small vistual file system since it invokes the go tool to actually modify files, etc. This test is going to be much more complicated than the tool itself. WDYT?

internal/testing/releasehelper.go
vangent: OK. I don't really find the `12:` that readable, maybe `TrimPrefix` there?

internal/testing/releasehelper.go
vangent: OK for now; may need to revisit as it gets more complicated.

internal/testing/releasehelper.go
```diff
@@ -0,0 +1,169 @@
...
+	"flag"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+// TODO(eliben): add bumpversion flag that bumps version by 1, or to a version
```
vangent: I have a feeling that when we add these it will be better to use commands rather than flags. Either commands that are high level like `prep` which package up a bunch of lower-level things, or low-level commands that you do one at a time (`releasehelper bumpversion`, `releasehelper addreplace`). Otherwise it may be tricky to define the semantics for various combinations of flags (e.g., these two existing flags already depend on each other in some way; it doesn't make sense to set both or neither. With more flags, that may get more complex).

But, we can revisit later.

internal/testing/releasehelper.go
eliben: Good idea, done

internal/testing/releasehelper.go
eliben: Yes, I see what you're saying, been thinking about this. It's hard to predict what future functionailty we'll want though and I don't want to over-YAGNI it too much right now. We'll enhance it as we go and learn the ropes of the new process

PR #2119, opened 2019-05-16
docstore: support queries on the key field

jba:
It should be valid to write a Where clause that uses the key field of the
collection.

This requires special treatment in Firestore.

Also made other minor tweaks to drivertest.


internal/docstore/firedocstore/query.go
```diff
@@ -277,7 +278,28 @@ func splitFilters(fs []driver.Filter) (sendToFirestore, evaluateLocally []driver
 	return sendToFirestore, evaluateLocally
 }
 
-func filterToProto(f driver.Filter) (*pb.StructuredQuery_Filter, error) {
+// // Convert filters on the key field to filters on "__name__".
```
shantuo: Delete these.

Issue #2110, opened 2019-05-15
testing: helper tool to summarize test results


eliben:
Followup on #2039 to track the final state we'd like to be in.

The default `go test` doesn't report how many tests were skipped, and we had some issues before with tests being skipped when they shouldn't have been. Therefore, we'd like to have a helper tool that will summarize test results.

Instead of running `go test` in our modules, we'll run a script which will invoke `go test -json` piped into a program that:

1. Incrementally reads stdin to produce similar output to `go test` - we want to see progress, which test executes, how long it took, outputs from tests, etc.
2. At the end, produces a summary specifying how many tests ran, skipped, passed, failed

eliben:
@jba assigning to you, since you mentioned in #2039 you want to keep working on it. I can do that too, if you're overloaded with other stuff.

jba:
I'll do it after I'm done preparing tomorrow's talk.

vangent:
This is done and working nicely.

PR #2121, opened 2019-05-16
test-summary: produce more output

jba:
Flags produce either full output (-vv), or one line per test (-v).

Suggest better flag names!

See #2110.


internal/testing/test-summary/test-summary.go
```diff
@@ -51,12 +58,35 @@ func main() {
 func run(r io.Reader) (msg string, failures bool, err error) {
 	counts := map[string]int{}
 	scanner := bufio.NewScanner(bufio.NewReader(r))
+	prevPkg := ""
```
eliben: Add a comment to explain why tracking `prevPkg` is needed

PR #2124, opened 2019-05-16
rabbitpubsub: use timer in ReceiveBatch, not ticker

### LINKED ISSUE #2103: pubsub: cost too many CPU
When I started my subscriber service, My computer's CPU usage is close to 100%.

I foud this code(https://github.com/google/go-cloud/blob/master/pubsub/pubsub.go#L583) will create goroutine continuously.

I don't think need to create goroutine in this case, And can reduce CPU consumption.

jba:
We were creating a ticker on each call to ReceiveBatch, but we weren't
stopping it.  Also, we only care about the first channel send (if
any). So switch to a timer instead.

In informal CPU profiling of a simplified program, this helped
significantly with CPU usage.

Fixes #2103.

### DIFF
--- a/pubsub/rabbitpubsub/rabbit.go
+++ b/pubsub/rabbitpubsub/rabbit.go
@@ -576,7 +576,7 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*dr
 
 	// Get up to maxMessages waiting messages, but don't take too long.
 	var ms []*driver.Message
-	maxTime := time.Tick(50 * time.Millisecond)
+	maxTime := time.NewTimer(50 * time.Millisecond)
 	for {
 		select {
 		case <-ctx.Done():
@@ -609,7 +609,7 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*dr
 				return ms, nil
 			}
 
-		case <-maxTime:
+		case <-maxTime.C:
 			// Timed out. Return whatever we have. If we have nothing, we'll get
 			// called again soon, but returning allows us to give up the lock.
 			return ms, nil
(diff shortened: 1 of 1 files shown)


pubsub/rabbitpubsub/rabbit.go
```diff
@@ -576,7 +576,7 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*dr
 
 	// Get up to maxMessages waiting messages, but don't take too long.
 	var ms []*driver.Message
-	maxTime := time.Tick(50 * time.Millisecond)
+	maxTime := time.NewTimer(50 * time.Millisecond)
```
vangent: Nice catch, thanks.

It doesn't look like we need the `for` loop here anymore; all 3 of the `select` clauses have a `return`. Can we get rid of the `for` loop, delete `maxTime` and use `time.After` in the last `select` clause, and move the `// Timed out` comment + `return` out of the `select`?

pubsub/rabbitpubsub/rabbit.go
```diff
@@ -609,7 +609,7 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*dr
 				return ms, nil
 			}
 
-		case <-maxTime:
+		case <-maxTime.C:
 			// Timed out. Return whatever we have. If we have nothing, we'll get
```
vangent: Also, can you update this comment to:

```
	// Timed out. Return whatever we have. If we have nothing, we'll get
	// called again soon, but returning allows us to give up the lock in
    // case there are acks/nacks to be sent.
```

pubsub/rabbitpubsub/rabbit.go
jba: The middle case doesn't always return (if `ok` and `len(ms) < maxMessages`)

vangent: Ah, you are right, that is subtle.

PR #2125, opened 2019-05-16
internal/cmd/gocdk: add success text to init

zombiezen:


internal/cmd/gocdk/init.go
```diff
@@ -81,6 +82,16 @@ func init_(ctx context.Context, pctx *processContext, args []string) error {
...
+	fmt.Fprintf(pctx.stdout, "Project created at %s with:\n", projectDir)
+	fmt.Fprintln(pctx.stdout, "- Go HTTP server")
+	fmt.Fprintln(pctx.stdout, "- Dockerfile")
+	fmt.Fprintln(pctx.stdout, "- 'dev' biome for local development settings")
+	fmt.Fprintln(pctx.stdout)
+	fmt.Fprintf(pctx.stdout, "Run `cd %s`, then run:\n", f.Arg(0))
+	fmt.Fprintln(pctx.stdout, "- `gocdk serve` to develop locally")
+	fmt.Fprintln(pctx.stdout, "- `gocdk demo` to test new APIs")
+	fmt.Fprintln(pctx.stdout, "- `gocdk build` to build a Docker container")
+	fmt.Fprintln(pctx.stdout, "- `gocdk biome` to create a biome for launch")
```
clausti: I don't think we can count on a user knowing what `launch` is, here. Maybe "to create a a new biome to hold settings for a deployment?" is that too long?

internal/cmd/gocdk/init.go
```diff
@@ -81,6 +82,16 @@ func init_(ctx context.Context, pctx *processContext, args []string) error {
 	if err := materializeTemplateDir(projectDir, "init", tmplValues); err != nil {
 		return xerrors.Errorf("gocdk init: %w", err)
 	}
+	fmt.Fprintf(pctx.stdout, "Project created at %s with:\n", projectDir)
+	fmt.Fprintln(pctx.stdout, "- Go HTTP server")
+	fmt.Fprintln(pctx.stdout, "- Dockerfile")
+	fmt.Fprintln(pctx.stdout, "- 'dev' biome for local development settings")
+	fmt.Fprintln(pctx.stdout)
+	fmt.Fprintf(pctx.stdout, "Run `cd %s`, then run:\n", f.Arg(0))
+	fmt.Fprintln(pctx.stdout, "- `gocdk serve` to develop locally")
```
clausti: Let's try and say something here about the live code reloading? (The `serve` output itself should also probably say something but if I'm a new user who read the docs sloppily or not at all I might not understand the core value prop of `serve`)

internal/cmd/gocdk/init.go
```diff
@@ -81,6 +82,16 @@ func init_(ctx context.Context, pctx *processContext, args []string) error {
 	if err := materializeTemplateDir(projectDir, "init", tmplValues); err != nil {
 		return xerrors.Errorf("gocdk init: %w", err)
 	}
+	fmt.Fprintf(pctx.stdout, "Project created at %s with:\n", projectDir)
```
vangent: WDYT about putting this content in `static`? It might be easier to edit that way.

zombiezen: It's short enough and a single message that I don't think the separation buys much.

internal/cmd/gocdk/init.go
zombiezen: Done.

internal/cmd/gocdk/init.go
zombiezen: Hm. I agree the user may not know what the `launch` command is, but my goal is:

a) Not to indicate `launch` or `deploy` yet, since launch to the dev biome (local launcher) is probably not going to meet the user's expectation of launching or deploying.
b) Communicate that creating a biome is the first step to getting to a successful launch.

"hold settings for a deployment" does communicate more about what a biome is, but IMO it emphasizes the wrong part: the configuration. It's also a bit imprecise: it holds info that is true for many deployments, not just one.

WDYT about:

```
- `gocdk biome` to configure launch settings
```

clausti: I think either the way it already is or `- `gocdk biome` to configure launch settings` is fine for now and we can keep thinking about it?

zombiezen: SGTM. Amended to reflect `gocdk biome add` change. 🎉

PR #2131, opened 2019-05-16
internal/cmd/gocdk: First pass at biome add command

### LINKED ISSUE #2082: internal/cmd/gocdk: minimal `biome add`
Implement a command for adding a new `biome`, as `gocdk biome add`.

clausti:
fixes #2082


internal/cmd/gocdk/biome_add.go
```diff
@@ -0,0 +1,101 @@
...
+	if err := f.Parse(args); xerrors.Is(err, flag.ErrHelp) {
+		return nil
+	} else if err != nil {
+		return usagef("%s: %w", usageMsg, err)
+	}
+	if f.Args()[0] != "add" {
+		return usagef("%s: expected add, got %v", usageMsg, f.Args()[0])
+	}
+	newName := f.Args()[1]
+	dstPath := path.Join(pctx.workdir, "biomes", newName)
```
zombiezen: biome.go has a function `findBiomeDir` that does this already.

(For future reference, the `path` package operates on slash-separated paths only, so this would not work correctly on Windows. `path/filepath` is what you want.)

internal/cmd/gocdk/biome_add.go
```diff
@@ -0,0 +1,101 @@
...
+
+func biomeAdd(ctx context.Context, pctx *processContext, args []string) error {
+	// TODO(clausti) interpolate launcher from one supplied as a flag
+	f := newFlagSet(pctx, "demo")
+	usageMsg := "gocdk biome add BIOME_NAME"
+	if err := f.Parse(args); xerrors.Is(err, flag.ErrHelp) {
+		return nil
+	} else if err != nil {
+		return usagef("%s: %w", usageMsg, err)
+	}
```
zombiezen: Missing a check for number of arguments:

```go
if f.NArg() != 2 {
  return usagef("gocdk biome add BIOME_NAME")
}
```

internal/cmd/gocdk/biome_add.go
```diff
@@ -0,0 +1,101 @@
...
+	"path/filepath"
+	"text/template"
+
+	"golang.org/x/xerrors"
+)
+
+func biomeAdd(ctx context.Context, pctx *processContext, args []string) error {
+	// TODO(clausti) interpolate launcher from one supplied as a flag
+	f := newFlagSet(pctx, "demo")
+	usageMsg := "gocdk biome add BIOME_NAME"
```
zombiezen: Remove the `BIOME_NAME` here to avoid it being spilled out into usage error messages for which it doesn't make sense. For instance, line 42 would read:

```
gocdk biome add BIOME_NAME: expected add, got xyzzy
```

as opposed to:

```
gocdk biome add: expected add, got xyzzy
```

```suggestion
	const usageMsg = "gocdk biome add"
```

internal/cmd/gocdk/biome_add.go
```diff
@@ -0,0 +1,101 @@
...
+	if err := f.Parse(args); xerrors.Is(err, flag.ErrHelp) {
+		return nil
+	} else if err != nil {
+		return usagef("%s: %w", usageMsg, err)
+	}
+	if f.Args()[0] != "add" {
```
zombiezen: nit: `FlagSet` has a method `Arg(i int) string` which lets you get the argument without going through the slice access.

```suggestion
	if f.Arg(0) != "add" {
```

internal/cmd/gocdk/biome_add.go
```diff
@@ -0,0 +1,101 @@
...
+		return usagef("%s: expected add, got %v", usageMsg, f.Args()[0])
+	}
+	newName := f.Args()[1]
```
zombiezen: nit: FlagSet has a method Arg(i int) string which lets you get the argument without going through the slice access.

```suggestion
	newName := f.Arg(1)
```

internal/cmd/gocdk/biome_add.go
```diff
@@ -0,0 +1,101 @@
...
+	dstPath := path.Join(pctx.workdir, "biomes", newName)
+
+	tmplDir, err := static.Open("biome_add")
```
zombiezen: In init.go, I added a function `materializeTemplateDir` which will copy a directory from `static` into a desired actual path. This code can be replaced by:

```go
err := materializeTemplateDir(dstPath, "biome_add", data)
```

internal/cmd/gocdk/biome_add.go
```diff
@@ -0,0 +1,101 @@
...
+		xerrors.Errorf("biome add: %w", err)
+	}
+	data := struct {
+		ProjectName string
+	}{
+		ProjectName: filepath.Base(projectDir),
+	}
+
+	logger := log.New(pctx.stderr, "gocdk: ", log.Ldate|log.Ltime)
+	logger.Printf("Adding new biome '%v'...", newName)
```
zombiezen: This isn't a lengthy operation, so I would recommend just printing directly to stdout once at the end. No need for a logger here.

internal/cmd/gocdk/main.go
```diff
@@ -79,6 +79,8 @@ func run(ctx context.Context, pctx *processContext, args []string, debug *bool)
 		return serve(ctx, pctx, cmdArgs)
 	case "demo":
 		return addDemo(ctx, pctx, cmdArgs)
+	case "biome":
+		return biomeAdd(ctx, pctx, cmdArgs)
 	// TODO(rvangent): Add "add-biome".
```
zombiezen: We can remove this TODO, since it is done. 🙂

internal/cmd/gocdk/_static/biome_add/outputs.tf
```diff
@@ -0,0 +1,10 @@
+# This file specifies Terraform module outputs.
+# See https://www.terraform.io/docs/configuration/outputs.html
+
+# The launch specifier sets options for the biome's launcher.
+# This is a local launcher, which runs on the local Docker daemon.
+output "launch_specifier" {
+	value {
+		host_port = 8080
```
zombiezen: As mentioned over IM, please set `project_id`, `location` (should always be `us-central1` in current Cloud Run beta), and `service_name`.

```suggestion
		project_id   = "" # fill this in
		location     = "us-central1"
		service_name = "" # fill this in
```

internal/cmd/gocdk/biome_add.go
```diff
@@ -0,0 +1,101 @@
+// Copyright 2019 The Go Cloud Development Kit Authors
```
vangent: Nit: I would consider naming this file `biome.go`; it'll probably make sense to put all of the biome-related commands in the same file, no?

internal/cmd/gocdk/_static/biome_add/secrets.auto.tfvars
```diff
@@ -0,0 +1,6 @@
+# This would be where you would put secrets like database passwords.
```
vangent: Nit: consider `biome` instead of `biome_add`; are any other biome commands going to need static templates? If so, maybe subdirectories (e.g., `biome/add`).

internal/cmd/gocdk/biome_add.go
```diff
@@ -0,0 +1,101 @@
...
+	slashpath "path"
+	"path/filepath"
+	"text/template"
+
+	"golang.org/x/xerrors"
+)
+
+func biomeAdd(ctx context.Context, pctx *processContext, args []string) error {
+	// TODO(clausti) interpolate launcher from one supplied as a flag
+	f := newFlagSet(pctx, "demo")
```
vangent: Not `demo`.

internal/cmd/gocdk/biome_add.go
vangent: Turns out there is already a `biome.go`, maybe add it there?

zombiezen: I had recommended to @clausti to use biome_add.go to avoid the conflict. biome.go has a bunch of biome-related support functions that are independent of any one command, and stuffing all the biome-related _commands_ in there will likely get messy.

internal/cmd/gocdk/biome_add.go
clausti: should we move it out of `init.go` if it's going to be used for multiple commands? (does `demo` use it?)

internal/cmd/gocdk/_static/biome_add/secrets.auto.tfvars
clausti: Related to @zombiezen's [comment](https://github.com/google/go-cloud/pull/2131#discussion_r285202267) about having suggested `biome_add.go` since `biome.go` already had a bunch of stuff in it, and I wanted to name it to match the name of the command file.

internal/cmd/gocdk/biome_add.go
zombiezen: `gocdk demo` does not use it. I don't have a strong preference as to where the function is located (but it seems fine where it is).

internal/cmd/gocdk/biome_add_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+		ctx := context.Background()
+		pctx := &processContext{
+			workdir: dir,
+			stdin:   strings.NewReader(""),
+			stdout:  ioutil.Discard,
+			stderr:  ioutil.Discard,
+		}
+		const newBiome = "foo"
+		want := &biomeConfig{
```
zombiezen: nit: Move this closer to its usage (in `readBiomeConfig`).

internal/cmd/gocdk/biome_add_test.go
```diff
@@ -0,0 +1,61 @@
...
+			ServeEnabled: configBool(false),
+			Launcher:     configString("cloudrun"),
+		}
+
+		if err := biomeAdd(ctx, pctx, []string{"add", newBiome}); err != nil {
+			t.Fatal(err)
+		}
+		got, err := readBiomeConfig(dir, newBiome)
```
zombiezen: Please add a comment documenting what this and the following code is verifying.

internal/cmd/gocdk/biome_add_test.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		const newBiome = "foo"
+		want := &biomeConfig{
+			ServeEnabled: configBool(false),
+			Launcher:     configString("cloudrun"),
+		}
+
+		if err := biomeAdd(ctx, pctx, []string{"add", newBiome}); err != nil {
+			t.Fatal(err)
+		}
```
zombiezen: We could also use the check from init_test.go that ensures at least one file in the biome has the extension ".tf".

clausti:
@vangent PTAL?

internal/cmd/gocdk/_static/biome_add/secrets.auto.tfvars
vangent: I meant the directory name under `static` here, e.g. `_static/biome/...` or `_static/biome/add/...` instead of `_static/biome_add/...`.

PR #2132, opened 2019-05-16
docstore: introduce OrderBy method

jba:
- Add an OrderBy method to docstore.Query.

- Add fields in driver.Query.

- Overhaul and improve query error checking.

OrderBy implementations TBD.


internal/docstore/query.go
```diff
@@ -112,58 +156,65 @@ func (q *Query) BeforeQuery(f func(asFunc func(interface{}) bool) error) *Query
...
 	wrapErr := func(err error) error { return wrapError(dcoll, err) }
-	if err := q.init(fps); err != nil {
+	if err := q.initGet(fps); err != nil {
 		return &DocumentIterator{err: wrapErr(err)}
 	}
 	it, err := dcoll.RunGetQuery(ctx, q.dq)
 	return &DocumentIterator{iter: it, wrapError: wrapErr, err: wrapErr(err)}
 }
 
+func (q *Query) initGet(fps []FieldPath) error {
```
shantuo: Should this also check whether `OrderByField` also appears in the filters?

jba: Done.

I also added implementations for mem, mongo and firestore.

PTAL.

shantuo: Forgot to push?

jba: Right :(

PTAL

Issue #1821, opened 2019-04-12
internal/cmd/gocdk: deal with Terraform output


clausti:
See code comment: https://github.com/google/go-cloud/pull/1805/files#diff-c6ae2ccc3e43cd1fd49425541507811dR49

zombiezen:
(Scope has increased since we call Terraform all over the place now.)

PR #2134, opened 2019-05-17
internal/cmd/gocdk: add apply -input flag

zombiezen:
Prevents serve from prompting user for unset Terraform variables. The user should create .tfvars files or set environment variables for serve.

Updates #1821

### DIFF
--- a/internal/cmd/gocdk/apply.go
+++ b/internal/cmd/gocdk/apply.go
@@ -20,12 +20,14 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 
 	"golang.org/x/xerrors"
 )
 
 func apply(ctx context.Context, pctx *processContext, args []string) error {
 	f := newFlagSet(pctx, "apply")
+	input := f.Bool("input", true, "ask for input for Terraform variables if not directly set")
 	if err := f.Parse(args); xerrors.Is(err, flag.ErrHelp) {
 		return nil
 	} else if err != nil {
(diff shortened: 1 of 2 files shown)


internal/cmd/gocdk/apply.go
```diff
@@ -20,12 +20,14 @@ import (
...
 	"os/exec"
 	"path/filepath"
+	"strconv"
 
 	"golang.org/x/xerrors"
 )
 
 func apply(ctx context.Context, pctx *processContext, args []string) error {
 	f := newFlagSet(pctx, "apply")
+	input := f.Bool("input", true, "ask for input for Terraform variables if not directly set")
```
vangent: `prompt`? `allow_input`?

internal/cmd/gocdk/apply.go
```diff
@@ -90,8 +93,10 @@ func ensureTerraformInit(ctx context.Context, pctx *processContext, moduleRoot,
 	}
 
 	// Biome exists but not initialized. Need to run terraform init.
-	c := exec.CommandContext(ctx, "terraform", "init")
+	c := exec.CommandContext(ctx, "terraform", "init", "-input="+strconv.FormatBool(input))
```
vangent: Curious: what might `terraform init` prompt for? I've never seen it ask for anything.

internal/cmd/gocdk/apply.go
zombiezen: I'm matching Terraform's `-input` flag. I'd like to avoid inventing our own name for this flag.

internal/cmd/gocdk/apply.go
zombiezen: It can happen in more complex Terraform configurations where the remote state backend configuration is only partially configured: https://www.terraform.io/docs/backends/config.html#partial-configuration

PR #2135, opened 2019-05-17
internal/cmd/gocdk: use launch_environment biome output in serve and launch

### LINKED ISSUE #1890: internal/cmd/gocdk: specify configuration mechanism

zombiezen:
Verified manually to work in serve, local launcher, and Cloud Run launcher.

Fixes #1890


internal/cmd/gocdk/_static/init/biomes/dev/outputs.tf
```diff
@@ -1,6 +1,15 @@
 # This file specifies Terraform module outputs.
 # See https://www.terraform.io/docs/configuration/outputs.html
 
+# The launch environment sets environment variables for the running server
+# which can be read inside your program using os.Getenv.
+output "launch_environment" {
```
zombiezen: @clausti FYI, we'll likely want to add a stanza like this to the created biome template as well. If you want to add it in #2131, it won't break anything to have it in there: it just won't be used anywhere. Otherwise, I can follow up with a small PR to add it back in.

clausti: I can paste it in real quick.

PR #2136, opened 2019-05-17
internal/cmd/gocdk: Docker ignore .git and biomes directories

zombiezen:
Ignoring these improves cacheability of Docker builds


internal/cmd/gocdk/_static/init/.dockerignore
```diff
@@ -1,3 +1,5 @@
+/biomes/
```
vangent: Why ignore `biomes`? Don't we expect that to be checked in?

zombiezen: Yes, but `.dockerignore` is about what data is copied from the working copy to the Docker daemon, not what is checked into version control. Biome files are never needed to produce the built Docker image.

vangent: Ah, sorry, I missed that this was `.dockerignore` not `.gitignore`. Updating title.

PR #2144, opened 2019-05-17
gocdk: use Cobra for command-line processing

### LINKED ISSUE #2147: gocdk: switch to using Cobra for command line and flag processing
After much analysis and discussion, we've decided to use Cobra.

vangent:
Fixes #2147.

There are a bunch of TODOs to fill in better help text etc., but I'd rather deal with them separately since this PR will have lots of conflicts if it lingers.

Sample output:

```
$ ./gocdk
$ ./gocdk
TODO more about gocdk

Usage:
  gocdk [command]

Available Commands:
  apply       TODO Apply Terraform for BIOME
  biome       TODO Manage biomes
  build       TODO Build a Docker image
  demo        TODO Manage demos
  deploy      TODO Deploy the biome
  help        Help about any command
  init        TODO: Initialize a new project
  launch      TODO Launch BIOME
  serve       TODO Run an auto-reloading local server

Flags:
      --debug   show verbose error messages
  -h, --help    help for gocdk

Use "gocdk [command] --help" for more information about a command.



$ ./gocdk help demo
TODO more about demos

Usage:
  gocdk demo [command]

Available Commands:
  add         TODO Add a demo
  list        TODO List available demos

Flags:
  -h, --help   help for demo

Global Flags:
      --debug   show verbose error messages

Use "gocdk demo [command] --help" for more information about a command.






$ ./gocdk help demo add
TODO more about adding a demo

Usage:
  gocdk demo add DEMO [flags]

Flags:
      --force   re-add even the demo even if it has already been added, overwriting previous files
  -h, --help    help for add

Global Flags:
      --debug   show verbose error messages






$ ./gocdk demo add
Error: accepts 1 arg(s), received 0
Usage:
  gocdk demo add DEMO [flags]

Flags:
      --force   re-add even the demo even if it has already been added, overwriting previous files
  -h, --help    help for add

Global Flags:
      --debug   show verbose error messages





$ ./gocdk init foo
Project created at /usr/local/google/home/rvangent/go/src/github.com/google/go-cloud/internal/cmd/gocdk/foo with:
- Go HTTP server
- Dockerfile
- 'dev' biome for local development settings

Run `cd foo`, then run:
- `gocdk serve` to run the server locally with live code reloading
- `gocdk demo` to test new APIs
- `gocdk build` to build a Docker container
- `gocdk biome add` to configure launch settings




$ ./gocdk init foo
Error: gocdk init: /usr/local/google/home/rvangent/go/src/github.com/google/go-cloud/internal/cmd/gocdk/foo already exists
```

Let me know if you want to see more examples.


internal/cmd/gocdk/demo_test.go
```diff
@@ -43,14 +43,14 @@ func TestAddDemo(t *testing.T) {
...
-	if err := run(ctx, pctx, []string{"init", "-m", "test", "--allow-existing-dir", dir}, new(bool)); err != nil {
+	if err := init_(ctx, pctx, dir, "testmodpath", true); err != nil {
 		t.Fatalf("run init error: %+v", err)
 	}
 
 	// Call the main package run function as if 'add-demo' were being called
 	// from the command line for each of the demos.
 	for _, demo := range allDemos {
-		if err := run(ctx, pctx, []string{"demo", demo.name}, new(bool)); err != nil {
+		if err := run(ctx, pctx, []string{"demo", "add", demo.name}, new(bool)); err != nil {
```
vangent: FYI, I used the pattern of calling the function directly in tests most of the time (e.g., `init_` above), but calling `run` for the function under test (e.g., `run` here). WDYT?

internal/cmd/gocdk/main.go
```diff
@@ -46,53 +46,35 @@ func main() {
...
+		Short: "TODO gocdk",
+		Long:  "TODO more about gocdk",
+		// We want to print usage for any command/flag parsing errors, but
+		// suppress usage for random errors returned by a command's RunE.
+		// This function gets called right before RunE, so suppress from now on.
+		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+			cmd.SilenceUsage = true
+		},
 	}
+	rootCmd.PersistentFlags().BoolVar(debug, "debug", false, "show verbose error messages")
```
vangent: IIUC, this flag doesn't do anything any more, so we could remove it.

Cobra has a `SilenceErrors` setting that we could use, and then print out the errors ourselves (including using the `debug` flag for verbose errors), but that has the down side of printing out errors after usage for cases where usage is printed, which IMHO looks weird.

internal/cmd/gocdk/biome_add.go
```diff
@@ -16,30 +16,39 @@ package main
...
-	f := newFlagSet(pctx, "biome")
-	const usageMsg = "gocdk biome add"
-	if err := f.Parse(args); xerrors.Is(err, flag.ErrHelp) {
-		return nil
-	} else if err != nil {
-		return usagef("%s: %w", usageMsg, err)
-	}
-	if f.Arg(0) != "add" {
-		return usagef("%s: expected add, got %v", usageMsg, f.Args()[0])
+// TODO(rvangent): This should probably not be in biome_add.go anymore.
```
clausti: Are you thinking we should revisit having a "biome" dir for the biome commands (`add` and `list` at least, maybe eventually `clone` or others) to live in? Or just rename the file?

internal/cmd/gocdk/build.go
```diff
@@ -17,42 +17,54 @@ package main
...
+			}
+			return build(ctx, pctx, ref)
+		},
 	}
+	buildCmd.Flags().BoolVar(&list, "list", false, "display Docker images of this project")
+	buildCmd.Flags().StringVarP(&ref, "tag", "t", defaultDockerTag, "name and/or tag in the form `name[:tag] OR :tag`")
+	rootCmd.AddCommand(buildCmd)
+}
+
+// TODO(rvangent): Rename ref -> dockerTag for consistency?
```
clausti: 👍

internal/cmd/gocdk/biome_add.go
vangent: We should discuss; it should be consistent for all commands, not just `biome`. But it's pretty clear that `biome_add.go` is not the right filename given that it will have other `biome` subcommands.

Options include subdirectories/packages, `biome.go` with all of the `biome` subcommands, `biome.go` that sets up the commands and separate `biome_foo.go` for each subcommands, ...?

internal/cmd/gocdk/build.go
zombiezen: ... It's complicated. `docker build`'s `-t` flag, which this flag is trying to keep consistency with, actually is specifying a [Docker image reference](https://godoc.org/github.com/docker/distribution/reference), not just the tag. I used to use the term "image name" to refer to this concept, but as @clausti has pointed out, that is imprecise. So throughout, I am trying to use "image reference" to mean the thing documented at the link above.

internal/cmd/gocdk/demo_test.go
zombiezen: Using `run` over time seems like it would be closer to how the user would experience it, which would make the tests less brittle over time. Can you help me understand the case for calling command functions directly in tests? I think I'd be more okay with using helper functions shared between tests and prod code that work on the parsed arguments.

internal/cmd/gocdk/main.go
zombiezen: That's unfortunate. The `%+v` output has been really useful to me for understanding failures, and I think will be even more so when we need to collect errors from folks in the field. I'd be okay removing it as long as we have an issue to track somehow regaining this functionality.

internal/cmd/gocdk/build.go
vangent: I updated the TODO with a link to this comment.

internal/cmd/gocdk/demo_test.go
vangent: Mostly just that it's simpler -- you get to call a function with named arguments that probably match what you have in hand. With `run`, you have to convert those arguments to flags etc.; it's harder to write and harder to read.

I left the use of `run` for the thing actually being tested; e.g., `demo_test.go` uses `run` to call `gocdk demo add`, as I agree it's important to test as close to what the user experience as possible. But I don't think it's necessary to do that for things that aren't directly under test.

We can revisit this if you want to discuss more, it's pretty easy to change.

vangent: Also I notice that I forgot to update this file to use a new test helper, `newTestProject`, which gets rid of the call to `init_` here anyway.

internal/cmd/gocdk/main.go
vangent: Filed https://github.com/google/go-cloud/issues/2173 and remove `--debug`.

PR #2146, opened 2019-05-18
postgres/rdspostgres: fix code and tests so tests pass; add URL tests

### LINKED ISSUE #2140: postgres/rdspostgres: TestOpenBadValues fails
Running via https://github.com/google/go-cloud/blob/master/internal/testing/prereleasechecks.sh, the Terraform scripts succeed, but the tests fail:

```
--- FAIL: TestOpenBadValues (0.04s)
    --- FAIL: TestOpenBadValues/user=foo (0.00s)
        rdspostgres_test.go:102: error = <nil>; want to contain "user"
    --- FAIL: TestOpenBadValues/password=foo (0.00s)
        rdspostgres_test.go:102: error = <nil>; want to contain "password"
    --- FAIL: TestOpenBadValues/dbname=foo (0.00s)
        rdspostgres_test.go:102: error = <nil>; want to contain "dbname"
    --- FAIL: TestOpenBadValues/host=localhost (0.00s)
        rdspostgres_test.go:102: error = <nil>; want to contain "host"
    --- FAIL: TestOpenBadValues/port=1234 (0.00s)
        rdspostgres_test.go:102: error = <nil>; want to contain "port"
FAIL
```

vangent:
Fixes #2140.


postgres/rdspostgres/rdspostgres.go
```diff
@@ -72,15 +74,17 @@ func init() {
 // OpenPostgresURL opens a new RDS database connection wrapped with OpenCensus instrumentation.
 func (uo *URLOpener) OpenPostgresURL(ctx context.Context, u *url.URL) (*sql.DB, error) {
 	cf := new(rds.CertFetcher)
-	vals := u.Query()
```
vangent: I am not sure if there was a point to this?

postgres/rdspostgres/rdspostgres.go
```diff
@@ -72,15 +74,17 @@ func init() {
...
-	u.RawQuery = vals.Encode()
 
+	database := u.EscapedPath()
```
vangent: It turns out `RawPath` was empty; IIUC, `EscapedPath()` uses `RawPath` if it's available, otherwise computes it from `Path`. Apparently the godoc is wrong when it says `RawPath` is always set.

postgres/rdspostgres/rdspostgres.go
```diff
@@ -97,9 +101,12 @@ func Open(ctx context.Context, provider rds.CertPoolProvider, params *Params) (*
 	vals := make(url.Values)
 	for k, v := range params.Values {
 		// Only permit parameters that do not conflict with other behavior.
-		if k == "sslmode" || k == "sslcert" || k == "sslkey" || k == "sslrootcert" {
```
vangent: IIUC, these were exactly backwards; the ones that are allowable were listed, while the ones that conflict with `Params` were not. Should be the other way around.

postgres/rdspostgres/rdspostgres.go
zombiezen: There was not, and the `u.RawQuery` mutation violates the contract for the `URLOpener` interface. I caught this in at least one PR, but it must have slipped into the codebase. Sorry.

postgres/rdspostgres/rdspostgres.go
zombiezen: @clausti filed this as golang/go#29662, which has been fixed at tip.

postgres/rdspostgres/rdspostgres.go
```diff
@@ -72,15 +74,17 @@ func init() {
 // OpenPostgresURL opens a new RDS database connection wrapped with OpenCensus instrumentation.
 func (uo *URLOpener) OpenPostgresURL(ctx context.Context, u *url.URL) (*sql.DB, error) {
 	cf := new(rds.CertFetcher)
-	vals := u.Query()
-	u.RawQuery = vals.Encode()
 
+	database := u.EscapedPath()
+	if strings.HasPrefix(database, "/") {
```
zombiezen: nit: use `strings.TrimPrefix`

postgres/rdspostgres/rdspostgres.go
zombiezen: I don't agree with this change: I believe this to be correct as-is.

There was an issue in one iteration of this (across PRs) that wouldn't permit host in particular, which made it difficult to connect to databases using Unix domain sockets. I think for consistency we should allow those parameters, but we definitely want to forbid specifying any SSL/TLS options.

vangent: Can you explain the intent then? IIUC, you can set the `user/password/dbname/host/port` fields:
* if you're using the explicit constructor, via `Params`
* if you're using a URL, via parts of the URL (which constructs a `Params`).

So why would you then want to also set those fields via a query parameter?
And, if we disallow query parameters like `sslmode` etc., how would you set them?

I interpreted the comment "Only permit parameters that do not conflict with other behavior" as "only permit parameters that don't conflict with what you can specify another way (via `Params`), but maybe I misunderstood.

postgres/rdspostgres/rdspostgres.go
vangent: Done.

postgres/rdspostgres/rdspostgres.go
vangent: Per discussion, reverting this change. The SSL-related query parameters will definitely not work, because the underlying dialer provides TLS and so we explicitly disable SSL. I had to update the test, which was testing for the behavior I changed it to.

The query parameters like `user` etc. are probably not needed, because they are settable via `Params`, but there might be scenarios where they are useful (e.g., if you are using a URL opener and it can't parse your host correctly), so we don't gain much by disallowing them.

PR #2153, opened 2019-05-20
docstore/internal/fields: interpret "-," tag like encoding/json.

jba:
We used to return an error if "-" was followed by options, but it
should be treated as a name instead.


internal/docstore/internal/fields/fields.go
```diff
@@ -459,22 +458,19 @@ func dominantField(fs []Field) (Field, bool) {
 
 // ParseStandardTag extracts the sub-tag named by key, then parses it using the
 // de facto standard format introduced in encoding/json:
-//   "-" means "ignore this tag". It must occur by itself. (parseStandardTag returns an error
-//       in this case, whereas encoding/json accepts the "-" even if it is not alone.)
+//   "-" means "ignore this tag", unless it has options (that is, is followed by a comma),
+//       which which case it is treated a name.
```
shantuo: in which case

internal/docstore/internal/fields/fields_test.go
```diff
@@ -348,14 +349,17 @@ func TestDominantField(t *testing.T) {
...
 	if err != nil {
 		t.Fatal(err)
 	}
-	if len(got) != 0 {
-		t.Errorf("got %d fields, want 0", len(got))
+	if len(got) != 1 {
+		t.Errorf("got %d fields, want 1", len(got))
 	}
+
+	fmt.Println(got[0])
```
shantuo: Remove.

PR #2158, opened 2019-05-22
pubsub/all: BREAKING_CHANGE add Message.Nackable, remove driver.AckFunc

vangent:
* Adds Message.Nackable
* Updates docs
* Updates `azuresb` to use a boolean instead of a func
* Updates `nats` to not use an ackfunc
* Drops the no-longer-used `AckFunc` on driver


pubsub/azuresb/azuresb.go
```diff
@@ -496,11 +496,14 @@ func (s *subscription) CanNack() bool {
...
 		return false
 	}
-	return s.opts.AckFuncForReceiveAndDelete == nil
+	return !s.opts.ReceiveAndDelete
 }
 
 // SendNacks implements driver.Subscription.SendNacks.
 func (s *subscription) SendNacks(ctx context.Context, ids []driver.AckID) error {
+	if !s.CanNack() {
+		panic("unreachable")
```
eliben: Is this something the user can get to? Now sure "unreachable" is the most informative error message

pubsub/pubsub.go
```diff
@@ -38,45 +38,23 @@
...
-// the message is essentially auto-acked when it is delivered.
-//
-// This package accommodates both kinds of systems. See the provider-specific
-// package documentation to see whether it is at-most-once or at-least-once.
-// Some providers support both modes.
-//
-// Application developers should think carefully about which kind of semantics
-// the application needs. Even though the application code may look similar, the
-// system-level characteristics are quite different.
+// The semantics of message deliver vary across PubSub providers.
```
jba: deliver -> delivery

pubsub/pubsub.go
```diff
@@ -168,8 +146,13 @@ type Message struct {
...
 // Ack acknowledges the message, telling the server that it does not need to be
-// sent again to the associated Subscription. It returns immediately, but the
-// actual ack is sent in the background, and is not guaranteed to succeed.
+// sent again to the associated Subscription. It will be a no-op for some
+// providers; see
+// https://godoc.org/gocloud.dev/pubsub#hdr-At_most_once_and_At_least_once_Delivery
+// for more info.
+//
+// Ack returns immediately, but the actual ack is sent in the background, and
+// s not guaranteed to succeed.
```
jba: s -> is

Also mention that if ack fails, that will eventually show up in the error from Receive.

pubsub/azuresb/azuresb.go
vangent: No, the user cannot get here. Since `CanNack` returns `false`, the concrete type should never call `SendNacks`. This is more of an assertion to make the code more clear.

pubsub/pubsub.go

pubsub/pubsub.go

PR #2163, opened 2019-05-23
gocdk: fix executable path for windows by adding .EXE suffix

### LINKED ISSUE #2155: travis: windows build is failing on a couple of non-core modules
`internal/contributebot`:

```
# gocloud.dev/internal/contributebot [gocloud.dev/internal/contributebot.test]
.\sync.go:177:25: undefined: unix.SIGTERM
FAIL	gocloud.dev/internal/contributebot [build failed]
?   	gocloud.dev/internal/contributebot/webhook	[no test files]
```

We ran into this before and decided to just not run them. Maybe that's still the right answer, but now that we run all tests for all modules in a loop (instead of bespoke code for each), it's less obvious how to do that.

`internal/website`:

```
--- FAIL: TestGather (55.07s)
    --- FAIL: TestGather/BlankImports (0.19s)
        gatherexamples_test.go:265: go [list -e -json -compiled=true -test=true -export=false -deps=true -find=false -- ./...]: exit status 1: error writing go.mod: rename C:\Users\travis\AppData\Local\Temp\TestGather_BlankImports835612617\primarymod\foo\go.mod562915362.tmp C:\Users\travis\AppData\Local\Temp\TestGather_BlankImports835612617\primarymod\foo\go.mod: Access is denied.
            
FAIL
FAIL	gocloud.dev/internal/website/gatherexamples	55.428s
```

This just looks like a bug where a file is read and not closed before it's being written.

vangent:
Fixes #2155.

On Windows, the executable must end with `.EXE`.

### DIFF
--- a/internal/cmd/gocdk/demo_test.go
+++ b/internal/cmd/gocdk/demo_test.go
@@ -24,6 +24,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 )
@@ -57,6 +58,9 @@ func TestAddDemo(t *testing.T) {
 
 	// Build the binary.
 	exePath := filepath.Join(dir, "add-demo-test")
+	if runtime.GOOS == "windows" {
+		exePath += ".EXE"
+	}
 	if err := buildForServe(ctx, pctx, dir, exePath); err != nil {
 		t.Fatal("buildForServe(...):", err)
 	}
--- a/internal/cmd/gocdk/serve.go
+++ b/internal/cmd/gocdk/serve.go
@@ -239,6 +239,7 @@ loop:
 }
 
 // buildForServe runs Wire and `go build` at moduleRoot to create exePath.
+// Note that on Windows, exePath must end with .EXE.
 func buildForServe(ctx context.Context, pctx *processContext, moduleRoot string, exePath string) error {
 	moduleEnv := overrideEnv(pctx.env, "GO111MODULE=on")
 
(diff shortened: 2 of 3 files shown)


internal/cmd/gocdk/serve.go
```diff
@@ -239,6 +239,7 @@ loop:
 }
 
 // buildForServe runs Wire and `go build` at moduleRoot to create exePath.
+// Note that on Windows, exePath must end with .EXE.
 func buildForServe(ctx context.Context, pctx *processContext, moduleRoot string, exePath string) error {
```
vangent: Could also modify `exePath` here as needed, and add it as a return value. It's a bit awkward for the existing callers since they stash `exePath` in a struct.

internal/cmd/gocdk/demo_test.go
```diff
@@ -57,6 +58,9 @@ func TestAddDemo(t *testing.T) {
 
 	// Build the binary.
 	exePath := filepath.Join(dir, "add-demo-test")
+	if runtime.GOOS == "windows" {
+		exePath += ".EXE"
```
eliben: Is it really uppercase `.EXE` these days?

vangent: Doesn't matter, Windows is case-insensitive.

zombiezen:
I believe this fixes the test, but IIUC this actually keeps `gocdk serve` broken on Windows. 🙁

vangent:
> I believe this fixes the test, but IIUC this keeps `gocdk serve` broken on Windows.

Doh, you are right; sending a PR.

PR #2164, opened 2019-05-25
docstore: add Key method to driver.Collection

jba:
Add a Key method to driver.Collection that extracts the key from a document.

Use it to check that the key is always present (except for Create).

Also implement the check for ActionList that all documents must be distinct.

This is a first step in implementing the new semantics for action lists.


internal/docstore/docstore.go
```diff
@@ -268,10 +270,45 @@ func (l *ActionList) BeforeDo(f func(asFunc func(interface{}) bool) error) *Acti
...
+func (l *ActionList) toDriverActions() ([]*driver.Action, error) {
 	var das []*driver.Action
 	var alerr ActionListError
+	// Create a set of (document key, is Get action) pairs for detecting duplicates:
+	// an action list can have at most one get and at most one write for each key.
+	type keyAndKind struct {
+		key   interface{}
+		isGet bool
+	}
+	seen := map[keyAndKind]bool{}
```
shantuo: If the `key` is a pointer type, will this cause a false negative match? Should we make sure the key is not a pointer type?

internal/docstore/docstore.go
```diff
@@ -399,6 +436,27 @@ func isIncNumber(x interface{}) bool {
...
+func (l *ActionList) String() string {
+	var as []string
+	for _, a := range l.actions {
+		as = append(as, a.String())
+	}
+	return "[" + strings.Join(as, ", ") + "]"
+}
+
+func (a *Action) String() string {
+	buf := &bytes.Buffer{}
```
shantuo: nit: `buf` can be a `strings.Builder`.

internal/docstore/driver/driver.go
```diff
@@ -24,6 +24,12 @@ import (
 
 // A Collection is a set of documents.
 type Collection interface {
+	// Key returns the document key, or nil if the document doesn't have one. Key
+	// should not return an error if the key is missing unless the collection cannot
```
shantuo: The `unless the collection ...` part reads a little bit hard, maybe separate into its own sentence and explain what it means by `cannot generate a fresh key`?

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -202,13 +202,27 @@ func newCollection(db *dyn.DynamoDB, tableName, partitionKey, sortKey string, op
...
+		return nil, nil // missing key is not an error
 	}
-	return []string{c.partitionKey, c.sortKey}
+	if c.sortKey != "" {
+		keys[1], _ = doc.GetField(c.sortKey) // ignore error since keys[1] is nil in that case
+	}
+	return keys, nil
 }
 
+// func (c *collection) KeyFields() []string {
```
shantuo: What is this method? Should it be removed?

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -239,62 +253,31 @@ func (c *collection) runActionsOrdered(ctx context.Context, actions []*driver.Ac
...
-				wm[keys] = true
-			}
-		}
 	}
 	collect()
 	return groups
 }
 
-func (c *collection) shouldSplit(curr, next *driver.Action, wm map[[2]interface{}]bool) bool {
+func (c *collection) shouldSplit(curr, next *driver.Action) bool {
```
shantuo: Add to TODO to group actions as many as possible.

internal/docstore/docstore.go

internal/docstore/docstore.go

internal/docstore/driver/driver.go
jba: Done.

internal/docstore/dynamodocstore/dynamo.go
jba: Done.

internal/docstore/dynamodocstore/dynamo.go
jba: This is all going to be obsolete soon anyway.

jba:
PTAL.

Issue #2038, opened 2019-05-09
internal/cmd/gocdk: create processContext logger


zombiezen:
There's a variety of places that require synchronous access to writing to stderr/logs. I'm currently using the standard library `log.Logger`, but we likely need to integrate non-log-entry data like raw stdout/stderr from subprocesses in `serve`.

zombiezen:
The requirements for this as I see them:

- It would be ideal to have method(s) on `processContext` that log a message.
- It is desirable to log information to either stdout or stderr. Generally, the guideline is that output that another process may want to parse (e.g. request logs or a URL endpoint for a running server) should be sent to stdout. stderr should be reserved for output that is informational but the format may change.
- For similar reasons, our tests should be able to read stdout after the command has completed.
- We must support concurrent writes to either stdout or stderr, even in an unstructured stream. For example, when running `gocdk serve`, the CLI process and the (up to) two server subprocesses may be trying to write to either stderr or stdout concurrently with each other. I believe it to be acceptable in such cases to interleave lines, but partial lines being interleaved would not be okay.

clausti:
@zombiezen, in your [comment](https://github.com/google/go-cloud/pull/2169#issuecomment-497049381) on @vangent's PR, in response to his comment that "Maybe we should add functions to ProcessContext for Log and Printf and use those everywhere?", you said 

> I'd like to see us address the larger problem rather than trying to optimize individual parts

but I'm not sure I understand if you're saying Robert's suggestion does that or doesn't do that... it didn't seem to me like the suggestion to consolidate logging onto `ProcessContext` says anything one way or the other about latter bullet points on this issue?

zombiezen:
My comment about addressing larger problem was aimed at PR #2169, not the particular statement. I don't see the PR moving us toward any of the bullets above, since it's just one more call site we have to migrate anyway. That was my intent.

I agree with Robert that we should consolidate logging onto `processContext` (that's bullet 1). I think any such effort should also address the other bullets, too.

vangent:
@zombiezen 	how do you see `Logger` fitting in with `stdout` vs `stderr`? Currently we use `Logger` in a bunch of places to insert a timestamp and `gocdk: ` timestamp. That doesn't sound useful for what you've described as `stdout` since it makes the output harder to process. I'm not sure it adds much value for `stderr` either...?

PR #2169, opened 2019-05-28
internal/gocdk: use a logger instead of straight fmt.Printf

vangent:


clausti:
@zombiezen told me not to use a logger on this the first time around?

vangent:
> @zombiezen told me not to use a logger on this the first time around?

Hmm, we appear to be pretty inconsistent. This uses `fmt.Printf`, `init.go` uses `fmt.Fprintf(pctx.stdout, ...`, and `demo.go` uses `logger`. We should be consistent, and it should be easier to be consistent.

BTW, I noticed this because when you run `go test`, this message gets printed out (and nothing else), which doesn't seem right.

Maybe we should add functions to `ProcessContext` for `Log` and `Printf` and use those everywhere?

clausti:
>Maybe we should add functions to ProcessContext for Log and Printf and use those everywhere?

That would definitely make it easier to be consistent.

zombiezen:
See #2038 for my original report (and I've added some more nuance). I'd like to see us address the larger problem rather than trying to optimize individual parts, but I'm happy to approve if this blocks something.

vangent:
> See #2038 for my original report (and I've added some more nuance). I'd like to see us address the larger problem rather than trying to optimize individual parts, but I'm happy to approve if this blocks something.

It doesn't block anything, I just find it annoying to see this one message every time I run tests.

@clausti , up to you whether to go ahead with this or drop it and wait for the larger fix.

clausti:
Yeah, let's go ahead and remove that annoyance

PR #2170, opened 2019-05-28
testing: enhance test summary tool to report failures

eliben:
It's quite aggravating to have to scroll through >1500 lines to see which test(s) failed, so we might as well report it explicitly


internal/testing/test-summary/test-summary.go
```diff
@@ -94,5 +99,20 @@ func run(r io.Reader) (msg string, failures bool, err error) {
 	p := counts["pass"]
 	f := counts["fail"]
 	s := counts["skip"]
-	return fmt.Sprintf("ran %d; passed %d; failed %d; skipped %d", p+f+s, p, f, s), f > 0, nil
+
+	summary := fmt.Sprintf("ran %d; passed %d; failed %d; skipped %d", p+f+s, p, f, s)
+	if len(failedTests) > 0 {
+		var sb strings.Builder
```
jba: A *strings.Builder implements io.Writer, so you can use fmt.FprintXXX on it. The main difference would be line 108, where could could use fmt.Fprintf instead of Writestring + fmt.Sprintf.

eliben: Nice, thanks! Done

PR #2171, opened 2019-05-28
testing: enable the test-summary tool for runchecks

eliben:
... looking at how Travis will take it, for now

### DIFF
--- a/internal/testing/runchecks.sh
+++ b/internal/testing/runchecks.sh
@@ -87,7 +87,7 @@ while read -r path || [[ -n "$path" ]]; do
   # codecov will only save the last one anyway.
   if [[ "${TRAVIS_OS_NAME:-}" == "linux" ]]; then
     echo "Running Go tests with coverage..."
-    go test -mod=readonly -race -coverpkg=./... -coverprofile=coverage.out ./... || result=1
+    go test -mod=readonly -json -race -coverpkg=./... -coverprofile=coverage.out ./... | go run "$rootdir"/internal/testing/test-summary/test-summary.go -progress || result=1
     if [ -f coverage.out ] && [ $result -eq 0 ]; then
       # Filter out test packages.
       grep -v test coverage.out > coverage2.out
@@ -103,7 +103,7 @@ while read -r path || [[ -n "$path" ]]; do
     if [[ "${TRAVIS_OS_NAME:-}" == "windows" ]] && ([[ "$path" == "internal/contributebot" ]] || [[ "$path" == "internal/website" ]]); then
       echo "  Skipping tests on Window"
     else
-      go test -mod=readonly -race ./... || result=1
+      go test -mod=readonly -json -race ./... | go run "$rootdir"/internal/testing/test-summary/test-summary.go -progress || result=1
     fi
   fi
   # Do these additional checks for the Linux build on Travis, or when running


eliben:
Travis logs look reasonable to me, WDYT?

vangent:
> Travis logs look reasonable to me, WDYT?

Looks OK I guess. I like the summary, but do you think the test output is too chatty? IIRC, it is longer than it was before (separate line for each test, normally it is a single line per package).

It would also be nice to see what it looks like if a single test is failing.

eliben:
> > Travis logs look reasonable to me, WDYT?
> 
> Looks OK I guess. I like the summary, but do you think the test output is too chatty? IIRC, it is longer than it was before (separate line for each test, normally it is a single line per package).
> 

I suppose we could report success per package, as in the original tester. @jba WDYT?

> It would also be nice to see what it looks like if a single test is failing.

With #2170 it's actually going to be nice since it will report which tests failed without having to hunt through the log. Excerpt from full output with one failing tests (last page of text or so):

```
pass     TestURLOpenerForParams/BadPrivateKeyPath (0.00s)
pass     TestURLOpenerForParams/PrivateKeyPath (0.00s)
pass     TestURLOpenerForParams (0.00s)
pass     TestOpenBucketFromURL (0.00s)
skip gocloud.dev/gcp/cloudsql (0.00s)
skip gocloud.dev/internal/docstore/drivertest (0.00s)
skip gocloud.dev/internal/testing (0.00s)
skip gocloud.dev/internal/testing/octest (0.00s)
skip gocloud.dev/internal/testing/setup (0.00s)
skip gocloud.dev/internal/testing/terraform (0.00s)
skip gocloud.dev/internal/useragent (0.00s)
skip gocloud.dev/pubsub/driver (0.00s)
skip gocloud.dev/pubsub/drivertest (0.00s)
skip gocloud.dev/runtimevar/driver (0.00s)
skip gocloud.dev/runtimevar/drivertest (0.00s)
skip gocloud.dev/secrets/driver (0.00s)
skip gocloud.dev/secrets/drivertest (0.00s)
skip gocloud.dev/server/driver (0.00s)
skip gocloud.dev/server/sdserver (0.00s)
skip gocloud.dev/server/xrayserver (0.00s)
skip gocloud.dev/tests/internal/testutil (0.00s)
Failures (reporting up to 10):
  gocloud.dev/blob/gcsblob/TestPreconditions
ran 1573; passed 1523; failed 1; skipped 49
exit status 1
```

jba:
> do you think the test output is too chatty

You shouldn't see a line per test if you omit the -progress flag.

vangent:
> > do you think the test output is too chatty
> 
> You shouldn't see a line per test if you omit the -progress flag.

+1 for no `-progress` then :-).

eliben:
> > do you think the test output is too chatty
> 
> You shouldn't see a line per test if you omit the -progress flag.

If you omit the `-progress` flag you don't see the test results running/reporting one by one either. All we'll see on Travis is "Checking XXX" and then ominous silence for up to minutes until all results are available. This is a strictly worse user experience than the default `go test` and a no-go, IMHO.

What I think we could do is only report `"pass"` events for packages, not individual tests. It's probably what the default test runner is doing too. These events don't have a `Test` key set, for example:

```
{"Time":"2019-05-28T14:41:18.937829289-07:00","Action":"pass","Package":"gocloud.dev/aws","Elapsed":0.001}
```

Then we should have the same level of verbosity as `go test`.

eliben:
PTAL at the strawman and sample output in #2174

vangent:
BTW I am fine with submitting this as-is and iterating on the output independently; I approved this PR.

PR #2178, opened 2019-05-29
docstore: new semantics for action lists

jba:
Action lists are always unordered, and execute actions independently.

This PR does most of the work, but not every provider does independent execution yet.


internal/docstore/mongodocstore/mongo.go
```diff
@@ -230,41 +230,40 @@ func (c *collection) Key(doc driver.Document) (interface{}, error) {
...
-// TODO(jba): use bulk RPCs.
 func (c *collection) RunActions(ctx context.Context, actions []*driver.Action, opts *driver.RunActionsOptions) driver.ActionListError {
-	if opts.Unordered {
-		return c.runActionsUnordered(ctx, actions, opts)
+	errs := make([]error, len(actions))
+	beforeGets, gets, writes, afterGets := driver.GroupActions(actions)
+	c.runGets(ctx, beforeGets, errs, opts)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() { defer wg.Done(); c.runGets(ctx, gets, errs, opts) }()
```
shantuo: I think you can just use a errCh and block under afterGets, seems clearer.

internal/docstore/mongodocstore/mongo.go
```diff
@@ -642,38 +460,74 @@ func (c *collection) unorderedBulkWrite(ctx context.Context, iactions []indexedA
...
-	for i, newID := range newIDs {
-		if err := iactions[i].action.Doc.SetField(c.idField, newID); err != nil {
-			alerr = append(alerr, indexedError{iactions[i].index, err})
+		reterrs = append(reterrs, gcerr.Newf(gcerr.NotFound, nil, "some writes failed (replaced %d, upserted %d, out of total %d)", res.MatchedCount, res.UpsertedCount, nNonCreateWrite))
+	}
+	return reterrs
+}
+
+func (c *collection) determineDeleteErrors(ctx context.Context, models []mongo.WriteModel, actions []*driver.Action, errs []error) {
+	// TODO(jba): do this concurrently.
```
shantuo: I'll also change this to use `Find` when I update get.

internal/docstore/mongodocstore/mongo.go
```diff
@@ -642,38 +460,74 @@ func (c *collection) unorderedBulkWrite(ctx context.Context, iactions []indexedA
...
-	}
-	for i, newID := range newIDs {
-		if err := iactions[i].action.Doc.SetField(c.idField, newID); err != nil {
-			alerr = append(alerr, indexedError{iactions[i].index, err})
+		reterrs = append(reterrs, gcerr.Newf(gcerr.NotFound, nil, "some writes failed (replaced %d, upserted %d, out of total %d)", res.MatchedCount, res.UpsertedCount, nNonCreateWrite))
+	}
+	return reterrs
+}
+
+func (c *collection) determineDeleteErrors(ctx context.Context, models []mongo.WriteModel, actions []*driver.Action, errs []error) {
```
shantuo: Just want to make sure mongo does report error on deleting items that don't exist.

internal/docstore/driver/util.go
```diff
@@ -48,3 +50,44 @@ func SplitActions(actions []*Action, split func(a, b *Action) bool) [][]*Action
...
+			writes[a.Key] = a
+		}
+	}
+
+	vals := func(m map[interface{}]*Action) []*Action {
+		var as []*Action
+		for _, v := range m {
+			as = append(as, v)
+		}
+		// Sort so the order is always the same for replay.
```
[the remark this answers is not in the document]
jba: Map iteration order is undefined and changes on every run. So the returned slices would be in a different order each time you run the test. So the replayer can't find the request that was recorded.

internal/docstore/driver/util.go
```diff
@@ -48,3 +50,44 @@ func SplitActions(actions []*Action, split func(a, b *Action) bool) [][]*Action
 	collect()
 	return groups
 }
+
```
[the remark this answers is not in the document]
jba: Done.

internal/docstore/firedocstore/fs.go
```diff
@@ -250,81 +250,77 @@ func (c *collection) Key(doc driver.Document) (interface{}, error) {
...
+					seen[a] = true
+				}
+			}
 		}
+		c.batchGet(ctx, toRun, errs, opts)
+	}
+}
+
+// Collect the Get actions into groups with the same set of field paths.
+func groupByFieldPath(gets []*driver.Action) [][]*driver.Action {
```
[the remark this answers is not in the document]
jba: I agree. Let's see if it's needed anywhere else first.

internal/docstore/mongodocstore/mongo.go
jba: Good idea. Done.

internal/docstore/mongodocstore/mongo.go
jba: It doesn't, but docstore doesn't consider that an error.

jba:
PTAL.

internal/docstore/mongodocstore/mongo.go
shantuo: Would this be correctly reported once the issue you reported got fixed? So would the mismatched `nNonCreateWrite` number?

jba:
> Would this be correctly reported once the issue you reported got fixed? So would the mismatched nNonCreateWrite number?

The delete errors wouldn't be correctly reported even if the issue was fixed, because a delete with a non-matching document is not an error. The creates would be fixed, however.

Issue #1860, opened 2019-04-19
all: consider unifying compound ID handling


zombiezen:
(This is a write-up of a cross-cutting problem that @vangent and I stumbled across, but with concrete data about affected APIs.)

**EDIT:** The purpose of this issue is to discuss the problem and determine what we want to do about it (even if the answer is "nothing").

/cc @clausti @eliben @jba @shantuo 

### Is your feature request related to a problem? Please describe.

Many cloud providers have resource IDs that are made up of subcomponents in some well-defined syntax. For example, [GCP KMS key IDs](https://cloud.google.com/kms/docs/object-hierarchy#key) take the form `projects/[PROJECT_ID]/locations/[LOCATION]/keyRings/[KEY_RING]/cryptoKeys/[KEY]`. For the purpose of this issue, let's call these _compound IDs_.

There are two broad compound ID usage patterns we have observed:

1)  Applications will keep resources in the same location or project, so the application will build the ID from the subcomponents rather than passing the entire resource ID around.
2)  Applications will pass a verbatim string from configuration down to the API, since this is what was easily copy-pasteable from the cloud console UI.

As of 2019-04-19, compound IDs show up in:

-  The MySQL and PostgreSQL packages. Every database has a collection of fields that build up their instance connection string. For example, Cloud SQL's instance syntax is `myproject:us-central1:myinstance`. The constructors take in parameter structs ([example](https://github.com/google/go-cloud/blob/af156be9f2ca89a3dae18338a4beabf65b7499cc/mysql/cloudmysql/cloudmysql.go#L50-L76)) that separate each field, but it has more than just an instance name.
-  [GCP Pub/Sub constructors](https://github.com/google/go-cloud/blob/af156be9f2ca89a3dae18338a4beabf65b7499cc/pubsub/gcppubsub/gcppubsub.go#L232) take in a project ID and a topic name as two separate strings. In the underlying GCP APIs, these have names like `projects/myproject/topics/mytopic`.
[...]

PR #2181, opened 2019-05-30
runtimevar/gcpruntimeconfigurator: BREAKING_CHANGE use full variable key string in URL and constructors

vangent:
Instead of a `ResourceName` struct.

Updates #1860.


eliben:
This is post the 0.14 tag, right? So it's for the next release? And also we're in freeze for 0.15 now, yes?

vangent:
> This is post the 0.14 tag, right? So it's for the next release? And also we're in freeze for 0.15 now, yes?

Yes, yes, and yes.

PR #2189, opened 2019-05-30
all: initial module-geddon commit for testing

eliben:


internal/testing/alldeps
```diff
@@ -41,28 +41,21 @@ github.com/gorilla/mux
 github.com/grpc-ecosystem/grpc-gateway
 github.com/hashicorp/errwrap
 github.com/hashicorp/go-cleanhttp
-github.com/hashicorp/go-hclog
```
vangent: Why would `alldeps` change in this way?

internal/website/data/examples.json
```diff
@@ -127,22 +127,6 @@
 		"imports": "import (\n\t\"context\"\n\n\t\"gocloud.dev/pubsub\"\n\t_ \"gocloud.dev/pubsub/gcppubsub\"\n)",
 		"code": "topic, err := pubsub.OpenTopic(ctx, \"gcppubsub://myproject/mytopic\")\nif err != nil {\n\treturn err\n}\ndefer topic.Shutdown(ctx)"
 	},
-	"gocloud.dev/pubsub/kafkapubsub.ExampleOpenSubscription": {
```
vangent: This change to `examples.json` looks wrong. We probably need to run `gatherexamples` over all modules or something...?

eliben: Yep, fixed the gathering script to use allmodules

internal/testing/alldeps
eliben: Good spot. Will revert the vault bump that caused this

eliben:
Travis is green.

Moving to the next step now -- removing all `replace` directives and bumping `require`s referring to our own modules to version 0.15, which was not tagged yet. Travis is expected to fail after this change.

eliben:
This should be ready to go with @vangent's LGTM

[removed Travis requirement from branch protection so I could merge this]

PR #2192, opened 2019-05-30
internal/tests: improve prerelease checks

### LINKED ISSUE #2180: testing: prerelease.sh should run -record for all record/replay tests
To make sure that the recorded interactions we have are still valid.

vangent:
* It will now run through all of its tests and report failures at the end, instead of stopping at the first failures (which was really annoying since they are all quite slow).
* There are some known failures with issues filed for them; they are skipped or failures are swallowed.
* It now runs `go test -record` on all packages that have `testdata`. This will verify that we still work against the live provider.
* Updated the release documentation with how to run `prerelease.sh`.
* `blob/gcsblob` now has a default for the `--private_key` flag so that we can put the file in the expected place and not have to special-case the call to it.

Fixes #2180.


internal/docs/release.md
```diff
@@ -4,6 +4,19 @@ We aim to do a release after each Sprint (roughly, every 3 weeks).
...
+        You can delete the branch and abandon all the changes when you are done.
+    -   Run `./internal/testing/prerelease.sh init` until it finishes with
+        `SUCCESS`. Note that there are some known failures that are logged but
+        don't prevent overall `SUCCESS`.
+    -   Run `./internal/testing/prerelease.sh run` Again, there are some known
+        failures. You can re-run any unexpected failures independently to
+        investigate; see the `prerelease.sh` script to see the exact command
+        line to use.
+    -   Run `./internal/testing/prerelease.sh cleanup` when you are done.
+    -   Commit the changes to the branch, and then delete the branch.
```
eliben: Why would you commit changes to a branch and then delete it?

vangent: There may be a better way, but that's how I do it. You can't delete a branch while you're in it, and if you move out of the branch without committing the changes, they come along with you and aren't dropped when you delete the branch.

PR #2194, opened 2019-05-30
firedocstore: run writes independently

jba:


internal/docstore/firedocstore/fs.go
```diff
@@ -372,51 +372,53 @@ func (c *collection) newGetRequest(gets []*driver.Action) (*pb.BatchGetDocuments
...
-		// TODO(jba): re-run actions so that it appears they happen independently.
-		setErr(err)
-		return
-	}
-	// Now that we've successfully done the action, set the names for newly created docs
-	// that weren't given a name by the caller.
-	for a, nn := range newNames {
-		_ = a.Doc.SetField(c.nameField, nn)
+	// Run the commit calls concurrently.
+	// TODO(jba): limit number of outstanding RPCs.
```
shantuo: We should have an issue to look into doing this at the docstore package level.

internal/docstore/firedocstore/fs.go
```diff
@@ -586,6 +588,65 @@ func processMods(mods []driver.Mod) (fields map[string]*pb.Value, maskPaths []st
...
+		if err := opts.BeforeDo(asFunc); err != nil {
+			return nil, err
+		}
+	}
+	res, err := c.client.Commit(withResourceHeader(ctx, req.Database), req)
+	if err != nil {
+		return nil, err
+	}
+	if len(res.WriteResults) != len(ws) {
+		return nil, gcerr.Newf(gcerr.Internal, nil, "wrong number of WriteResults from firestore commit")
```
shantuo: This error will be confusing when the `doCommitCall` method assign it to the no-precondition calls, as those writes can be separate. I don't think in practice this is a big issue but since we are making many assumptions on how the user is using the API, maybe we should 1) document the behavior; 2) provide a way to still let user run things in batches when they need that.

internal/docstore/firedocstore/fs.go
jba: Filed #2228.

internal/docstore/firedocstore/fs.go
jba: Really, this is an impossible error. It would mean that firestore is very broken. I only added this so that we don't panic. I think we can basically ignore it.

PR #2202, opened 2019-05-31
docstore: improve package documentation

jba:


internal/docstore/doc.go
```diff
@@ -13,38 +13,142 @@
...
+// memdocstore, an in-process, in-memory implementation suitable for testing and
+// development.
 //
+// Your application should import one of these provider-specific subpackages and use
+// its exported functions to create a *Collection; do not use the NewCollection
+// function in this package. For example:
 //
-// Representing Times
+//   coll, err := memdocstore.OpenBucket("SSN", nil)
+//     if err != nil {
```
vangent: Nit: something weird about indentation here.

internal/docstore/doc.go
```diff
@@ -13,38 +13,142 @@
...
+//
+//
+// Documents
+//
+// Docstore allows you to represent documents as either map[string]interface{} or
+// struct pointers. Using structs is recommended, because it enforces some structure
+// on your data. Docstore mimics the encoding/json package in its treatment of
+// structs: by default, a struct's exported fields are the fields of the document.
+// You can alter this default mapping by using a struct tag beginning with
+// "docstore:". Docstore struct tags support renaming, omitting fields
```
vangent: Side note: weren't we going to also obey the 'json:' tag?

internal/docstore/doc.go
```diff
@@ -13,38 +13,142 @@
...
+//
+// The value stored in a document field can be any of a wide range of types. All
+// primitive types except for complex numbers are supported, as well as slices and
+// maps (the map key type must be a string, an integer, or a type that implements
+// encoding.TextMarshaler). In addition, any type that implements
+// encoding.BinaryMarshaler or encoding.TextMarshaler is permitted. This set of types
+// closely matches the encoding/json package.
+//
+// Times deserve special mention. Docstore can store and retrieve values of type
+// time.Time, with two caveats. First, the timezone may not be preserved. Second,
```
vangent: Maybe we should make this "will not be preserved"?

internal/docstore/doc.go
```diff
@@ -13,38 +13,142 @@
...
+//
+// Here, the "SSN" field of the document is used as the key. Some providers let you
+// supply a function to extract the key from the document, which can be useful if the
+// key is composed of more than field.
+//
+//
+// Actions
+//
+// Docstore supports six actions on documents:
+// - Get retrieves a document.
```
vangent: I suggest you run `godoc` locally already and make sure the formatting of this looks OK. I think you might need to indent these by a space to make sure they are on separate lines...

internal/docstore/doc.go
```diff
@@ -13,38 +13,142 @@
...
 // revision field is unspecified. When defining a struct for storing docstore data,
 // define the field to be of type interface{}. For example,
-//    type User { Name string; DocstoreRevision interface{} }
+//
+//    type User {
+//      Name             string
+//      DocstoreRevision interface{}
+//    }
+//
 // If a struct doesn't have a DocstoreRevision field, then the logic described above
```
vangent: I feel like this caveat should be moved up to the beginning of the section. Something more like (please wordsmith, just suggesting the order):

Docstore supports revisions of documents based on a document field. By default it uses the field "DocstoreRevision" to hold the revision. Some providers give you the option....

If a struct doesn't have a revision field, no revisioning is performed. Otherwise, every document is given a revision when created .... etc.

internal/docstore/doc.go
jba: Yes. Filed #2219.

internal/docstore/doc.go
jba: done

internal/docstore/doc.go
jba: Done.

internal/docstore/doc.go
jba: Fixed.

Issue #2154, opened 2019-05-20
docstore: make revision field a collection option


jba:
When a collection is opened, let the user specify the name of the field that holds the document revision as an option.

I originally thought we could make this a struct tag, but we can't: we need for the entire collection to use the same field, so we can get and set it. As an example of what could go wrong if it were a struct tag, consider saving a document using a struct with a revision tag, then reading it back into a `map[string]interface{}`. How would we know which of the document fields is the revision? We need that information so we can set up the right precondition on subsequent writes.

PR #2203, opened 2019-05-31
memdocstore: provide an option for a different revision field

jba:
See #2154.


internal/docstore/memdocstore/mem.go
```diff
@@ -91,35 +94,40 @@ func OpenCollection(keyField string, _ *Options) (*docstore.Collection, error) {
...
 	return docstore.NewCollection(c), nil
 }
 
-func newCollection(keyField string, keyFunc func(docstore.Document) interface{}) (driver.Collection, error) {
+func newCollection(keyField string, keyFunc func(docstore.Document) interface{}, opts *Options) (driver.Collection, error) {
 	if keyField == "" && keyFunc == nil {
 		return nil, gcerr.Newf(gcerr.InvalidArgument, nil, "must provide either keyField or keyFunc")
 	}
+	if opts == nil {
+		opts = &Options{RevisionField: docstore.RevisionField}
```
vangent: I think you should create `&Options{}`, then check if `opts.RevisionField == ""` to set the default, in cases someone passes in a zero `Options`.

internal/docstore/memdocstore/mem.go
```diff
@@ -73,14 +73,17 @@ func (*URLOpener) OpenCollectionURL(ctx context.Context, u *url.URL) (*docstore.
 }
 
 // Options are optional arguments to the OpenCollection functions.
-type Options struct{}
+type Options struct {
+	// The name of the field holding the document revision.
```
vangent: Add "Defaults to docstore.RevisionField."

internal/docstore/memdocstore/mem.go
jba: Done.

internal/docstore/memdocstore/mem.go
jba: done.

PR #2204, opened 2019-05-31
internal/cmd/gocdk: re-tag image with gcr.io for Cloud Run

### LINKED ISSUE #2130: internal/cmd/gocdk: re-tag and publish image to GCR as part of Cloud Run launch

The Cloud Run launcher takes in a Docker image reference as input and it [must be accessible to the Cloud Run service](https://cloud.google.com/run/docs/deploying#images). The two ways to do this are either by having an image in Google Container Registry (GCR) or on a public Docker registry like Docker Hub (but only allowed when running Cloud Run in Kubernetes Engine).

In order to push a local Docker image to GCR, the Docker image needs to be [tagged with one of the `gcr.io/` prefixes](https://cloud.google.com/container-registry/docs/pushing-and-pulling#tag_the_local_image_with_the_registry_name). The Dockerfile we generate in init deliberately uses a vendor-neutral tag, and the Docker image may be used in several different contexts, so it would be nice to avoid polluting the biome-independent Docker image with biome-dependent Docker tags.

When the Cloud Run launcher runs, the Docker image passed to Cloud Run should be the first of:

1. If the launch specifier contains an `image_name` argument, that will be used. `gocdk launch` should return an error if the `image_name` does not start with one of the `gcr.io/` prefixes.
2. If the image name provided to `gocdk launch` starts with one of the `gcr.io/` prefixes, then it will be used directly.
2. Otherwise, `gcr.io/PROJECT_ID/` will be prepended to the image name given, where `PROJECT_ID` is the project given in the launch specifier `project_id` parameter.

When executing the launch, if the image reference given to `gocdk launch` does not have the same name as the one chosen above, then `gocdk launch` will re-tag the image given to `gocdk launch` with the desired `gcr.io/`-based image name on the local Docker daemon, then push the image from the local Docker daemon to GCR. This assumes the user has already [configured `gcloud` as a Docker credentials helper](https://cloud.google.com/container-registry/docs/advanced-authentication).

As mentioned above, we could kludge around this by making the Dockerfile magic comment include a gcr.io prefix, but I'd rather avoid that to keep the user's application image free of vendor prefixes. (e.g. in the case where they want to launch on multiple clouds).

### Additional context

This is a follow-up to [#2106](https://github.com/google/go-cloud/pull/2106#discussion_r284441333).

zombiezen:
Fixes #2130

### DIFF
--- a/internal/cmd/gocdk/launch.go
+++ b/internal/cmd/gocdk/launch.go
@@ -235,12 +235,11 @@ func (crl *cloudRunLauncher) Launch(ctx context.Context, input *LaunchInput) (*u
 	}
 
 	// Push to GCR.
-	localImage := input.DockerImage
-	remoteImage, err := cloudRunImage(localImage, input.Specifier)
+	imageRef, err := crl.tagForCloudRun(ctx, input.DockerImage, input.Specifier)
 	if err != nil {
 		return nil, xerrors.Errorf("cloud run launch: %w", err)
 	}
-	if err := crl.dockerPush(ctx, localImage, remoteImage); err != nil {
+	if err := crl.dockerPush(ctx, imageRef); err != nil {
 		return nil, xerrors.Errorf("cloud run launch: %w", err)
 	}
 
@@ -269,7 +268,7 @@ func (crl *cloudRunLauncher) Launch(ctx context.Context, input *LaunchInput) (*u
 				RevisionTemplate: &cloudrun.RevisionTemplate{
 					Spec: &cloudrun.RevisionSpec{
 						Container: &cloudrun.Container{
-							Image: remoteImage,
+							Image: imageRef,
 							Env:   env,
 							// TODO(light): Add liveness and readiness probes.
 						},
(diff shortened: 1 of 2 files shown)


internal/cmd/gocdk/launch.go
```diff
@@ -364,6 +377,75 @@ func (crl *cloudRunLauncher) dockerPush(ctx context.Context, imageName string) e
...
+			return "", xerrors.New("determine image name for Cloud Run: launch specifier project_id empty")
+		}
+		// TODO(light): This can be wrong for ORG:PROJECT project IDs, but those
+		// are deprecated anyway.
+		name = "gcr.io/" + project + "/" + name
+	}
+	return name + tag + digest, nil
+}
+
+// parseImageRef parses a Docker image reference, as documented in
```
vangent: Isn't there a library that does this kind of stuff?
E.g.,
  https://github.com/docker/distribution/blob/master/reference/reference.go
  https://github.com/mafintosh/docker-parse-image

zombiezen: Yes, although as noted, this one intentionally implements a looser version of that spec. I did verify that this would not add too many "go.mod-finding" deps, so what you're saying is an option. Overall, I don't think there's much benefit in using a separate library for this single function.

(The second link is a JavaScript library.)

zombiezen:
@clausti I have moved the tagging+rewriting operations together as suggested offline. PTAL.

PR #2207, opened 2019-05-31
docstore: examples

jba:


internal/docstore/docstore.go
```diff
@@ -182,8 +182,9 @@ func (l *ActionList) Get(doc Document, fps ...FieldPath) *ActionList {
 // mod "a.b": 2, then either Update will fail, or it will succeed with the result
 // {a: {b: 2}}.
 //
-// Update does not modify its doc argument. To obtain the new value of the document,
-// call Get after calling Update.
+// Update does not modify its doc argument, except to set the new revision. To obtain
+// the new value of the document, call Get after calling Update.
```
vangent: Should this be "the new revision value"?

jba: No, I meant the entire contents of the document on the server after the update.

Reworded.

PR #2223, opened 2019-05-31
docstore: support "json:" struct tags

jba:


vangent:
Update the godoc that mentions `docstore` as well.

PR #2226, opened 2019-06-03
internal/docstore/dynamodocstore: use BatchGet API

### LINKED ISSUE #2150: docstore/dynamo: compare perfomance of TransactGetItems and BatchGetItem
Right now we issue multiple Gets with TransactGetItems. Would it be faster to use BatchGetItem? If so, we should, since we don't promise atomicity anyway.

shantuo:
Fixes #2150


shantuo:
Benchmark numbers for 100-gets:

```
100-Gets (Transaction)

goos: darwin
goarch: amd64
pkg: gocloud.dev/internal/docstore/dynamodocstore
BenchmarkConformance/BenchmarkActionListGet-8                 10         182787465 ns/op

100-Gets(BatchGet)

goos: darwin
goarch: amd64
pkg: gocloud.dev/internal/docstore/dynamodocstore
BenchmarkConformance/BenchmarkActionListGet-8                 30          41605736 ns/op
```

PR #2234, opened 2019-06-03
internal/gocdk: go through pctx for stderr/stdout

### LINKED ISSUE #2038: internal/cmd/gocdk: create processContext logger

vangent:
Fixes #2038.


internal/cmd/gocdk/main.go
```diff
@@ -124,6 +138,38 @@ func (pctx *processContext) gcpCredentials(ctx context.Context) (*google.Credent
...
+
+// ApplyToCmd updates fields of cmd for this processContext.
+//
+// dir may be empty; it defaults to pctx.workdir.
+//
+// Note that this function sets cmd.Stdout, and that Cmd.Output requires
+// Stdout to be nil; you may need to set it back to nil explicitly (or
+// don't use this function). Similarly for Cmd.CombinedOutput, for both
+// Stdout and Stderr.
+func (pctx *processContext) ApplyToCmd(cmd *exec.Cmd, dir string) {
```
zombiezen: Nice, I've been thinking of creating a similar function for a while.

"Apply" didn't evoke for me what this function actually does. What if we rearranged this to have the same signature as `os/exec.CommandContext`, but filling in all the fields like you have? Then this would be `pctx.NewCommand(ctx, "terraform", "output")` etc.

internal/cmd/gocdk/main.go
```diff
@@ -95,11 +98,22 @@ func osProcessContext() (*processContext, error) {
...
 }
 
+// newTestProcessContext returns a processContext for use in tests.
+func newTestProcessContext(workdir string) *processContext {
+	return &processContext{
+		workdir: workdir,
+		env:     os.Environ(),
+		stdin:   strings.NewReader(""),
+		outlog:  log.New(ioutil.Discard, "", 0),
+		errlog:  log.New(ioutil.Discard, "gocdk: ", 0),
```
zombiezen: nit: AFAICT, no prefix required for a log that will be discarded.

internal/cmd/gocdk/main.go
```diff
@@ -81,12 +84,12 @@ type processContext struct {
...
-	stdout  io.Writer
-	stderr  io.Writer
+	outlog  *log.Logger
+	errlog  *log.Logger
 }
 
-// osProcessContext returns the default process context from global variables.
-func osProcessContext() (*processContext, error) {
+// newDefaultProcessContext returns the default processContext from global variables.
+func newDefaultProcessContext() (*processContext, error) {
```
zombiezen: Why the change from `osProcessContext`? It uses the operating system process handles, so the name seems fitting.

internal/cmd/gocdk/main.go
vangent: Nice, I like it. Done.

internal/cmd/gocdk/main.go
vangent: OK, done.

internal/cmd/gocdk/main.go

PR #2238, opened 2019-06-04
internal/docstore/mongodocstore: use find to do bulk get

### LINKED ISSUE #2143: docstore/mongodocstore: use Find for unordered gets
Benchmark numbers for 50-get actions:
```
BenchmarkConformance/BenchmarkActionListGet-8           200	  10493466 ns/op
```

I experimented with using `Find` and an `$in` filter, the number is:
```
BenchmarkConformance/BenchmarkActionListGet-8          2000        565110 ns/op
```

The performance boosted from ~10 ms to ~0.5 ms, plus we don't need to manage plenty of single get requests.

There are some problems I still need to figure out:
1. Find takes an option with field paths selections, which means if the user wants different field paths, one Find results cannot fulfill that request.
   - We can split into groups by requested field paths. I suspect the user would likely to write those gets in a loop and the field paths would be the same. So this approach doesn't look that bad than what it appears to be.
   - We can request all field paths and decode later. If we go with the previous assumption, this would be wasting the effort.
   - We can simply give up and fall back to use single gets when we found those gets are not towards the same set of field paths.
1. The find will return a cursor and we need to match the documents with the order the user requests.
   - We can create a list with indices of the actions sorted by the IDs and pass the same order option into Find, then map the order to the original action list.

cc @jba  @vangent

shantuo:
Fixes #2143 

```
With FindOne:

goos: darwin
goarch: amd64
pkg: gocloud.dev/internal/docstore/mongodocstore
BenchmarkConformance/BenchmarkActionListGet-8           	     200	  10493466 ns/op

With Find:

goos: darwin
goarch: amd64
pkg: gocloud.dev/internal/docstore/mongodocstore
BenchmarkConformance/BenchmarkActionListGet-8                       2000            935609 ns/op
```


internal/docstore/mongodocstore/mongo.go
```diff
@@ -251,47 +251,78 @@ type indexedError = struct {
...
+		return
 	}
-	var m map[string]interface{}
-	if err := res.Decode(&m); err != nil {
-		return err
+	defer cursor.Close(ctx)
+
+	found := make([]bool, len(gets))
+	am := mapActionIndices(gets)
+	n := 0
```
jba: Is `n` ever used?

internal/docstore/mongodocstore/mongo.go
```diff
@@ -251,47 +251,78 @@ type indexedError = struct {
...
+		setErr(err)
+		return
 	}
-	var m map[string]interface{}
-	if err := res.Decode(&m); err != nil {
-		return err
+	defer cursor.Close(ctx)
+
+	found := make([]bool, len(gets))
+	am := mapActionIndices(gets)
```
jba: Could be a map from ID to *driver.Action.

jba: Then `found` would be a `map[*driver.Action]bool`.

shantuo: Done.

Also I encode the `Action.Key` when doing the encoding otherwise there might be type mismatch.

PR #2241, opened 2019-06-04
internal/gocdk: consolidate some helpers && verify project root dir

### LINKED ISSUE #2210: internal/cmd/gocdk: demo add proceeds even outside a project directory
### Describe the bug

The `gocdk demo add` command unconditionally adds a file to the current directory, regardless of whether it is in a project or not.

/cc @vangent 

### To Reproduce

```
cd random/dir
gocdk demo add blob
```

creates a demo_blob.go file.

## Expected behavior

An error is returned: the file should only be inserted at the root of the current module, of which we are not in one.

vangent:
Fixes #2210.

I started fixing #2210 (about "demo add" happening even when you're not in the project directory) and kind of went down a rabbit hole:

* I ended up modifying `findModuleRoot` to do the check for whether we're in a Go CDK directory, since we pretty much want this everywhere. I also renamed it to `ModuleRoot` on `processContext` since we always call it with `pctx.workdir`.

* I also renamed `findBiomeDir()` to `biomeDir()` and added `biomeRootDir()`. The `find` seems unnecessary since it's just constructing the path, not finding anything.

* I got rid of the `newTestModule` helper in favor of using `newTestProject`, which uses `gocdk init` to create a Go CDK project. This simplified a few tests where we were having to create `biomes` files etc. manually. In particular, `newTestBiome` is no longer needed.

* Fixed a bug in `biome add` where an error wasn't being returned.

* Simplified list of available demos; we just need the demo name, everything else can be computed.

### DIFF
--- a/internal/cmd/gocdk/biome.go
+++ b/internal/cmd/gocdk/biome.go
@@ -36,14 +36,14 @@ type biomeConfig struct {
 	Launcher     *string `json:"launcher,omitempty"`
 }
 
-// biomeRootDir returns the path to the biomes directory.
-func biomeRootDir(moduleRoot string) string {
+// biomesRootDir returns the path to the biomes directory.
+func biomesRootDir(moduleRoot string) string {
 	return filepath.Join(moduleRoot, "biomes")
 }
 
 // biomeDir returns the path to the named biome.
 func biomeDir(moduleRoot, name string) string {
-	return filepath.Join(biomeRootDir(moduleRoot), name)
+	return filepath.Join(biomesRootDir(moduleRoot), name)
 }
 
 // readBiomeConfig reads and parses the biome configuration from the filesystem.
--- a/internal/cmd/gocdk/biome.go
+++ b/internal/cmd/gocdk/biome.go
@@ -36,16 +36,21 @@ type biomeConfig struct {
 	Launcher     *string `json:"launcher,omitempty"`
 }
 
-// findBiomeDir returns the path to the named biome.
-func findBiomeDir(moduleRoot, name string) string {
-	return filepath.Join(moduleRoot, "biomes", name)
+// biomeRootDir returns the path to the biomes directory.
+func biomeRootDir(moduleRoot string) string {
+	return filepath.Join(moduleRoot, "biomes")
+}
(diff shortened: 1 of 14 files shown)


internal/cmd/gocdk/biome.go
```diff
@@ -36,16 +36,21 @@ type biomeConfig struct {
 	Launcher     *string `json:"launcher,omitempty"`
 }
 
-// findBiomeDir returns the path to the named biome.
-func findBiomeDir(moduleRoot, name string) string {
-	return filepath.Join(moduleRoot, "biomes", name)
+// biomeRootDir returns the path to the biomes directory.
+func biomeRootDir(moduleRoot string) string {
```
clausti: The root directory is named `biomes` rather than `biome`, should we make that match here?

internal/cmd/gocdk/demo.go
```diff
@@ -89,52 +67,52 @@ func registerDemoCmd(ctx context.Context, pctx *processContext, rootCmd *cobra.C
...
 // instantiateDemo does all of the work required to add a demo of a
 // portable API to the user's project.
 // TODO(rvangent): It currently copies a single source code file. It should
 // additionally iterate over existing biomes, adding a config entry and possibly
 // Terraform files.
-func instantiateDemo(pctx *processContext, demo *demoInfo, force bool) error {
+func instantiateDemo(pctx *processContext, moduleRoot, demo string, force bool) error {
 	logger := log.New(pctx.stderr, "gocdk: ", log.Ldate|log.Ltime)
-	logger.Printf("Adding %q...", demo.name)
+	logger.Printf("Adding %q...", demo)
```
clausti: should this be using the `pctx` logger?

internal/cmd/gocdk/biome.go
vangent: Sure, done.

internal/cmd/gocdk/demo.go
vangent: Yes, but that is not checked in yet. I will resolve conflicts like this before submitting depending on the submit order.

PR #2246, opened 2019-06-04
testing: refactor release helper and switch flags to commands

eliben:
Refactoring prepares for adding new commands, such as bumping versions
for release testing.

### DIFF
--- a/internal/testing/releasehelper.go
+++ b/internal/testing/releasehelper.go
@@ -12,13 +12,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Helper tool for creating new releases of the Go CDK. See -help for details.
+// Helper tool for creating new releases of the Go CDK. Run without arguments
+// or with 'help' for details.
 package main
 
 import (
 	"bufio"
 	"encoding/json"
-	"flag"
 	"fmt"
 	"log"
 	"os"
@@ -27,7 +27,7 @@ import (
 	"strings"
 )
 
-// TODO(eliben): add bumpversion flag that bumps version by 1, or to a version
+// TODO(eliben): add bumpversion command that bumps version by 1, or to a version
 // given on the command line
 var helpText string = `
 Helper tool for creating new releases of the Go CDK.
@@ -39,16 +39,27 @@ The tool processes all modules listed in the 'allmodules'
 file. For each module it handles all dependencies on
 other gocloud.dev modules.
 
--replace adds 'replace' directives to point to local versions
-for testing. -dropreplace removes these directives.
+Run it from the root directory of the project, as follows:
+
+$ %s <command>
+
+Where command is one of the following:
+
+  addreplace    adds 'replace' directives to point to local versions
+                for testing.
+
+  dropreplace   removes these directives.
+
+  help          prints this usage message
 
 Run it from the root directory of the project.
-
-Usage:
 `
 
-var doReplace = flag.Bool("replace", false, "add replace directives")
-var doDropReplace = flag.Bool("dropreplace", false, "drop replace directives")
+func printHelp() {
+	_, binName := filepath.Split(os.Args[0])
+	fmt.Printf(helpText, binName)
+	fmt.Println()
+}
 
 // cmdCheck invokes the command given in s, echoing the invocation to stdout.
 // It checks that the command was successful and returns its standard output.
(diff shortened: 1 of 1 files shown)


internal/testing/releasehelper.go
```diff
@@ -132,24 +153,32 @@ func runOnGomod(path string) {
...
-			} else if *doDropReplace {
-				cmdCheck(fmt.Sprintf("go mod edit -dropreplace=%s %s", r.Path, gomodPath))
-			}
+			reqHandler(gomodPath, r.Path, rel)
 		}
 	}
 }
 
 func main() {
-	flag.Usage = func() {
```
vangent: Consider using `github.com/google/subcommands`. It does a better job at `help` than you, and will set up the commands as struct that can define their own flags, etc.

https://github.com/google/go-cloud/blob/master/samples/gocdk-secrets/main.go has a simple example.

eliben: I considered it, and I'd rather not. Right now these simple helper tools are single-file. If they grow dependencies they'll need to be split into separate modules (or some unified `tools` module). 

Moreover, it's not clear to me what the advantage of `subcommands` is for this specific tool. It seems pretty trivial to me :man_shrugging:  Commands can have their own flags with `flag.NewFlagSet` if needed.

Besides, we use cobra for the gocdk which makes sense, but how many different subcommand packages we really need in a single project? I'd vote to only use cobra for complicated stuff like the CLI and the standard library for everything else.

internal/testing/releasehelper.go
```diff
@@ -39,16 +39,27 @@ The tool processes all modules listed in the 'allmodules'
...
+
+  addreplace    adds 'replace' directives to point to local versions
+                for testing.
 
-Usage:
+  dropreplace   removes these directives.
+
+  help          prints this usage message
+
+Run it from the root directory of the project.
```
vangent: project -> repository ?

internal/testing/releasehelper.go
```diff
@@ -39,16 +39,27 @@ The tool processes all modules listed in the 'allmodules'
...
 `
 
-var doReplace = flag.Bool("replace", false, "add replace directives")
-var doDropReplace = flag.Bool("dropreplace", false, "drop replace directives")
+func printHelp() {
+	_, binName := filepath.Split(os.Args[0])
```
vangent: Sorry, what does this do? Why not just `fmt.Println(helpText)`? And maybe `fmt.Fprintf(os.Stderr` ?

internal/testing/releasehelper.go
```diff
@@ -132,24 +153,32 @@ func runOnGomod(path string) {
...
+	case "addreplace":
+		handlerFunc = func(gomodPath string, mod string, modPath string) {
+			cmdCheck(fmt.Sprintf("go mod edit -replace=%s=%s %s", mod, modPath, gomodPath))
+		}
+	case "dropreplace":
+		handlerFunc = func(gomodPath string, mod string, modPath string) {
+			cmdCheck(fmt.Sprintf("go mod edit -dropreplace=%s %s", mod, gomodPath))
+		}
+	default:
+		log.Fatal("Expected 'addreplace' or 'dropreplace'")
```
vangent: Print usage here and exit(1)?

internal/testing/releasehelper.go
```diff
@@ -132,24 +153,32 @@ func runOnGomod(path string) {
...
-	flag.Parse()
-	if *doReplace && *doDropReplace {
-		log.Fatal("Expected only one of -replace/-dropreplace")
+	var handlerFunc reqHandlerFunc
+	switch os.Args[1] {
+	case "help":
+		printHelp()
+		os.Exit(0)
+	case "addreplace":
+		handlerFunc = func(gomodPath string, mod string, modPath string) {
```
vangent: Nit: `func(gomodPath, mod, modPath string)` here and below.

internal/testing/releasehelper.go

internal/testing/releasehelper.go
eliben: Without this the command is shown as `/tmp/go-build736844560/b001/exe/releasehelper` when running with `go run`.

Just figured it's nice to have the command name there for usage.

internal/testing/releasehelper.go

internal/testing/releasehelper.go

internal/testing/releasehelper.go
vangent: Ah, I see, I missed that `helpText` had a '%s' in it.

Issue #2046, opened 2019-05-10
internal/cmd/gocdk: factor out Launcher into its own package


zombiezen:
The code for supporting `gocdk launch` is likely going to grow a bit beyond the support needed for other CLI subcommands. For code organization, it would be helpful to move the `Launcher` interface and its implementations into a separate package.

zombiezen:
I've spent a little bit of time over last week thinking about this. In terms of code motion, I think the main goal is to move utility code away from command-line/argument-parsing code. As such, my plan is to:

1. Factor out all Docker subprocess logic together into a separate package. (This also moves us closer for #1804, should we choose to go that way.) There's a bunch of cross-cutting code (especially in launch) that would benefit from having a type-safe wrapper around the Docker CLI invocations, and would address the concern around passing `dockerEnv []string` variables around in a neater way.
2. Have one PR that is primarily focused at moving the relevant functions from launch.go into its own package. This should not change the API where possible (might need to export fields temporarily for visibility), but largely be a no-op code shuffle.
3. Have a final PR that makes API changes that we think are the right ones, now that the code is mostly in the right place.

@clausti, @vangent, WDYT?

vangent:
SGTM. For, 3), I still think it's interesting to think about whether you could write an implementation of the `Launcher` API based on a plug-in CLI. I think that's mostly about make the arguments relatively simple to serialize.

clausti:
SGTM, too.

zombiezen:
> For, 3), I still think it's interesting to think about whether you could write an implementation of the `Launcher` API based on a plug-in CLI. I think that's mostly about make the arguments relatively simple to serialize.

Ah, I think this is actually the intuitive reason I structured `LaunchInput` the way I did: these are all the arguments to a launch (i.e. a Docker image reference, `launch_environment`, and `launch_specifier`) and come in through deserialization.

We could definitely add such a structure later, but it's unclear to me what benefit it would provide over a CLI that reads these inputs from files or `terraform output` itself (as that is the same contract we are holding ourselves to).

zombiezen:
I've split out the redesign of the launcher API into a separate issue (#2376) and am marking this issue closed: `Launcher` is now in package `launcher`.

Issue #1804, opened 2019-04-11
internal/cmd/gocdk: investigate use of Docker client API


zombiezen:
Is there a way to interact w/ docker via a client API? That could be easier to test/mock

_Originally posted by @eliben in https://github.com/google/go-cloud/pull/1796_

zombiezen:
Random observation: using the HTTP API might be okay for most operations, but it would be hard to replicate `docker build`, since the [build context input](https://docs.docker.com/engine/api/v1.39/#operation/ImageBuild) must be a TAR file of the directory, respecting the `.dockerignore` file.

zombiezen:
Ultimately, I don't think this makes sense to implement. Shelling out to the Docker CLI works fine for our use case and does not add undue burden to our users.

PR #2248, opened 2019-06-04
internal/cmd/gocdk: factor Docker interactions into separate package

zombiezen:
The first in the sequence of PRs I outlined on #2046.

Updates #1804
Updates #2046


internal/cmd/gocdk/internal/docker/docker.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// New creates a new client with the given environment variables.
+// Calling docker.New(nil) is the same as new(docker.Client) and uses the
+// current process's environment.
+func New(env []string) *Client {
+	return &Client{env: env}
+}
+
+// Image stores metadata about a Docker image.
```
vangent: Can you document the fields, or maybe link to Docker documentation that explains them?

internal/cmd/gocdk/internal/docker/docker.go
```diff
@@ -0,0 +1,211 @@
...
+type Image struct {
+	ID         string
+	Repository string
+	Tag        string
+	Digest     string
+	CreatedAt  time.Time
+}
+
+// ListImages list images the Docker daemon has stored locally that match filterRef.
```
vangent: nit: list -> lists

internal/cmd/gocdk/internal/docker/docker.go
```diff
@@ -0,0 +1,211 @@
...
+			args = append(args, "--publish", p)
+		}
+	}
+	args = append(args, "--", imageRef)
+	if opts != nil {
+		args = append(args, opts.Args...)
+	}
+
+	cmd := exec.CommandContext(ctx, "docker", args...)
+	cmd.Env = c.env
```
vangent: Nit: consider factoring these lines out? I think this is exactly the same as `ListImages` above (lines 134-143).

internal/cmd/gocdk/internal/docker/docker.go
```diff
@@ -0,0 +1,211 @@
...
+	stderr := new(bytes.Buffer)
+	cmd.Stderr = stderr
+	out, err := cmd.Output()
+	if err != nil {
+		if stderr.Len() > 0 {
+			return nil, xerrors.Errorf("docker images:\n%s", stderr)
+		}
+		return nil, xerrors.Errorf("docker images: %w", err)
```
vangent: Do we need these prefixes on the errors? E.g., the actual error will now look something like:
`gocdk build: docker images: blah blah`. Isn't `gocdk build:` enough? If anything, it might be useful to log the exact command line.

internal/cmd/gocdk/build.go
```diff
@@ -80,10 +83,17 @@ func listBuilds(ctx context.Context, pctx *processContext) error {
...
 	}
-	c := pctx.NewCommand(ctx, "", "docker", "images", imageName)
-	if err := c.Run(); err != nil {
+	images, err := docker.New(pctx.env).ListImages(ctx, imageName)
+	if err != nil {
 		return xerrors.Errorf("list builds: %w", err)
 	}
+	for _, image := range images {
+		if image.Repository == "" || image.Tag == "" {
+			fmt.Fprintf(pctx.stdout, "@%-60s  %s\n", image.Digest, image.CreatedAt.Local().Format(time.Stamp))
```
vangent: Use `pctx.Printf` now.

internal/cmd/gocdk/build.go
```diff
@@ -80,10 +83,17 @@ func listBuilds(ctx context.Context, pctx *processContext) error {
...
 		return xerrors.Errorf("list builds: %w", err)
 	}
-	c := pctx.NewCommand(ctx, "", "docker", "images", imageName)
-	if err := c.Run(); err != nil {
+	images, err := docker.New(pctx.env).ListImages(ctx, imageName)
+	if err != nil {
 		return xerrors.Errorf("list builds: %w", err)
 	}
+	for _, image := range images {
+		if image.Repository == "" || image.Tag == "" {
```
vangent: Just for my understanding, this is reformatting the output, right? Before we just dumped whatever `docker images` outputted, now we are parsing that and then reformatting? Is that necessary/desirable?

internal/cmd/gocdk/build.go
```diff
@@ -64,8 +67,8 @@ func build(ctx context.Context, pctx *processContext, ref string) error {
 		}
 		ref = imageName + ref
 	}
-	c := pctx.NewCommand(ctx, moduleRoot, "docker", "build", "--tag", ref, ".")
-	if err := c.Run(); err != nil {
+	err = docker.New(pctx.env).Build(ctx, ref, moduleRoot, pctx.stderr)
```
vangent: Nit: `if err := docker.New(...); err != nil {}`

internal/cmd/gocdk/internal/docker/docker.go
```diff
@@ -0,0 +1,211 @@
...
+type Image struct {
+	ID         string
+	Repository string
+	Tag        string
+	Digest     string
+	CreatedAt  time.Time
+}
+
+// ListImages list images the Docker daemon has stored locally that match filterRef.
+func (c *Client) ListImages(ctx context.Context, filterRef string) ([]Image, error) {
```
vangent: `*Image`? Not sure there's any advantage to copying.

internal/cmd/gocdk/internal/docker/docker.go
```diff
@@ -0,0 +1,211 @@
...
+type RunOptions struct {
+	Args         []string
+	Env          []string
+	RemoveOnExit bool
+	Publish      []string
+}
+
+// Start starts a Docker container. It does not wait for the container to stop.
+// Start returns the container's ID on success.
+func (c *Client) Start(ctx context.Context, imageRef string, opts *RunOptions) (string, error) {
```
vangent: `Run`? It's a bit weird for `Start` to take `RunOptions` and call `docker run`.

internal/cmd/gocdk/launch.go
```diff
@@ -172,30 +169,16 @@ func (local *localLauncher) Launch(ctx context.Context, input *LaunchInput) (*ur
...
-	}
-	for _, v := range input.Env {
-		dockerArgs = append(dockerArgs, "--env", v)
-	}
-	dockerArgs = append(dockerArgs, "--env", "PORT=8080")
-	dockerArgs = append(dockerArgs, input.DockerImage)
-
-	c := exec.CommandContext(ctx, "docker", dockerArgs...)
-	c.Env = local.dockerEnv
-	c.Dir = local.dockerDir
```
vangent: IIUC, we lost this (setting `exec.Cmd.Dir` to `pctx.workdir`, previously via `dockerDir`. Was it not needed?

internal/cmd/gocdk/launch.go
```diff
@@ -238,7 +220,7 @@ func (crl *cloudRunLauncher) Launch(ctx context.Context, input *LaunchInput) (*u
 	if err != nil {
 		return nil, xerrors.Errorf("cloud run launch: %w", err)
 	}
-	if err := crl.dockerPush(ctx, imageRef); err != nil {
+	if err := crl.dockerClient.Push(ctx, imageRef, ioutil.Discard); err != nil {
```
vangent: Does `Push` need the arg then? IIUC this is the only place it's used.

internal/cmd/gocdk/build.go
zombiezen: Done.

internal/cmd/gocdk/build.go
zombiezen: Correct. Parsing this output is necessary for some of the other logic we have been discussing for the future (more reliable Cloud Run pushes, for instance). I don't really want to have two code paths for listing Docker images, so I figured that I'd make the more structured one here.

internal/cmd/gocdk/build.go
zombiezen: Thanks! Done.

internal/cmd/gocdk/internal/docker/docker.go
zombiezen: Done.

internal/cmd/gocdk/internal/docker/docker.go
zombiezen: Done.

internal/cmd/gocdk/internal/docker/docker.go
zombiezen: Done.

internal/cmd/gocdk/internal/docker/docker.go
zombiezen: It is helpful for `gocdk launch` for instance, where there is a `docker tag` followed by a `docker push`, to know the operation that failed. Exact command line can be a little verbose, so this seemed like a reasonable compromise.

For this exact operation, I am seeing that `docker images` isn't as helpful as `list docker images`, so changed to that.

internal/cmd/gocdk/internal/docker/docker.go
zombiezen: Done.

internal/cmd/gocdk/launch.go
zombiezen: It was not needed. The working directory that we run the `docker` CLI in is only relevant for `docker build`.

internal/cmd/gocdk/launch.go
zombiezen: This function doesn't have direct access to the `processContext` and I didn't want to plumb through the `io.Writer` since I was intending to airlift this code to another package anyway. Adding a TODO for now.

internal/cmd/gocdk/build.go
vangent: SGTM.

internal/cmd/gocdk/internal/docker/docker.go
vangent: OK.

internal/cmd/gocdk/internal/docker/docker.go
vangent: Ping. Either way is fine.

go.mod
```diff
@@ -40,9 +40,9 @@ require (
 	github.com/google/wire v0.2.2
 	github.com/googleapis/gax-go v2.0.2+incompatible
 	github.com/lib/pq v1.1.1
-	github.com/mitchellh/go-homedir v1.1.0 // indirect
 	github.com/opentracing/opentracing-go v1.1.0 // indirect
 	github.com/pkg/errors v0.8.1 // indirect
+	github.com/spf13/cobra v0.0.4 // indirect
```
vangent: These changes to go.mod and go.sum don't belong here.

internal/cmd/gocdk/internal/docker/docker.go
zombiezen: Missed this one, sorry. `Start` seems better to me because it implies the right semantics about not waiting for the container to stop. You could imagine a `Run` operation that does not set the `--detach` flag, but we don't have a need for one right now. Both of these would take in `RunOptions`.

internal/cmd/gocdk/internal/docker/docker.go
```diff
@@ -0,0 +1,228 @@
...
+
+// RunOptions holds optional parameters for starting a Docker container.
+type RunOptions struct {
+	Args         []string
+	Env          []string
+	RemoveOnExit bool
+	Publish      []string
+}
+
+// Start starts a Docker container. It does not wait for the container to stop.
```
clausti: As a reader of this documentation comment, I am very confused about the `It does not wait for the container to stop.` Should I be worried about it waiting for something to stop?/I don't understand why it's telling me this.

go.mod
zombiezen: Thanks. Sigh... I'm not sure what keeps updating these things, and I have to keep actively fighting to keep these from being checked in.

internal/cmd/gocdk/internal/docker/docker.go
zombiezen: Removed the sentence.

internal/cmd/gocdk/internal/docker/docker.go
```diff
@@ -0,0 +1,227 @@
...
+	if len(out) == 0 {
+		return "", xerrors.Errorf("docker run %s: docker did not return a container ID", imageRef)
+	}
+	return string(out), nil
+}
+
+// Build builds a Docker image from a source directory, optionally tagging it
+// with the given image reference. Output is streamed to the buildOutput writer.
+func (c *Client) Build(ctx context.Context, imageRef string, dir string, buildOutput io.Writer) error {
+	cmd := exec.CommandContext(ctx, "docker", "build", "-t="+imageRef, "--", dir)
```
clausti: What is the `--` (5th argument) doing here?  If I understand correctly, this constructs something that would look on a command like like `docker build -t=myimageref:tag -- path/somewhere, but I don't see an example of that usage on https://docs.docker.com/engine/reference/commandline/build/,  so I'm a little confused.

zombiezen: It prevents the subsequent arguments being interpreted as command-line flags, so if for instance the user tries to build a directory that starts with a hyphen, the `docker build` command would understand it properly.

I don't have a Docker-specific reference that documents this behavior, but is a behavior emulated by [GNU getopt](https://www.gnu.org/software/libc/manual/html_node/Using-Getopt.html)-compatible flag-parsing libraries.

internal/cmd/gocdk/internal/docker/docker.go
```diff
@@ -0,0 +1,227 @@
...
+	cmd.Env = c.env
+	if err := cmd.Run(); err != nil {
+		return xerrors.Errorf("docker build: %w", err)
+	}
+	return nil
+}
+
+// Tag adds a new tag to a given image reference.
+func (c *Client) Tag(ctx context.Context, srcImageRef, dstImageRef string) error {
+	cmd := exec.CommandContext(ctx, "docker", "tag", "--", srcImageRef, dstImageRef)
```
clausti: Similar question here about what is `--` doing? Is it something for Cobra?

Issue #2249, opened 2019-06-05
internal/docstore/mongodocstore: panic during conformance test


vangent:
Got this failure on Travis:

```
=== RUN   TestConformance/Create
panic: reflect: call of reflect.Value.Bytes on array Value
	panic: reflect: call of reflect.Value.Bytes on array Value [recovered]
	panic: reflect: call of reflect.Value.Bytes on array Value
goroutine 16 [running]:
testing.tRunner.func1(0xc00012c400)
	/home/travis/.gimme/versions/go1.12.5.linux.amd64/src/testing/testing.go:830 +0x69d
panic(0x1ed94c0, 0xc00000fb40)
	/home/travis/.gimme/versions/go1.12.5.linux.amd64/src/runtime/panic.go:522 +0x1b5
reflect.flag.mustBe(0x91, 0x17)
	/home/travis/.gimme/versions/go1.12.5.linux.amd64/src/reflect/value.go:207 +0xee
reflect.Value.Bytes(0x1fc9360, 0xc000044800, 0x91, 0x2834fc0, 0x1fc9360, 0x1cc4acc)
	/home/travis/.gimme/versions/go1.12.5.linux.amd64/src/reflect/value.go:260 +0x46
gocloud.dev/internal/docstore/driver.encodeList(0x1fc9360, 0xc000044800, 0x91, 0x2826c20, 0xc00000fb20, 0x0, 0x0)
	/home/travis/gopath/src/gocloud.dev/internal/docstore/driver/codec.go:206 +0x2d3
gocloud.dev/internal/docstore/driver.encode(0x1fc9360, 0xc000044800, 0x91, 0x2826c20, 0xc00000fb20, 0x20, 0x1fa6900)
	/home/travis/gopath/src/gocloud.dev/internal/docstore/driver/codec.go:173 +0xd7a
gocloud.dev/internal/docstore/driver.Encode(0x1fc9360, 0xc000044800, 0x91, 0x2826c20, 0xc00000fb20, 0x0, 0x0)
	/home/travis/gopath/src/gocloud.dev/internal/docstore/driver/codec.go:115 +0x7d
gocloud.dev/internal/docstore/mongodocstore.encodeValue(0x1fc9360, 0xc000044800, 0xc0001446c0, 0x0, 0x0, 0x0)
	/home/travis/gopath/src/gocloud.dev/internal/docstore/mongodocstore/codec.go:44 +0xe5
[...]

PR #2250, opened 2019-06-05
mongodocstore: be more precise about encoding keys

jba:
- Fix bug in driver codec where byte arrays are mistakenly encoded as byte slices.

- Distinguish between encoded and unencoded document IDs.

Also, remove driver code for len(mods) == 0 because that can't happen.

And some other minor changes.


jba:
Fixes #2249.

PR #2253, opened 2019-06-05
docstore: queries should always return the key and revision fields

jba:


internal/docstore/driver/util.go
```diff
@@ -160,3 +160,8 @@ func fpEqual(fp1, fp2 []string) bool {
 	}
 	return true
 }
+
+// FieldPathEqualsField reports whether a field path equals a field.
+func FieldPathEqualsField(fp []string, s string) bool {
```
shantuo: Why do we not care `fp` and `s` have multiple elements?

jba: This case comes up a lot: we want to compare a field path to a single field (basically a field path with exactly one element). For example, the revision field has to be a single field.

jba: PTAL

jba: Note that this function is just a convenience for
```
FieldPathsEqual(fp, []string{s})
```

shantuo: I see, please add a comment to not be used wrong.

jba: in next PR

PR #2255, opened 2019-06-05
release: refactor releasehelper for easier testing and add a test.

eliben:
The refactoring is to test both runOnGomod and the handlers in a black-box
way. The test creates a temporary directory simulating a small subset of
our repository.

### DIFF
--- a/internal/testing/releasehelper.go
+++ b/internal/testing/releasehelper.go
@@ -156,25 +156,33 @@ func runOnGomod(path string, reqHandler reqHandlerFunc) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


internal/testing/releasehelper.go
```diff
@@ -156,25 +156,33 @@ func runOnGomod(path string, reqHandler reqHandlerFunc) {
 	}
 }
 
+func gomodAddReplace(path string) {
```
vangent: Nit: I would consider dropping the `reqHandlerFunc` type. It is only used in one place and just adds a level of indirection.

PR #2259, opened 2019-06-05
dynamodocstore: support OrderBy

jba:
It turns out this is harder than I realized because table scans don't
support ordering. The best solution I could come up with is to let the
developer handle those queries themselves, by providing a function.


internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -173,8 +173,22 @@ type Options struct {
...
+	// If set, call this function on queries that we cannot execute at all (for
+	// example, a query with an OrderBy clause that lacks an equality filter on a
+	// partition key). The function should execute the query however it wishes, and
+	// return an iterator over the results. It can use the RunQueryFunc passed as its
+	// third argument to have the DynamoDB driver run a query, for instance a
+	// modified version of the original query.
+	//
+	// If RunQueryFallback is nil, queries that cannot be executed will fail with a
+	// error that has code Unimplemented.
+	RunQueryFallback func(context.Context, *driver.Query, RunQueryFunc) (driver.DocumentIterator, error)
```
shantuo: Is this fallback method necessary? How about a boolean like `AllowScanWithoutOrdering`, which let the query returns the result and the user can sort on their end.

internal/docstore/dynamodocstore/query.go
```diff
@@ -65,6 +66,9 @@ func (c *collection) checkPlan(qr *queryRunner) error {
 	return nil
 }
 
+// From https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html:
```
shantuo: Is this comment intended to be not attached with any code block?

internal/docstore/dynamodocstore/dynamo.go
jba: I don't like the solution, but at least it's general. This way the user can choose to sort on the client, or to issue a query for each partition key and merge the results (which is what I suggest in a TODO later in the file) or maybe something else that I haven't thought of.

I don't think returning the results out of order is an option, because where would they sort it? Their docstore code is supposedly portable, meaning it gives the same answer for all providers. The only non-portable part is the driver Open method and its options, which is why something like this has to happen there.

Let's live with this for now and ask @vangent if he has any thoughts.

internal/docstore/dynamodocstore/query.go
jba: Removed in the next PR.

PR #2262, opened 2019-06-06
docstore: clarify error codes and behavior with revision fields

### LINKED ISSUE #1652: docstore: Create should return AlreadyExists if doc exists
The Create action should return an error with code AlreadyExists if the document exists.

Currently the conformance test checks that we get an error, but it doesn't check the code.

Also, I don't think any of the drivers return the right code.

jba:
- Create on an existing document is an AlreadyExists error.

- Create on a document with a revision field is an InvalidArgument error.

- Put on a document with a revision field is the same as replace.

- Replace on a document without a revision field always returns NotFound.

Fixes #1652.


internal/docstore/docstore.go
```diff
@@ -138,7 +144,10 @@ func (l *ActionList) Replace(doc Document) *ActionList {
 
 // Put adds an action that adds or replaces a document to the given ActionList, and returns the ActionList.
 // The key fields must be set.
-// The document may or may not already exist.
+//
+// If the revision field is non-nil, then Put behaves exactly like Replace. Otherwise,
+// Put will create the document if it does not already exist.
```
shantuo: What if the document exists?

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -455,13 +457,22 @@ func (c *collection) put(ctx context.Context, k driver.ActionKind, doc driver.Do
...
+			}
 		}
-		_ = doc.SetField(docstore.RevisionField, rev)
+		return err
 	}
-	return err
+	if newPartitionKey != "" {
+		_ = doc.SetField(c.partitionKey, newPartitionKey)
+	}
+	_ = doc.SetField(docstore.RevisionField, rev)
```
shantuo: nit: just return this line.

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -455,13 +457,22 @@ func (c *collection) put(ctx context.Context, k driver.ActionKind, doc driver.Do
...
-	if err == nil {
-		if newPartitionKey != "" {
-			_ = doc.SetField(c.partitionKey, newPartitionKey)
+	if err != nil {
+		if ae, ok := err.(awserr.Error); ok && ae.Code() == dyn.ErrCodeConditionalCheckFailedException {
+			if k == driver.Create {
+				return gcerr.Newf(gcerr.AlreadyExists, err, "document already exists")
+			}
+			if rev, _ := doc.GetField(docstore.RevisionField); rev == nil && k == driver.Replace {
+				return gcerr.Newf(gcerr.NotFound, nil, "document not found")
```
shantuo: Should this be `FailedPrecondition`? I feel it is confusing to return "document not found" when I try to put something into a db.

internal/docstore/docstore.go
jba: Clarified.

internal/docstore/dynamodocstore/dynamo.go
jba: But it's a replace. You're effectively saying: "look up the document with this ID, then replace its contents". If there is no such document, then I think NotFound is the right code, just as it would be with a Get.

internal/docstore/dynamodocstore/dynamo.go
jba: That would return an error if there were no revision field. We want to support structs that have no revision field.

jba:
PTAL.

internal/docstore/dynamodocstore/dynamo.go
shantuo: Ah, I see.

internal/docstore/dynamodocstore/dynamo.go
shantuo: I was talking about the scenario that you change the kind to replace when the action is a put with the revision ID.

Also, are we doing the same thing for update? Also other providers?

I'll leave this up to you.

PR #2269, opened 2019-06-06
internal/gocdk: move static assets into its own package

vangent:
I'm moving the `static` stuff into a sub-package so that in future PRs I can provide a better API than just raw access to the files.

More details:

Today, we have the following uses of `static`:
* `demo and` copies a single file directly.
* `init` and `biome add` use `materializeTemplateDir` to recursively copy files, treating them like Go templates.

For `provision add`, I also need to:
* insert into an existing file based on a marker comment, with some check to see if it has already been done.
* append into an existing file, similarly checking to see if it has already been done.

My plan is to move `materializeTemplateDir` to the being-created `static` package, and have it support all of those use cases.


zombiezen:
One suggested improvement: this `static` package should live in `internal/cmd/gocdk/internal/static` so that when we move to `cmd/gocdk` that the package is not globally importable.

vangent:
> One suggested improvement: this `static` package should live in `internal/cmd/gocdk/internal/static` so that when we move to `cmd/gocdk` that the package is not globally importable.

Good point -- will do.

PR #2270, opened 2019-06-06
internal/gocdk: avoid bytes.ReplaceAll (1.12 only); fix test on Windows

vangent:
The `go 1.11` test failed :-(.

### DIFF
--- a/internal/cmd/gocdk/main_test.go
+++ b/internal/cmd/gocdk/main_test.go
@@ -299,5 +299,9 @@ func doList(dir string, out io.Writer, arg, indent string) error {
 // scrub removes dynamic content from recorded files.
 func scrub(rootDir string, b []byte) []byte {
 	const scrubbedRootDir = "[ROOTDIR]"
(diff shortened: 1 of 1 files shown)


internal/cmd/gocdk/main_test.go
```diff
@@ -299,5 +299,9 @@ func doList(dir string, out io.Writer, arg, indent string) error {
 // scrub removes dynamic content from recorded files.
 func scrub(rootDir string, b []byte) []byte {
 	const scrubbedRootDir = "[ROOTDIR]"
-	return bytes.ReplaceAll(b, []byte(rootDir), []byte(scrubbedRootDir))
+	rootDirWithSeparator := rootDir + string(filepath.Separator)
```
vangent: The problem here was that the record file started with something like `/tmp/whatever/myproj` and was scrubbed to `[ROOTDIR]/myproj`, but the `/` is a `\` on Windows.

PR #2271, opened 2019-06-06
internal/gocdk: fill in the "provision" command

vangent:
There are still a lot of TODOs here, but I think this is a good checkpoint.

After this PR, you can `gocdk init foo && cd foo && gocdk demo add blob && gocdk provision add dev blob/fileblob && gocdk serve` and it will work.

You can also `gocdk provision add dev blob/gcsblob` or `s3blob` and they will also work, with a few caveats:
* The projectID variable has a default of our test project; this needs to change to prompt the user instead.
* If you add to an existing biome, you need to manually run `terraform init` again.
* If you add multiple blob providers to a biome, the first one wins (but you can easily change that by commenting out lines in `outputs.tf`.

There's also a big TODO in there about providing a more general purpose way to instantiate static files; we now have 3 and we should consolidate (and this one is pretty hacky).

I'd be happy to give a demo again if that's useful.


internal/cmd/gocdk/provision.go
```diff
@@ -54,12 +60,170 @@ func registerProvisionCmd(ctx context.Context, pctx *processContext, rootCmd *co
...
+	dir, err := static.Open(provisionStaticRootDir)
+	if err != nil {
+		return xerrors.Errorf("provision list: couldn't open static resources: %w", err)
+	}
+	ptInfos, err := dir.Readdir(-1)
+	if err != nil {
+		return xerrors.Errorf("provision list: couldn't read static resources: %w", err)
+	}
+	// The entries at this level are portable APIs, e.g., "blob".
+	for _, ptInfo := range ptInfos {
```
zombiezen: We likely want sort this list. I'm not sure if `Readdir`'s order is guaranteed.

internal/cmd/gocdk/provision.go
```diff
@@ -54,12 +60,170 @@ func registerProvisionCmd(ctx context.Context, pctx *processContext, rootCmd *co
...
+	}
+	return nil
 }
 
 // The "provision add" command.
+// TODO(rvangent): Can we support adding a particular type more than once?
+// TODO(rvangent): If things fail in the middle, we are in an undefined state.
+//                 Unclear how to handle that....
+// TODO(rvangent): Modifying Terraform files in place means that we need to run
+//                 "terraform init" again; currently we don't.
```
zombiezen: Likely the user won't feel the effects of this most of the time because `gocdk apply` runs `terraform init`. The places where it doesn't right now are a bug: #2291.

internal/cmd/gocdk/provision.go
```diff
@@ -54,12 +60,170 @@ func registerProvisionCmd(ctx context.Context, pctx *processContext, rootCmd *co
...
+// TODO(rvangent): Currently all of the variables.tf files have default values, so
```
zombiezen: As discussed offline, I don't think we should be creating Terraform variables.

internal/cmd/gocdk/provision.go
```diff
@@ -54,12 +60,170 @@ func registerProvisionCmd(ctx context.Context, pctx *processContext, rootCmd *co
...
+		return xerrors.Errorf("provision add: %q is not a supported type; use 'gocdk provision list' to see available types", typ)
+	}
+
+	moduleDir, err := pctx.ModuleRoot(ctx)
+	if err != nil {
+		return xerrors.Errorf("provision add: %w", err)
+	}
+	dstPath := biomeDir(moduleDir, biome)
+
+	srcRoot := path.Join(provisionStaticRootDir, typ)
```
zombiezen: I am concerned that a user could specify a string that allows them to use a different directory inside the static resources. Consider using an explicit allowlist.

internal/cmd/gocdk/static/_assets/provision/blob/fileblob/fileblob.tf
```diff
@@ -0,0 +1,11 @@
+# TODO(rvangent): Add comments explaining.
```
zombiezen: nit: These files are not `terraform fmt`-ed

internal/cmd/gocdk/provision.go
vangent: It is "in directory order", which is at least stable, so I think we're OK.
https://golang.org/pkg/os/#File.Readdir

internal/cmd/gocdk/provision.go
vangent: Updated to reference the bug.

internal/cmd/gocdk/provision.go
vangent: Done. I added a new `mainlocal_` prefix for static files, which inserts into a "locals" section of `main.tf`.

internal/cmd/gocdk/static/_assets/provision/blob/fileblob/fileblob.tf
vangent: I did the best I could, but `terraform fmt` is not idempotent and gradually screws up comments :-(.

internal/cmd/gocdk/provision.go
vangent: Added a TODO. This will be easier when the other TODO about a helper for reading the dirs is done.

internal/cmd/gocdk/provision.go
```diff
@@ -54,12 +59,164 @@ func registerProvisionCmd(ctx context.Context, pctx *processContext, rootCmd *co
...
+		// -- A config setting means "append to this file".
+		// -- Options for insert/append to tell whether it should be skipped
+		//    (i.e., it's been added already), or whether to error.
+		// "demo add" and materializeTemplateDir could be updated to use this as well.
+		if strings.HasPrefix(name, "main_") {
+			if err := insertIntoFile(filepath.Join(dstPath, "main.tf"), "", srcBytes); err != nil {
+				return xerrors.Errorf("provision add: %w", err)
+			}
+		} else if strings.HasPrefix(name, "mainlocal_") {
+			const marker = "# DO NOT REMOVE: GO CDK LOCALS WILL BE INSERTED BELOW HERE"
```
clausti: If I saw this, I'm not sure if I'd know whether it referred to "don't remove *this comment*" vs "don't remove whatever code is immediately below this" `DO NOT REMOVE THIS COMMENT` would be unambiguous.

internal/cmd/gocdk/provision.go
```diff
@@ -54,12 +59,164 @@ func registerProvisionCmd(ctx context.Context, pctx *processContext, rootCmd *co
...
+			if err := insertIntoFile(filepath.Join(dstPath, "main.tf"), marker, srcBytes); err != nil {
+				return xerrors.Errorf("provision add: %w", err)
+			}
+		} else if strings.HasPrefix(name, "outputs_") {
+			const marker = "# DO NOT REMOVE: DEMO URLs WILL BE INSERTED BELOW HERE"
```
clausti: same comment as above, let's specify what is not to be removed (the comment?)

internal/cmd/gocdk/static/_assets/provision/blob/gcsblob/gcsblob.tf
```diff
@@ -0,0 +1,30 @@
+# TODO(rvangent): Add comments explaining.
```
clausti: Is there a specific issue tracking this todo already? If not, can you make an issue (or a few) for filling in this group of explanatory comments (or other chunks of commentary that make sense to tackle together?

internal/cmd/gocdk/static/_assets/biome/main.tf
```diff
@@ -4,3 +4,7 @@
 terraform {
   required_version = "~> 0.11"
 }
+
+locals {
+  DO NOT REMOVE: GO CDK LOCALS WILL BE INSERTED BELOW HERE
```
zombiezen: I believe this should be a comment?

vangent: Yes, done.

internal/cmd/gocdk/provision.go
```diff
@@ -54,12 +59,164 @@ func registerProvisionCmd(ctx context.Context, pctx *processContext, rootCmd *co
...
-func provisionList(pctx *processContext) {
-	pctx.Logf("NOT YET IMPLEMENTED")
+func provisionList(pctx *processContext) error {
+	// TODO(rvangent): There must be an easier way to get a listing... add
+	// a helper to static.
+	dir, err := static.Open(provisionStaticRootDir)
+	if err != nil {
+		return xerrors.Errorf("provision list: couldn't open static resources: %w", err)
+	}
+	ptInfos, err := dir.Readdir(-1)
```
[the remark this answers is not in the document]
vangent: Portable type.

internal/cmd/gocdk/provision.go
vangent: Done.

internal/cmd/gocdk/provision.go
vangent: Done.

internal/cmd/gocdk/static/_assets/provision/blob/gcsblob/gcsblob.tf
vangent: Done.

PR #2273, opened 2019-06-06
internal/cmd/gocdk: infer module path if multiple GOPATH entries exist

### LINKED ISSUE #2016: internal/cmd/gocdk: Add tests for init inferring module path
As a followup to https://github.com/google/go-cloud/pull/1945, we need more tests of the behavior of `init` when the `--module-path` option is not supplied.

### LINKED ISSUE #2157: internal/cmd/gocdk: init does not perform module path inference if GOPATH has multiple entries
If I have a `GOPATH` like:

```
/home/light/gopath:/foo/gopath
```

Then running `gocdk init /home/light/gopath/src/myproject` produces:

```
gocdk init: infer module path: /home/light/gopath/src/myproject not in GOPATH
```

I expect it to succeed with module path `myproject`.

clausti:
fixes #2157 
fixes #2016

### DIFF
--- a/internal/cmd/gocdk/init.go
+++ b/internal/cmd/gocdk/init.go
@@ -17,7 +17,6 @@ package main
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"io/ioutil"
 	"os"
 	slashpath "path"
@@ -149,10 +148,6 @@ func inferModulePath(ctx context.Context, pctx *processContext, projectDir strin
 	// so Output can collect stdout.
 	cmd.Stdout = nil
 	gopath, err := cmd.Output()
-	fmt.Println("FROM PCTX-- " + pctx.env[0])
-	fmt.Println("FROM CMD-- " + cmd.Env[0])
-	fmt.Println("WHAT IS HAPPENING HERE-- " + string(gopath))
-	fmt.Println(err)
 	if err != nil {
 		return "", xerrors.Errorf("infer module path: %w", err)
 	}
--- a/internal/cmd/gocdk/init.go
+++ b/internal/cmd/gocdk/init.go
@@ -152,15 +152,19 @@ func inferModulePath(ctx context.Context, pctx *processContext, projectDir strin
 	if err != nil {
 		return "", xerrors.Errorf("infer module path: %w", err)
 	}
-	// Check if the projectDir is relative to GOPATH.
-	rel, err := filepath.Rel(strings.TrimSuffix(string(gopath), "\n"), projectDir)
-	if err != nil {
-		return "", xerrors.Errorf("infer module path: %w", err)
+
+	gopathEntries := filepath.SplitList(strings.TrimSuffix(string(gopath), "\n"))
+	for _, entry := range gopathEntries {
+		// Check if the projectDir is relative to GOPATH.
+		rel, err := filepath.Rel(entry, projectDir)
+		if err != nil {
+			return "", xerrors.Errorf("infer module path: %w", err)
+		}
+		inGOPATH := !strings.HasPrefix(rel, ".."+string(filepath.Separator))
+		if inGOPATH {
+			return filepath.ToSlash(rel), nil
+		}
 	}
-	inGOPATH := !strings.HasPrefix(rel, ".."+string(filepath.Separator))
-	if !inGOPATH {
-		// If the project dir is outside of GOPATH, we can't infer the module import path.
-		return "", xerrors.Errorf("infer module path: %s not in GOPATH", projectDir)
-	}
-	return filepath.ToSlash(rel), nil
+	// If the project dir is outside of GOPATH, we can't infer the module import path.
+	return "", xerrors.Errorf("infer module path: %s not in GOPATH", projectDir)
 }
--- a/internal/cmd/gocdk/init.go
+++ b/internal/cmd/gocdk/init.go
@@ -17,6 +17,7 @@ package main
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"io/ioutil"
 	"os"
 	slashpath "path"
@@ -149,6 +150,10 @@ func inferModulePath(ctx context.Context, pctx *processContext, projectDir strin
 	// so Output can collect stdout.
 	cmd.Stdout = nil
 	gopath, err := cmd.Output()
+	fmt.Println("FROM PCTX-- " + pctx.env[0])
+	fmt.Println("FROM CMD-- " + cmd.Env[0])
+	fmt.Println("WHAT IS HAPPENING HERE-- " + string(gopath))
+	fmt.Println(err)
 	if err != nil {
 		return "", xerrors.Errorf("infer module path: %w", err)
 	}
--- a/internal/cmd/gocdk/init.go
+++ b/internal/cmd/gocdk/init.go
@@ -144,7 +144,6 @@ func materializeTemplateDir(dst string, srcRoot string, data interface{}) error 
 // inferModulePath will check the default GOPATH to attempt to infer the module
 // import path for the project.
 func inferModulePath(ctx context.Context, pctx *processContext, projectDir string) (string, error) {
-	// TODO(issue #2016): Add tests for init behavior when module-path is not given.
 	cmd := pctx.NewCommand(ctx, "", "go", "env", "GOPATH")
 	// Since we're going to call Output, we need to make sure cmd.Stdout is nil
 	// so Output can collect stdout.
--- a/internal/cmd/gocdk/init_test.go
+++ b/internal/cmd/gocdk/init_test.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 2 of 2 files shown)


internal/cmd/gocdk/init_test.go
```diff
@@ -147,6 +148,83 @@ func TestInit(t *testing.T) {
 		}
 	})
 }
+func TestInferModulePath(t *testing.T) {
```
vangent: Since most of the content of each test is the same, this is a good opportunity to use a table-driven test.

https://blog.golang.org/subtests

internal/cmd/gocdk/init_test.go
```diff
@@ -147,6 +148,83 @@ func TestInit(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer func() {
+			if err := os.RemoveAll(dir); err != nil {
+				t.Error(err)
+			}
+		}()
+		ctx := context.Background()
+		pctx := &processContext{
```
vangent: You can use `newTestProcessContext` for this now. You can override the `env` field to be what you need after that.

internal/cmd/gocdk/init_test.go
```diff
@@ -147,6 +148,83 @@ func TestInit(t *testing.T) {
...
+			stdin:   strings.NewReader(""),
+			stdout:  ioutil.Discard,
+			stderr:  ioutil.Discard,
+		}
+
+		fmt.Println(pctx.env)
+
+		const projectName = "myspecialproject"
+		if err := run(ctx, pctx, []string{"init", projectName}); err == nil {
+			t.Errorf("inferModulePath should error when --module-path is not" +
```
vangent: Nit: you don't need to explain all this in the error; the name of the test provides most of the context. Just report what didn't meet expectations, e.g. `got nil error, want non-nil error`.

internal/cmd/gocdk/init.go
```diff
@@ -152,15 +152,19 @@ func inferModulePath(ctx context.Context, pctx *processContext, projectDir strin
...
-	rel, err := filepath.Rel(strings.TrimSuffix(string(gopath), "\n"), projectDir)
-	if err != nil {
-		return "", xerrors.Errorf("infer module path: %w", err)
-	}
-	inGOPATH := !strings.HasPrefix(rel, ".."+string(filepath.Separator))
-	if !inGOPATH {
-		// If the project dir is outside of GOPATH, we can't infer the module import path.
-		return "", xerrors.Errorf("infer module path: %s not in GOPATH", projectDir)
+
+	gopathEntries := filepath.SplitList(strings.TrimSuffix(string(gopath), "\n"))
```
vangent: The test is failing due to missing import for `strings`.

internal/cmd/gocdk/init.go
```diff
@@ -152,15 +152,19 @@ func inferModulePath(ctx context.Context, pctx *processContext, projectDir strin
 	if err != nil {
 		return "", xerrors.Errorf("infer module path: %w", err)
 	}
-	// Check if the projectDir is relative to GOPATH.
-	rel, err := filepath.Rel(strings.TrimSuffix(string(gopath), "\n"), projectDir)
-	if err != nil {
-		return "", xerrors.Errorf("infer module path: %w", err)
```
vangent: Do your tests cover both this error and the one below? You can check with `codecov`...

internal/cmd/gocdk/init_test.go
```diff
@@ -147,6 +148,83 @@ func TestInit(t *testing.T) {
...
+		}
+		defer func() {
+			if err := os.RemoveAll(dir); err != nil {
+				t.Error(err)
+			}
+		}()
+		ctx := context.Background()
+		pctx := &processContext{
+			workdir: dir,
+			env:     []string{"GOPATH=" + dir + ":~/some/other/path"},
```
zombiezen: I'm not sure whether it's significant for this test, but `~/some/other/path` isn't a valid path for Windows (different path separators) or Unix (tilde expansion occurs at the shell, not in path resolution).

internal/cmd/gocdk/init_test.go
```diff
@@ -147,6 +148,83 @@ func TestInit(t *testing.T) {
...
+		ctx := context.Background()
+		pctx := &processContext{
+			workdir: dir,
+			env:     []string{},
+			stdin:   strings.NewReader(""),
+			stdout:  ioutil.Discard,
+			stderr:  ioutil.Discard,
+		}
+
+		fmt.Println(pctx.env)
```
zombiezen: Here and elsewhere: I'm not sure if this is left over from debugging, but use `t.Log` in tests if there should be diagnostic information display during a test. (It might be good to include a little more context for the log.)

internal/cmd/gocdk/init_test.go
```diff
@@ -147,6 +148,83 @@ func TestInit(t *testing.T) {
...
+		pctx := &processContext{
+			workdir: dir,
+			env:     []string{"GOPATH=" + dir},
+			stdin:   strings.NewReader(""),
+			stdout:  ioutil.Discard,
+			stderr:  ioutil.Discard,
+		}
+		fmt.Println(pctx.env)
+		const projectName = "myspecialproject"
+		if err := run(ctx, pctx, []string{"init", projectName}); err != nil {
```
zombiezen: IIUC, this creates a project at `$GOPATH/myspecialproject`, but I believe it should be trying to create the project at `$GOPATH/src/myspecialproject`.

clausti:
Ok I am stumped. The multiple GOPATH entries test fails and I know why-- which is that it's not seeing anything from the output of the `go env GOPATH` call in `inferModulePath`, but I have no idea why it's not getting anything there.

This is a screenshot of running the tests on verbose, with the code as of `9e70bcc`. In the `no_GOPATH_entry` subtest, everything is as we expect (empty). In the `single_GOPATH_entry` subtest, everything is as we expect, and `"WHAT IS HAPPENING HERE-- " + string(gopath)` (where `gopath` is the output of the command) is set. 

In the `multiple_GOPATH_entries` subtest, the `GOPATH` is set everywhere it should be set... EXCEPT from the output of the command.... even though if you query the `cmd` struct directly, the `ENV` is as we expect. 

What is going on?

Screen Shot 2019-06-07 at 2 09 47 PM

internal/cmd/gocdk/init_test.go
clausti: Sorry, not sure what you're suggesting here. Why "should" it be adding `src` in the middle? Is this a problem with the test or `init`?

zombiezen: I am realizing this is a bug in `gocdk init`. As per [GOPATH](https://golang.org/cmd/go/#hdr-GOPATH_environment_variable) documentation, the `src` subdirectory is where source code lives. You can see the difference between `go mod init` and `gocdk init` like this:

```
mkdir $GOPATH/src/xyzzy
cd $GOPATH/src/xyzzy
go mod init
grep '^module' go.mod
```

yields `module xyzzy`, whereas:

```
gocdk init $GOPATH/src/xyzzy
grep '^module' $GOPATH/src/xyzzy/go.mod
```

yields `src/xyzzy`. I can reproduce this at master, so it's not a regression in this PR.

internal/cmd/gocdk/init_test.go
```diff
@@ -147,6 +148,69 @@ func TestInit(t *testing.T) {
...
+			"single GOPATH entry",
+			func(dir string) string {
+				return "GOPATH=" + dir
+			},
+			false,
+		},
+		{
+			"multiple GOPATH entries",
+			func(dir string) string {
+				multiPath := "GOPATH=" + dir + ":" + filepath.Join("", "some", "other", "path")
```
zombiezen: Use `filepath.ListSeparator` to keep portable:

```suggestion
				multiPath := "GOPATH=" + dir + string(filepath.ListSeparator) + filepath.Join("", "some", "other", "path")
```

internal/cmd/gocdk/init.go
```diff
@@ -149,18 +150,26 @@ func inferModulePath(ctx context.Context, pctx *processContext, projectDir strin
 	// so Output can collect stdout.
```
zombiezen: Not related to this line but above: let's remove the TODO about #2016 since this PR fixes it.

internal/cmd/gocdk/init.go
```diff
@@ -149,18 +150,26 @@ func inferModulePath(ctx context.Context, pctx *processContext, projectDir strin
...
 	cmd.Stdout = nil
 	gopath, err := cmd.Output()
+	fmt.Println("FROM PCTX-- " + pctx.env[0])
+	fmt.Println("FROM CMD-- " + cmd.Env[0])
+	fmt.Println("WHAT IS HAPPENING HERE-- " + string(gopath))
+	fmt.Println(err)
 	if err != nil {
 		return "", xerrors.Errorf("infer module path: %w", err)
```
zombiezen: I'm not 100% sure what is going on in the error you posted, but it seems to be coming from this line and it's reporting that the `go` command exited with a failure status. See if you can pull the stderr from the subprocess and see if that gives you any information.

(My hunch is that because the set of environment variables being given to the `go` command is limited, it might be complaining, but I have no basis for this.)

vangent: My guess is that it can't find the `go` command because the env does not have a `PATH` in it...?

internal/cmd/gocdk/init_test.go
```diff
@@ -147,6 +148,69 @@ func TestInit(t *testing.T) {
...
+			}
+			defer func() {
+				if err := os.RemoveAll(dir); err != nil {
+					t.Error(err)
+				}
+			}()
+
+			ctx := context.Background()
+			pctx := newTestProcessContext(dir)
+			pctx.env = []string{tc.testGOPATH(dir)}
```
vangent: Instead of clobbering the existing `env`, try appending to it:
`pctx.env = append(pctx.env, tc.testGOPATH(dir))`

internal/cmd/gocdk/init_test.go
```diff
@@ -147,6 +148,69 @@ func TestInit(t *testing.T) {
...
+			fmt.Println("FROM SET ENV " + pctx.env[0])
+
+			err = run(ctx, pctx, []string{"init", "myspecialproject"})
+
+			switch tc.wantErr {
```
vangent: Nit: shorter way to write this is something like:

```
  if (err != nil) != tc.wantErr {
    t.Errorf("got err %v but wantErr is %v", err, tc.wantErr)
  }
```

internal/cmd/gocdk/init_test.go
clausti: for "systematically eliminating possibilities" purposes, this comment is to note that this results in exactly the same errors.

internal/cmd/gocdk/init_test.go
clausti: This seems like it warrants its own issue, which I can make, but so I can be a little clearer on it: Inferring the module path ought to rely on the given project dir being not just relative to `$GOPATH` but relative to `$GOPATH/src`? (As opposed to current behavior, which is that it only requires it be relative to `$GOPATH`)

zombiezen: Opened #2307 to track.

internal/cmd/gocdk/init.go
clausti: Not sure if you mean the error on 160 or 168... looks like the former is "no" and the later is "yes".
Screen Shot 2019-06-11 at 11 37 12 AM

vangent: I meant that the tests should try to cover all of the error conditions :-).

clausti: reprising in-person discussion: not going to add another test for this. can you PTAL, on the PR as a whole?

clausti:
tests are passing!! PTAL @zombiezen

internal/cmd/gocdk/init_test.go
```diff
@@ -147,6 +147,71 @@ func TestInit(t *testing.T) {
...
+			if err != nil {
+				t.Fatal(err)
+			}
+			defer func() {
+				if err := os.RemoveAll(dir); err != nil {
+					t.Error(err)
+				}
+			}()
+
+			ctx := context.Background()
```
vangent: Nit: you could define the `ctx` outside the loop since it doesn't change.

clausti:
fixes #2157
fixes #2016

PR #2279, opened 2019-06-07
docstore: add a RevisionField option to all providers

### LINKED ISSUE #2154: docstore: make revision field a collection option

jba:
Fixes #2154


internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -207,6 +211,9 @@ func newCollection(db *dyn.DynamoDB, tableName, partitionKey, sortKey string, op
 	if opts == nil {
 		opts = &Options{}
 	}
+	if opts.RevisionField == "" {
+		opts.RevisionField = docstore.RevisionField
```
shantuo: Maybe rename the `docstore.RevisionField` to `docstore.DefaultRevisionField`?

internal/docstore/firedocstore/codec.go
```diff
@@ -142,7 +141,7 @@ func floatval(x float64) *pb.Value { return &pb.Value{ValueType: &pb.Value_Doubl
 ////////////////////////////////////////////////////////////////
 
 // decodeDoc decodes a Firestore document into a driver.Document.
-func decodeDoc(pdoc *pb.Document, ddoc driver.Document, nameField string) error {
+func decodeDoc(pdoc *pb.Document, ddoc driver.Document, nameField, revField string) error {
```
shantuo: I suggest wrap `nameField` and `revField` into an options struct.

internal/docstore/drivertest/driverbenchmark.go
```diff
@@ -39,7 +39,6 @@ func RunBenchmarks(b *testing.B, coll *docstore.Collection) {
 	b.Run("BenchmarkActionListGet", func(b *testing.B) {
 		benchmarkActionListGet(100, b, coll)
 	})
-	cleanUpTable(b, coll)
```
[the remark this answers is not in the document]
jba: Restored.

internal/docstore/dynamodocstore/dynamo.go
jba: Good idea. Done.

internal/docstore/firedocstore/codec.go
jba: I see your point. Maybe if we add one more thing that will be worth it.

PR #2280, opened 2019-06-07
internal/cmd/gocdk: add default VSCode tasks to initial project

zombiezen:
Allows users to run common gocdk commands via Run Task in VSCode without manual configuration. See https://code.visualstudio.com/docs/editor/tasks for an overview.


internal/cmd/gocdk/_static/init/.vscode/tasks.json
```diff
@@ -0,0 +1,53 @@
...
+            "command": "gocdk apply ${input:biome}",
+            "problemMatcher": []
+        }
+    ],
+    "inputs": [
+        {
+            "id": "biome",
+            "type": "pickString",
+            "description": "Which biome to use",
+            // You can add other biome names here to have them show up in the
```
zombiezen: Note to reviewers: my understanding is that if we had a VSCode extension, we could populate this menu automatically (likely using `gocdk biome list` or some such).

zombiezen:
Argh, I realize that #2269 changed the static assets layouts in the meantime. Content won't change, but exact file paths will.

PR #2281, opened 2019-06-07
internal/docs: document compound IDs decision

zombiezen:
Updates #1860

### DIFF
--- a/internal/docs/design.md
+++ b/internal/docs/design.md
@@ -213,15 +213,16 @@ We call these _compound IDs_.
 
 There are two broad compound ID usage patterns we have observed:
 
-1. Applications will keep resources in the same location or project, so the
-   application will build the ID from the subcomponents rather than passing the
-   entire resource ID around.
+1. Applications will keep resources in the same location, so the application
+   will build the ID from the subcomponents rather than passing the entire
+   resource ID around.
 2. Applications will pass a verbatim string from configuration down to the
    API, since this is what was easily copy-pasteable from the cloud console UI.
 
 Go CDK constructors that take in compound IDs should take in a `string` with the
-full compound ID. Helper functions to build these compound IDs may be provided
-as needed.
+full compound ID. Helper functions to build these compound IDs from
+subcomponents may be provided as needed. URL openers (described below) should
+prefer to use the full compound ID in their URL format.
 
 [GCP KMS key IDs]: https://cloud.google.com/kms/docs/object-hierarchy#key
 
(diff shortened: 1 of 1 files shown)


internal/docs/design.md
```diff
@@ -204,6 +204,27 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
+
+There are two broad compound ID usage patterns we have observed:
+
+1. Applications will keep resources in the same location or project, so the
+   application will build the ID from the subcomponents rather than passing the
+   entire resource ID around.
+2. Applications will pass a verbatim string from configuration down to the
+   API, since this is what was easily copy-pasteable from the cloud console UI.
+
+Go CDK constructors that take in compound IDs should take in a `string` with the
```
vangent: Nit: add "and URLOpeners".

zombiezen: But URLOpeners take in a `*url.URL`, not a `string`. Can you elaborate on the change you want to see here?

internal/docs/design.md
```diff
@@ -204,6 +204,27 @@ needed). However, this would result in inconsistent names over time (e.g., some
...
+### Compound IDs
+
+Many cloud providers have resource IDs that are made up of subcomponents in
+some well-defined syntax. For example, [GCP KMS key IDs][] take the form
+`projects/[PROJECT_ID]/locations/[LOCATION]/keyRings/[KEY_RING]/cryptoKeys/[KEY]`.
+We call these _compound IDs_.
+
+There are two broad compound ID usage patterns we have observed:
+
+1. Applications will keep resources in the same location or project, so the
```
eliben: Nit: not sure what the first part of the sentence (up to "project") means and how it helps explain the pattern. Consider dropping it

internal/docs/design.md
vangent: I mean that the URL used to open the resource should have the full compound ID in them. So, in your example, the URL should have `projects/[PROJECT_ID]/locations/[LOCATION]/keyRings/[KEY_RING]/cryptoKeys/[KEY]` in it somewhere.

internal/docs/design.md
zombiezen: Done.

internal/docs/design.md
zombiezen: Gotcha. Added a slightly different wording to be more precise, PTAL.

vangent: SGTM.

Issue #2284, opened 2019-06-07
release: make release helper tool set git tags for releases automatically


eliben:
Following up on a TODO in https://github.com/google/go-cloud/blob/master/internal/docs/release.md

The process currently requires manual tagging of all released modules. This could be automated, for example in our `releashelper` tool.

We have a list of all modules in `allmodules`, but not all modules on this list should be released/tagged - we don't want our `samples` module to be importable and appear on godoc.org, for example.

Therefore, this may require to solve #2111 first - `allmodules` should mark which modules are tagged/released and which aren't.

PR #2285, opened 2019-06-07
testing: Add option to have extra fields in allmodules

eliben:
This PR teaches all our shell scripts that procses allmodules to ignore
any additional whitespace-separated fields.

The next step will be to add special purpose fields for selective
tagging -- #2284

Updates #2111


eliben:
Note: I'm using `awk` because `cut` is not great for multiple spaces. It would need another `tr` piped into it, so `awk` ends up much simpler (see https://stackoverflow.com/a/19069428/8206)

Issue #2133, opened 2019-05-17
docstore: limit ongoing RPC's when necessary


shantuo:
When running benchmarks against the DynamoDB backend, I got errors like:
```
at XXX: RequestError: send request failed
        caused by: Post https://dynamodb.us-east-2.amazonaws.com/: dial tcp: lookup dynamodb.us-east-2.amazonaws.com: no such host;
```

I think this is because we have too many open request when doing the plenty of unordered actions. I'm running on macOS, which has a relative small default file descriptor limit so this is relatively easy to trigger.

jba:
Are we really dialing a TCP connection for every RPC? Is there a way to keep connections alive for a while?

PR #2292, opened 2019-06-08
docstore: add option to limit outstanding RPCs for action lists

jba:
This does everything but dynamo, which is forthcoming.

See #2133.

### DIFF
--- a/internal/docstore/driver/util.go
+++ b/internal/docstore/driver/util.go
@@ -17,6 +17,7 @@ package driver
 import (
 	"reflect"
 	"sort"
+	"sync"
 
 	"github.com/google/uuid"
 )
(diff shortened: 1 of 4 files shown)


internal/docstore/driver/util.go
```diff
@@ -166,3 +167,44 @@ func FieldPathsEqual(fp1, fp2 []string) bool {
...
+
+// Release releases a token obtained by Acquire.
+func (t *Throttle) Release() {
+	if t.c != nil {
+		<-t.c
+	}
+	t.wg.Done()
+}
+
+func (t *Throttle) Wait() {
```
shantuo: Add a comment here.

jba: Done in next PR.

PR #2295, opened 2019-06-10
dynamodocstore: add RPC throttling

### LINKED ISSUE #2133: docstore: limit ongoing RPC's when necessary

jba:
Fixes #2133.

Also, rewrite handling of actions. Two reasons:

- Factor out common code between transactional and non-transactional
  execution.

- Pick all random values (new partition keys and revisions)
  deterministically, before starting goroutines. This makes replay
  work. Even though the random sequence is the same on every replay,
  if random values are picked in goroutines, then different actions
  may end up with different values, which means that different
  requests could be created on each run.

### DIFF
--- a/internal/docstore/dynamodocstore/dynamo.go
+++ b/internal/docstore/dynamodocstore/dynamo.go
@@ -189,6 +189,10 @@ type Options struct {
 	// If RunQueryFallback is nil, queries that cannot be executed will fail with a
 	// error that has code Unimplemented.
 	RunQueryFallback func(context.Context, *driver.Query, RunQueryFunc) (driver.DocumentIterator, error)
+
+	// The maximum number of concurrent goroutines started for a single call to
+	// ActionList.Do. If less than 1, there is no limit.
+	MaxOutstandingActionRPCs int
 }
 
 // RunQueryFunc is the type of the function passed to RunQueryFallback.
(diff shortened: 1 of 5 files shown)


internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -445,74 +461,104 @@ func (c *collection) put(ctx context.Context, k driver.ActionKind, doc driver.Do
...
+				TableName:                 dput.TableName,
+				Item:                      dput.Item,
+				ConditionExpression:       dput.ConditionExpression,
+				ExpressionAttributeNames:  dput.ExpressionAttributeNames,
+				ExpressionAttributeValues: dput.ExpressionAttributeValues,
 			}
-			if rev, _ := doc.GetField(c.opts.RevisionField); rev == nil && k == driver.Replace {
-				return gcerr.Newf(gcerr.NotFound, nil, "document not found")
+			_, err := c.db.PutItemWithContext(ctx, in)
+			if err != nil {
```
shantuo: This `if` could be removed and L512 can just `return err`.

jba: OK, done in next PR.

Issue #2260, opened 2019-06-05
docstore: test BeforeDo and BeforeQuery


jba:
Make sure these callbacks are called with the right values.

I don't think we have tests for that now (in fact, dynamodb doesn't even implement BeforeDo).

PR #2299, opened 2019-06-11
docstore: test BeforeDo

jba:
Also add BeforeDo support and fix some bugs related to it.

Part of #2260.


internal/docstore/drivertest/drivertest.go
```diff
@@ -1382,40 +1381,86 @@ func testUnorderedActions(t *testing.T, coll *ds.Collection, revField string) {
 	}
 }
 
-func testAs(t *testing.T, coll *ds.Collection, st AsTest) {
-	docs := []*HighScore{
-		{game3, "steph", 24, date(4, 25), nil},
-		{game3, "mia", 99, date(4, 26), nil},
+// Verify that BeforeDo is invoked, and its as function behaves as expected.
+func testBeforeDo(t *testing.T, newHarness HarnessMaker) {
```
jba: I think this is a more straightforward and comprehensive way to test escape hatches like `BeforeDo` than the way we do it elsewhere, where each driver must implement an `AsTest` interface and write its own tests.

@vangent WDYT?

internal/docstore/memdocstore/mem.go
```diff
@@ -151,6 +154,15 @@ func (c *collection) RunActions(ctx context.Context, actions []*driver.Action, o
 		t.Wait()
 	}
 
+	if opts.BeforeDo != nil {
```
shantuo: This doesn't do anything, can we remove it?

internal/docstore/drivertest/drivertest.go
shantuo: I think this way you save the work of implementing `AsTest` by writing the `XXXTypes` function.

internal/docstore/memdocstore/mem.go
jba: It lets the user implement a `BeforeDo` that aborts the execution of the list. But yeah, that is probably not useful. But I need it for this test, so I guess we should leave it? If I take it out, how do I get memdocstore to pass the test?

jba:
PTAL

internal/docstore/docstore.go
```diff
@@ -260,10 +260,14 @@ func (e ActionListError) Unwrap() error {
 	return nil
 }
 
-// BeforeDo takes a callback function that will be called before the ActionList
-// is executed by the underlying provider's action functionality. The callback
-// takes a parameter, asFunc, that converts its argument to provider-specific
-// types. See https://gocloud.dev/concepts/as/ for background information.
+// BeforeDo takes a callback function that will be called before the ActionList is
+// executed by the underlying provider. It may be invoked multiple times for a
+// single call to ActionList.Do. If the callback returns an error, ActionList.Do
```
vangent: When would it be invoked multiple times? Consider explaining this.

internal/docstore/drivertest/drivertest.go
```diff
@@ -175,6 +172,8 @@ func RunConformanceTests(t *testing.T, newHarness HarnessMaker, ct CodecTester,
 	t.Run("DeleteQuery", func(t *testing.T) { withTwoKeyCollection(t, newHarness, testDeleteQuery) })
 	t.Run("UpdateQuery", func(t *testing.T) { withTwoKeyCollection(t, newHarness, testUpdateQuery) })
 
+	t.Run("BeforeDo", func(t *testing.T) { testBeforeDo(t, newHarness) })
```
vangent: I suspect you will end up wanting to combine this with other `As` tests, but you can worry about that later.

internal/docstore/drivertest/drivertest.go
vangent: It seems fine. You lose the ability to do an end-to-end test of `As` -- for example, `gcsblob` uses `As` to set a field that's not supported by `blob` in the `BeforeWrite`, and then reads it back using the `As` for read. If you don't want to do that, this is more succinct.

internal/docstore/drivertest/drivertest.go
```diff
@@ -53,6 +54,12 @@ type Harness interface {
 	// that the revision field should be drivertest.AlternateRevisionField.
 	MakeAlternateRevisionFieldCollection(context.Context) (driver.Collection, error)
 
+	// BeforeDoTypes should return a list of values whose types are valid for the as
+	// function given to BeforeDo. For example, if the provider converts Get actions
```
vangent: It should be clearer from this whether `As` should work for all of the types, or just one of them. The test should probably support both, as there are cases where a single `As` will support multiple types (not one or the other, but both).

internal/docstore/dynamodocstore/dynamo.go
```diff
@@ -30,7 +30,8 @@
 //
 // dynamodocstore exposes the following types for As:
 //  - Collection.As: *dynamodb.DynamoDB
-//  - ActionList.BeforeDo: *dynamodb.BatchGetItemsInput or *dynamodb.TransactWriteItemsInput
+//  - ActionList.BeforeDo: *dynamodb.BatchGetItemInput, *dynamodb.PutItemInput, *dynamodb.DeleteItemInput,
```
vangent: I think this should say "or" to make it more clear that each call will support one of these, not all.

internal/docstore/drivertest/drivertest.go
jba: I try BeforeDo on all six action types, and each action type must support at least one of the BeforeDo types.

I thought about making this more fine-grained, but it's tricky. A Get may have one type, a Put another, and an Update two others. It's going to get messy to specify all that. I figured this was a good start.

How about I put a TODO to make it more action-specific?

internal/docstore/drivertest/drivertest.go
jba: Agreed.

internal/docstore/docstore.go

jba:
PTAL.

internal/docstore/dynamodocstore/dynamo.go
vangent: please fix

jba: Done.

internal/docstore/drivertest/drivertest.go
jba: I added the TODO.

PR #2302, opened 2019-06-11
internal/cmd/gocdk: factor out launcher code to new package

zombiezen:
Not changing API yet, that will come in a follow-up PR.

Updates #2046


zombiezen:
> Is there a way to teach git about the rename so that I can see what actually changed?

Nope, Git proudly [doesn't track this info](https://www.git-scm.com/book/en/v2/Git-Basics-Recording-Changes-to-the-Repository#_git_mv).

> Or did nothing actually change...?

This is why I tried to do as pure of a code motion as I could. The only real differences are some minor renames to export symbols that needed to cross API boundaries. If you see anything that substantially changed, that's unintentional.

PR #2311, opened 2019-06-11
samples: add a sample command for docstore

### LINKED ISSUE #2245: docstore: add sample
Perhaps a simple CLI like the other APIs' samples.

shantuo:
Fixes #2245


samples/gocdk-docstore/main.go
```diff
@@ -0,0 +1,205 @@
...
+	subcommands.Register(&putCmd{}, "")
+	subcommands.Register(&deleteCmd{}, "")
+	log.SetFlags(0)
+	log.SetPrefix("gocdk-docstore: ")
+	flag.Parse()
+	os.Exit(int(subcommands.Execute(context.Background())))
+}
+
+type Message struct {
+	Date             string
```
jba: Why are these both strings and not `time.Time`s?

samples/gocdk-docstore/main.go
```diff
@@ -0,0 +1,205 @@
...
+type listCmd struct {
+	date string
+}
+
+func (*listCmd) Name() string     { return "ls" }
+func (*listCmd) Synopsis() string { return "List items in a collection" }
+func (*listCmd) Usage() string {
+	return `ls [-d <date>] <collection URL>
+
+  List the docstores in <collection URL>.
```
jba: docstores => documents

samples/gocdk-docstore/main.go
```diff
@@ -0,0 +1,205 @@
...
+	if f.NArg() != 1 {
+		f.Usage()
+		return subcommands.ExitUsageError
+	}
+	collectionURL := f.Arg(0)
+
+	// Open a *docstore.Collection using the collectionURL.
+	collection, err := docstore.OpenCollection(ctx, collectionURL)
+	if err != nil {
+		log.Printf("Failed to open collection: %v\n", err)
```
jba: Remove ending newline.

samples/gocdk-docstore/main.go
```diff
@@ -0,0 +1,205 @@
...
+		return subcommands.ExitFailure
+	}
+
+	q := collection.Query()
+	if cmd.date != "" {
+		q = q.Where("Date", "=", cmd.date)
+	}
+	iter := q.Get(ctx)
+	for {
```
jba: defer iter.Stop()

samples/gocdk-docstore/main.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (*putCmd) Name() string     { return "put" }
+func (*putCmd) Synopsis() string { return "Put an item from stdin" }
+func (*putCmd) Usage() string {
+	return `put <collection URL> <message>
+
+  Read from stdin and put an message with the current timestamp in <collection URL>.
+
+  Example:
+    gocdk-docstore put firestore://mycollection "hello docstore` + helpSuffix
```
jba: Not the right firestore URL (but it may change anyway, see #2312).

samples/gocdk-docstore/main.go
```diff
@@ -0,0 +1,205 @@
...
+	// Open a *docstore.Collection using the collectionURL.
+	collection, err := docstore.OpenCollection(ctx, collectionURL)
+	if err != nil {
+		log.Printf("Failed to open collection: %v\n", err)
+		return subcommands.ExitFailure
+	}
+
+	t := time.Now()
+	if err := collection.Put(ctx, &Message{
+		Date:      t.Format("2006-01-02"),
```
jba: It looks like date is redundant -- you can get it from Timestamp.

samples/gocdk-docstore/main.go
```diff
@@ -0,0 +1,205 @@
...
+	q := collection.Query()
+	if cmd.date != "" {
+		q = q.Where("Date", "=", cmd.date)
+	}
+	if err := q.Delete(ctx); err != nil {
+		log.Printf("Failed to delete: %v", err)
+		return subcommands.ExitFailure
+	}
+	return subcommands.ExitSuccess
+}
```
jba: Maybe something that shows action lists, like: modify a document (with Update) and then Get and print the new document.

shantuo: Good idea, add an updateCmd.

samples/gocdk-docstore/main.go
shantuo: Changed.

samples/gocdk-docstore/main.go
shantuo: I use `Timestamp` as the key, firestore only takes string for the name field. Filed #2331.

jba: Oh, right.

samples/gocdk-docstore/main.go
jba: OK.

samples/gocdk-docstore/main.go
jba: On second thought, I think you should use something else as the primary key. Timestamp isn't good because it's not guaranteed to be unique. It also has the two problems I mentioned—it is redundant with date, and people will wonder why it's not a time.Time (I don't know if we'll fix that for Firestore, because I don't think timestamps work as primary keys).

So how about generating a unique ID using the `github.com/google/uuid` package? 
```
type Message {
    Key string // unique random key
    Date string // date message was added, as a string for easy searching by date
   Content string
   ...
}
```

samples/gocdk-docstore/main.go
```diff
@@ -0,0 +1,261 @@
...
+
+func (*putCmd) Name() string     { return "put" }
+func (*putCmd) Synopsis() string { return "Put an item from stdin" }
+func (*putCmd) Usage() string {
+	return `put <collection URL> <message>
+
+  Read from stdin and put an message with the current timestamp in <collection URL>.
+
+  Example:
+    gocdk-docstore put "firestore://projects/myproject/databases/(default)/documents/mycollection?name_field=myID" "hello docstore` + helpSuffix
```
jba: Doesn't name_field have to match what you have? (`Timestamp` in your version, or `Key` in my suggestion)

jba: Maybe show a simpler example, like for mongo.

samples/gocdk-docstore/main.go
shantuo: Done.

samples/gocdk-docstore/main.go
shantuo: Done.

shantuo:
I did manual test.

Issue #2148, opened 2019-05-18
gocdk: add a "biome provision" command


vangent:
For adding Terraform to provision various types of resources.

* Add a new Terraform file and import it from `main.tf` (or, append to `main.tf`).
* Add something to `outputs.tf` with the resulting URL.

The "demo add" command and/or "biome add" commands could be updated to use this to fill in the grid of [demo x biome].

vangent:
Done!

PR #2313, opened 2019-06-11
internal/gocdk: consolidate materialize mechanisms, with a JSON config

vangent:
Updates #2148.


internal/cmd/gocdk/internal/static/static.go
```diff
@@ -14,12 +14,285 @@
...
+	// Logger is used to log information about changes.
+	Logger *log.Logger
+	// TemplateData is passed when executing templates marked with
+	// fileOptions.IsTemplate.
+	TemplateData interface{}
+}
+
+// TODO(rvangent): There are several cases where multiple places under
+// static/assets_ are repeating the same data:
+// 1. For places that we ask the user questions; currently, we are hardcoding
```
vangent: As mentioned today, I plan to tackle this in a separate PR that adds a "prompt" package; the `.tf` files etc. in the existing static resource directories would get removed in favor of some sort of signal in the `.json` file that the prompt is needed.

internal/cmd/gocdk/internal/static/static.go
```diff
@@ -14,12 +14,285 @@
...
+//    their "answer" (e.g., in provision/blob/gcsblob/mainlocal_gcpproject.tf).
+//    But each place that wants to ask the question would need to repeat the
+//    .tf file and the .json config for it. Instead, we should have something
+//    like a "prompt" package that handles asking questions and (if needed)
+//    saving them. Each "provision" provider should just specify which
+//    questions it needs answered.
+// 2. Similarly, there are shared resources like the "random" provider that are
```
vangent: I think this can be tackled by something like a `shared` directory in `static`, which the `.json` file can reference. E.g., `Path = ../shared/random_provider.tf`.

internal/cmd/gocdk/testdata/recordreplay/Provision/out.txt
```diff
@@ -17,6 +17,9 @@ $ cd myproj
 
 $ gocdk provision add dev blob/fileblob
 gocdk: Adding "blob/fileblob" to "dev"...
+gocdk:   added a new file "fileblob.tf" to your project
+gocdk:   added a Terraform provider "local" for creating a scratch directory to "main.tf"
+gocdk:   added a BLOB_BUCKET_URL variable to "outputs.tf"
```
vangent: Let me know what you think about these messages. I think they might be useful to users, to see what changes we are making for them, but we could also remove them and let them use `git diff` or whatever. Currently the caller of `Materialize` can decide whether to log them, by setting (or not) `MaterializeOptions.Logger` -- so, `gocdk init` does not, `gocdk provision add` does.

internal/cmd/gocdk/internal/static/static.go
```diff
@@ -14,12 +14,285 @@
...
+// dstRoot.
+//
+// By default, it copies files. If a file already exists at the destination,
+// Materialize will return an error unless force is true.
+//
+// Additional behavior can be configured by adding a JSON file to the static
+// data, at srcRoot + ".json". The JSON file should consist of a sequence of
+// JSON objects that will be deserialized into fileOptions structs; see
+// the struct docstring for more info.
+func Materialize(dstRoot, srcRoot string, opts *MaterializeOptions) error {
```
vangent: An alternative to the `.json` files with configuration would be to do that in Go, and pass it here. So, each caller would construct a map of path -> file options and pass it to `Materialize`. I'm starting to lean more toward that approach because it will probably work better for things in the TODOs above like shared resources and prompts. Feedback welcome.

vangent:
Dropping this. I think I'm going to go with the more code-based approach than the `.json` one here.

zombiezen:
> An alternative to the `.json` files with configuration would be to do that in Go, and pass it here. So, each caller would construct a map of path -> file options and pass it to `Materialize`. I'm starting to lean more toward that approach because it will probably work better for things in the TODOs above like shared resources and prompts.

That sounds like a good idea. I was going to suggest something like that offline yesterday, but I didn't have an intuitive sense for which way would work better.

PR #2317, opened 2019-06-12
docstore: add Collection.Close

### LINKED ISSUE #2310: docstore: add Collection.Close

jba:
Fixes #2310.

### DIFF
--- a/internal/docstore/docstore.go
+++ b/internal/docstore/docstore.go
@@ -21,6 +21,7 @@ import (
 	"reflect"
 	"sort"
 	"strings"
+	"sync"
 	"unicode/utf8"
 
 	"gocloud.dev/gcerrors"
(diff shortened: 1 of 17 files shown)


jba:
PTAL.

internal/docstore/docstore.go
```diff
@@ -52,7 +54,21 @@ var NewCollection = newCollection
 
 // newCollection makes a Collection.
 func newCollection(d driver.Collection) *Collection {
-	return &Collection{driver: d}
+	c := &Collection{driver: d}
+	_, file, lineno, ok := runtime.Caller(1)
+	runtime.SetFinalizer(c, func(c *Collection) {
```
vangent: Do you want to check this in? We don't have it for other top-level structs like `blob.Bucket` (just for smaller-scoped things like `blob.Reader`). But maybe we should...?

jba: In a later CL I'll make a note to remove it. We'll be adding more tests, so it may still be useful.

PR #2318, opened 2019-06-12
firedocstore: accept the full collection resource ID as a single string

### LINKED ISSUE #2312: firedocstore: constructor should take a single resource string
According to https://github.com/google/go-cloud/blob/master/internal/docs/design.md#compound-ids, constructors should take a single string denoting the resource.

The actual Google Cloud Firestore resource string would be `projects/$PROJECT/databases/(default)/documents/$COLLECTION`. If I understand the design doc right, the `OpenCollection` function would take this string, and we could provide a helper function (named what?) that would take a project ID and collection and return this string. So one would write
```
firedocstore.OpenCollection(client, firedocstore.ResourceString(projectID, collPath), ...)
```
which seems like quite a mouthful.

Alternatively, `OpenCollection` could accept the string `$PROJECT/$COLLECTION`. That seems more reasonable, but I don't know if it satisfies the design rule.

jba:
Fixes #2312.

### DIFF
--- a/internal/docstore/firedocstore/example_test.go
+++ b/internal/docstore/firedocstore/example_test.go
@@ -87,7 +87,7 @@ func Example_openCollectionFromURL() {
 	ctx := context.Background()
 
 	// docstore.OpenCollection creates a *docstore.Collection from a URL.
-	url := "firestore://projects/my-project/databases/(default)/documents/my-collection?name_field=userID"
+	const url = "firestore://projects/my-project/databases/(default)/documents/my-collection?name_field=userID"
 	coll, err := docstore.OpenCollection(ctx, url)
 	if err != nil {
 		log.Fatal(err)
--- a/internal/docstore/firedocstore/example_test.go
+++ b/internal/docstore/firedocstore/example_test.go
@@ -37,7 +37,8 @@ func ExampleOpenCollection() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	coll, err := firedocstore.OpenCollection(client, "my-project", "my-collection", "userID", nil)
+	resourceID := firedocstore.CollectionResourceID("my-project", "my-collection")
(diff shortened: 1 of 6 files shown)


internal/docstore/firedocstore/example_test.go
```diff
@@ -86,7 +87,8 @@ func Example_openCollectionFromURL() {
 	ctx := context.Background()
 
 	// docstore.OpenCollection creates a *docstore.Collection from a URL.
-	coll, err := docstore.OpenCollection(ctx, "firestore://my-project/my-collection?name_field=userID")
+	url := "firestore://projects/my-project/databases/(default)/documents/my-collection?name_field=userID"
```
vangent: nit: const url?

internal/docstore/firedocstore/fs.go
```diff
@@ -93,17 +93,23 @@ type Options struct {
...
+// See the OpenCollection example for use.
+func CollectionResourceID(projectID, collPath string) string {
+	return fmt.Sprintf("projects/%s/databases/(default)/documents/%s", projectID, collPath)
+}
+
 // OpenCollection creates a *docstore.Collection representing a Firestore collection.
 //
-// collPath is the path to the collection, starting from a root collection. It may
-// refer to a top-level collection, like "States", or it may be a path to a nested
+// collResourceID must be of the form "project/<projectID>/databases/(default)/documents/<collPath>".
```
vangent: If you can find a link documenting this form, please add. Especially if it includes stuff like advice on how to copy/paste from the UI.

internal/docstore/firedocstore/example_test.go
jba: done.

internal/docstore/firedocstore/fs.go
jba: Done, as best I could.

PR #2319, opened 2019-06-12
docstore: reorganize files

jba:
- Move things to different files.

- Delete redundant example.


internal/docstore/docstore.go
```diff
@@ -18,7 +18,7 @@ import (
 	"context"
 	"fmt"
 	"log"
-	"net/url"
+
```
shantuo: Remove the blank line.

jba: OK, in next PR.

PR #2323, opened 2019-06-13
mongodocstore: add cosmos test

### LINKED ISSUE #2088: docstore: test mongo driver on Azure Cosmos
Make sure we can use our mongoDB driver against Azure Cosmos. If necessary, add code to simplify the experience.

jba:
To enable the test, pass the -cosmos flag to `go test` with a connection string.

Fixes #2088.

### DIFF
--- a/internal/docstore/mongodocstore/cosmos_test.go
+++ b/internal/docstore/mongodocstore/cosmos_test.go
@@ -16,26 +16,33 @@ package mongodocstore
 
 import (
 	"context"
-	"flag"
+	"os"
 	"testing"
 
 	"gocloud.dev/internal/docstore/drivertest"
+	"gocloud.dev/internal/testing/setup"
 )
 
 // Run conformance tests on Azure Cosmos.
-// We cannot record this interaction, so this test must be enabled with a flag.
 
-var cosmosConnectionString = flag.String("cosmos", "", "connection string for Azure Cosmos")
+var (
+	// See https://docs.microsoft.com/en-us/azure/cosmos-db/connect-mongodb-account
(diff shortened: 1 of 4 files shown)


jba:
Makes sense to have it work automatically with the prelease script. Changed to use -record and env var.

The problem is that it won't run, and neither will azuresb, because there is no testdata directory. I added almost-empty testdata directories for both.

jba:
PTAL.

vangent:
> The problem is that it won't run, and neither will azuresb, because there is no testdata directory. I added almost-empty testdata directories for both.

Why doesn't it run if there's no `testdata` directory? They aren't actually recording...?

jba:
> Why doesn't it run if there's no `testdata` directory? They aren't actually recording...?

prereleasechecks.sh filters on that:
```
while read -r TESTDIR; do
...
done < <( find . -name testdata -printf "%h\n" )
```

jba:
By the way, out of curiosity, why do you write the loop like that instead of
```
for TESTDIR in $(find...); do
```
?

vangent:
> By the way, out of curiosity, why do you write the loop like that instead of
> 
> ```
> for TESTDIR in $(find...); do
> ```
> 
> ?

Because I suck at bash? Feel free to fix :-).

vangent:
> > Why doesn't it run if there's no `testdata` directory? They aren't actually recording...?
> 
> prereleasechecks.sh filters on that:
> 
> ```
> while read -r TESTDIR; do
> ...
> done < <( find . -name testdata -printf "%h\n" )
> ```

Ah OK.

PR #2329, opened 2019-06-13
internal/docstore: test struct types in actions

shantuo:
First step, test struct types in all actions.

Some fixes:
1. dynamodb: properly detects missing key for struct types.
2. dynamodb: properly chains increment expressions.
3. firestore: properly handles missing key for structs.


internal/docstore/firedocstore/fs.go
```diff
@@ -168,7 +168,7 @@ func newCollection(client *vkit.Client, collResourceID, nameField string, nameFu
 func (c *collection) Key(doc driver.Document) (interface{}, error) {
 	if c.nameField != "" {
 		name, err := doc.GetField(c.nameField)
-		if err != nil {
+		if err != nil || name == nil {
```
jba: I'm not sure about this.

If you wrote
```
map[string]interface{}{"name": nil}
```
then that's probably a bug.

If instead you had a struct and you left the name field blank, then shouldn't we be checking for the zero value? It could be a `string` or `int` field, for example. 

I guess if we check for the zero value, we will have to treat the above situation with a map as a missing key instead of a bug. But that's not too bad.

shantuo: We could also check for zero value besides nil, as we use `interface{}` in the tests.

I was trying to distinguish the two case:
```go
m := map[string]interface{}{"name": nil} // InvalidArgument
s := struct{name interface{}}{} // ok (generate key)
```

Will that be too confusing?

jba: I've been trying to ignore the differences between maps and structs, which is why everything happens with `driver.Document`. So I don't think we should distinguish those cases, even though the first is probably a bug. But I'm also worried about
```
struct { Name string }
```
where it won't be nil, it will be `""`

So that's why I think the right thing is to check for the zero value. That will be nil for interface{} and "" for string.

shantuo: Done.

docstore/driver/codec.go
```diff
@@ -715,8 +715,8 @@ func wrap(err error, code gcerr.ErrorCode) error {
 var fieldCache = fields.NewCache(parseTag, nil, nil)
 
 // Copied from encoding/json, go 1.12.
-func isEmptyValue(v reflect.Value) bool {
-	switch v.Kind() {
+func IsEmptyValue(v reflect.Value) bool {
```
jba: Unfortunately encoding/json's definition of "empty value" is not identical to "zero value," and I think we want "zero value". The difference is that maps and slices must be nil, not just have length zero.

jba: Never mind, changed my mind.

jba:
Just document somewhere what we mean by a key being present.

shantuo:
I added some comments to `docstore.Create` and `driver.Collection.Key` explaining empty "key". I didn't include all types in the explanation since I think only a few types are commonly used as keys. PTAL

docstore/docstore.go
```diff
@@ -136,8 +136,9 @@ func (l *ActionList) add(a *Action) *ActionList {
 // AlreadyExists is returned if it does. (See gocloud.dev/gcerrors for more on error
 // codes.)
 //
-// If the document doesn't have key fields, key fields with unique values will be
-// created and doc will be populated with them.
+// If the document doesn't have key fields, or the key fields is empty, meaning
```
jba: is => are
meaning as => meaning
and any => or any

shantuo: Done.

PR #2330, opened 2019-06-13
postgres: remove constructors in favor of URLOpeners

zombiezen:
Fixed consistency issues as I found them.

Updates #1175


zombiezen:
@shantuo, I'd like you to review since @vangent is out next week. I'll have follow-up PRs to do the same for MySQL, but I'm going to take that one slower since it affects guestbook.

I have manually tested each of the packages as I made changes.

postgres/cloudpostgres/cloudpostgres.go
```diff
@@ -100,100 +100,50 @@ func (uo *URLOpener) OpenPostgresURL(ctx context.Context, u *url.URL) (*sql.DB,
...
 	}
-	params.TraceOpts = uo.TraceOpts
-	db, _, err := Open(ctx, uo.CertSource, params)
-	return db, err
+
+	query := u.Query()
+	for k := range query {
+		// Only permit parameters that do not conflict with other behavior.
+		if k == "sslmode" || k == "sslcert" || k == "sslkey" || k == "sslrootcert" {
+			return nil, fmt.Errorf("cloudpostgres: open: extra parameter %s not allowed; use Params fields instead", k)
```
vangent: "use Params fields instead" is stale.

postgres/rdspostgres/rdspostgres.go
```diff
@@ -33,31 +32,12 @@ import (
...
-	// TraceOpts contains options for OpenCensus.
-	TraceOpts []ocsql.TraceOption
-}
-
 // URLOpener opens RDS PostgreSQL URLs
 // like "rdspostgres://user:password@myinstance.borkxyzzy.us-west-1.rds.amazonaws.com:5432/mydb".
 type URLOpener struct {
+	// CertSource specifies how the opener will obtain the RDS Certificate
+	// Authority. If nil, it will use the default *rds.CertFetcher.
+	CertSource rds.CertPoolProvider
```
vangent: Why not use the "lazy" pattern used elsewhere here? Is getting the default CertFetcher just super cheap for some reason, while for `cloudpostgres` it is not?

postgres/rdspostgres/rdspostgres_test.go
```diff
@@ -104,6 +61,7 @@ func TestOpenWithURL(t *testing.T) {
 		{fmt.Sprintf("rdspostgres://%s:%s@%s/wrongdbname", username, password, endpoint), false, true},
 		{fmt.Sprintf("rdspostgres://%s:%s@%s/%s?foo=bar", username, password, endpoint, databaseName), false, true},
```
vangent: It looks like you deleted all of the tests verifying that `sslmode` etc. are rejected, can you add some back here?

zombiezen: They are already covered by other test cases, see ~10 lines above.

postgres/cloudpostgres/cloudpostgres.go
zombiezen: Done.

postgres/rdspostgres/rdspostgres.go
zombiezen: Yes. `rds.CertFetcher` does not require credentials (which is expensive to fetch), it purely fetches a root CA and can do that on-demand.

postgres/rdspostgres/rdspostgres_test.go
vangent: Oh sorry, GitHub being clever and hiding lines.

PR #2337, opened 2019-06-14
internal/gocdk: add code for prompting the user and optionally saving the response

vangent:


internal/cmd/gocdk/internal/prompt/prompt.go
```diff
@@ -0,0 +1,117 @@
...
+// TODO(rvangent): Add support for validation?
+// TODO(rvangent): Ctrl-C doesn't work; maybe replace cancel.
+func String(in io.Reader, out io.Writer, msg, dflt string) (string, error) {
+	fmt.Fprintln(out)
+	fmt.Fprintln(out, msg)
+	if dflt != "" {
+		fmt.Fprintln(out, "  default:", dflt)
+	}
+	const prompt = `Enter a value, or "cancel": `
+	reader := bufio.NewReader(in)
```
zombiezen: This can be tricky: if `in` is a `*bufio.Reader`, this function will simply pass it through, but otherwise this can read more than just the particular line and then discard the data. This can result in unexpected behavior if `String` is called repeatedly.

Probably best to take in a `*bufio.Reader` directly and push responsibility to the caller for arranging for the correct thing to happen.

internal/cmd/gocdk/internal/prompt/prompt.go
```diff
@@ -0,0 +1,117 @@
...
+	"io"
+	"strings"
+
+	"gocloud.dev/internal/cmd/gocdk/internal/static"
+)
+
+// String prompts the user to input a string. It outputs msg to out, then
+// prompts the user to enter a value. If the user just hits return, the
+// value defaults to dflt.
+// The function only returns an error if the user enters "cancel".
```
zombiezen: And on I/O errors.

internal/cmd/gocdk/internal/prompt/prompt.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	const prompt = `Enter a value, or "cancel": `
+	reader := bufio.NewReader(in)
+	for {
+		fmt.Fprintf(out, prompt)
+		s, err := reader.ReadString('\n')
+		if err != nil {
+			return "", err
+		}
+		s = strings.Replace(s, "\n", "", -1)
```
zombiezen:
```suggestion
    s = strings.TrimSuffix(s, "\n")
```

internal/cmd/gocdk/internal/prompt/prompt.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"gocloud.dev/internal/cmd/gocdk/internal/static"
+)
+
+// String prompts the user to input a string. It outputs msg to out, then
+// prompts the user to enter a value. If the user just hits return, the
+// value defaults to dflt.
+// The function only returns an error if the user enters "cancel".
+// TODO(rvangent): Add support for validation?
+// TODO(rvangent): Ctrl-C doesn't work; maybe replace cancel.
```
zombiezen: Ctrl-C will work if you accept a `context.Context`.

internal/cmd/gocdk/internal/prompt/prompt.go
```diff
@@ -0,0 +1,117 @@
...
+func String(in io.Reader, out io.Writer, msg, dflt string) (string, error) {
+	fmt.Fprintln(out)
```
zombiezen: This prompt seems verbose to me. IIUC the result is:

```
Enter a GCP region. default: us-central1
Enter a value, or "cancel":
```

What about:

```
Enter a GCP region [default: us-central1], or "cancel": 
```

(We could remove the `or "cancel"` if we support Ctrl-C, see above.)

internal/cmd/gocdk/internal/static/static.go
```diff
@@ -132,8 +133,27 @@ func AddProvider(provider string) *Action {
 	}
 }
 
+// ReadLocal reads a local Terraform variable from main.tf.
+// The local variable was likely added using an Action returned from AddLocal.
+// If the variable is not found, it returns the empty string and no error.
+func ReadLocal(biomeDir, key string) (string, error) {
```
zombiezen: Organization: This function doesn't feel like it conceptually belongs under the heading "static files" to me. Perhaps we should change the package name to reflect, but probably doesn't belong in this PR.

internal/cmd/gocdk/internal/prompt/prompt.go
vangent: Done.

internal/cmd/gocdk/internal/prompt/prompt.go
vangent: Done.

internal/cmd/gocdk/internal/prompt/prompt.go
vangent: Done.

internal/cmd/gocdk/internal/prompt/prompt.go
vangent: Updated TODO.

internal/cmd/gocdk/internal/prompt/prompt.go
vangent: I was trying to follow what Terraform does. We can revisit how we want prompts to look later; added a TODO.

internal/cmd/gocdk/internal/static/static.go
vangent: Agreed; added a TODO.

vangent:
Thanks for the quick review; PTAL.

PR #2340, opened 2019-06-14
internal/gocdk: prompt for launcher in biome add; use biome add in init

### LINKED ISSUE #2216: internal/cmd/gocdk: init should use `biome add`
Currently, the `init` command and the `biome add` command use distinct templates. They should use the same templates, and `init` should call `biome add` to create the `dev` biome.

### LINKED ISSUE #2217: internal/cmd/gocdk: `biome add` flag for specifying launcher
The `biome add` command should have a flag allowing the user to specify a launcher.

vangent:
Fixes #2216.
Fixes #2217.


internal/cmd/gocdk/biome.go
```diff
@@ -28,37 +32,111 @@ func registerBiomeCmd(ctx context.Context, pctx *processContext, rootCmd *cobra.
...
+		reader := bufio.NewReader(pctx.stdin)
+		projectID, err := prompt.GCPProjectID(reader, pctx.stderr)
+		if err != nil {
+			return xerrors.Errorf("biome add: %w", err)
+		}
+		saveActions = append(saveActions, static.AddLocal(prompt.GCPProjectIDTfLocalName, projectID))
+		// TODO(rvangent): Eventually we should ask for the GCP region; currently
+		// cloudrun is singly homed in us-central1.
+		region := "us-central1"
+		/*
```
zombiezen: Delete commented out code?

internal/cmd/gocdk/init.go
```diff
@@ -87,6 +88,12 @@ func doInit(ctx context.Context, pctx *processContext, dir, modpath string, allo
 	if err := static.Do(projectDir, nil, actions...); err != nil {
 		return xerrors.Errorf("gocdk init: %w", err)
 	}
+	// Add a "dev" biome with a local launcher, using a new pctx that discards
+	// the normal output of "gocdk biome add".
+	subpctx := newProcessContext(projectDir, pctx.stdin, ioutil.Discard, ioutil.Discard)
+	if err := biomeAdd(ctx, subpctx, "dev", "local"); err != nil {
```
zombiezen: Is there a way we can factor out the code in biome add such that we don't have to create a new process context?

internal/cmd/gocdk/biome.go
```diff
@@ -28,37 +32,111 @@ func registerBiomeCmd(ctx context.Context, pctx *processContext, rootCmd *cobra.
...
 		Short: "TODO Add BIOME_NAME",
 		Long:  "TODO more about adding biomes",
 		Args:  cobra.ExactArgs(1),
 		RunE: func(_ *cobra.Command, args []string) error {
-			return biomeAdd(ctx, pctx, args[0])
+			return biomeAdd(ctx, pctx, args[0], launcher)
 		},
 	}
+	// TODO(rvangent): There should be a way to specify answer to launcher-specific
+	// prompts via flags. Maybe a JSON snippet?
```
zombiezen: This should definitely have an issue opened for it, since we've been talking about this for a while.

internal/cmd/gocdk/biome.go
```diff
@@ -28,37 +32,111 @@ func registerBiomeCmd(ctx context.Context, pctx *processContext, rootCmd *cobra.
...
+		if err != nil {
+			return err
+		}
+	}
+
+	data := &launcherData{Launcher: launcher}
+	var launchSpecifiers []*launchSpecifier
+	var saveActions []*static.Action
+	switch launcher {
+	case "local":
```
zombiezen: (Optional) Create named constants to share with the `newLauncher` function?

internal/cmd/gocdk/biome.go
vangent: Sure, OK.

internal/cmd/gocdk/biome.go
vangent: Filed https://github.com/google/go-cloud/issues/2346.

internal/cmd/gocdk/biome.go
vangent: I'm not going to do that here. IMHO there should be a larger refactor where you specify everything for a launcher in one place, and this would just call it.

internal/cmd/gocdk/init.go
vangent: Yes, but not easily; it uses a bunch of `pctx` fields so we'd end up needing to pass almost all of them. Not sure that's an improvement.

Note that the working directory of this `subpctx` is also different than the base one.

vangent:
PTAL

internal/cmd/gocdk/init.go
zombiezen: Mind opening an issue? It feels like there should be some way of mitigating this, but I don't have a concrete suggestion right now.

vangent: https://github.com/google/go-cloud/issues/2348

PR #2343, opened 2019-06-14
docstore/firedocstore: explain uses for name function

### LINKED ISSUE #2297: docstore/firedocstore: convert non-string name field to string if possible

The Firestore driver returns an error when the name field is not a string, while other drivers don't have that requirement.

Convert the name field to a string is possible, otherwise return an InvalidArgument error.

jba:
Fixes #2297.

### DIFF
--- a/docstore/firedocstore/fs.go
+++ b/docstore/firedocstore/fs.go
@@ -106,9 +106,9 @@ func CollectionResourceID(projectID, collPath string) string {
 // collection, like "States/Wisconsin/Cities".
 // See https://cloud.google.com/firestore/docs/reference/rest/ for more detail.
 //
-// firedocstore requires that a single field, nameField, be designated the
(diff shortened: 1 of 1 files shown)


docstore/firedocstore/fs.go
```diff
@@ -106,9 +106,9 @@ func CollectionResourceID(projectID, collPath string) string {
 // collection, like "States/Wisconsin/Cities".
 // See https://cloud.google.com/firestore/docs/reference/rest/ for more detail.
 //
-// firedocstore requires that a single string field, nameField, be designated the
-// primary key. Its values must be unique over all documents in the collection, and
-// the primary key must be provided to retrieve a document.
+// firedocstore requires that a single field, nameField, be designated the
```
shantuo: nit: 110 seems much longer than 109

PR #2344, opened 2019-06-14
internal/gocdk: install Terraform on Travis

### LINKED ISSUE #2342: internal/gocdk: install Terraform on Travis so we don't have to skip tests
`apply_test` is skipped, but likely there will be more.

vangent:
Fixes #2342.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ before_install:
   - unzip /tmp/terraform.zip -d /tmp
   - mkdir -p ~/bin
   - mv /tmp/terraform ~/bin
-  - export PATH="~/bin:$PATH"
+  - export PATH="$HOME/bin:$PATH"
   - terraform -version
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,18 @@ before_install:
       HOMEBREW_NO_AUTO_UPDATE=1 brew install bash grep;
       mv $(brew --prefix)/bin/ggrep $(brew --prefix)/bin/grep;
     fi
+  # Install Terraform.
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -14,6 +14,18 @@ before_install:
...
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+      TERRAFORM_OS="darwin";
+    else
+      TERRAFORM_OS="$TRAVIS_OS_NAME";
+    fi
+  - curl -sLo /tmp/terraform.zip https://releases.hashicorp.com/terraform/0.11.11/terraform_0.11.11_"${TERRAFORM_OS}"_amd64.zip
+  - unzip /tmp/terraform.zip -d /tmp
+  - mkdir -p ~/bin
+  - mv /tmp/terraform ~/bin
+  - export PATH="~/bin:$PATH"
```
zombiezen: Use `$HOME` instead of `~` (shell magic character).

```suggestion
  - export PATH="$HOME/bin:$PATH"
```

internal/cmd/gocdk/apply_test.go
```diff
@@ -33,10 +32,6 @@ func TestApply(t *testing.T) {
 	// terraform never previously applied
 	// terraform apply repeat
 
-	if _, err := exec.LookPath("terraform"); err != nil {
```
zombiezen: Why remove the skip? If we're running on a dev's machine that doesn't have Terraform installed, then `go test ./...` should still succeed IMO.

vangent: Sure, OK.

vangent:
PTAL.

PR #2347, opened 2019-06-14
internal/gocdk: upgrade Terraform version to v0.12

### LINKED ISSUE #2341: internal/gock: our terraform files don't work with the latest version of Terraform
I tried upgrading to Terraform `v0.12` and got a bunch of errors:

```
Warning: Skipping backend initialization pending configuration upgrade
The root module configuration contains errors that may be fixed by running the
configuration upgrade tool, so Terraform is skipping backend initialization.
See below for more information.
Terraform has initialized, but configuration upgrades may be needed.
Terraform found syntax errors in the configuration that prevented full
initialization. If you've recently upgraded to Terraform v0.12, this may be
because your configuration uses syntax constructs that are no longer valid,
and so must be updated before full initialization is possible.
Terraform has installed the required providers to support the configuration
upgrade process. To begin upgrading your configuration, run the following:
    terraform 0.12upgrade
To see the full set of errors that led to this message, run:
    terraform validate
Error: Missing required argument
  on outputs.tf line 6, in output "launch_environment":
   6: output "launch_environment" {
The argument "value" is required, but no definition was found.
Error: Unsupported block type
  on outputs.tf line 7, in output "launch_environment":
   7:   value {
Blocks of type "value" are not expected here. Did you mean to define argument
"value"? If so, use the equals sign to assign it a value.
Error: Missing required argument
  on outputs.tf line 17, in output "launch_specifier":
  17: output "launch_specifier" {[0m
The argument "value" is required, but no definition was found.
Error: Unsupported block type
  on outputs.tf line 18, in output "launch_specifier":
  18:   value {
Blocks of type "value" are not expected here. Did you mean to define argument
"value"? If so, use the equals sign to assign it a value.
```

vangent:
Updates #2341.


internal/cmd/gocdk/internal/static/_assets/biome/versions.tf
```diff
@@ -0,0 +1,3 @@
+terraform {
```
zombiezen: This should go in main.tf, not a new file.

internal/cmd/gocdk/biomeutil.go
```diff
@@ -132,7 +132,7 @@ func tfReadOutput(ctx context.Context, dir string, env []string) (map[string]*tf
 
 // tfOutput describes a single output value.
 type tfOutput struct {
-	Type      string      `json:"type"` // one of "string", "list", or "map"
+	Type      interface{} `json:"type"`
```
zombiezen: I don't understand this change. What breaks here?

internal/cmd/gocdk/internal/static/_assets/provision/blob/fileblob.tf
```diff
@@ -1,7 +1,7 @@
 # TODO(rvangent): Add comments explaining.
 
 locals {
-  blob_bucket_path    = "${path.module}/fileblob_scratchdir"
+  blob_bucket_path    = "${path.cwd}/fileblob_scratchdir"
```
zombiezen: This is a behavior change and I don't believe it is correct: `${path.module}` stores the directory inside the biome whereas `${path.cwd}` stores it wherever the `terraform apply` is run from.

internal/cmd/gocdk/internal/static/_assets/biome/versions.tf
vangent: OK. The upgrade process created a new file, but I missed that we already had a version section in `main.tf`.

Done.

internal/cmd/gocdk/biomeutil.go
vangent: `terraform output -json` doesn't return "map" for the type anymore:

```
  "launch_specifier": {
    "sensitive": false,
    "type": [
      "object",
      {
        "host_port": "number"
      }
    ],
    "value": {
      "host_port": 8080
    }
  }
```

Since we don't actually use it, I just changed it to `interface{}`.

internal/cmd/gocdk/internal/static/_assets/provision/blob/fileblob.tf
vangent: Hmm, I'm not sure what to do then. The output of `${path.module}` changed in this Terraform version; it is no longer giving an absolute path, but a relative one, resulting in:

```
  "BLOB_BUCKET_URL" = "file://./fileblob_scratchdir"
```

which does not work. I haven't been able to find a Terraform function that makes a path absolute.

vangent: Given that we run Terraform, `cwd` seems like it works...

internal/cmd/gocdk/biomeutil.go
zombiezen: Interesting. Let's remove the field since we don't use it. :)

internal/cmd/gocdk/internal/static/_assets/provision/blob/fileblob.tf
zombiezen: That's.... really unfortunate. That feels like a Terraform bug.

internal/cmd/gocdk/biomeutil.go
vangent: OK. Removed `Sensitive` as well, since it's also unused.

zombiezen:
Discussed offline. This seems disruptive (it affects our other tests), so let's hold off the upgrade until we have a clearer picture of what breaks.

internal/cmd/gocdk/internal/static/_assets/provision/blob/fileblob.tf
vangent: Maybe. I don't think they really want to support local files since they have weird behavior for multiple users sharing `.tfstate`.

vangent:
> Discussed offline. This seems disruptive (it affects our other tests), 
> so let's hold off the upgrade until we have a clearer picture of what breaks.

Up to you -- tests are passing now. I imagine we may also need to update `guestbook` and the various other places where we use Terraform, but that could happen as separate PRs.

vangent:
Do you think I should update Guestbook and all the other Terraform files in our repo as part of this PR?

zombiezen:
Yes, I think that having parts of the repo being on one version of Terraform and the others not is going to be more confusing.

vangent:
OK I will reopen when I have everything upgraded.

PR #2370, opened 2019-06-17
internal/cmd/gocdk: Check for project dir relative to GOPATH/src

### LINKED ISSUE #2307: internal/cmd/gocdk: inferred modules in init prefixed with "src"
### Describe the bug

The current logic for inferring the module path based on GOPATH evaluates relative to `$GOPATH` instead of `$GOPATH/src`, thus producing `src/example.com/foo` for a directory `$GOPATH/src/example.com/foo`. The desired module path is `example.com/foo`.

### To Reproduce

```
gocdk init $GOPATH/src/xyzzy
grep '^module' $GOPATH/src/xyzzy/go.mod
```

## Expected behavior

```
module xyzzy
```

This should always match the result of running:

```
mkdir $GOPATH/src/xyzzy
cd $GOPATH/src/xyzzy
go mod init
grep '^module' go.mod
```

### Version

master (197d5f6625b911fb268cb583e5534d0457621711)

clausti:
fixes #2307


internal/cmd/gocdk/init_test.go
```diff
@@ -198,8 +198,9 @@ func TestInferModulePath(t *testing.T) {
 					t.Error(err)
 				}
 			}()
-
-			pctx := newTestProcessContext(dir)
+			srcDir := filepath.Join(dir, "src")
+			os.Mkdir(srcDir, 0777)
```
zombiezen: We should fail the test if this returns `err != nil`.

internal/cmd/gocdk/init_test.go
```diff
@@ -198,8 +198,9 @@ func TestInferModulePath(t *testing.T) {
 					t.Error(err)
 				}
 			}()
```
zombiezen: General comment: it would be nice to test for both the "src" and non-"src" variant as a regression test.

internal/cmd/gocdk/init_test.go
```diff
@@ -198,8 +198,9 @@ func TestInferModulePath(t *testing.T) {
...
-
-			pctx := newTestProcessContext(dir)
+			srcDir := filepath.Join(dir, "src")
+			os.Mkdir(srcDir, 0777)
+			pctx := newTestProcessContext(srcDir)
 			gopath, cleanup := tc.testGOPATH(dir)
 			defer cleanup()
 			pctx.env = []string{gopath}
```
zombiezen: Later down: does this test not assert for the module name? That's the part we care about for this fix, and I'm not sure if we're testing for it.

clausti: It's not currently asserting for the module name, no. I can add it.

clausti:
@zombiezen PTAL?

internal/cmd/gocdk/init_test.go
```diff
@@ -149,27 +149,42 @@ func TestInit(t *testing.T) {
 }
 func TestInferModulePath(t *testing.T) {
 	testCases := []struct {
-		name       string
-		testGOPATH func(string) (string, func())
-		wantErr    bool
+		name    string
+		setup   func(string) (string, string, func())
```
zombiezen: Please name the return values: gopath, initDir.

internal/cmd/gocdk/init_test.go
```diff
@@ -149,27 +149,42 @@ func TestInit(t *testing.T) {
...
+			},
+			true,
+		},
+		{
+			"multiple GOPATH entries, project in GOPATH/src",
+			func(dir string) (string, string, func()) {
+				srcDir := filepath.Join(dir, "src")
+				if err := os.Mkdir(srcDir, 0777); err != nil {
+					t.Error(err)
+				}
```
zombiezen: The second temporary directory could be rolled inside this one, which would obviate the need for a cleanup function.

PR #2373, opened 2019-06-17
mongodocstore: make its own module

### LINKED ISSUE #2240: docstore: put mongo into its own module
Let's do it after we make it alpha.

jba:
Also, change docstore examples so they don't depend on mongodocstore.

Fixes #2240.


allmodules
```diff
@@ -12,6 +12,7 @@
 
 # module-directory           released
 .                            yes
+docstore/mongodocstore       no
```
eliben: I believe this should be "yes". This information is only used by the release tool, and this module will be tagged

docstore/mongodocstore/go.mod
```diff
@@ -0,0 +1,28 @@
...
+
+require (
+	github.com/go-stack/stack v1.8.0 // indirect
+	github.com/golang/snappy v0.0.1 // indirect
+	github.com/google/go-cmp v0.3.0
+	github.com/tidwall/pretty v1.0.0 // indirect
+	github.com/xdg/scram v0.0.0-20180814205039-7eeb5667e42c // indirect
+	github.com/xdg/stringprep v1.0.0 // indirect
+	go.mongodb.org/mongo-driver v1.0.3
+	gocloud.dev v0.0.0-00010101000000-000000000000
```
eliben: v0.15?

samples/go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	github.com/gorilla/mux v1.7.2
 	go.opencensus.io v0.22.0
 	gocloud.dev v0.15.0
+	gocloud.dev/docstore/mongodocstore v0.0.0-00010101000000-000000000000
```
shantuo: This version looks weird, is it because it is a complete new module and gomod cannot find any revision?

allmodules
jba: Done.

docstore/mongodocstore/go.mod
jba: Done.

samples/go.mod
jba: I guess?

Issue #2371, opened 2019-06-17
all: upgrade to Wire 0.3.0


zombiezen:
Wire 0.3.0 has one notable breaking change and a new feature that causes deprecation notices to be emitted:

1. The semantics of `wire.Bind` have changed, so a bunch of callsites break.
2. Struct providers have been changed to use `wire.Struct`.

Before Wire 0.3.0 goes out, we should upgrade Go CDK.

/cc @shantuo

zombiezen:
This was fixed in ea7a522dac81dd35f2c89d93e445e52ab14e51a8.

PR #2374, opened 2019-06-17
all: upgrade to Wire 0.3.0 prerelease BREAKING_CHANGE_OK

zombiezen:
Updates struct providers and wire.Bind calls.

Updates #2371


blob/s3blob/s3blob.go
```diff
@@ -88,8 +88,7 @@ func init() {
 
 // Set holds Wire providers for this package.
 var Set = wire.NewSet(
-	Options{},
-	URLOpener{},
+	wire.Struct(new(URLOpener), "ConfigProvider"),
```
shantuo: My understanding of having these `Set`s types provides convenient ways for the user to compose their provider sets. I think it is ok to have some defined required fields to use them, however, I'm not sure what the criteria we would use. Is it that the field is required for the type to work at all, or maybe 80% of the time the user would want to use the "sets" of fields we defined?

I think most of the changes here look OK since a lot of them only have one field and it often makes sense to provide it. Maybe the solution here is document it in our design.md.

go.mod
```diff
@@ -37,7 +37,7 @@ require (
 	github.com/google/go-cmp v0.3.0
 	github.com/google/martian v2.1.1-0.20190517191504-25dcb96d9e51+incompatible // indirect
 	github.com/google/uuid v1.1.1
-	github.com/google/wire v0.2.2
+	github.com/google/wire v0.3.0-0.20190611223505-93b1ce745f8d
```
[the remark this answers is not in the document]
zombiezen: Already done.

blob/s3blob/s3blob.go
zombiezen: Right. I don't see it as something beta-blocking (we can just add more sets over time), and I'd rather we get usage information before proceeding with it full-scale. My priority was to keep existing sets working in a reasonable fashion.

PR #2377, opened 2019-06-18
secrets: improve tests on malformed message

### LINKED ISSUE #1692: secrets: verify that Decrypt does not permit using the wrong key

In the [upcoming secrets how-to guide](https://github.com/google/go-cloud/blob/535dce4559dbb584c8dea26fa3141f218fa4cade/internal/website/content/howto/secrets/crypt.md#decrypting-data), I write that:

> The `Decrypt` method will return an error if the input data did not come from the same secret keeper.

I am actually not sure whether we guarantee this or whether we should guarantee it. It would be useful for an end-user to have this property, as it avoids the scenario where they encrypt data with key A, accidentally decrypt with key B, and potentially obtain garbage data.

We should determine whether our current set of providers (AWS KMS, GCP KMS, Vault, and Local Secrets) all raise errors. If they do (which @clausti and I have a sneaking suspicion of), then we should document that all drivers should do this, and create a conformance test that enforces the property.

shantuo:
Fixes #1692


secrets/drivertest/drivertest.go
```diff
@@ -238,8 +238,39 @@ func testDecryptMalformedError(t *testing.T, newHarness HarnessMaker) {
...
+		name      string
+		malformed []byte
+	}{
+		{
+			name:      "wrong first byte",
+			malformed: append([]byte{encryptedMsg[0] + 1}, encryptedMsg[1:]...),
+		},
+		{
+			name:      "missing second byte",
+			malformed: append(encryptedMsg[:1], encryptedMsg[2:]...),
```
zombiezen: This removes the byte for the rest of the test, though. Each of these slices should operate on a copy of `encryptedMsg`. The most concise way I can think of to do the copy is by creating a small helper function above:

```go
encryptedMsg, err := keeper.Encrypt(ctx, msg)
if err != nil {
	t.Fatal(err)
}
copyEncryptedMsg := func() []byte {
  return append([]byte(nil), encryptedMsg...)
}
```

shantuo: Done.

Issue #2357, opened 2019-06-14
mysql/postgres: reconsider package names


vangent:
E.g., maybe `cloudsql` should be `gcpcloudsql`.

https://github.com/google/go-cloud/blob/master/internal/docs/design.md#package-naming-conventions

zombiezen:
I think the GCP Cloud databases are the only ones that need changing. I'm thinking `cloudmysql` should be `gcpmysql` (the product name isn't too strong here). I'll make this change once the dust clears from #1175.

polinasok:
According to https://cloud.google.com/sql/docs/, it is Google Cloud SQL for MySQL and Google Cloud SQL for PostgreSQL, so if we want to keep them as separate APIs, it makes sense to go with Ross' suggestion and use gcpmysql and gcppostgresql.

I would also suggest change rds to aws for consistency.
Amazon RDS for MySQL (https://aws.amazon.com/rds/mysql/) => awsmysql
Amazon RDS for PostgreSQL (https://aws.amazon.com/rds/postgresql/) => awspostresql

zombiezen:
From meeting this morning: the decision is to make both the s/cloud/gcp/ and s/rds/aws/ changes for the SQL packages.

PR #2379, opened 2019-06-19
mysql/cloudmysql: substitute Open for URLOpener BREAKING_CHANGE_OK

zombiezen:
This package didn't previously have a URLOpener, so it adds one. Updated
guestbook injection to match new URLOpener usage.

Updates #1175

### DIFF
--- a/mysql/cloudmysql/cloudmysql.go
+++ b/mysql/cloudmysql/cloudmysql.go
@@ -13,6 +13,18 @@
 // limitations under the License.
 
 // Package cloudmysql provides connections to managed MySQL Cloud SQL instances.
+// See https://cloud.google.com/sql/docs/mysql/ for more information.
+//
+// URLs
+//
+// For mysql.Open, cloudmysql registers for the scheme "cloudmysql".
+// The default URL opener will create a connection using the default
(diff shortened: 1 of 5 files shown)


mysql/cloudmysql/cloudmysql.go
```diff
@@ -13,92 +13,130 @@
...
 )
 
-// CertSourceSet is a Wire provider set that binds a Cloud SQL proxy
-// certificate source from an GCP-authenticated HTTP client.
-//
-// Deprecated: Use cloudsql.CertSourceSet.
-var CertSourceSet = cloudsql.CertSourceSet
+// Scheme is the URL scheme cloudmysql registers its URLOpener under on
+// mysql.DefaultMux.
+const Scheme = "cloudmysql"
```
shantuo: I wonder whether we should rethink the name of this package, maybe `gcpmysql`, `googlecloudmysql`, or `gcpcloudmysql`.

zombiezen: @polinasok suggested that first in #2357; I'd rather do that in a separate PR (this already has a bunch going on). My vote is for `gcpmysql`.

Issue #1676, opened 2019-03-26
samples/guestbook: use URL openers


zombiezen:
Now that we've implemented URL openers across all of our APIs, I think we should be using the URL openers instead of a [single global "environment" flag](https://github.com/google/go-cloud/blob/a247047ee622c794cd160a1785af510f68c9436b/samples/guestbook/main.go#L61). This would make Guestbook more representative of usage that we want Go CDK users to use. Guestbook could still use Wire, but would construct URL openers through the graph.

zombiezen:
IMO, this issue encompasses two broad changes:

1. The current set of command-line flags to guestbook for a backend are split apart into the individual components. For example, the database connection takes the flags `--db_host`, `--db_user`, etc. and then builds it up into a single database connection. Now that we have URLs (and especially after fixing #1860), we want to encourage a single resource URL to connect to.
2. We probably want the sample to stop using Wire (I realize this is counter to my original comment). Especially before the default URL muxes, manually generated setup for each cloud resource was much more complex, so it wasn't feasible for a sample application to not use Wire. Guestbook's control flow will likely be much easier to understand if each services was opened with the corresponding `foo.OpenWidget` call, and blank importing all the drivers it could possibly use.

Sketching here of what the database server startup code would look like:

```go
import (
  "gocloud.dev/mysql"
  _ "gocloud.dev/mysql/azuremysql"
  _ "gocloud.dev/mysql/cloudmysql"
  _ "gocloud.dev/mysql/rdsmysql"
)

func main() {
  databaseURL := flag.String("db", "mysql://localhost/guestbook", "database URL, see https://gocloud.dev/howto/sql/")
  // ...
  flag.Parse()

  ctx := context.Background()
  db, err := mysql.Open(ctx, *databaseURL)
  // ...
}
```

PR #2380, opened 2019-06-19
mysql/rdsmysql: remove constructor in favor of URLOpener BREAKING_CHANGE_OK

zombiezen:
Update Guestbook Wire injection to match.

Updates #1175

### DIFF
--- a/mysql/rdsmysql/rdsmysql.go
+++ b/mysql/rdsmysql/rdsmysql.go
@@ -46,7 +46,7 @@ import (
 // Set is a Wire provider set that provides a *sql.DB given
 // *Params and an HTTP client.
 var Set = wire.NewSet(
-	// DO NOT SUBMIT: This is broken now.
+	wire.Struct(new(URLOpener), "CertSource"),
 	rds.CertFetcherSet,
 )
 
--- a/mysql/rdsmysql/rdsmysql.go
+++ b/mysql/rdsmysql/rdsmysql.go
@@ -46,32 +46,15 @@ import (
 // Set is a Wire provider set that provides a *sql.DB given
 // *Params and an HTTP client.
 var Set = wire.NewSet(
-	Open,
+	// DO NOT SUBMIT: This is broken now.
 	rds.CertFetcherSet,
 )
-
-// Params specifies how to connect to an RDS database.
-type Params struct {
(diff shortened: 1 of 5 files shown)


mysql/rdsmysql/rdsmysql.go
```diff
@@ -87,45 +70,36 @@ func init() {
...
-		provider: provider,
-		params:   *params,
-		sem:      make(chan struct{}, 1),
-		ready:    make(chan struct{}),
+		addr:     u.Host,
+		user:     u.User.Username(),
+		password: password,
+		dbName:   strings.TrimPrefix(u.Path, "/"),
+		// Make a copy of TraceOpts to avoid caller modifying.
+		traceOpts: append([]ocsql.TraceOption(nil), uo.TraceOpts...),
```
shantuo: Why do we need to prevent this being modified particularly? Other parts can be modified, too.

samples/guestbook/inject_aws.go
```diff
@@ -60,16 +62,19 @@ func awsBucket(ctx context.Context, cp awsclient.ConfigProvider, flags *cliFlags
...
-// awscloud.AWS use the parameters to construct a *sql.DB.
-func awsSQLParams(flags *cliFlags) *rdsmysql.Params {
-	return &rdsmysql.Params{
-		Endpoint: flags.dbHost,
-		Database: flags.dbName,
-		User:     flags.dbUser,
-		Password: flags.dbPassword,
+// openAWSDatabase is a Wire provider function that connects to RDS based on
+// the command-line flags.
+func openAWSDatabase(ctx context.Context, opener *rdsmysql.URLOpener, flags *cliFlags) (*sql.DB, func(), error) {
```
shantuo: Should this function call the portable open url function so that it looks clear since the parameter is a url string?

mysql/rdsmysql/rdsmysql.go
zombiezen: The slice contains a reference, so the backing array could be mutated before the connection is created.

Might be overly defensive programming, but Hyrum's Law or whatever.

samples/guestbook/inject_aws.go
zombiezen: Yes, but I there's a larger cleanup issue open for that: #1676

PR #2383, opened 2019-06-19
secrets: rename vault to hashivault BREAKING_CHANGE_OK

### LINKED ISSUE #2382: secrets/vault: consider package name vaultsecrets
In the spirit of #2357, cc @polinasok

https://github.com/google/go-cloud/blob/master/internal/docs/design.md#package-naming-conventions

shantuo:
Fixes #2382 

This doesn't change the URL scheme `vault://`


shantuo:
It was failing since vaultsecrets is a module, changed the check_api_change script to not fail when a top-level module is not found.

shantuo:
PTAL

internal/testing/check_api_change.sh
```diff
@@ -53,12 +52,19 @@ trap cleanup EXIT
 
 git clone -b "$UPSTREAM_BRANCH" . "$MASTER_CLONE_DIR" &> /dev/null
 
+# Copy the current branch into the temparory directory to avoid changing go.mod's
+cp -R . "$BRANCH_DIR"
```
zombiezen: I'm not following why this is now needed over the old script.

internal/testing/check_api_change.sh
```diff
@@ -35,14 +35,13 @@ echo "Checking for incompatible API changes relative to ${UPSTREAM_BRANCH}..."
 
 INSTALL_DIR="$(mktemp -d)"
 MASTER_CLONE_DIR="$(mktemp -d)"
+BRANCH_DIR="$(mktemp -d)"
```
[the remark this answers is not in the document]
shantuo: This change is to fix a bug that cause the Travis build failure for this PR. The bug is that we were using the incoming changes' module list to check changes, where we should use the upstream's one.

vangent:
I think we decided today to rename this to `hashivault` instead of `vaultsecrets`, including the URL scheme.

internal/testing/check_api_change.sh
shantuo: Turns out I might be doing something wrong. Removed, PTAL.

internal/website/content/secrets/vault/_index.md
```diff
@@ -1,4 +0,0 @@
----
```
zombiezen: Don't delete this file: this could break `go get`s that operate on the current released version.

internal/testing/check_api_change.sh
```diff
@@ -53,12 +51,16 @@ trap cleanup EXIT
 
 git clone -b "$UPSTREAM_BRANCH" . "$MASTER_CLONE_DIR" &> /dev/null
 
+# Run the following checks in the master directory
+ORIG_DIR=$(pwd)
```
zombiezen: Surprised shellcheck didn't catch this:

```suggestion
ORIG_DIR="$(pwd)"
```

internal/website/content/secrets/vault/_index.md
shantuo: Done.

shantuo:
PTAL

polinasok:
Thanks,  @shantuo.

Issue #2365, opened 2019-06-17
internal/website: add how-to guide for runtimevar


zombiezen:
Follow-on from #1576.

The `runtimevar` package is missing a how-to guide. The guide should walk the user through:

- Opening a runtime variable on various providers
- Encoding/decoding concerns
- Watching for changes synchronously
- Fetching the latest value from a variable

As per @vangent's request, this is marked as P1/beta-blocking.

zombiezen:
Also: it's important for all guides, but especially for `runtimevar`, to explain what the purpose of the package is (i.e. reading dynamic configuration data). How-to guides should be framed in terms of solving a user's problem, not in terms of the solution we have.

eliben:
Fixed by #2384

PR #2384, opened 2019-06-19
website: add how-to guide for runtimevar

eliben:
For #2365


internal/website/content/howto/runtimervar/runtimevar.md
```diff
@@ -0,0 +1,108 @@
...
+Working with runtime variables using the Go CDK takes two steps:
+
+1. Open the variable with the `runtimevar` provider of your choice.
+2. As many times as needed, use the `Latest` or `Watch` methods to fetch the
+   value of the variable.
+
+## Constructors versus URL openers
+
+The easiest way to open a variable is using [`runtimevar.OpenVariable`][] and a
+URL pointing to the variable. See the [documentation on URLs][] for more
```
cnguyen83:
```suggestion
URL pointing to the variable. See [Concepts: URLs][] for more
```

internal/website/content/howto/runtimervar/runtimevar.md
```diff
@@ -0,0 +1,108 @@
...
+details. If you need fine-grained control over the connection settings, you can
+call the constructor function in the driver package directly (like
+`awsparamstore.OpenVariable`). This guide will show how to use both forms for
+each storage provider.
+
+[documentation on URLs]: {{< ref "/concepts/urls.md" >}}
```
cnguyen83:
```suggestion
[Concepts: URLs]: {{< ref "/concepts/urls.md" >}}
```

internal/website/content/howto/runtimervar/runtimevar.md
```diff
@@ -0,0 +1,108 @@
...
+2. As many times as needed, use the `Latest` or `Watch` methods to fetch the
+   value of the variable.
+
+## Constructors versus URL openers
+
+The easiest way to open a variable is using [`runtimevar.OpenVariable`][] and a
+URL pointing to the variable. See the [documentation on URLs][] for more
+details. If you need fine-grained control over the connection settings, you can
+call the constructor function in the driver package directly (like
+`awsparamstore.OpenVariable`). This guide will show how to use both forms for
```
cnguyen83:
```suggestion
`awsparamstore.OpenVariable`). This guide shows how to use both forms for
```

internal/website/content/howto/runtimervar/runtimevar.md
```diff
@@ -0,0 +1,108 @@
...
+each storage provider.
+
+[documentation on URLs]: {{< ref "/concepts/urls.md" >}}
+
+[`runtimevar.OpenVariable`]:
+https://godoc.org/gocloud.dev/runtimevar#OpenVariable
+
+## GCP Runtime Configurator {#rc-url}
```
zombiezen: Optional: Consider using GCP/AWS/Azure for these fragments, since these acronyms aren't well-accepted.

```suggestion
## GCP Runtime Configurator {#gcp-url}
```

internal/website/content/howto/runtimervar/runtimevar.md
```diff
@@ -0,0 +1,108 @@
+---
```
zombiezen: This could all be placed into _index.md.

internal/website/content/howto/runtimervar/_index.md
```diff
@@ -0,0 +1,10 @@
+---
+title: "Runtime Variables"
```
zombiezen:
```suggestion
title: "Runtime Configuration"
```

zombiezen:
LGTM from technical side.

internal/website/content/howto/runtimervar/runtimevar.md
eliben: Good point. I was thinking we could add more stuff later, but maybe YAGNI

eliben:
Thank you very much for the review and comments, Christina and Ross!

I applied the suggestions and based on Ross's advice moved the content to `_index.md`.

@cnguyen83 please Approve the PR once it LGTY

internal/website/content/howto/runtimervar/_index.md
```diff
@@ -0,0 +1,108 @@
+---
+title: "Runtime Configuration"
+date: 2019-06-18T14:51:12-07:00
+showInSidenav: true
+---
+
+Working with runtime variables using the Go CDK takes two steps:
+
+1. Open the variable with the `runtimevar` provider of your choice.
+2. As many times as needed, use the `Latest` or `Watch` methods to fetch the
```
vangent: This is not quite right for `Watch`. You would either use `Latest` as many times as needed, OR you would use `Watch`, probably once in a loop.

I would consider just focusing on `Latest` here, and somewhere down below mention that instead of `Latest`, you can use `Watch` to get edge-based notifications.

internal/website/content/howto/runtimervar/_index.md
```diff
@@ -0,0 +1,108 @@
...
+   value of the variable.
+
+## Constructors versus URL openers
+
+The easiest way to open a variable is using [`runtimevar.OpenVariable`][] and a
+URL pointing to the variable. See [Concepts: URLs][] for more
+details. If you need fine-grained control over the connection settings, you can
+call the constructor function in the driver package directly (like
+`awsparamstore.OpenVariable`). This guide shows how to use both forms for
+each storage provider.
```
vangent: Not storage here.

internal/website/content/howto/runtimervar/_index.md
```diff
@@ -0,0 +1,108 @@
...
+[Concepts: URLs]: {{< ref "/concepts/urls.md" >}}
+
+[`runtimevar.OpenVariable`]:
+https://godoc.org/gocloud.dev/runtimevar#OpenVariable
+
+## GCP Runtime Configurator {#gcp-url}
+
+To open a variable stored in [GCP Runtime Configurator][] via a URL, you can use
+the `runtimevar.OpenVariable` function as follows. It uses the `string`
+[decoder][].
```
vangent: Decoders are not introduced yet, I would add a section on them above.

internal/website/content/howto/runtimervar/_index.md
```diff
@@ -0,0 +1,108 @@
...
+
+[`awsparamstore.OpenVariable`]:
+https://godoc.org/gocloud.dev/runtimevar/awsparamstore#OpenVariable
+
+## etcd {#etcd-url}
+
+The Go CDK supports using [etcd][] for storing variables locally or
+remotely. To open a variable stored in `etcd` via a URL, you can use the
+`runtimevar.OpenVariable` function as follows.
+
```
vangent: Mention the required environment variable here? Or is that obvious from the example (I can't easily see the example during the review).

internal/website/content/howto/runtimervar/_index.md
```diff
@@ -0,0 +1,108 @@
...
+[`etcdvar.OpenVariable`]:
+https://godoc.org/gocloud.dev/runtimevar/etcdvar#OpenVariable
+
+{{< goexample "gocloud.dev/runtimevar/etcdvar.Example_openVariableHowto" >}}
+
+## Fetching the latest value {#latest}
+
+Once we have an open variable, we can use the [`Variable.Latest`][] method to
+fetch its latest value. This method returns the latest good [`Snapshot`][] of
+the variable value, blocking if no good value has ever been received. To avoid
```
vangent: I would emphasize this even more, that it only blocks if no good value has *ever* been received. In particular, it will not block if there's a good value and then there's a bad value (it will keep returning the good value).

Users might want to call `Snapshot` once during startup, or have `healthz` only succeed, based on getting a first good value. In fact, `Variable` implements `health.Checker` to make that easy if they happen to be using `server`.

internal/website/content/howto/runtimervar/_index.md
```diff
@@ -0,0 +1,108 @@
...
+blocking, you can pass an already-`Done` context.
+
+{{< goexample src="gocloud.dev/runtimevar.Example_latestStringVariableHowto"
+imports="0" >}}
+
+Type [`Variable`][] also has a [`Watch`][] method for obtaining the value of
```
vangent: Good. I would put this under a new heading, and just drop the reference at the top of the page.

runtimevar/gcpruntimeconfig/example_test.go
```diff
@@ -76,11 +76,59 @@ func ExampleOpenVariable() {
 	_ = cfg
 }
 
+func Example_openVariableHowto() {
```
vangent: Why not just fix the existing example? I'd rather not have "Howto"-specific examples.

runtimevar/example_test.go
```diff
@@ -77,6 +77,20 @@ func ExampleVariable_Latest_stringVariable() {
 	// "hello world"
 }
 
+func Example_latestStringVariableHowto() {
```
vangent: Why the `Howto` suffix?

runtimevar/etcdvar/example_test.go
```diff
@@ -57,16 +57,45 @@ func ExampleOpenVariable() {
 	_ = cfg
 }
 
+func Example_openVariableHowto() {
```
vangent: Same as below, why not fix the existing example?

runtimevar/awsparamstore/example_test.go
```diff
@@ -58,14 +58,44 @@ func ExampleOpenVariable() {
 	_ = cfg
 }
 
+func Example_openVariableHowto() {
```
vangent: Same as below.

internal/website/content/howto/runtimervar/_index.md
eliben: Removed

internal/website/content/howto/runtimervar/_index.md
eliben: Good catch, done

internal/website/content/howto/runtimervar/_index.md
eliben: Right, so I linked "decoder" to the doc. I think it's sufficient at this point, since the reference documentation (linked) describes decoders fairly well

internal/website/content/howto/runtimervar/_index.md
eliben: Yes, it's a comment in the example that describes the env var

internal/website/content/howto/runtimervar/_index.md
eliben: I added emphasis to "ever", I think the rest of the first paragraph is implied already. I suggest we leave the server/health reference out for now - until we have a howto for server. Then we can see where's the best place to describe this

internal/website/content/howto/runtimervar/_index.md
eliben: Dropped from the top.

Why new heading? How would you call it? This section is pretty short as is

runtimevar/example_test.go
eliben: The previous example uses `constantvar` and has an actual output expectation. I don't want these to show up. The pattern of having 

```
var v *PortableType
```

Is already used elsewhere in examples, since we're only compiling them.

runtimevar/gcpruntimeconfig/example_test.go
eliben: In this case, because the existing example is doing too much. It's also calling `Latest`, which I don't want to show here.

Perhaps we could enhance the example gathering script to snip off everything from some point on? We kinda have it with `// Ignore unused variables in example` now, but this is confusing when people are just looking at the example.

In general, using exactly a single source for examples and howtos is challenging because users reach this code in two ways - (1) reading the howto and (2) looking at the example code in godoc.org

I could perhaps rename this, if having "howto" in it doesn't seem good?

internal/website/content/howto/runtimervar/_index.md
vangent: IMHO it would be good to have a section on Decoders in here; this introduction with just a link to godoc is too abrupt. We should explain that you can tell `runtimevar` to interpret your config as JSON, a string, or `[]byte`, and that that determines the resulting type in the `Snapshot`. It's not obvious.

internal/website/content/howto/runtimervar/_index.md
vangent: `Watch`? Up to you.

runtimevar/gcpruntimeconfig/example_test.go
vangent: Let's just drop the `Latest` from the existing example then, and maybe replace it with a comment like `// ... use the runtimevar.Variable`, or nothing.

runtimevar/example_test.go
vangent: What happens if you do the "variables set up elsewhere" above the `constantvar.NewBytes`?

internal/website/content/howto/runtimervar/_index.md
eliben: Beefed it up a bit, without a separate section. Should be much less abrupt now, while still linking to the reference doc now.

The reason I'm objecting to a separate section is that it (1) breaks the flow of just showing a simple end-to-end example, and even (2) if placed in the end, would repeat a lot of what the linked reference describes. I think the current text is a reasonable compromise. The user gets a quick idea of what decoders mean and a link for more info if needed.

internal/website/content/howto/runtimervar/_index.md
eliben: Done

eliben:
Robert - PTAL. Moved things around as discussed

internal/website/content/howto/runtimervar/_index.md
```diff
@@ -0,0 +1,110 @@
...
+title: "Runtime Configuration"
+date: 2019-06-18T14:51:12-07:00
+showInSidenav: true
+---
+
+Working with runtime variables using the Go CDK takes two steps:
+
+1. Open the variable with the `runtimevar` provider of your choice. You can
+   provide a [decoder][] parameter to specify how the value stored in the
+   variable is interpreted.
```
vangent: Consider: "...to specify whether the raw value stored in the variable is interpreted as a `string`, a `[]byte`, or as JSON.

internal/website/content/howto/runtimervar/_index.md
```diff
@@ -0,0 +1,110 @@
...
+2. Use the `Latest` method to fetch the value of the variable.
```
vangent: Consider "value" -> "interpreted value"?

internal/website/content/howto/runtimervar/_index.md
```diff
@@ -0,0 +1,110 @@
...
+[`etcdvar.OpenVariable`]:
+https://godoc.org/gocloud.dev/runtimevar/etcdvar#OpenVariable
+
+{{< goexample "gocloud.dev/runtimevar/etcdvar.ExampleOpenVariable" >}}
+
+## Fetching the latest value {#latest}
+
+Once we have an open variable, we can use the [`Variable.Latest`][] method to
+fetch its latest value. This method returns the latest good [`Snapshot`][] of
+the variable value, blocking if no good value has *ever* been received. To avoid
```
vangent: Consider adding: "The dynamic type of `Snapshot.Value` depends on the `Decoder` you provided when creating the `Variable`."

internal/website/content/howto/runtimervar/_index.md
eliben: Done

internal/website/content/howto/runtimervar/_index.md
eliben: I prefer the existing text to keep it simple. The previous sentence already describes "interpreted" so no reason to repeat it here, IMO

internal/website/content/howto/runtimervar/_index.md
eliben: Done

PR #2385, opened 2019-06-20
internal/testing/cmdtest: package for CLI testing

jba:
This package generalizes the gocdk CLI tests
(internal/cmd/gocdk/main_test.go) for use with any command-line
program.

We will use it to test our samples, and it will replace the gocdk test
framework. All that work is done and will be mailed when this PR is
submitted.

All the features here will be justified in that later work. Briefly:

- All the builtin commands are used in sample tests, and in testing
  this package itself.

- Input redirection is needed for sample sample CLI commands that
  take input from stdin.

- Variable substitution is needed in some samples to refer to
  the root directory (temporary test working directory).

- The Setup option is used by gocdk to set GOPATH.


zombiezen:
Thanks for writing this! Just a heads-up: this is a larger block of new code and I want to make sure I review it properly. Please give me a couple days so I can ensure quality review.

jba:
@zombiezen Thanks for setting expectations! I fully expect that it will take a couple of days; no rush. Also I'm sure @vangent will want to take a pass.

internal/testing/cmdtest/cmdtest_test.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Fatal(err)
+	}
+	// Set PATH to the current directory so test files can find echo-stdin.
+	os.Setenv("PATH", cwd)
+
+	ret := m.Run()
+	_ = os.Remove(filepath.Join(cwd, "echo-stdin"))
```
zombiezen: On Windows, this path will be `echo-stdin.exe`.

internal/testing/cmdtest/cmdtest.go
```diff
@@ -0,0 +1,616 @@
...
+}
+
+var varRegexp = regexp.MustCompile(`\$\{([^${}]+)\}`)
+
+// expandVariables replaces variable references in s with their values. A reference
+// to a variable V looks like "${V}".
+// lookup is called on a variable's name to find its value. Its second return value
+// is false if the variable doesn't exist.
+// expandVariables fails if s contains a reference to a non-existent variable.
+func expandVariables(s string, lookup func(string) (string, bool)) (string, error) {
```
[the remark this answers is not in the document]
jba: Clarified in the comment.

internal/testing/cmdtest/cmdtest.go
```diff
@@ -0,0 +1,616 @@
...
+		if _, err := fmt.Fprintf(w, "$ %s\n", c); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func writeLines(w io.Writer, lines []string) error {
+	for _, l := range lines {
+		if _, err := fmt.Fprintf(w, "%s\n", l); err != nil {
```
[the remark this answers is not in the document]
jba: Done, but it's not pretty. Is there a nicer way?

internal/testing/cmdtest/cmdtest.go
```diff
@@ -0,0 +1,616 @@
...
+// set environment variable
+func setenvCmd(args []string) ([]byte, error) {
+	if len(args) != 2 {
+		return nil, fatal{errors.New("need exactly 2 arguments")}
+	}
+	return nil, os.Setenv(args[0], args[1])
+}
+
+func checkPath(path string) error {
+	if strings.ContainsRune(path, '/') {
```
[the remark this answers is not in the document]
jba: Actually, I think we should disallow both separators. Otherwise this check could pass on Linux and fail on Windows, or vice versa.

jba:
@vangent, since @zombiezen is out next week, can you take over this review?

internal/testing/cmdtest/cmdtest.go
```diff
@@ -0,0 +1,610 @@
...
+//   echof FILE ARG1 ARG2 ...
+//
+// These all have their usual Unix shell meaning, except for echof, which writes its
+// arguments to a file (output redirection is not supported). All file and directory
+// arguments must refer to the current directory; that is, they cannot contain
+// slashes.
+//
+// If the first word of a command line is not one of the above built-ins, or any
+// additional commands added to the TestFile.Commands map before execution, then
+// cmdtest runs the command using the exec package. Recall that the exec package
```
vangent: Do we have a use case for this (exec'ing arbitrary executables)?

I started out with this in the `gocdk` tests but ended up removing it because I didn't need it, and because it's difficult to make portable.

internal/testing/cmdtest/cmdtest.go
```diff
@@ -0,0 +1,610 @@
...
+// and can also update a file with new "golden" output that is deemed correct.
+//
+// Start using cmdtest by writing a test file with commands and expected output.
+// See the TestFile documentation for the syntax.
+//
+// To test, first read the file:
+//
+//    tf, err := cmdtest.Read("test.ct")
+//
+// Then configure the resulting TestFile by adding commands or enabling
```
vangent: I didn't see any example of this (doing anything with `TestFile` other than `Compare` or `Update`). Is there a use case?

Related: would it be simpler to have `cmdtest` define the `--record` flag, and just have the interface to be something more like "run the tests in this file", and let it manage reading/writing the expected output based on `--record`?

jba:
> If you have it waiting, can you send the PR that updates `internal/cmd/gocdk`'s tests to use this? It will help make it more concrete. Are there any other places that we'd update?

I can't make a dependent PR, but you can see the changes at https://github.com/jba/go-cloud/blob/samples-test/internal/cmd/gocdk/main_test.go.

No other updates, but new tests for all the CLI-based samples. You can find those on the same branch as above.

internal/testing/cmdtest/cmdtest.go
jba: Each of the sample CLIs is an executable, and I can make them available just by setting PATH. However, I could also make it easy to install a custom command that invokes that particular binary.

I tested on windows (by pushing a commit that hacked .travis.yml to run windows on PRs) and I _thought_ it just worked, but maybe I messed up. Do you have to change the binary name to end in ".exe" or does the exec package handle that for you? What else is non-portable?

internal/testing/cmdtest/cmdtest.go
jba: See the branch in my fork for examples.

It's bad practice in Go for packages to define flags. A package should have a purely programmatic API. (Also, `-record` is the wrong name; it should be `-update` or something like that, IMO.)

jba:
PTAL (No code changes, but I've responded to all your comments.)

internal/testing/cmdtest/cmdtest.go
vangent: Yes, you have to change the binary name to end in `.exe`.

If you're using this primarily as a way to run the executable under test, I'd suggest requiring the test to install it in the `Commands` map; maybe a helper for doing so would make it easy. The non-portable thing you've exposed here is that I can easily write tests that call non-portable things like `ls`.

internal/testing/cmdtest/cmdtest.go
```diff
@@ -0,0 +1,610 @@
...
+// The cases of a test file are executed in order, starting in a freshly
+// created temporary directory.
+//
+// The built-in commands (initial contents of the Commands map) are:
+//
+//   cd DIR
+//   cat FILE
+//   mkdir DIR
+//   setenv VAR VALUE
+//   echo ARG1 ARG2 ...
```
vangent: Why not add `ls`? It seems like it might be common, and it somewhat subtle due to platform-specific restrictions.

internal/testing/cmdtest/cmdtest.go
vangent: (and then remove the ability to `exec` arbitrary executables; everything should have to be built-in).

internal/testing/cmdtest/cmdtest.go
vangent: Maybe you have a better sense of this than I do at this point, but it still feels like you're requiring the caller to do more than necessary (i.e., there will be a lot of repeated code in callers).

Can't use of this package look something like this?

```
record := false // usually would be defined as a flag
tf := cmdtest.New(testDataDir)
tf.Setup := func... // optional
tf.AddBinary(myBinaryName, pathToMyBinaryExecutable) // optional
tf.AddCommand("foo", func() { whatever }) // optional
if err := tf.Run(record); err != nil {
  t.Fatal(err)
}
```

internal/testing/cmdtest/cmdtest.go
jba: Omitting it for now, as per our discussion.

jba:
PTAL.

internal/testing/cmdtest/cmdtest.go
jba: > Yes, you have to change the binary name to end in `.exe`.

You don't have to do this. I logged into a Windows GCP machine and ran the following program (and I deserve a medal for it):
```
package main

import "fmt"
import "os/exec"

func main() {
	out, err := exec.Command("go", "version").CombinedOutput()
	fmt.Println(err)
	if err == nil {
		fmt.Println(string(out))
	}
}
```
Notice that it doesn't say `go.exe`. It still worked. I also tried the absolute path `c:\go\bin\go` without the `.exe`. That also worked.

internal/testing/cmdtest/cmdtest.go
jba: Added `Run`. Left `Compare` and `Update`; should I unexport them?

vangent: IIUC, you did about half of what I was suggesting -- you combined `Compare`/`Update` into `Run`, but you didn't respond to the part about changing `cmdtest` to operate on a directory. Note that in my pseudocode, you provide a directory, configuration (once), and then `Run` does the work of finding tests, `Read`ing them, etc.

This seems noticeably easier to me than having every client do the work of iterating over all the tests. It also feels weird to me to have to re-configure each `TestFile` (e.g., to add `Command`s).

Is there a reason to keep the interface as a single test rather than as a test suite like that?

jba: Sorry, I missed all that. Makes sense.

Done. I kept `Commands` as an exported map, and I use `Read` instead of `New` (because I think of `New` as an in-memory thing), but otherwise it matches your snippet.

PTAL

jba:
Awaiting @clausti's approval.

jba:
I also need @zombiezen's approval.

internal/testing/cmdtest/cmdtest.go
```diff
@@ -0,0 +1,616 @@
...
+	if err != nil {
+		return fmt.Errorf("%s: %v", tf.filename, err)
+	}
+	if tf.KeepRootDir {
+		fmt.Printf("test root directory: %s\n", rootDir)
+	} else {
+		defer os.RemoveAll(rootDir)
+	}
+
+	if err := os.Setenv("ROOTDIR", rootDir); err != nil {
```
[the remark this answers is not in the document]
zombiezen: Thanks for the explanation. This is fine.

For the CLI, we ended up passing around everything explicitly, but that's probably a non-starter for general infrastructure.

PR #2393, opened 2019-06-21
dynamodocstore: provide a general query callback for in-memory sorting

jba:
Following up on a suggestion of @vangent to provide an easy way
to construct a fallback function for doing in-memory sorting.

Also, split out url code to a separate file.


docstore/driver/document.go
```diff
@@ -150,6 +150,21 @@ func (d Document) SetField(field string, value interface{}) error {
 	return nil
 }
 
+// FieldNames returns names of the top-level fields of d.
+func (d Document) FieldNames() []string {
```
shantuo: Can you just use the fieldsCache?

docstore/dynamodocstore/query.go
```diff
@@ -547,3 +553,190 @@ func (c *collection) runActionQuery(ctx context.Context, q *driver.Query, mods [
...
+// InMemorySortFallback returns a query fallback function for Options.RunQueryFallback.
+// The function accepts a query with an OrderBy clause. It runs the query without that clause,
+// reading all documents into memory, then sorts the documents according to the OrderBy clause.
+//
+// Only string, numeric, time and binary ([]byte) fields can be sorted.
+//
+// createDocument should create an empty document to be passed to DocumentIterator.Next.
+// The DocumentIterator returned by the FallbackFunc will also expect the same type of document.
+// If nil, then a map[string]interface{} will be used.
+func InMemorySortFallback(createDocument func() interface{}) FallbackFunc {
```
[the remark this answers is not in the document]
jba: I don't think it's needed by other providers. Both mongo and firestore can sort a table scan.

I'm not sure what you mean by having the user provide the sort type. We want the ordering to be the same that DynamoDB itself would use, for consistency. For example, 
```
Query().OrderyBy("field")
```
should produce documents in the same order as
```
Query().Where("field", ">", 3).OrderBy("field")
```
even though the second would use an index.

docstore/driver/document.go
jba: Can you sketch out what the code would be? I'm not sure I understand.

shantuo: nvm, I didn't see `Document.fields` contains only struct fields, I thought it was both.

docstore/dynamodocstore/query.go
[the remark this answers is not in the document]
shantuo: I see, yeah let's leave it here.

PR #2394, opened 2019-06-21
CONTRIBUTING: add acceptable licenses

zombiezen:
Also add some supporting references and wording about global state
that I had been meaning to add for a while.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -102,8 +102,9 @@ categorized under [`notice`][notice license],
 [`permissive`][permissive license], or [`unencumbered`][unencumbered license]
 in the [Google Open Source Licenses][] documentation. This is enforced with a
 Travis build check that verifies that every module depended on is in the
-[`alldeps` file][]. Do not add new lines to the file unless you have verified
-that the dependency is released under an acceptable license.
+[`alldeps` file][]. Do not add new direct or indirect dependencies to the Go CDK
+unless you have verified that the dependency is released under an acceptable
+license.
 
 [`alldeps` file]: https://github.com/google/go-cloud/blob/master/internal/testing/alldeps
 [notice license]: https://opensource.google.com/docs/thirdparty/licenses/#notice
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,6 +95,22 @@ cloud resources. Replay files scrub sensitive information.
 [Send your PR](#making-a-pull-request) without the replay files, and we can
 generate new ones to be used by others.
 
+### Dependencies
+
+The Go CDK has a policy to only depend on code that is under a license that is
+categorized under [`notice`][notice license],
+[`permissive`][permissive license], or [`unencumbered`][unencumbered license]
+in the [Google Open Source Licenses][] documentation. This is enforced with a
+Travis build check that verifies that every module depended on is in the
+[`alldeps` file][]. Do not add new lines to the file unless you have verified
+that the dependency is released under an acceptable license.
+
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -95,6 +95,22 @@ cloud resources. Replay files scrub sensitive information.
...
 generate new ones to be used by others.
 
+### Dependencies
+
+The Go CDK has a policy to only depend on code that is under a license that is
+categorized under [`notice`][notice license],
+[`permissive`][permissive license], or [`unencumbered`][unencumbered license]
+in the [Google Open Source Licenses][] documentation. This is enforced with a
+Travis build check that verifies that every module depended on is in the
+[`alldeps` file][]. Do not add new lines to the file unless you have verified
```
eliben: "Do not add new lines..." is a little confusing since we don't directly add lines to this file. Maybe something like "Do not add new (direct or indirect) dependencies unless ..."?

internal/docs/design.md
```diff
@@ -108,16 +108,19 @@ https://godoc.org/github.com/google/go-cloud/blob#Bucket.NewWriter
 ## Minimize Global State
 
 As a library, the Go CDK should not introduce global state. Global state is
-difficult to reason about in large codebases, where it can be necessary for
-different parts of the application to use different states. Instead of adding
-global state, push responsibility to the application to inject the state where
-it is needed.
+[difficult to reason about in large codebases][Singletons Considered Harmful],
```
eliben: Do we really want to link to an OOP-oriented article here? I personally find the existing text sufficient. If anything, I'd make it even shorter

CONTRIBUTING.md
zombiezen: Done.

internal/docs/design.md
zombiezen: This was a to-do I had from a while back: a few folks on the team found the cited source helpful. Wire vs. non-Wire is a principle that we haven't written down anywhere else, and this seemed like the best place to put it.

polinasok: +1 since Go is not strictly object-oriented (and the first sentence of the article is "Singletons are possibly the single greatest obstacle to good object-oriented design today"). Also, wouldn't you want to discourage other types of global state (even just plain simple global variables)? A go-specific or more general anti-global-state article would be better here if we could find one...

CONTRIBUTING.md
```diff
@@ -95,6 +95,22 @@ cloud resources. Replay files scrub sensitive information.
 [Send your PR](#making-a-pull-request) without the replay files, and we can
 generate new ones to be used by others.
 
+### Dependencies
+
+The Go CDK has a policy to only depend on code that is under a license that is
```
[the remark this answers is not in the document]
zombiezen: Done.

CONTRIBUTING.md
```diff
@@ -95,6 +95,22 @@ cloud resources. Replay files scrub sensitive information.
...
+categorized under [`notice`][notice license],
+[`permissive`][permissive license], or [`unencumbered`][unencumbered license]
+in the [Google Open Source Licenses][] documentation. This is enforced with a
+Travis build check that verifies that every module depended on is in the
```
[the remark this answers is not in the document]
zombiezen: Done.

internal/docs/design.md
zombiezen: Reverted.

eliben:
> I've reverted the design.md change by popular demand.

Ross, up to you, but FTR my comment was only w.r.t. the link to OOP documentation. The added part mentioning wire seemed good to me

zombiezen:
Added "Wire vs. non-Wire" verbage back. Merging.

PR #2398, opened 2019-06-24
pubsub/azuresb: group ack/nack RPCs by partition

### LINKED ISSUE #2362: pubsub/azuresb: batching lock tokens for different partitions is not supported for an entity
### Describe the bug

LockIDs are not grouped by partitionID when completing a batch of messages.

### To Reproduce

The service bus must be set up with partitioning enabled...

In a loop
Call `Receive` on an azuresb implementation of the `*pubsub.Subscription` object.
Call `Ack` on the message object returned from `Receive`

## Expected behavior

We expect no issues, however in some cases messages come down from different partitions and it seems that when batching the `Ack` calls for messages from different partitions the below error gets returned.

```
pubsub (code=Unknown): unhandled error link {{id}}: status code 400 and description: Batching lock tokens for different partitions is not supported for an entity
``` 

The partition key is available as part of the underlying [message model](https://github.com/Azure/azure-service-bus-go/blob/master/message.go#L74)

Would it be possible to modify [`driver.Message`](https://github.com/google/go-cloud/blob/master/pubsub/driver/driver.go#L38) to have a hook to sort on these things?

It may be easier modify the abstraction [here](https://github.com/google/go-cloud/blob/master/pubsub/azuresb/azuresb.go#L512) to have a mechanism to catch when it is illegal for the service bus to `Ack` these since it will have the partition keys [here](https://github.com/google/go-cloud/blob/master/pubsub/azuresb/azuresb.go#L446)

### Version

0.15.0

vangent:
Fixes #2362.

Instead of using the LockToken itself as the AckID, we now use a struct with the LockToken and the partitionID. When handling a batch of AckIDs for ack/nack, we group by the partitionID.


pubsub/azuresb/azuresb.go
```diff
@@ -506,16 +525,32 @@ func (s *subscription) updateMessageDispositions(ctx context.Context, ids []driv
...
+			lockTokenBytes := [16]byte(*pid.LockToken)
+			partitions[pid.PartitionID] = append(partitions[pid.PartitionID], amqp.UUID(lockTokenBytes))
 		}
 	}
+
+	// Update all partitions in parallel.
+	g, ctx := errgroup.WithContext(ctx)
+	for _, lockTokens := range partitions {
+		g.Go(func() error {
+			return s.updateMessageDispositionsInPartition(ctx, lockTokens, disposition)
```
zombiezen: This will have the behavior that one RPC failing will cancel all the other acks.. is this desirable behavior?

pubsub/azuresb/azuresb.go
```diff
@@ -506,16 +525,32 @@ func (s *subscription) updateMessageDispositions(ctx context.Context, ids []driv
...
+		if pid, ok := ackID.(*partitionAckID); ok {
+			lockTokenBytes := [16]byte(*pid.LockToken)
+			partitions[pid.PartitionID] = append(partitions[pid.PartitionID], amqp.UUID(lockTokenBytes))
 		}
 	}
+
+	// Update all partitions in parallel.
+	g, ctx := errgroup.WithContext(ctx)
+	for _, lockTokens := range partitions {
+		g.Go(func() error {
```
zombiezen: Do we ever expect the number of partitions to be large? The parallelism here is unbounded.

(Also, does this end up being an issue with HTTP replay due to nondeterminism?)

pubsub/azuresb/azuresb.go
vangent: No, I guess not; we might as well ack as many as possible. Fixed by no longer using `errgroup`.

pubsub/azuresb/azuresb.go
vangent: I doubt it, but better safe than sorry. Fixed to max out at 5 goroutines.

There's no HTTP replay for `azuresb` :-(.

vangent:
PTAL

pubsub/azuresb/azuresb.go
```diff
@@ -506,16 +524,44 @@ func (s *subscription) updateMessageDispositions(ctx context.Context, ids []driv
...
+	var mu sync.Mutex
+	var lastErr error
+	for _, lockTokens := range partitions {
+		sem <- struct{}{}
+		go func() {
+			defer func() { <-sem }() // Release the semaphore.
+			if err := s.updateMessageDispositionsInPartition(ctx, lockTokens, disposition); err != nil {
+				mu.Lock()
+				defer mu.Unlock()
+				lastErr = err
```
jakebjorkekato: Would it be possible to maintain the errors if multiple occur?  

Combine the strings maybe?

My only concern is multiple messages acks fail, and not seeing which ones/or their errors in the logs may be an issue.

vangent: Sure; done. Although:

* (nit) It's not multiple acks, it's multiple batches of acks. So even if you get an error, it's not obvious how many acks actually failed (some of the batch may have succeeded). In our benchmark, we regularly see batches of > 100 acks going out at once.

* You'll get this error back from `Receive` or `Shutdown`, not from `Message.Ack`.

vangent:
Note that it is hard to test this. I tried enabling partitioning on the topic for our benchmark test, but all of the messages still came from the same partition.

pubsub/azuresb/azuresb.go
jakebjorkekato: Thank you.

Correct, I misstated that the failure would be from multiple batchs of ack messages.

Yes, since it happens async the next call to `Receive` (in our use case from #2362 ) we would see this error message.  I do like having all of the errors that occurred since the service bus usually has detailed error strings.

vangent:
@vsaroopchand 	please review when you have a chance and send a PR or file issue if you see any problems.

Issue #2305, opened 2019-06-11
gocdk: fill in TODOs for comments explaining demos


vangent:
The Go code for demos have a TODO at the top of each file to provide comments explaining the code. Remove the TODOs and add comments.

```
TODO(rvangent): This file is user-visible, add many comments explaining
```

vangent:
Done!

PR #2402, opened 2019-06-25
internal/gocdk: in demos, don't wait forever

### LINKED ISSUE #2358: internal/gocdk: demos should have a deadline on the context used to initially connect

vangent:
Fixes #2358.

This will wait 5 seconds at startup, including getting a good `Snapshot` for `runtimevar`.

We could consider trying to wire these into the healthcheckers for the server, but I think that's too complicated for what we're trying to show here; probably it should be explained in comments instead (as part of #2305).


zombiezen:
> We could consider trying to wire these into the healthcheckers for the server, but I think that's too complicated for what we're trying to show here; probably it should be explained in comments instead (as part of #2305).

Wiring into health checks would make this behave much more nicely with the rest of the CDK tooling, especially since if the server takes too long to start up the process will be killed.

Issue #2301, opened 2019-06-11
internal/cmd/gocdk: add AWS ECS launcher


zombiezen:
Based on past research, AWS's [Elastic Container Service](https://aws.amazon.com/ecs/) appears to be the rough equivalent of Cloud Run, although it seems there is an explicit choice about whether to have the instance backed by EC2 or [Fargate](https://aws.amazon.com/fargate/). I do not know enough about the choice to make an informed suggestion or design.

It pulls Docker images from [Elastic Container Registry](https://aws.amazon.com/ecr/), which can be provisioned with an [`aws_ecr_repository`](https://www.terraform.io/docs/providers/aws/r/ecr_repository.html) Terraform resource. There might be some auth details to iron out, as [each repository requires a separate login](https://docs.aws.amazon.com/AmazonECR/latest/userguide/repository-create.html). The [GetAuthorizationToken ECR API](https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_GetAuthorizationToken.html) may be of use.

Links:
- [ECS Developer Guide](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/Welcome.html)
- [ECS API (Control Plane) Reference](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/Welcome.html)
- [ECR Guide](https://docs.aws.amazon.com/AmazonECR/latest/userguide/what-is-ecr.html)

zombiezen:
Update: I have the control plane aspect of this working on my local machine but with no prompts on `biome add` yet. Discoveries:

- The biome will need to create an ECR repository, an ECS cluster, a VPC, and an EC2 instance template.
- Fargate does not appear to have a free tier, so EC2 is likely the better option for our users.
- It is unclear to me from the [documentation](https://www.terraform.io/docs/providers/aws/r/ecs_cluster.html) how to provision an ECS cluster with EC2 instances with Terraform. It is also unclear how to provision an ECS cluster with Fargate.

Overall, this was much smoother to implement than I expected.

zombiezen:
Once #2416 is in, the remaining work is to:

- Add the template and prompting for an ECS biome to `gocdk biome add`
- Add logic to the launcher to authenticate Docker to the ECR repository the user chose

zombiezen:
Found a blog post that seems to describe what's required for `gocdk biome add`: http://blog.shippable.com/setup-a-container-cluster-on-aws-with-terraform-part-2-provision-a-cluster

[...]

PR #2416, opened 2019-06-26
internal/cmd/gocdk: add ECS launcher

zombiezen:
This adds "ecs" as a launcher type recognized by `gocdk launch`, but doesn't add it as an option to `gocdk biome add` yet. The launcher pushes an image to ECR then starts a task on an EC2-based ECS cluster. I have tested this manually. The logic mostly follows the same steps that the Cloud Run launcher does.

The launch process is effectively everything needed, but there is a notable TODO to set up Docker auth for the ECR repository. This will be addressed in a subsequent PR.

Updates #2301

### DIFF
--- a/internal/cmd/gocdk/launch.go
+++ b/internal/cmd/gocdk/launch.go
@@ -21,6 +21,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/spf13/cobra"
 	"gocloud.dev/gcp"
 	"gocloud.dev/internal/cmd/gocdk/internal/docker"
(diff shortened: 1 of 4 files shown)


internal/cmd/gocdk/launch.go
```diff
@@ -126,6 +127,16 @@ func newLauncher(ctx context.Context, pctx *processContext, launcherName string)
 			Client:       runService,
 			DockerClient: docker.New(pctx.env),
 		}, nil
+	case "ecs":
```
vangent: These names (including the existing `cloudrun` are a bit obscure. It's fine if they are used as an internal enum, which I think these are, but we need more explanation (e.g., maybe links to product pages) when presenting to users.

zombiezen: Ack. I agree, but I can't come up with an action to take on this feedback.

clausti: the launcher interface could specify an "info" method, which would return a short paragraph, and it could be used in conjunction with the map of specifier strings to launcher types to display info to the user when the list of options is displayed?

zombiezen: Great idea! Opened #2431 to track.

PR #2421, opened 2019-06-26
doc: update howto "constructors versus URL openers" to be consistent

eliben:


eliben:
> I like the direction of the new wording, but I think it's important to include verbage about the blank imports to explain why they're there. Can you add that back?

Done. Also added it to the runtimevar doc which already had the new wording

PR #2423, opened 2019-06-26
doc: make ref headings consistent with howto headings

eliben:


zombiezen:
I'm not sure I agree with this change. I think it would be better to have the reference section be titled after the packages each section represents. WDYT?

eliben:
> I'm not sure I agree with this change. I think it would be better to have the reference section be titled after the packages each section represents. WDYT?

They aren't right now, or at least the capitalization is wrong, so it just seems like an inconsistency. An alternative would be to rename the ref headings to be `fixed width` package names, which would give a better clue to their difference from the other sections?

zombiezen:
> An alternative would be to rename the ref headings to be `fixed width` package names, which would give a better clue to their difference from the other sections?

That sounds good. I'd be willing to take a shot at that, if you like. Main thing stopping us is we'd have to poke at the Hugo layout templates to change the sidebar styling.

eliben:
> > An alternative would be to rename the ref headings to be `fixed width` package names, which would give a better clue to their difference from the other sections?
> 
> That sounds good. I'd be willing to take a shot at that, if you like. Main thing stopping us is we'd have to poke at the Hugo layout templates to change the sidebar styling.

Sure, go for it.  I figured it would *just work* but maybe styling tweaks are required too

eliben:
#2430 is a better fix

Issue #2350, opened 2019-06-14
internal/cmd/gocdk: create unique tag for each build


zombiezen:
### Is your feature request related to a problem? Please describe.

When trying to identify Docker images, it is frequently helpful to include a unique, parseable tag beyond just the digest to be able to identify the image.

### Describe the solution you'd like

Running `gocdk build` should apply multiple tags: `foo:latest` and `foo:20190614024745_abcdef01` where `foo` is the base image name, `20190614024745` is the timestamp that the build started, and `abcdef01` are 4 hex-encoded random bytes (since these tags need to be globally unique).

### Additional context

There's an example of applying multiple tags to a `docker build` in [the docs](https://docs.docker.com/engine/reference/commandline/build/#tag-an-image--t).

PR #2424, opened 2019-06-26
internal/cmd/gocdk: validate implicit image name

zombiezen:
Prevents potential bugs where a tag may be added in a surprising context.


zombiezen:
(This is cleanup prep for #2350.)

PR #2425, opened 2019-06-27
docstore/memdocstore: remove unnecessary Option BREAKING_CHANGE_OK

### LINKED ISSUE #2411: docstore/memdocstore: MaxOutstandingActionRPCs makes no sense for memdocstore
https://godoc.org/gocloud.dev/docstore/memdocstore#Options

vangent:
Fixes #2411.


docstore/memdocstore/mem.go
```diff
@@ -140,9 +136,9 @@ func (c *collection) ErrorCode(err error) gcerr.ErrorCode {
 func (c *collection) RunActions(ctx context.Context, actions []*driver.Action, opts *driver.RunActionsOptions) driver.ActionListError {
 	errs := make([]error, len(actions))
 
-	// Run the actions concurrently with each other.
+	// Run the actions concurrently with each other, with a max of 10.
 	run := func(as []*driver.Action) {
-		t := driver.NewThrottle(c.opts.MaxOutstandingActionRPCs)
+		t := driver.NewThrottle(10)
```
shantuo: If we want to remove the max, why do we still need to throttling here?

jba:
Please include me as a reviewer on all significant docstore PRs, especially API changes like this one.

docstore/memdocstore/mem.go
vangent: So, should I just remove this entirely? I wasn't sure if doing so would result in unbounded goroutines.

jba:
Even though the name of the option may be wrong (because of `RPC`), I still think it's valuable. 

For one thing, it is educational; when someone playing with `memdocstore` moves on to use "real" drivers, they'll understand the concept.

But it's also useful when constructing very long action lists. It's reasonable even in a "toy" driver like this to put a limit on the number of goroutines created at once.

docstore/memdocstore/mem.go
jba: As per my comment in "Conversation," can we just rename the option to `MaxOutstandingActions`?

vangent: Done.

PR #2434, opened 2019-06-27
docstore: add more details to an example comment about id_field

eliben:


docstore/mongodocstore/example_test.go
```diff
@@ -78,6 +78,12 @@ func Example_openCollectionFromURL() {
 	ctx := context.Background()
 
 	// docstore.OpenCollection creates a *docstore.Collection from a URL.
+	// MongoDB requires a unique ID key in each document. If the default
```
shantuo: There is a section describing this in the how-to guide (https://github.com/google/go-cloud/blob/master/internal/website/content/howto/docstore/open-collection.md), I think we shouldn't repeat here, we can comment the link maybe.

Also, there is another way for struct, the user can use a `docstore:"_id"` struct tag.

eliben:
Good point, I'll close the PR then.

Issue #2366, opened 2019-06-17
internal/website: add server how-to guide


zombiezen:
Follow-on from #1576.

~The `server` package is missing a how-to guide. The guide should walk the user through:~

- ~Starting a server~
- ~Shutting down a server~
- ~Adding a health check and explain the difference between `/healthz/liveness` and `/healthz/readiness`~
- ~Setting up request logging (probably part of starting a server)~
- Links to OpenCensus docs for setting up trace exporters

~As per @vangent's request, this is marked as P1/beta-blocking.~

OpenCensus and Stackdriver sections still need to be written, but this is no longer a P1/beta-blocking bug since a basic guide has been committed. 

[edits by @clausti]

PR #2437, opened 2019-06-28
internal/website/content/howto: add server examples

clausti:
updates #2366  (Updates rather than fixes because this PR doesn't contain examples for OpenCensus trace exporting.)


internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,67 @@
...
+
+{{< goexample src="gocloud.dev/server.ExampleServer_RequestLogger" >}}
+
+### Adding health checks
+
+The Go CDK `server` package affords a hook for you to define health checks for your application and see the results at `/healthz/readiness`. Health checks are an imortant part of application monitoring.
+
+
+Because each application may have a different definition of what it means to be "healthy", you will need to define a concrete type to implment the `health.Checker` interface and define a `CheckHealth` method specific to your application.
+```go
```
clausti: This code I had to copy because the example importer at the moment seems to only does actual example tests. (I tried adding the magic comment to both the struct and the function and importing them and using them like the example test and it did not work. Turns out you also cannot define a struct type and give it methods from within another function.)

internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,67 @@
+---
+title: "Server"
+date: 2019-06-21T10:36:43-07:00
+draft: true
+showInSidenav: true
+---
+
+The Go CDK's `server` package provides a pre-configured HTTP server with diagnostic hooks for request logging, health checks, and trace exporting via OpenCensus. These guides will show you how to start up and shut down the server, as well as how to work with the request logging, health checks, and trace exporting.
```
eliben: nit: wrap to 80 columns here and elsewhere

eliben: s/These guides/This guide/

internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,67 @@
...
+
+The Go CDK includes a server package because??
```
eliben: I suggest you turn this into a real paragraph for now (I'm concerned we will forget it in its current state - there's a nonzero chance...) and reviewers will suggest fixes if interested

internal/website/content/howto/server/_index.md
eliben: SG, I saw the docstore howto does the same

internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,67 @@
...
+You can use the `server.Options` struct to specify a request logger.
+
+The example is shown with the Go CDK [`requestlog`](https://godoc.org/gocloud.dev/requestlog) package's `NCSALogger`. To get logs in the Stackdriver JSON format, use `NewStackdriverLogger` in place of `NewNCSALogger`.
+
+{{< goexample src="gocloud.dev/server.ExampleServer_RequestLogger" >}}
+
+### Adding health checks
+
+The Go CDK `server` package affords a hook for you to define health checks for your application and see the results at `/healthz/readiness`. Health checks are an imortant part of application monitoring.
+
```
eliben: nit: one empty line between paragraphs

internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,67 @@
...
+### Trace exporting with OpenCensus
+
+- default behavior
+- how to specify something
+
+
+## Shutting down the server
+
+Like all Go CDK [portable types](https://gocloud.dev/concepts/structure/#portable-types-and-drivers), `server.Server`  has a `driver` field with an interface type (in this case, `driver.Server`). By default, the `server` package uses `http.Server` to satisfy its [driver interface](https://godoc.org/gocloud.dev/server/driver), but you can use other implementations if you prefer. Calling `Shutdown` on the server calls the driver's `Shutdown` method.
```
eliben: blob.Bucket doesn't have a `driver` field

I'd suggest phrasing it in a less implementation-oriented way.

server/example_test.go
```diff
@@ -0,0 +1,133 @@
...
+	defer h.mu.RUnlock()
+	if !h.healthy {
+		return errors.New("not ready yet!")
+	}
+	return nil
+}
+func ExampleServer_HealthChecks() {
+	// This example is used in https://gocloud.dev/howto/server/
+
+	// Create a health.Checker from the type we definied for our application.
```
eliben: defined

server/example_test.go
```diff
@@ -0,0 +1,133 @@
...
+	// The server.Options struct takes a collection of health checks, because you
+	// may need to check several things.
+	srvOptions := &server.Options{
+		HealthChecks: []health.Checker{healthCheck},
+	}
+
+	// Pass the options to the Server constructor.
+	srv := server.New(http.DefaultServeMux, srvOptions)
+
+	// Register routes and start the server.
```
eliben: Maybe fill this in? Examples should be reasonably self contained (as much as possible) so that people could copy paste them out and run them.

You can use magic comments to exclude this code from the howto page

internal/website/content/howto/server/_index.md
vangent: Strongly suggest using `mdformat`:

http://go/mdformat

There's no external alternative that we could find, but it's pretty handy.

internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,67 @@
...
+The example is shown with the Go CDK [`requestlog`](https://godoc.org/gocloud.dev/requestlog) package's `NCSALogger`. To get logs in the Stackdriver JSON format, use `NewStackdriverLogger` in place of `NewNCSALogger`.
+
+{{< goexample src="gocloud.dev/server.ExampleServer_RequestLogger" >}}
+
+### Adding health checks
+
+The Go CDK `server` package affords a hook for you to define health checks for your application and see the results at `/healthz/readiness`. Health checks are an imortant part of application monitoring.
+
+
+Because each application may have a different definition of what it means to be "healthy", you will need to define a concrete type to implment the `health.Checker` interface and define a `CheckHealth` method specific to your application.
```
vangent: implment -> implement

internal/website/content/howto/server/_index.md
vangent: I would consider leaving the part about the driver out. I don't think we expect most users to provider their own driver.

Honestly, I'm not even sure you need a section on `Shutdown` explicitly; just make the examples call it appropriately.

server/example_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	// Use the constructor function to create the server.
+	srv := server.New(http.DefaultServeMux, nil)
+
+	// Register a route.
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, "Hello, World!")
+	})
+
+	// Start the server.
```
vangent: This function blocks forever, right? Worth mentioning.

server/example_test.go
```diff
@@ -0,0 +1,133 @@
...
+	srv := server.New(http.DefaultServeMux, srvOptions)
+
+	// Register a route.
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, "Hello, World!")
+	})
+
+	// Start the server. You will see requests logged to STDOUT.
+	if err := srv.ListenAndServe(":8080"); err != nil {
+		fmt.Fprintln(os.Stderr, err)
```
vangent: Consider using `log.Exitf` instead of `fmt.Fprint` and `os.Exit`.

server/example_test.go
```diff
@@ -0,0 +1,133 @@
...
+	// Use the constructor function to create the server.
+	srv := server.New(http.DefaultServeMux, nil)
+
+	// Register a route.
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		fmt.Fprintln(w, "Hello, World!")
+	})
+
+	// Start the server.
+	srv.ListenAndServe(":8080")
```
vangent: Check for error here as you do below.

server/example_test.go
```diff
@@ -0,0 +1,133 @@
...
+	// after startup, and as healthy henceforth. Check the /healthz/readiness
+	// HTTP path to see readiness.
+	healthCheck := new(customHealthCheck)
+	time.AfterFunc(10*time.Second, func() {
+		healthCheck.mu.Lock()
+		defer healthCheck.mu.Unlock()
+		healthCheck.healthy = true
+	})
+
+	// The server.Options struct takes a collection of health checks, because you
```
vangent: collection -> slice

server/example_test.go
vangent: +1, and call `ListenAndServe`.

server/example_test.go
```diff
@@ -0,0 +1,133 @@
...
+		HealthChecks: []health.Checker{healthCheck},
+	}
+
+	// Pass the options to the Server constructor.
+	srv := server.New(http.DefaultServeMux, srvOptions)
+
+	// Register routes and start the server.
+}
+
+func ExampleServer_Shutdown() {
```
vangent: I would leave this one out.

server/example_test.go
```diff
@@ -0,0 +1,133 @@
...
+	// This example is used in https://gocloud.dev/howto/server/
+
+	// Create a logger, and assign it to the RequestLogger field of a
+	// server.Options struct.
+	srvOptions := &server.Options{
+		RequestLogger: requestlog.NewNCSALogger(os.Stdout, func(error) {}),
+	}
+
+	// Pass the options to the Server constructor.
+	srv := server.New(http.DefaultServeMux, srvOptions)
```
vangent: Probably should add `defer srv.Shutdown()` here and elsewhere, right?

server/example_test.go
clausti: this one? https://godoc.org/github.com/golang/glog#Exitf  

(https://godoc.org/log doesn't seem to have anything called `Exitf`?)

vangent: Sorry, `Fatalf`: https://golang.org/pkg/log/#Fatalf

eliben: FWIW, we're using `log.Fatal` for this in other examples. Let's use the same here? There's also `log.Fatalf` if needed (not here)

eliben:
On Wed, Jul 3, 2019 at 2:55 PM Robert van Gent <notifications@github.com>
wrote:

> *@vangent* commented on this pull request.
> ------------------------------
>
> In server/example_test.go
> <https://github.com/google/go-cloud/pull/2437#discussion_r300167284>:
>
> > +	// server.Options struct.
> +	srvOptions := &server.Options{
> +		RequestLogger: requestlog.NewNCSALogger(os.Stdout, func(error) {}),
> +	}
> +
> +	// Pass the options to the Server constructor.
> +	srv := server.New(http.DefaultServeMux, srvOptions)
> +
> +	// Register a route.
> +	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
> +		fmt.Fprintln(w, "Hello, World!")
> +	})
> +
> +	// Start the server. You will see requests logged to STDOUT.
> +	if err := srv.ListenAndServe(":8080"); err != nil {
> +		fmt.Fprintln(os.Stderr, err)
>
> Sorry, Fatalf: https://golang.org/pkg/log/#Fatalf
>

Crossed comments :)

server/example_test.go
clausti: 🙃

clausti: I added the defer calls to `Shutdown`, not sure if they're in the right place. I stuck it right after creating the server.

server/example_test.go
clausti: I'm leaving it for now because I don't expect to merge this until next week, but if I'm adding shutdown calls to the other examples it does seem kind of redundant.

zombiezen: 👍

internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,65 @@
...
+draft: true
+showInSidenav: true
+---
+
+The Go CDK's `server` package provides a pre-configured HTTP server with 
+diagnostic hooks for request logging, health checks, and trace exporting via 
+OpenCensus. This guide will show you how to start up and shut down the server,
+as well as how to work with the request logging and health checks.
+
+The Go CDK includes a server package to provide some reasonable defaults, such
```
eliben: I'm actually not sure we need this paragraph at all. The first paragraph is pretty good now, Curious to hear what other reviewers think here

internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,65 @@
...
+as timeouts, or the Apache standard format logger, and also to share best
+practices like health checks.
+
+## Starting up the server
+
+The GO CDK Server constructor takes an `http.Handler` and an `Options` struct. 
+The simplest way to start the server is to use the `http.DefaultServeMux` and
```
eliben: nit: maybe "the" is unnecessary before `http.DefaultServeMux`?

internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,65 @@
...
+### Adding health checks
+
+The Go CDK `server` package affords a hook for you to define health checks for
+your application and see the results at `/healthz/readiness`. Health checks are
+an imortant part of application monitoring.
+
+Because each application may have a different definition of what it means to be
+"healthy", you will need to define a concrete type to implement the `health.Checker`
+interface and define a `CheckHealth` method specific to your application.
+```go
```
eliben: Add empty line before the code sample

internal/website/content/howto/server/_index.md
zombiezen: I think the first paragraph is strong enough to stand on its own.

internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,65 @@
...
+OpenCensus. This guide will show you how to start up and shut down the server,
+as well as how to work with the request logging and health checks.
+
+The Go CDK includes a server package to provide some reasonable defaults, such
+as timeouts, or the Apache standard format logger, and also to share best
+practices like health checks.
+
+## Starting up the server
+
+The GO CDK Server constructor takes an `http.Handler` and an `Options` struct.
```
zombiezen: Typo:

```suggestion
The Go CDK Server constructor takes an `http.Handler` and an `Options` struct. 
```

internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,65 @@
...
+To get logs in the Stackdriver JSON format, use `NewStackdriverLogger` in place
+of `NewNCSALogger`.
+
+{{< goexample src="gocloud.dev/server.ExampleServer_RequestLogger" >}}
+
+### Adding health checks
+
+The Go CDK `server` package affords a hook for you to define health checks for
+your application and see the results at `/healthz/readiness`. Health checks are
+an imortant part of application monitoring.
```
zombiezen: Spelling:

```suggestion
an important part of application monitoring.
```

internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,65 @@
...
+The example is shown with the Go CDK [`requestlog`](https://godoc.org/gocloud.dev/requestlog) package's `NCSALogger`.
+To get logs in the Stackdriver JSON format, use `NewStackdriverLogger` in place
+of `NewNCSALogger`.
+
+{{< goexample src="gocloud.dev/server.ExampleServer_RequestLogger" >}}
+
+### Adding health checks
+
+The Go CDK `server` package affords a hook for you to define health checks for
+your application and see the results at `/healthz/readiness`. Health checks are
```
zombiezen: It would also be good to mention the `/healthz/liveness` path that server runs too. There's no programmer control over it, but it will always respond with HTTP 200, assuming that the server is not somehow wedged. This is used by Kubernetes liveness checks and by `gocdk serve` to know when the server has started up, independently of when the server is ready to receive traffic.

server/example_test.go
zombiezen: Sorry, missed this on first pass:

`defer srv.Shutdown()` is probably not right since `ListenAndServe` blocks forever. The more correct code would be:

```go
srv := server.New(http.DefaultServeMux, srvOptions)
go func() {
	interrupt := make(chan os.Signal, 1)
	signal.Notify(interrupt, os.Interrupt)
	for { // in a loop, because the interrupt could be sent before ListenAndServe starts
		<-interrupt
		srv.Shutdown()
	}
}()
```

But stepping back, most "cloud-native"™️ deployments will likely behind a load balancer that will handle the graceful shutdown of requests. I think the hard interrupt is fine, so I think we can leave `Shutdown` out of these examples.

Because this is subtle, I recommend we add my above snippet to the `ExampleServer_Shutdown` godoc function, but not discuss it in the how-to guide, since it doesn't seem like an 80% use case. WDYT?

internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,67 @@
...
+The simplest way to start the server is to use `http.DefaultServeMux` and
+pass `nil` for the options.
+
+{{< goexample src="gocloud.dev/server.ExampleServer_New" >}}
+
+### Adding a request logger
+
+You can use the `server.Options` struct to specify a request logger.
+
+The example is shown with the Go CDK [`requestlog`](https://godoc.org/gocloud.dev/requestlog) package's `NCSALogger`.
```
vangent: Update to `server/requestlog` now.

internal/website/content/howto/server/_index.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Adding health checks
+
+The Go CDK `server` package affords a hook for you to define health checks for
+your application and see the results at `/healthz/readiness`. The server also
+runs an endpoint at `/healthz/liveness` which will always return `HTTP 200` if
+the server is up, which is also the endpoint Kubernetes will use to check liveness.
+
+Health checks are an important part of application monitoring, and readiness
+checks are subtly different than liveness checks. The liveness check will always
```
vangent: "always and simply" sounds a bit odd, maybe "will always return" or "will simply return"?

vangent: Or just drop this sentence, you said it above?

server/example_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+	"sync"
+	"time"
+
+	"gocloud.dev/health"
```
vangent: These need to be updated to `server/health` and `server/requestlog` now (and sorted under `server`).

server/example_test.go
vangent: SGTM. Sorry for my first misleading comment.

server/example_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// The server.Options struct takes a slice of health checks, because you
+	// may need to check several things.
+	srvOptions := &server.Options{
+		HealthChecks: []health.Checker{healthCheck},
+	}
+
+	// Pass the options to the Server constructor.
+	srv := server.New(http.DefaultServeMux, srvOptions)
+	// Don't forget to shut down the server.
```
vangent: Remove here too.

clausti:
ok, to the best of my knowledge I have addressed all comments-to-date, and the recent test failures are known flakes in pubsub. PTAL?

Issue #2415, opened 2019-06-26
docstore: provide better error message when Get fieldpaths have wrong case


vangent:
I have a struct with a field `Key`.

When I do a `Get` and pass it `xxxx` as a field path, I get a good error message, something like: `no field matching xxxx`.

When I do the same with `key`, I get:
`cannot set type string to <nil>`

Why the difference?

shantuo:
I think this is intended and follows the Go rules, like in `encoding/json`. Our godoc says "Match looks for an exact match first, then falls back to a case-insensitive comparison", in the second case, it got a match, you got the error because of the type mismatch.

vangent:
What type mismatch?

shantuo:
It looks like the `key` is a string type but it got a `nil`? Can I see your example? It might be a bug in somewhere else in the decoder.

vangent:
Try adding this to `memdocstore/mem_test.go`:

```
type MyDoc struct {
        Key string
        Value string
        DocstoreRevision interface{}
}

func TestWrongKey(t *testing.T) {
        ctx := context.Background()

        coll, err := OpenCollection("Key", nil)
        if err != nil {
                t.Fatal(err)
        }
        defer coll.Close()

        doc := MyDoc{Key: "foo", Value: "testvalue"}
        if err := coll.Create(ctx, &doc); err != nil {
                t.Fatal(err)
        }

        getdoc := MyDoc{Key: "foo"}
        if err := coll.Get(ctx, &getdoc, "Value"); err != nil {
                t.Fatal(err)
        }
        if getdoc.Value != "testvalue" {
                t.Errorf("got %v want %v", getdoc.Value, "testvalue")
        }

        err = coll.Get(ctx, &getdoc, "BadKey")
        t.Log(err)

        err = coll.Get(ctx, &getdoc, "value")
        t.Log(err)
}
```

and then run with `-v`. I get:

```
--- PASS: TestWrongKey (0.00s)
    mem_test.go:86: no field matching "BadKey" in memdocstore.MyDoc (code=InvalidArgument)                                                           
    mem_test.go:89: cannot set type string to <nil> (code=InvalidArgument)
```

The first error looks fine.
The second error is weird. It's not obvious where the `nil` is coming from, it's the same document that worked above.

shantuo:
The driver's decoder find the match but the memdocstore doesn't do the same.

shantuo:
Actually, I was wrong. I think the second one should also be an error, we could improve the error message.

[...]

PR #2438, opened 2019-06-28
docstore/internal/fields: support both case-sensitive and case-insensitive matching

jba:
This is something we would want if this package were ever made available for general use.

I did it so we could easily experiment with having docstore do case-sensitive field matching.

### DIFF
--- a/docstore/internal/fields/fields.go
+++ b/docstore/internal/fields/fields.go
@@ -178,16 +178,32 @@ func (c *Cache) Fields(t reflect.Type) (List, error) {
 // A List is a list of Fields.
 type List []Field
 
-// Match returns the field in the list whose name best matches the supplied
+// MatchExact returns the field in the list with the given name, or nil if there is
+// none.
+func (l List) MatchExact(name string) *Field {
+	return l.MatchExactBytes([]byte(name))
+}
+
+// MatchExactBytes is identical to MatchExact, except that the argument is a byte slice.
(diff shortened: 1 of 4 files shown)


docstore/internal/fields/fields.go
```diff
@@ -178,16 +178,32 @@ func (c *Cache) Fields(t reflect.Type) (List, error) {
 // A List is a list of Fields.
 type List []Field
 
-// Match returns the field in the list whose name best matches the supplied
+// MatchExact returns the field in the list with the given name, or nil if there is
```
shantuo: My only concern is that we should document when to use which "match" somewhere, maybe here in the comment is fine.

jba: What do you mean by "when to use which 'match'"? Can you give me an example?

shantuo: I meant there are `MatchExact` and `MatchFold`, we are using `MatchFold` at the moment, it would be helpful to document why we need both for future reference.

jba: Oh, I think of this as a general-purpose package. Someday I want to move it out of here, maybe to its own repo. For now, I don't want to expose it to the world so it lives in internal. But it isn't specific to docstore.

So I think that comment is a good idea, but instead of putting it here I'd put it in driver/codec.go, where we call the match function. You can do that as part of #2415.

Issue #2435, opened 2019-06-28
pubsub/kafkapubsub: consumer group not rebalancing


burdiyan:
### Describe the bug

Having multiple consumers within the same consumer group, rebalance is not handled properly.

### To Reproduce

Having Kafka running: 

1. Create a topic with 2 or more partitions (below `backblaze_smart`).
2. Create a `main` package with the following code:

```go
package main

import (
	"context"
	"fmt"
	"os"
	"os/signal"
	"syscall"

	"gocloud.dev/pubsub/kafkapubsub"
)

var (
	brokers = []string{"localhost:9092"}
	topics  = []string{"backblaze_smart"}
)

func main() {
	if err := run(); err != nil {
		panic(err)
	}
}

func run() error {
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	go func() {
		ch := make(chan os.Signal, 1)
		signal.Notify(ch, os.Interrupt, os.Kill, syscall.SIGTERM)
		<-ch
		cancel()
		signal.Stop(ch)
	}()

	sub, err := kafkapubsub.OpenSubscription(
		brokers,
		kafkapubsub.MinimalConfig(),
		"pubsub-group-test",
		topics,
		&kafkapubsub.SubscriptionOptions{
			KeyName: "key",
		},
	)
	if err != nil {
		return err
	}
	defer sub.Shutdown(context.Background())

	for {
		msg, err := sub.Receive(ctx)
		if err != nil {
			return err
		}

		fmt.Printf("%s:%s\n", msg.Metadata["key"], string(msg.Body))
		msg.Ack()
	}
}
```

3. Start one instance of the program and see messages being printed in the console.
4. Start another instance of the program, and see messages being printed in the console of the second instance, but first instance stops.

## Expected behavior

Messages being printed in both consoles, since this topic has 2 partitions, and messages should spread across 2 instances.

### Version

https://github.com/google/go-cloud/commits/2e278191e064c358d4fee5cc10de4076878ad974

vangent:
Hi, thanks for the report. Confirmed that this is a bug; fix is on the way.

burdiyan:
@vangent Could you leave the issue open until the PR is merged? I guess otherwise it's a little bit confusing. If that goes against your contributing guidelines, sorry for that and nevermind :)

vangent:
[...]

PR #2451, opened 2019-07-02
pubsub/kafkapubsub: add a test for rebalancing and multiple partitions

vangent:
Updates #2435.


pubsub/kafkapubsub/kafka_test.go
```diff
@@ -276,6 +276,120 @@ func TestKafkaKey(t *testing.T) {
...
+	// Send some messages.
+	send := func() {
+		for i := 0; i < nMessages; i++ {
+			m := &pubsub.Message{
+				Metadata: map[string]string{
+					keyName: fmt.Sprintf("key%d", i),
+				},
+				Body: []byte("hello world"),
+			}
+			err = topic.Send(ctx, m)
```
jba: `if err := topic.Send(..); err != nil ..`

Or at the very least, don't re-use the `err` from the outer scope.

vangent: Done.

PR #2457, opened 2019-07-02
docstore: revise package doc for clarity

polinasok:


docstore/doc.go
```diff
@@ -236,26 +239,26 @@
...
+//       fmt.Println(m)
+//   }
 //
 //
 // Errors
 //
 // The errors returned from this package can be inspected in several ways:
 //
-// The Code function from gocloud.dev/gcerrors will return an error code, also
+// The Code function from https://gocloud.dev/gcerrors will return an error code, also
```
jba: Cool, that link redirects to godoc. TIL!

docstore/doc.go
```diff
@@ -12,42 +12,42 @@
...
 // certain criteria.
 //
-// Subpackages contain distinct implementations ("drivers") of docstore for various
-// providers, including MongoDB, Google Cloud Firestore, and Amazon DynamoDB.
-// You can use Azure Cosmos DB and Amazon DocumentDB via the MongoDB driver.
-// There is also memdocstore, an in-process, in-memory implementation suitable
-// for testing and development.
-//
-// Your application should import one of these provider-specific subpackages and use
+// Your application should import one of these service-specific subpackages and use
```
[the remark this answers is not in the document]
polinasok: Where would the "In docstore" paragraph would go then? I rearranged this way, so the high level blurb about use cases goes first, and then "In docstore" paragraph introduces the more low level concept of collections right before we talk about the specifics of creating one. I don't feel strongly about this and can be convinced to restore the original order. For now, I changed "these" for this to read better as is.

docstore/doc.go
```diff
@@ -109,78 +109,77 @@
...
 //
-//    type Book struct {
+//   type Book struct {
 //       Title            string `docstore:"title"`
 //       Author           Name   `docstore:"author"`
-//       PublicationYears []int `docstore:"pub_years,omitempty"`
-//       PublicationCount string `docstore:"-"` // number of publications
-//    }
+//       PublicationYears []int  `docstore:"pub_years,omitempty"`
+//       NumPublications  string `docstore:"-"`
```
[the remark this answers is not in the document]
polinasok: You are right! I missed this as well :(. Fixed.

docstore/doc.go
```diff
@@ -109,78 +109,77 @@
...
 // Keys never appear alone in the docstore API, only as part of a document. For
-// instance, to retrieve a document by key, you pass the Get method a document whose
-// key field is populated, and docstore populates the rest of that argument with the
-// stored document.
+// instance, to retrieve a document by key, you pass the Collection.Get method
+// a document struct or map with the key field populated, and docstore
+// populates the rest of that argument with the stored contents.
 //
-// When you open a collection using one of the subpackages' OpenCollection methods or
+// When you open a collection using OpenCollection method of the service-specific driver or
```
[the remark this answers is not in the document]
polinasok: Done

docstore/doc.go
```diff
@@ -109,78 +109,77 @@
...
 //
 // Actions
 //
-// Docstore supports six actions on documents:
+// Docstore supports six actions on documents as methods on the Collection type:
 //   - Get retrieves a document.
 //   - Create creates a new document.
 //   - Replace replaces an existing document.
-//   - Put puts a document into a collection whether or not it is already present.
+//   - Put puts a document into a collection, replacing it if already present.
```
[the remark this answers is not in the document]
polinasok: Done.

docstore/doc.go
polinasok: I was excited to learn this too :)

docstore/doc.go
```diff
@@ -109,78 +109,77 @@
...
-// Keys
+// Representing Keys
 //
 // The key of a docstore document is some function of its contents, usually a field.
 // Keys never appear alone in the docstore API, only as part of a document. For
-// instance, to retrieve a document by key, you pass the Get method a document whose
-// key field is populated, and docstore populates the rest of that argument with the
-// stored document.
+// instance, to retrieve a document by key, you pass the Collection.Get method
+// a document struct or map with the key field populated, and docstore
```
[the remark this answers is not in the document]
polinasok: Reworded. PTAL.

docstore/doc.go
```diff
@@ -12,42 +12,42 @@
...
-// NoSQL databases, document stores are schemaless.
+// Package docstore provides a portable way of interacting with a document store.
+// A document store is a service that stores data in semi-structured JSON-like
+// documents. Like other NoSQL databases, document stores are schemaless.
 // See https://en.wikipedia.org/wiki/Document-oriented_database for more information.
 //
+// Subpackages contain distinct implementations ("drivers") of docstore for public
+// cloud services like Google Cloud Firestore and Amazon DynamoDB as well as any
+// MongoDB service hosted by a cloud provider or on-premise. For example, you can
+// use Azure Cosmos DB and Amazon DocumentDB via the MongoDB driver. There is also
```
shantuo: I think Cosmos and DocDB are not "service hosted by a cloud provider", I think they are just compatible with MongoDB's API.

docstore/doc.go
```diff
@@ -12,42 +12,42 @@
...
-// reconfigure your initialization code to choose a different provider.
+// reconfigure your initialization code to choose a different driver.
 // You can develop your application locally using memdocstore, or deploy it to
 // multiple Cloud providers. You may find http://github.com/google/wire useful
 // for managing your initialization code.
 //
-// Alternatively, you can construct a *Collection via a URL and the OpenCollection
-// function in this package. See https://gocloud.dev/concepts/urls for more
+// Alternatively, you can construct a *Collection via a URL and the
+// docstore.OpenCollection function. See https://gocloud.dev/concepts/urls for more
```
shantuo: Do we need to add `docstore` here since it is in this package. When the user looks at the godoc, this would be in the same page.

polinasok: Done

docstore/doc.go
polinasok: I reworded a bit. I believe these are services, and they are hosted by a provider, but they are not MongoDB services, strictly speaking, so I called out MongoDB-compatible separately (although this page does not: https://www.mongodb.com/cloud/atlas/compare). Wouldn't our mongo driver work with MongoDB Atlas as well?

jba: That's right. MongoDB Atlas is something like that -- I think it's a multi-cloud product built by Mongo that will start mongo instances in the cloud.

As Tuo said, Cosmos and DocumentDB are their own products with their own implementations, and they just talk the mongo protocol.

jba: Your rewording LGTM.

PR #2463, opened 2019-07-03
secrets/azurekeyvault: add a way to use az CLI auth; provide a better resource default

### LINKED ISSUE #2462: secrets/azurekeyvault: add a way to use CLI auth
We're currently using environment variables defined by the Azure auth library to authenticate (in `Dial`). However, there's another function provided that gets creds from the `az` CLI, which may also be useful. Let's add an environment variable that determines which to use.

Also, we should provide a default `AZURE_AD_RESOURCE` since the regular default is wrong (the Azure `auth` library defaults it to the "management" resource, which is different than the "keyvault" one).

vangent:
Fixes #2462.

1. If you set `AZURE_KEYVAULT_AUTH_VIA_CLI` to a truthy value, the default `URLOpener` will now get authorization from the `az` CLI instead of from the environment (there are 2 different functions in the Azure `auth` library for this). We also expose both forms via the existing `Dial` and a new `DialUsingCLIAuth`.

2. If you provide a `AZURE_AD_RESOURCE`, we use it, but if you don't, we use `https://vault.azure.net`. The Azure `auth` library defaults it to the "management" resource, which is wrong for KeyVault.

3. Fixed a bug in URL parsing where the URL was parsed incorrectly if the version was included.


secrets/azurekeyvault/akv.go
```diff
@@ -20,7 +20,10 @@
 //
 // For secrets.OpenKeeper, azurekeyvault registers for the scheme "azurekeyvault".
 // The default URL opener will use Dial, which gets default credentials from the
-// environment.
+// environment, unless the AZURE_KEYVAULT_AUTH_VIA_CLI environment variable is
```
shantuo: nit: `AZURE_KEYVAULT_AUTH_VIA_CLI` is defined by us, so it might be helpful to differentiate it from azure's env names. Maybe starts with `GOCDK`.

vangent: We don't do that with any of the other GOCDK environment variables, so I don't think we should do it here. There was some discussion earlier about namespacing them that way, but we decided not to.

PR #2470, opened 2019-07-04
samples/order: frontend implementation

jba:
The front-end web server for the order sample.

As I expected, writing the frontend caused me to rethink the
backend. So that has also changed: everything about the order is in
the database, and the pubsub response has just the order ID. I also
refactored order processing to make it clearer (hopefully).

Still to be done:

- Figure out how to use the response topic. Ideally this would
  cause the frontend server to signal the web page that an order
  is ready.  I'm not sure how to write that.

- Some minimal CSS to make the site presentable.


samples/order/frontend.go
```diff
@@ -12,19 +12,193 @@
...
+
+var (
+	listTemplate      = template.Must(template.ParseFiles("list.htmlt"))
+	orderFormTemplate = template.Must(template.ParseFiles("order-form.htmlt"))
+)
+
+// run starts the server on port and runs it indefinitely.
+func (f *frontend) run(ctx context.Context, port int) error {
+	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) { http.ServeFile(w, r, "index.html") })
+	http.HandleFunc("/orders/", wrap(f.listOrders))
```
eliben: I'd rather not have this abstraction in tutorial-y code. As an alternative consider just spelling the error handling out, or have `listOrders` et.al. do `http.Error` directly. If you really want the wrapper, at least rename it to be more expressive like `wrapHttpError`.

samples/order/frontend.go
```diff
@@ -12,19 +12,193 @@
...
+	}
+	defer file.Close()
+	if _, err := f.doCreateOrder(r.Context(), email, file, time.Now()); err != nil {
+		return err
+	}
+	fmt.Fprintln(w, "Order received. Thank you.")
+	return nil
+}
+
+// doCreateOrder returns the order ID it generates, for testing.
```
eliben: beef up the comment to explain what it actually does, not just what it returns

samples/order/frontend.go
```diff
@@ -12,19 +12,193 @@
...
+		Email:      email,
+		CreateTime: now,
+	}
+
+	// Copy the uploaded input file to the bucket.
+	w, err := f.bucket.NewWriter(ctx, req.InImage, nil)
+	if err != nil {
+		return "", err
+	}
+	defer func() {
```
eliben: This is a bit hard to follow because the flow of control is inverted :-/

samples/order/frontend.go
```diff
@@ -12,19 +12,193 @@
...
+			return err
+		}
+		orders = append(orders, &ord)
+	}
+	return executeTemplate(listTemplate, orders, w)
+}
+
+// newID creates a new unique ID for an incoming order. It uses the current
+// second, formatted in a readable way. The resulting IDs sort nicely and are
+// easy to read, but of course are not suitable for production because there
```
eliben: why not append a running integer to the second? will still sort nicely and likely won't take much more code than this comment :)

samples/order/frontend.go
jba: I think it's a useful technique, and not very complicated, so I kept it but renamed it.

samples/order/frontend.go
jba: Done.

samples/order/frontend.go
jba: You'll get dups on a fast restart or with clock resets. My concern is that we'd be giving the illusion that we're dealing with dups and making the code (slightly) more complex, but we haven't come any closer to a production-quality implementation. (In fact it's worse, because it will fail more rarely.)

I'd be open to a simple production-quality idea (like using a UUID) or something even dumber than what I have (an in-memory counter).

jba:
PTAL

samples/order/frontend.go
eliben: Fair enough.

samples/order/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta http-equiv="content-type" content="text/html; charset=utf-8">
+  </head>
+  <body>
+    <h1>Go CDK Order Sample</h1>
```
vangent: WDYT about embracing the actual functionality a bit more? I.e., instead of "Place an Order", "Convert an Image"?

And maybe the frontend could have a link to the converted image once the "order" is complete.

@eliben

samples/order/frontend.go
```diff
@@ -12,19 +12,196 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// TODO(jba): use the response subscription to dynamically update a "My Orders" page.
```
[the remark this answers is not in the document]
jba: done

samples/order/index.html
jba: Can I implement the link by calling Bucket.SignedURL with nil options? It says it's not supported on all providers, but does it work on the big 3? 

I'd rather not implement the link myself (by reading from the bucket and writing to the browser) if all the providers have a way to link to a blob. That would indicate something missing from our API.

vangent: > Can I implement the link by calling Bucket.SignedURL with nil options?
> It says it's not supported on all providers, but does it work on the big 3?

Yes.

Although, for at least one provider (GCS) you need to provide some more info at construction time (e.g., use an `access_id` query parameter).  You can likely ignore that for now.

vangent: Feel free to add a `TODO` for this as well, and do it in a separate PR.

jba: Added the TODO.

samples/order/frontend.go
jba: Rewrote.

samples/order/order.go
```diff
@@ -33,25 +34,59 @@ var (
 	responseSubURL   = flag.String("response-sub", "mem://responses", "gocloud.dev/pubsub URL for response subscription")
 	bucketURL        = flag.String("bucket", "", "gocloud.dev/blob URL for image bucket")
 	collectionURL    = flag.String("collection", "mem://orders/ID", "gocloud.dev/docstore URL for order collection")
-	// TODO(jba): uncomment after adding frontend
-	// runProcessor     = flag.Bool("processor", true, "run the image processor")
+
+	port         = flag.Int("port", 10538, "port for frontend")
```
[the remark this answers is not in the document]
jba: done

samples/order/order.go
```diff
@@ -33,25 +34,59 @@ var (
...
+	}
+	frontend, processor, cleanup, err := setup(conf)
 	if err != nil {
 		log.Fatal(err)
 	}
 	defer cleanup()
-	if err := processor.run(context.Background()); err != nil {
-		log.Fatal(err)
+
+	errc := make(chan error, 2)
```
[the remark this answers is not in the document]
jba: done

jba:
PTAL.

PR #2472, opened 2019-07-08
memdocstore: support saving to a file

### LINKED ISSUE #1760: docstore: make the local driver stateful
We want to let the local dev driver keep the record when the user wants to do some local dev work without writing new items every time they restart it. There are two options:

1. Do a file based docstore.
1. Dump the mem store table when mem store exit.

jba:
Add a Filename option that will load a collection from a filename if it exists, and save it
when the collection is closed.

Fixes #1760.

### DIFF
--- a/docstore/memdocstore/mem.go
+++ b/docstore/memdocstore/mem.go
@@ -64,7 +64,7 @@ type Options struct {
 	// The filename associated with this collection.
 	// When a collection is opened with a non-nil filename, the collection
 	// is loaded from the file if it exists. Otherwise, an empty collection is created.
-	// When the collection is closed, is contents are saved to the file.
+	// When the collection is closed, its contents are saved to the file.
 	Filename string
 }
 
@@ -431,8 +431,8 @@ func (c *collection) As(i interface{}) bool { return false }
 func (c *collection) ErrorAs(err error, i interface{}) bool { return false }
 
 // Close implements driver.Collection.Close.
-// If the collection was created with  Filename option, Close writes the collection's
-// documents to the file.
+// If the collection was created with a Filename option, Close writes the
+// collection's documents to the file.
 func (c *collection) Close() error {
 	return saveDocs(c.opts.Filename, c.docs)
 }
--- a/docstore/memdocstore/mem.go
+++ b/docstore/memdocstore/mem.go
@@ -39,6 +39,8 @@ package memdocstore // import "gocloud.dev/docstore/memdocstore"
 
 import (
 	"context"
+	"encoding/gob"
+	"os"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


docstore/memdocstore/mem.go
```diff
@@ -58,6 +60,12 @@ type Options struct {
 	// The maximum number of concurrent goroutines started for a single call to
 	// ActionList.Do. If less than 1, there is no limit.
 	MaxOutstandingActions int
+
+	// The filename associated with this collection.
+	// When a collection is opened with a non-nil filename, the collection
+	// is loaded from the file if it exists. Otherwise, an empty collection is created.
+	// When the collection is closed, is contents are saved to the file.
```
shantuo: nit: s/is/its

docstore/memdocstore/mem.go
```diff
@@ -419,4 +431,48 @@ func (c *collection) As(i interface{}) bool { return false }
 func (c *collection) ErrorAs(err error, i interface{}) bool { return false }
 
 // Close implements driver.Collection.Close.
-func (c *collection) Close() error { return nil }
+// If the collection was created with  Filename option, Close writes the collection's
```
shantuo: nit: ... created with a filename, Close writes ...?

docstore/memdocstore/mem.go
```diff
@@ -419,4 +431,48 @@ func (c *collection) As(i interface{}) bool { return false }
...
+	}
+	f, err := os.Create(filename)
+	if err != nil {
+		return err
+	}
+	if err := gob.NewEncoder(f).Encode(m); err != nil {
+		_ = f.Close()
+		return err
+	}
+	return f.Close()
```
shantuo: Is this error important to return? Maybe it can just be simplified as

```go
defer f.Close()
return gob.NewEncoder(f).Encode(m)
```

docstore/memdocstore/urls.go
```diff
@@ -37,7 +37,11 @@ const Scheme = "mem"
 // The URL's host is the name of the collection.
 // The URL's path is used as the keyField.
 //
-// No query parameters are supported.
+// The revision_field query parameter specifies the name of the revision field.
```
shantuo: Just want to be consistent with other provider packages:

```go
// The following query parameters are supported:
//
//   - revision_field:
//   - filename:
```

docstore/memdocstore/mem.go
jba: done

docstore/memdocstore/mem.go
jba: When writing, `Close` might return an important error (because it may be flushing a write buffer, for example). You should also check errors from `Close` when writing.

docstore/memdocstore/urls.go
jba: done

PR #2473, opened 2019-07-08
samples/gocdk-blob,-pubsub: add tests

jba:
These test use the new cmdtest package to test the binary's output
directly.

Also, in gocdk-blob, make the upload command check for errors when
closing the writer.


samples/gocdk-blob/blob.ct
```diff
@@ -0,0 +1,35 @@
...
+
+$ gocdk-blob ls file://${ROOTDIR}/bucket
+hw
+
+$ gocdk-blob download file://${ROOTDIR}/bucket hw
+hello world
+
+# Error downloading a bucket that doesn't exist.
+$ gocdk-blob download file://${ROOTDIR}/bucket noexist --> FAIL
+gocdk-blob: Failed to read "noexist": blob (code=NotFound): stat ${ROOTDIR}/bucket/noexist: no such file or directory
```
vangent: The exact error string is not going to be portable.

jba: This is a hard problem. We could easily come up with some sort of approximate match (e.g. look only at the prefix of the line; or allow regexps). But then how would updating work? 

I'm going to omit this test case pending a decent solution.

vangent: SGTM. Can you file an issue?

samples/gocdk-blob/main_test.go
```diff
@@ -0,0 +1,42 @@
...
+	if err := exec.Command("go", "build").Run(); err != nil {
+		t.Fatal(err)
+	}
+	defer os.Remove("gocdk-blob")
+
+	ts, err := cmdtest.Read(".")
+	if err != nil {
+		t.Fatal(err)
+	}
+	ts.Commands["gocdk-blob"] = cmdtest.Program("gocdk-blob")
```
vangent: By executing the binary this way, we lose any coverage data (look at the coverage comment on the PR -- it thinks coverage hasn't changed, when really it has). Is it possible to make the `gocdk-blob` hook call `main` instead of build the binary + execing it?

jba: Almost, but it's not pretty.

I just did it for blob. If you're okay with it, I'll move `execInProcess` to `cmdtest` for reuse.

vangent: Yes, LGTM.

jba: Done, PTAL.

internal/testing/cmdtest/cmdtest_test.go
```diff
@@ -37,6 +39,19 @@ func setup() {
 	}
 }
 
+// echoStdin contains the same code as the main functio of
```
vangent: nit: functio -> function

PR #2474, opened 2019-07-08
internal/gocdk: add support for provisioning docstore

vangent:
Fixes #2474.


internal/cmd/gocdk/internal/static/_assets/resource/docstore/awsdynamodb.tf
```diff
@@ -0,0 +1,16 @@
+# TODO(rvangent): Add comments explaining.
+
+locals {
+  awsdynamodb_url = "dynamodb://${aws_dynamodb_table.table.name}?region=${local.aws_region}&partition_key=Key&sort_key=Key"
```
jba: I don't think you can use the same field for the partition and sort key. I think you can omit the sort key.

internal/cmd/gocdk/resource.go
```diff
@@ -113,6 +113,47 @@ var provisionableTypes = map[string]func(*processContext, string) ([]*static.Act
...
+	},
+	"docstore/azurecosmos": func(pctx *processContext, biomeDir string) ([]*static.Action, error) {
+		reader := bufio.NewReader(pctx.stdin)
+		_, err := prompt.AzureLocationIfNeeded(reader, pctx.stderr, biomeDir)
+		if err != nil {
+			return nil, err
+		}
+		return []*static.Action{
+			static.AddProvider("azurerm"),
+			static.AddProvider("random"),
```
jba: What does "random" mean here?

internal/cmd/gocdk/internal/static/_assets/resource/docstore/awsdynamodb.tf
vangent: OK, done.

internal/cmd/gocdk/resource.go
vangent: It means it will add the `random` provider (if it hasn't already been added yet); it is used to generate a random name for various things (e.g., the AWS table, the Azure Cosmos account, the Firestore database name).

```
  name         = local.gocdk_random_name
```

is a reference to that provider.

Issue #2443, opened 2019-07-01
all: move server-related packages under server/


vangent:
Including:

1. `health`
2. `requestlog`

PR #2476, opened 2019-07-08
all: move health and requestlog packages under server

vangent:
Updates #2443.


zombiezen:
@clausti, heads-up for #2366.

Issue #2291, opened 2019-06-07
internal/cmd/gocdk: always run terraform init during apply


zombiezen:
### Describe the bug

If the user adds a new provider into their biome's Terraform configuration after the first `gocdk apply` and runs `gocdk apply`, then `gocdk apply` will exit with an error.

/cc @clausti @vangent

### To Reproduce

1. `gocdk init myproject && cd myproject`
2. `echo 'provider "random" {}' >> biomes/dev/main.tf && gocdk apply dev` to force a terraform.tfstate file to be created.
4. `echo 'provider "aws" {}' >> biomes/dev/main.tf` to add a new required plugin
5. `gocdk apply dev`

### Expected behavior

`gocdk apply dev` succeeds.

### Actual behavior

`gocdk apply dev` exits with the following error:

```
Plugin reinitialization required. Please run "terraform init".
Reason: Could not satisfy plugin requirements.

Plugins are external binaries that Terraform uses to access and manipulate
resources. The configuration provided requires plugins which can't be located,
don't satisfy the version constraints, or are otherwise incompatible.

1 error(s) occurred:

* provider.aws: no suitable version installed
  version requirements: "(any version)"
  versions installed: none

Terraform automatically discovers provider requirements from your
configuration, including providers used in child modules. To see the
requirements and constraints from each module, run "terraform providers".


Error: error satisfying plugin requirements


Error: apply dev: exit status 1
```

### Version

1a8f982087f6a2a0ae09c2cae31fa48cf497c9bb

### Additional context

The reason we don't run `terraform init` as often is mostly because of the line noise it makes on first run, related to the output control mentioned in #1821. (Originally, I thought that `terraform init` might not be safe to run multiple times, but I later learned this is [explicitly false](https://www.terraform.io/docs/commands/init.html#usage).) We probably want to squelch successful output from `terraform init`.

PR #2477, opened 2019-07-08
internal/gocdk: by default, check that biomes exist

vangent:
Updates #2291.


internal/cmd/gocdk/biomeutil.go
```diff
@@ -41,27 +41,32 @@ func biomesRootDir(moduleRoot string) string {
...
+	if _, err := os.Stat(filepath.Join(dir, biomeConfigFileName)); err != nil {
+		// TODO(light): Wrap error for formatting chain but not unwrap chain.
+		return "", &biomeNotFoundError{moduleRoot: moduleRoot, biome: name, frame: xerrors.Caller(0), detail: err}
+	}
+	return dir, nil
+}
+
+// newBiomeDir returns the path to the named biome, without checking to see
+// if the biome exists.
+func newBiomeDir(moduleRoot, name string) string {
```
zombiezen: The name `newBiomeDir` here seems confusing, as it implies it would create the biome directory.

I see two options:
1) Call `os.MkdirAll` in this function.
2) Make `biomeDir`'s first return the correct path, and just ignore the error where appropriate.

While option 2 is slightly inefficient (1 unneeded syscall), IMO it makes the intent of the caller more clear.

vangent: Did #2.

vangent:
PTAL.

vangent:
@googlebot please recheck

Issue #2481, opened 2019-07-09
pubsub/kafkapubsub: often seems to hang on Travis


vangent:
Or maybe just take too long? I've seen a lot of builds end on the Kafka tests with the "no response received for 10 minutes" message from Travis.

vangent:
```
******************************
* Running Go tests for module: pubsub/kafkapubsub
******************************
go: finding github.com/pierrec/lz4 v2.2.3+incompatible
go: finding github.com/Shopify/sarama v1.22.1
go: finding github.com/DataDog/zstd v1.4.0
go: finding github.com/pierrec/lz4 v0.0.0-20190327172049-315a67e90e41
go: finding github.com/DataDog/zstd v1.3.6-0.20190409195224-796139022798
go: finding golang.org/x/crypto v0.0.0-20190404164418-38d8ce5564a5
go: finding github.com/schollz/progressbar/v2 v2.12.1
go: finding code.cloudfoundry.org/bytefmt v0.0.0-20180906201452-2aa6f33b730c
go: finding github.com/pkg/profile v1.2.1
go: finding golang.org/x/sys v0.0.0-20190403152447-81d4e9dc473e
go: finding github.com/mitchellh/colorstring v0.0.0-20190213212951-d06e56a500db
go: downloading github.com/Shopify/sarama v1.22.1
go: extracting github.com/Shopify/sarama v1.22.1
go: downloading github.com/rcrowley/go-metrics v0.0.0-20181016184325-3113b8401b8a
go: downloading github.com/davecgh/go-spew v1.1.1
go: downloading github.com/eapache/go-resiliency v1.1.0
go: downloading github.com/DataDog/zstd v1.4.0
go: downloading github.com/eapache/queue v1.1.0
go: downloading github.com/pierrec/lz4 v2.2.3+incompatible
go: downloading github.com/eapache/go-xerial-snappy v0.0.0-20180814174437-776d5712da21
go: extracting github.com/davecgh/go-spew v1.1.1
go: extracting github.com/rcrowley/go-metrics v0.0.0-20181016184325-3113b8401b8a
go: extracting github.com/eapache/go-resiliency v1.1.0
go: extracting github.com/eapache/go-xerial-snappy v0.0.0-20180814174437-776d5712da21
go: extracting github.com/eapache/queue v1.1.0
go: extracting github.com/DataDog/zstd v1.4.0
go: extracting github.com/pierrec/lz4 v2.2.3+incompatible
=== RUN   TestConformance
[...]

PR #2482, opened 2019-07-09
internal/travis: various runchecks improvements

vangent:
Updates #2481.

This refactors `runchecks.sh` to:

1. run Go tests for all modules, then `go mod tidy` for all modules, then `wire diff`. I think this is clearer to read both in the bash code and in the output.

2. build the arguments for `go test` in a bash array so that we can just call it from one place (previously we had one call for coverage, another for non-coverage, with mostly the same command line).

3. add `-v` to `go test` for `pubsub/kafkapubsub`, which seems to time out a lot and needs more investigation.

4. built `test-summary` once instead of multiple `go run` calls.

5. got rid of the use of "pushd/popd" in a couple of places.

6. fixed a `shellcheck` warning about unescaped `\n` in `prereleasechecks.sh`.

### DIFF
--- a/internal/testing/runchecks.sh
+++ b/internal/testing/runchecks.sh
@@ -76,8 +76,9 @@ fi
 result=0
 rootdir="$(pwd)"
 
-# Run Go tests.
-(cd internal/testing/test-summary && go build) || result=1
+# Build the test-summary app, which is used inside the loop to summarize results
+# from Go tests.
+(cd internal/testing/test-summary && go build)
 while read -r path || [[ -n "$path" ]]; do
   echo
   echo "******************************"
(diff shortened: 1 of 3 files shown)


internal/testing/runchecks.sh
```diff
@@ -74,47 +74,47 @@ if [[ "${TRAVIS_OS_NAME:-}" == "linux" ]]; then
 fi
 
 result=0
-
 rootdir="$(pwd)"
+
+# Run Go tests.
```
eliben: Move this comment to the loop or remove entirely. Add a separate comment to the test-summary line to explain what it's doing

vangent: Done.

vangent: Also removed the `|| result=1` for this line; if test-summary doesn't build, we should just exit.

vangent:
@googlebot please rescan

PR #2484, opened 2019-07-09
pubsub/tests: ensure tests fail faster by using a ctx with timeout

vangent:
Updates #2481.

This PR makes sure that the `ctx` we pass to `Receive` in tests always has a 30 second timeout.

It also makes sure that `kafkapubsub`'s `TestMultiplePartionsWithRebalancing` doesn't hang when the `ctx` times out.

### DIFF
--- a/pubsub/kafkapubsub/kafka_test.go
+++ b/pubsub/kafkapubsub/kafka_test.go
@@ -385,7 +385,6 @@ func TestMultiplePartionsWithRebalancing(t *testing.T) {
 	}
 	defer func() {
 		if err := sub2.Shutdown(ctx); err != nil {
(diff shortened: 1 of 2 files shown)


pubsub/kafkapubsub/kafka_test.go
```diff
@@ -374,6 +385,7 @@ func TestMultiplePartionsWithRebalancing(t *testing.T) {
 	}
 	defer func() {
 		if err := sub2.Shutdown(ctx); err != nil {
+			fmt.Printf("shutting down sub2\n")
```
jba: Do you want this? Or left over from debugging?

vangent: Left over; removed.

PR #2485, opened 2019-07-09
samples/order: add output image links

jba:


jba:
To get this to work with `fileblob`, and `file:` URLs, I had to parse the URL myself. We could consider adding `base_url` and `secret_key_path` query parameters to `fileblob` URLs to make this less hacky.

eliben:
Could we just grab the image from the bucket and display it to the user in the HTML page somehow? This would save us from having to deal with signed URLs here

jba:
> Could we just grab the image from the bucket and display it to the user in the HTML page somehow?

We could, but I think using URLs is what you'd do on a real site. I wish they didn't have to be signed.

I'm open to changing it so that the links are all to the frontend, which reads the file and then displays it. That's what I have to do for fileblob anyway, as you can see in `showImage`. @vangent?

vangent:
> I'm open to changing it so that the links are all to the frontend, which reads the file and then displays it. That's what I have to do for fileblob anyway, as you can see in `showImage`. @vangent?


Yes, let's just do that and leave off the signing. It makes it so that all output images are visible to anyone, but that's OK.

jba:
PTAL

PR #2498, opened 2019-07-10
sample/runtimevar: add test

jba:
Also, stop the "watch" command when the context is canceled.

This test was complicated by the fact that the "watch" subcommand
needs to run concurrently with other test actions.

### DIFF
--- a/samples/gocdk-runtimevar/runtimevar.ct
+++ b/samples/gocdk-runtimevar/runtimevar.ct
@@ -14,10 +14,6 @@
 
 # Tests of the gocdk-runtimevar program.
 
-# We can't test watch in this test framework, because it runs indefinitely in the
-# foreground. We would need to add some job control (background, kill). Probably
-# not worth it.
-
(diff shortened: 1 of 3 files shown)


samples/gocdk-runtimevar/runtimevar.ct
```diff
@@ -0,0 +1,39 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Tests of the gocdk-runtimevar program.
+
+# We can't test watch in this test framework, because it runs indefinitely in the
```
vangent: Stale?

jba: Done

jba:
PTAL

PR #2499, opened 2019-07-10
samples/order: fix on windows

jba:
WIP do not review


jba:
This is ready for review.

samples/gocdk-runtimevar/runtimevar.ct
```diff
@@ -22,12 +22,15 @@ $ echof fvar contents
 $ gocdk-runtimevar cat file:///${ROOTDIR}/fvar?decoder=string
 (string) contents
 
+$ echo ROOTDIR is '${ROOTDIR}'
```
vangent: This can be removed?

samples/gocdk-blob/blob.ct
```diff
@@ -18,19 +18,19 @@
 $ mkdir bucket
 
 # List an empty bucket: valid, but no output.
-$ gocdk-blob ls file://${ROOTDIR}/bucket
+$ gocdk-blob ls file://${SLASHDIR}/bucket
```
vangent: This took me a bit to follow. Might it be more clear to use `ROOTDIR_URL` as the environment variable that is added in `Setup`, where the value is the full URL? Then this line, for example, would be:

```
$ gocdk-blob ls ${ROOTDIR_URL}/bucket
```

Up to you.

jba: done

samples/gocdk-runtimevar/runtimevar.ct
jba: done

PR #2501, opened 2019-07-10
samples/gocdk-docstore: add test

jba:


samples/gocdk-docstore/main.go
```diff
@@ -39,21 +38,23 @@ import (
 const helpSuffix = `
 
   See https://gocloud.dev/concepts/urls/ for more background on
-  Go CDK URLs, and sub-packages under gocloud.dev/docstore
-  (https://godoc.org/gocloud.dev/docstore#pkg-subdirectories)
+  Go CDK URLs, and sub-packages under gocloud.dev/internal/docstore
```
vangent: Why did you add `internal` here?

samples/gocdk-docstore/main_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Requires mongo to be running. Run docstore/mongodocstore/localmongo.sh.
+
+var update = flag.Bool("update", false, "replace test file contents with output")
+
+func Test(t *testing.T) {
+	if !setup.HasDockerTestEnvironment() {
+		t.Skip("do not have docker test environment")
+	}
+	ts, err := cmdtest.Read(".")
```
vangent: Why not use `memdocstore`?

jba: Originally I couldn't because  I was exec'ing the binary on every command, and there was no persistence.

I added persistence, there is still a problem: each command in the CLI binary closes the collection when it's done (as it should). But once you've closed a collection, you can't reopen a collection with the same name—it's permanently closed. `memdocstore` is the only driver with that property.

The root cause is that the mapping from names to collections is in the `URLOpener`, which doesn't know when a collection is closed. So when you open the same URL again, you get the same `*collection`, which is closed, so any attempt to use it returns an error. (See https://github.com/google/go-cloud/blob/master/docstore/memdocstore/urls.go#L82.)

I believe the right fix is to move the map from name to collection into a package-level variable in `memdocstore`. Then opening with a URL works like calling `OpenCollection` (both require a name), and `*collection.Close` can delete itself from the map.

Let me know if you're okay with that and I'll do it. I know it goes against a principle we have about global state, but it's effectively the same global state we have now, since the default `URLOpener` is registered in some global registry.

vangent: Hmm, interesting. I don't think moving the map to the package-level is the right thing to do; the default `URLOpener` is just that, a default, and someone could be using another one, or multiple, and they should stay independent.

Perhaps the `URLOpener` could set something (either a cleanup func, or a reference to itself) on the collection, so that it is able to remove itself from the map during `Close`?

jba: I think the generality of the URL design is leading you in a wrong direction. 

For no other provider is the binding from name to collection at the whim of the URLOpener: collections are intrinsically named. Even if you wrote another URLOpener for `gcpfirestore`, it couldn't do anything about the names of Firestore collections. (It might support URLs that use different names, in which case it would have state, but it would ultimately need to use the real names.)

A collection shouldn't have to know anything about a `URLOpener`. I guess we could do it as a generic "call this function when you get closed" option, but that still feels weird.

jba: Meh, I guess it's not that important to me. Added the generic callback. PTAL.

vangent:
You will need to add a `BREAKING_CHANGE_OK` commit:

```
Incompatible changes:
- Options: old is comparable, new is not
```

PR #2502, opened 2019-07-10
internal/website: add automatic table of contents

zombiezen:
This is enabled per-page using `toc: true` in the front matter. It will float a sidebar to the right on desktop and show the list inline with content after the title on tablet and smaller screens.

### DIFF
--- a/internal/website/static/css/style.css
+++ b/internal/website/static/css/style.css
@@ -75,7 +75,6 @@ body {
 }
 .MainContent-bounds {
   box-sizing: border-box;
-  grid-area: main;
   margin: 0;
   padding: 0;
   max-width: 50rem;
@@ -125,6 +124,39 @@ body {
 .Sidenav-pageLink:hover,
 .Sidenav-pageLink:active {
   text-decoration: underline;
+}
+.PageTOC {
+  border-left: 2px solid #dbd9d6;
+  float: right;
+  margin: 0.25rem 0 1rem 2rem;
+  max-width: 25%;
+  min-width: 15%;
+  padding: 0.5rem 0 0.5rem 1rem;
+}
+@media (max-width: 1024px) {
+  .PageTOC {
+    border: none;
+    float: none;
+    margin: 1.5rem 0 0.5rem;
+    max-width: none;
+    min-width: none;
+    padding: 0;
+  }
+}
+.PageTOC-heading {
+  font: 700 1rem/1.25 'Work Sans', 'Roboto', sans-serif;
+  margin: 0 0 0.5rem;
+  text-rendering: optimizeLegibility;
+}
+.PageTOC ul {
+  display: block;
+  margin: 0;
+  padding: 0;
+}
+.PageTOC li {
+  display: block;
+  margin: 0.3rem 0;
+  padding: 0;
 }
 .PageFooter {
   background: #555759;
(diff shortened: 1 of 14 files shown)


zombiezen:
Screenshot of this in action:

Screenshot of Secrets page with TOC sidebar

vangent:
@zombiezen is there documentation on how to run `hugo` locally to preview the site? @polinasok was looking for that as well.

internal/website/static/css/style.css
```diff
@@ -126,6 +125,39 @@ body {
...
+    border: none;
+    float: none;
+    margin: 1.5rem 0 0.5rem;
+    max-width: none;
+    min-width: none;
+    padding: 0;
+  }
+}
+.PageTOC-heading {
+  font: 700 1rem/1.25 'Work Sans', 'Roboto', sans-serif;
```
vangent: WDYT about making the font smaller? Maybe everywhere (not just TOC), but that's a separate question :-). At least in the screenshot you gave, the TOC is too long to fit on a page and wraps strangely, the font looks too big for me.

zombiezen:
> @zombiezen is there documentation on how to run `hugo` locally to preview the site? @polinasok was looking for that as well.

https://github.com/google/go-cloud/blob/master/internal/website/README.md#local-development

internal/website/static/css/style.css
zombiezen: At any legible font size, the page TOC is going to spill off the page for any of our larger (read as: opening) how-to guides, especially on smaller screens. I put in a fair amount of care to make sure the layout doesn't break when this happens.

If you want the font smaller, I'd like to do that as a separate global change. I think some of the issue is coming from the fact that the top header pushes content below the fold.

vangent:
> @vangent
> 
> > @zombiezen is there documentation on how to run `hugo` locally to preview the site? @polinasok was looking for that as well.
> 
> [/internal/website/README.md@`master`#local-development](https://github.com/google/go-cloud/blob/master/internal/website/README.md#local-development) [](/google/go-cloud/blob/HEAD@{2019-07-10T18:13:30Z}/internal/website/README.md#local-development)

Excellent, thanks!

internal/website/static/css/style.css
zombiezen: Opened #2504 for the font size change.

vangent:
> Excellent, thanks!

Hmm, I pulled this PR and ran `hugo server -D`, but I don't see any TOCs on the pages...?

vangent:
> > Excellent, thanks!
> 
> Hmm, I pulled this PR and ran `hugo server -D`, but I don't see any TOCs on the pages...?

Oh never mind, I wasn't looking at the right pages.

zombiezen:
@vangent Can you confirm that `git rev-parse HEAD` is 3b4c96fbf229e7487fec0800b13ee93682bf687e? And it's not on all pages, just how-to leaf pages (like `/howto/pubsub/subscribe/`).

internal/website/static/css/style.css
polinasok: + 1 to size change for the content. Would TOC size need to be a separate change as well? Seems like it is visually independent of content and can be done here. Also, I think it is important to preserve the nesting structure of the headers.

zombiezen: Done.

zombiezen:
New screenshot (also with a slightly larger window):

Screenshot of secrets how-to guide after changes

eliben:
How does it look on mobile?

zombiezen:
Mobile:

Screenshot of top of secrets page on mobile

Issue #2376, opened 2019-06-18
internal/cmd/gocdk/internal/launcher: reconsider launcher API


zombiezen:
(Follow-on from #2046)

The `launcher` package API should probably have some more conscious design applied to it, but the exact requirements are unclear (e.g. support for exec'ing a subprocess, whether it should be structured as a portable type, whether we want external folks to import the package). FWIW, I think we could likely implement each of the launchers as simple functions, but keep the `LaunchInput` type for sake of consistency.

zombiezen:
Decision from in-person meeting: Structure is probably fine as-is, pending results of a draft PR to add an "exec" launcher. I will write this up.

zombiezen:
Draft PR out for review and the structure is fine &mdash; adding an `exec` launcher did not require changes to the `Launcher` interface. Closing this out.

PR #2503, opened 2019-07-10
internal/cmd/gocdk/internal/launcher: create launcher that calls to a subprocess

zombiezen:
This is primarily a proof-of-concept. The launch input is sent as stdin to the subprocess and the subprocess must print out a valid URL on success. The process to execute is given as the `argv` key in launch specifier, so this could also use `go run` or the like to run an arbitrary Go "script".

As an example, the following script implements the same functionality as `launcher.Local`:

  ```bash
  #!/bin/bash
  set -euo pipefail

  log() {
    echo "$@" 1>&2
  }

  # Save stdin to $launch_input.
  launch_input="$(cat)"
  log "Got launch input: $launch_input"

  # Convert environment variables into Docker argument list.
  mapfile -t docker_env < <(echo "$launch_input" | jq -r '.env[]')
  docker_env_args=()
  for e in "${docker_env[@]}"; do
    docker_env_args+=( "-e" "$e" )
  done
  docker_env_args+=( "-e" "PORT=8080" )

  # Start Docker image.
  docker_image="$(echo "$launch_input" | jq -r '.docker_image')"
  host_port="$(echo "$launch_input" | jq -r '.specifier.host_port')"
  container_id="$(docker run -d -p "$host_port":8080 "${docker_env_args[@]}" "$docker_image")"
  log "Container running, stop it with docker stop $container_id"

  # Print URL back to CDK.
  echo "http://localhost:$host_port/"
  ```

Updates #2376

### DIFF
--- a/internal/cmd/gocdk/internal/launcher/exec.go
+++ b/internal/cmd/gocdk/internal/launcher/exec.go
@@ -1,0 +1,62 @@
+// Copyright 2019 The Go Cloud Development Kit Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package launcher
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"io"
+	"net/url"
+	"os/exec"
+
+	"golang.org/x/xerrors"
+)
+
+// Exec runs an arbitrary subprocess to launch.
+type Exec struct {
+	// Stderr is where the subprocess's stderr will be written to.
+	Stderr io.Writer
+
+	// Dir is the directory the subprocess should be run from. For now, this
+	// should always be the module root.
+	Dir string
+}
+
+// Launch implements Launcher.Launch.
+func (e *Exec) Launch(ctx context.Context, input *Input) (*url.URL, error) {
+	argv := specifierStringArrayValue(input.Specifier, "argv")
+	if len(argv) == 0 {
+		return nil, xerrors.Errorf("exec launch: specifier argv must be an array of strings with at least one element")
+	}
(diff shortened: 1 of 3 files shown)


clausti:
@zombiezen was this for review or just POC to look at?

zombiezen:
Proof of concept. IMO a more full solution must consider how to have the custom launcher subprocess feed questions into `gocdk biome add`, but that would require more thoughtful design work.

vangent:
> Proof of concept. IMO a full solution must consider how to have the custom launcher subprocess feed questions into `gocdk biome add`, but that would require more thoughtful design work.

Thanks Ross.

It would be interesting to see:
1. A "null" launcher that just echoes the args it is passed.
2. A "local" launcher that does the same thing as the existing in-code local launcher, but via a shell script.
3. Same for Cloud Run and ECS launchers.

Possible?

zombiezen:
> 1. A "null" launcher that just echoes the args it is passed.

```bash
#!/bin/bash
cat 1>&2
exit 1
```

> 2. A "local" launcher that does the same thing as the existing in-code local launcher, but via a shell script.

See the first comment in the PR.

> 3. Same for Cloud Run and ECS launchers.
> 
> Possible?

Possible! I'm not terribly keen on reimplementing the string manipulation code in bash, so if you're looking for more answers than a simple feasibility check, then I'm going to push back. All of the information is available in the subprocess to implement it.

vangent:
> See the first comment in the PR.

Ah! I missed that; cool.

> I'm not terribly keen on reimplementing the string manipulation code in bash

It is true, there does seem to be quite a lot of string manipulation going on in the launchers.

Is it manipulation of the launcher inputs (e.g., the image name, etc.)? If so, perhaps we could provide the parsed values as additional inputs to make it easier for individual launchers (e.g., the image name parts split by `:`).

Or is it manipulation of strings that are generated as part of the launch process itself? If so, that's surprising; I would expect that the providers should make it relatively easy for a user on a command line to go through a similar set of steps using CLI(s) to deploy a Docker image.

zombiezen:
It's a little of both: there's extraction of individual elements of a Docker image reference, and then there's launcher-specific manipulations to build up `docker push` invocations and requests to the orchestration system (and then finally to generate the URL).

I haven't been terribly surprised at the need for this, as most deploy scripts would likely just hard-code a bunch more of these parameters to be right for their organization, whereas our tools don't have that luxury. They're relatively easy for a human to copy and paste or generate, but hard for a tool like ours that has to deal with arbitrary input and make decisions from it.

PR #2507, opened 2019-07-10
docstore: perform optimistic locking when the data tell us to

### LINKED ISSUE #2308: docstore: support structs without a revision field
1. If I have a struct with a revision field, calling `Create` on it will fail. I think we should allow that since the user may want to use the same struct for other writes and want the revision check.
2. If I query a collection and get them into struct type without revision field, dynamo and mongo fails, I think this is a bug.

shantuo:
Fixes #2308


docstore/driver/document.go
```diff
@@ -182,3 +182,15 @@ func (d Document) Decode(dec Decoder) error {
...
+
+// RevisionOn returns whether or not we should perform optimistic locking for
+// d. It could be either a map, or a struct with a revision field named after
+// revField.
+func (d Document) RevisionOn(revField string) bool {
+	switch d.Origin.(type) {
+	case map[string]interface{}:
+		return true
+	default:
+		return d.fields.MatchFold(revField) != nil
```
jba: Why not an exact match? Don't we do everything with an exact match now (except for mongo when LowercaseFields = true)?

docstore/driver/document.go
```diff
@@ -182,3 +182,15 @@ func (d Document) Decode(dec Decoder) error {
 	}
 	return decodeStruct(d.s, dec)
 }
+
+// RevisionOn returns whether or not we should perform optimistic locking for
+// d. It could be either a map, or a struct with a revision field named after
+// revField.
+func (d Document) RevisionOn(revField string) bool {
+	switch d.Origin.(type) {
```
jba: Use `d.m != nil`. (See the other functions in this file.)

docstore/driver/document.go
```diff
@@ -182,3 +182,15 @@ func (d Document) Decode(dec Decoder) error {
...
+	case map[string]interface{}:
+		return true
```
jba: I thought we decided that in all cases, we only write the revision when there is a revision field in the document?

docstore/driver/document.go
```diff
@@ -182,3 +182,15 @@ func (d Document) Decode(dec Decoder) error {
 	}
 	return decodeStruct(d.s, dec)
 }
+
+// RevisionOn returns whether or not we should perform optimistic locking for
```
jba: Specifically, this tells you whether you should write a revision value when you write a document. Let's make that clear.

docstore/awsdynamodb/dynamo.go
```diff
@@ -349,9 +349,12 @@ func (c *collection) newPut(a *driver.Action, opts *driver.RunActionsOptions) (*
 		// It doesn't make sense to generate a random sort key.
 		return nil, fmt.Errorf("missing sort key %q", c.sortKey)
 	}
-	rev := driver.UniqueString()
-	if av.M[c.opts.RevisionField], err = encodeValue(rev); err != nil {
-		return nil, err
+	var rev string
+	if a.Doc.RevisionOn(c.RevisionField()) {
```
jba: Why `c.RevisionField()` here but `c.opts.RevisionField` just below?

docstore/drivertest/drivertest.go
```diff
@@ -1674,6 +1675,33 @@ func testUnorderedActions(t *testing.T, coll *ds.Collection, revField string) {
 	}
 }
 
+func testActionsOnStructWithoutRevision(t *testing.T, ctx context.Context, h Harness, coll *ds.Collection) {
```
jba: You don't need `h` so use `withCollection`.

docstore/memdocstore/codec.go
```diff
@@ -102,7 +102,7 @@ func decodeDoc(m map[string]interface{}, ddoc driver.Document, fps [][]string, r
 	} else {
 		// Make a document to decode from that has only the field paths and the revision field.
 		// (We don't need the key field because ddoc must already have it.)
-		m2 = map[string]interface{}{revField: m[revField]}
+		m2 = map[string]interface{}{}
```
jba: Change above comment to say "and possibly the revision field".

docstore/driver/document.go
jba: Please write a test to catch this.

docstore/memdocstore/codec.go
```diff
@@ -115,6 +115,9 @@ func decodeDoc(m map[string]interface{}, ddoc driver.Document, fps [][]string, r
 				return err
 			}
 		}
+		if ddoc.RevisionOn(revField) {
```
jba: I thought we said we're going to require the revision field explicitly?

docstore/driver/document.go
```diff
@@ -182,3 +182,14 @@ func (d Document) Decode(dec Decoder) error {
...
+
+// RevisionOn returns whether or not we should write a revision value when we
+// write a document. It could be either a map or a struct with a revision key or
+// field named after revField.
+func (d Document) RevisionOn(revField string) bool {
+	if d.m != nil {
+		_, ok := d.m[revField]
+		return ok
+	}
+	return d.fields.MatchFold(revField) != nil
```
jba: This will break if you weren't in lowercase mode had two fields, `Revision` and `revision`. (You probably shouldn't do that, but we do say we'll match everything exactly except for mongo lowercase mode, so we should do that.) 

I think this should do an exact match, and maybe in the mongo driver, when you call it, you lowercase the argument to it first. I think `mongodocstore.collection.revisionField` is the right version of the field name to use.

docstore/doc.go
```diff
@@ -186,23 +186,27 @@
...
-// docstore will compare the revision of the stored document to that of the
-// given document and return an error with code FailedPrecondition on mismatch.
-// (See https://gocloud.dev/gcerrors for information about error codes.)
-// If you call Get to retrieve a document with a revision, then later perform a
-// write action with that same document, it will fail if the document was changed
-// since the Get. Create, Put, Replace and Update all populate the revision
-// field of the argument document with the new revision value.
+// Docstore gives every document a revision at creation time and updates that
+// revision whenever the document is modified if and only if the document struct
+// or map has a revision field/key matching the revision field name. If you
```
jba: Remove the word "key" here. We  call them "fields" even if they're map keys. The word "key" is used for the document's unique ID.

jba: Also remove the first "revision" from this line:  "... or map has a field matching the ..."

docstore/gcpfirestore/fs.go
```diff
@@ -554,8 +554,9 @@ func (c *collection) doCommitCall(ctx context.Context, call *commitCall, errs []
 	j := 0
 	for i, wr := range wrs {
 		if _, ok := call.writes[i].Operation.(*pb.Write_Transform); !ok {
-			// Ignore errors. It's fine if the doc doesn't have a revision field.
-			call.actions[j].Doc.SetField(c.opts.RevisionField, wr.UpdateTime)
+			if call.actions[j].Doc.RevisionOn(c.opts.RevisionField) {
+				call.actions[j].Doc.SetField(c.opts.RevisionField, wr.UpdateTime)
```
jba: You shouldn't ignore the error from `SetField` any more.

docstore/memdocstore/mem.go
```diff
@@ -352,10 +355,14 @@ func (c *collection) changeRevision(doc map[string]interface{}) {
 }
 
 func (c *collection) checkRevision(arg driver.Document, current map[string]interface{}) error {
-	if current == nil {
-		return nil // no existing document
+	if current == nil || !arg.RevisionOn(c.opts.RevisionField) {
```
jba: The RevisionOn check is redundant with the check after GetField below.

I'd leave the RevisionOn check, and instead return an error if GetField returns err != nil.

docstore/driver/document.go
shantuo: That won't work, `collection.revisionField` is the right version in the DB, not necessarily the right one for the user-provided document, which is the `d` here.

The easiest way is to have another `HasFieldFold` for such use case, which is what I choose for now. The advantage is that it is easy to implement and less confusion for the user to use.

Another way I've tried is to store the field user passed in and the lower them when we need to (e.g. encoding, writing) but they all end up very confusing both for implementing and using the API with lowercaseFields on.

docstore/awsdynamodb/dynamo.go
```diff
@@ -200,7 +200,7 @@ func (c *collection) batchGet(ctx context.Context, gets []*driver.Action, errs [
 		// We need to add the key fields if the user doesn't include them. The
 		// BatchGet API doesn't return them otherwise.
 		var hasP, hasS bool
-		nbs := []expression.NameBuilder{expression.Name(c.opts.RevisionField)}
+		nbs := []expression.NameBuilder{}
```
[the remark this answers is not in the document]
shantuo: Done.

docstore/doc.go
```diff
@@ -186,23 +186,27 @@
...
-// checks are performed, and changes are forced blindly. Otherwise, whenever
-// you pass a document with a revision to Put, Replace, Update or Delete,
-// docstore will compare the revision of the stored document to that of the
-// given document and return an error with code FailedPrecondition on mismatch.
-// (See https://gocloud.dev/gcerrors for information about error codes.)
-// If you call Get to retrieve a document with a revision, then later perform a
-// write action with that same document, it will fail if the document was changed
-// since the Get. Create, Put, Replace and Update all populate the revision
-// field of the argument document with the new revision value.
+// Docstore gives every document a revision at creation time and updates that
```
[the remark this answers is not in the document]
shantuo: Reworded.

PR #2526, opened 2019-07-11
internal/gocdk: simplify and add comments to blob demo

vangent:
Updates #2305.

This PR simplifies the `demo_blob.go` page a bit, and adds explanatory comments to it.

The simplifications/improvements other than comments are:
1. Changed the `blob` link to point to the `gocloud.dev/howto` page instead of godoc.
2. Removed the "view" page with a dropdown to select a blob to view; it's not really necessary since the "list" page already demonstrates listing blob keys, and the "list" page links directory to "view".
3. Removed TODOs, there are bugs filed for them.


internal/cmd/gocdk/internal/static/_assets/demo/demo_blob.go
```diff
@@ -17,31 +23,32 @@ import (
...
-
-var bucketURL string
-var bucket *blob.Bucket
-var bucketErr error
-
-func init() {
-	ctx := context.Background()
 
+	// Initialize the blob.Bucket using a URL from the environment, defaulting
+	// to an in-memory bucket provider.
```
clausti: Can we add a _n.b_ here that if the user runs `serve`, it will empty the default mem blob? It's obvious-if-you-think-about-it but it might be confusing the first time they observe it.

vangent: Sure, done.

PR #2527, opened 2019-07-11
docs: make runtimevar howto page consistent with the new howto template

### LINKED ISSUE #2525: internal/website: make single-page runtimevar how-to consistent
#2506 for runtimevar
The how-to is already a single page, but doesn't follow the new pattern.

vangent:
Fixes #2525.

Top of page looks like: https://screenshot.googleplex.com/MP5b405o59j.

Note that I named the sub-headings for constructors just `Constructor` instead of repeating the service name -- otherwise they are really long.

Separately, WDYT about always having a `URL` subheader as well?

E.g.,

```
## Provider
### URL
...
### Constructor
...
```

I think it might make the nav-bar more clear, as currently `URL` doesn't appear there.


polinasok:
I had the same thoughts when working on blob:
- It bothered me that service header and constructor subheader repeat the same strings, but Ross pointed out that if you just look at the header in the middle of the page, you no longer see the parent header to remind you which constructor this is for.
- I also noticed that we don't have a section for URLs. But I prefer to do things incrementally, merging and making things consistent with minimal reshuffling first, then experimenting with new formats. In particular, on the background I have been thinking about a way to present his information in a table format that would eliminate the subheaders and make things more compact and easier to contrast/compare.

internal/website/content/howto/runtimevar/_index.md
```diff
@@ -1,53 +1,91 @@
...
 To open a variable stored in [GCP Runtime Configurator][] via a URL, you can use
 the `runtimevar.OpenVariable` function as follows.
 
+[GCP Runtime Configurator]: https://cloud.google.com/deployment-manager/runtime-configurator/
+
 {{< goexample
 "gocloud.dev/runtimevar/gcpruntimeconfig.Example_openVariableFromURL" >}}
 
-### GCP Runtime Configurator Constructor {#rc-ctor}
+#### Constructor {#gcprc-ctor}
```
polinasok: Maybe the compromise would be GCP Constructor, AWS Constructor?

vangent: OK, done.

vangent:
> * It bothered me that service header and constructor subheader repeat the same strings, but Ross pointed out that if you just look at the header in the middle of the page, you no longer see the parent header to remind you which constructor this is for.

OK, fair enough. I used `GCP` and `AWS` as you suggested.

> * I also noticed that we don't have a section for URLs. But I prefer to do things incrementally, merging and making things consistent with minimal reshuffling first, then experimenting with new formats. In particular, on the background I have been thinking about a way to present his information in a table format that would eliminate the subheaders and make things more compact and easier to contrast/compare.

SGTM.

internal/website/content/howto/runtimevar/_index.md
```diff
@@ -116,7 +116,7 @@ open a variable stored in `etcd` via a URL, you can use the
 
 [etcd]: https://etcd.io/
 
-#### Constructor {#etcd-ctor}
+#### Etcd Constructor {#etcd-ctor}
```
polinasok: The spell it with lower case letter even in the beginning of the sentence or in the header:
https://github.com/etcd-io/etcd#etcd-tcp-ports
https://etcd.io/

PR #2533, opened 2019-07-11
samples/order: add genericy Go-brand styling

zombiezen:
Jumbo text, since there's not much page content. Work still remaining is to turn the "order created" response into an actual styled page.

/cc @jba FYI, but I know you're OOO. Hopefully you like it!


samples/order/index.html
```diff
@@ -2,12 +2,26 @@
...
+      <a href="/" class="PageTitle-link"><h1 id="PageTitle">Go CDK Image Conversion Sample</h1></a>
+    </header>
+    <main id="PageContent">
+      <div class="FrontPageActions">
+        <button class="FrontPageActions-button" data-href="/orders/new">Convert an Image</button>
+        <button class="FrontPageActions-button" data-href="/orders">List Conversions</button>
+      </div>
+    </main>
+    <script type="text/javascript">
+      document.addEventListener('click', function(e) {
```
eliben: Nifty. Took me a bit of Googling to figure out what this does :) Can you please add a comment explaining it for posterity?

jba:
Thanks for doing this!

samples/order/index.html
zombiezen: Done.

PR #2537, opened 2019-07-12
internal/testing/cmdtest: support running a subset of test files

jba:
Add RunMatch, which lets the caller provide a regexp to run
only matching files.

I found this invaluable in debugging the gocdk tests.


jba:
That seems reasonable.

PR #2539, opened 2019-07-12
internal/gocdk: add comments to the pubsub demo

vangent:
Updates #2305.

I also changed the "send" page to use `POST` instead of `GET`, and to not carry over the message (so, after sending a message, the "send" form is blank again).


internal/cmd/gocdk/internal/static/_assets/demo/demo_pubsub.go
```diff
@@ -124,30 +129,42 @@ var (
...
+	}
+
+	// POST. Send the message.
+	data.Err = topic.Send(req.Context(), &pubsub.Message{Body: []byte(req.FormValue("msg"))})
+	if data.Err == nil {
+		data.Success = true
 	}
 }
 
+// pubsubSendHandler is the handler for /demo/pubsub/receive.
```
clausti: should this be `pubsubReceiveHandler`?

vangent: Yes; done.

PR #2541, opened 2019-07-12
internal/gocdk: add comments to the docstore demo

vangent:
Updates #2305, last one!

Again, similar to the other PRs, this is mostly adding comments to the docstore demo. I'll leave some comments inline explaining other changes.


internal/cmd/gocdk/demo_test.go
```diff
@@ -244,7 +244,10 @@ func TestAddDemo(t *testing.T) {
 			description: "create: empty key",
 			urlPaths:    []string{"/demo/docstore/edit"},
 			op:          "POST",
-			urlValues:   map[string][]string{"create": {"true"}, "value": {"foo"}},
+			urlValues: map[string][]string{
```
vangent: This is just a formatting change to make the code clearer.

internal/cmd/gocdk/internal/static/_assets/demo/demo_docstore.go
```diff
@@ -135,9 +140,6 @@ const (
       <br/>
       <input type="text" name="value" value="{{ .Value }}">
     </label></p>
-    {{if not .Create}}
-      <p>Revision: {{ .Revision }}</p>
```
vangent: I dropped the revision stuff from the demo, as it adds more complication.

I still needed to keep `DocstoreRevision` in the `MyDocument` struct, because docstore is still adding that field automatically and the `Get` fails if it's not in the struct (Tuo is fixing this behavior, but it's not in yet).

A possible alternative for later is to make the demo include revisioning, which would include grabbing the revision from the document when initializing the "edit" form, putting it in a "hidden" field in the form, and then putting it back on the `MyDocument` before calling `Replace`.

vangent:
@googlebot PTAL

PR #2543, opened 2019-07-12
internal/cmd/gocdk: add ECS launcher support to `biome add`

### LINKED ISSUE #2301: internal/cmd/gocdk: add AWS ECS launcher

zombiezen:
This adds resources to a ECS-launching biome to create an ECS cluster. It's ported from the CloudFormation template that the ECS console applies when the "Create Cluster" UI is used. Sadly, there is no convenient "single API" that creates an ECS cluster like GKE does on GCP. Broadly, this configuration creates a new VPC with an EC2 autoscaling group that will report back to the ECS orchestrator. If this breaks in the future, you can create a new ECS cluster from the AWS console, look at the resulting CloudFormation template, and make the corresponding changes to the Terraform template.

I have manually tested this configuration and it is repeatable and results in a working service.

Fixes #2301

(@clausti FYI)


internal/cmd/gocdk/biome.go
```diff
@@ -142,6 +143,30 @@ func biomeAdd(ctx context.Context, pctx *processContext, biome, launcher string)
...
+		if err != nil {
+			return xerrors.Errorf("biome add: %w", err)
+		}
+		saveActions = append(saveActions,
+			static.AddLocal(prompt.AWSRegionTfLocalName, region),
+			static.AddProvider("aws"),
+			static.AddProvider("random"),
+			&static.Action{
+				SourcePath:   "/launchers/ecs.tf",
+				DestRelPath:  "main.tf",
```
vangent: WDYT about creating this as a new file rather than appending it? It seems big enough.

internal/cmd/gocdk/internal/static/_assets/launchers/ecs.tf
```diff
@@ -0,0 +1,170 @@
+# Create an ECR Docker repository.
+# The name used here will be part of the deployed Docker image name.
+resource "aws_ecr_repository" "default" {
+  name = "{{.ProjectName}}"
+}
+
+# ECS Cluster
+
+resource "aws_ecs_cluster" "default" {
+  name = random_id.cluster_name.hex
```
vangent: I have been using `local.gocdk_random_name` for random string everywhere, so that it's always the same...? It's currently using `gocdk-` as the prefix instead of `{{.ProjectName}}`, but we could fix that independently.

internal/cmd/gocdk/internal/static/_assets/launchers/ecs.tf
```diff
@@ -0,0 +1,170 @@
+# Create an ECR Docker repository.
+# The name used here will be part of the deployed Docker image name.
+resource "aws_ecr_repository" "default" {
+  name = "{{.ProjectName}}"
+}
+
+# ECS Cluster
```
vangent: WDYT about a comment here saying something like "The resources below are needed to create a working ECS cluster. The AWS UI creates them for you automatically, but the `aws` CLI does not." ?

zombiezen:
PTAL

internal/cmd/gocdk/internal/static/_assets/launchers/ecs.tf
zombiezen: It would be nice to have the include the project name instead of `gocdk`, but I agree, that's a separate issue.

internal/cmd/gocdk/internal/static/_assets/launchers/ecs.tf
```diff
@@ -1,17 +1,21 @@
...
 }
 
 # ECS Cluster
 
 resource "aws_ecs_cluster" "default" {
-  name = random_id.cluster_name.hex
+  name = local.gocdk_random_name
 }
 
 resource "random_id" "cluster_name" {
```
vangent: You can delete this resource, it is no longer used.

zombiezen: Done.

PR #2547, opened 2019-07-12
internal/testing/cmdtest: run each file as a subtest

### LINKED ISSUE #2542: internal/cmdtest: consider running testcases as subtests
What do you think about changing `cmdtest` to take a `testing.T`, and running each `testCase` as a sub-test? Then the normal go test command lines would work, the "RunMatch" function could be deleted, tand `go test -v` would also be more useful

For example, right now it takes almost a minute to run the gocdk CLI tests (not sure why, investigating), and there's no output at all until it's completely finished.

vangent:
Fixes #2542.

Before this PR, `cmdtest` runs the entire test suite as a single test, and returns the output in a single error. This is unwieldy for several reasons:

* There's a function `RunMatch` to run a subset of tests, but there's no way to access it from the command line; you have to update the code under test to call it instead of `Run`.

* If there are a lot of tests, it can be quite slow; the `gocdk` CLI tests take almost a minute (separate issue!), and I assumed that they were hung because `go test -v` didn't emit anything for so long.

* If a bunch of tests failed, the output could be unwieldy.

This PR:

* Adds `CompareT` and `UpdateT` functions that are similar to the existing `Compare` and `Update` functions, but are passed a `*testing.T`, run each test file in a separate subtest, and fail the test if there are errors or differences.
* Updates `Run` to take a `*testing.T` and call one of the new functions.
* Deletes `RunMatch`; it is no longer needed because you can use the `go test --run` flag to select tests.
* Updates all existing callers to pass `t` to `Run`.

@jba -- I could also remove the `Verbose` field and figure out a way to call `t.Log` instead, WDYT?


jba:
> I could also remove the `Verbose` field and figure out a way to call `t.Log` instead, WDYT?

SGTM

internal/testing/cmdtest/cmdtest.go
```diff
@@ -269,57 +270,24 @@ func (tf *testFile) addCase(tc *testCase) []string {
 	return suffix
 }
 
-// Run runs Compare or Update, depending on the value of the argument.
-func (ts *TestSuite) Run(update bool) error {
+// Run runs CompareT or UpdateT, depending on the value of update.
```
jba: I think it's OK to remove remove Compare and Update and the *T versions, and just have a single unexported function for each. We don't have a use for them so far, and if we ever do we can add them back easily enough.

internal/testing/cmdtest/cmdtest.go
```diff
@@ -329,6 +297,18 @@ func (ts *TestSuite) compare(files []*testFile) error {
...
+			if s := tf.compare(); s != "" {
+				t.Error(s)
+			}
+		})
+	}
+}
+
 func (tf *testFile) compare() string {
 	if err := tf.execute(); err != nil {
 		return fmt.Sprintf("%v\n", err)
```
jba: nit: I think you can remove the terminating newline here and below, since the output is passed to `t.Error`.

internal/testing/cmdtest/cmdtest.go
```diff
@@ -364,6 +340,23 @@ func (ts *TestSuite) update(files []*testFile) error {
 	return nil
 }
 
+// UpdateT is similar to Update, but runs each test in a subtest. It only
+// fails t if there are errors running the test suite or updating the output
+// files.
+func (ts *TestSuite) UpdateT(t *testing.T) {
+	for _, tf := range ts.files {
+		t.Run(tf.filename, func(t *testing.T) {
```
jba: nit: use `strings.TrimSuffix(tf.filename, ".ct")`.

internal/testing/cmdtest/cmdtest.go
vangent: The negative tests need the version that returns an error string. The best way I could think of to keep that was like this. I guess we could make that version unexported?

internal/testing/cmdtest/cmdtest.go
vangent: Done.

internal/testing/cmdtest/cmdtest.go
vangent: Done.

internal/testing/cmdtest/cmdtest.go
vangent: Done. I just left a separate unexported function `compareReturningError` that's used in the test.

Issue #2548, opened 2019-07-12
godoc: use "driver" instead of "provider" to refer to implementation packages


vangent:

vangent:
I'm done updating everything except for `internal/website` -- @polinasok is going to handle those pages once the churn in re-doing how-tos is over.

polinasok:
Into/Opening updates made in PRs: #2622, #2628, #2633

PR #2550, opened 2019-07-13
docs: update godoc uses of "provider" to "driver" or "service" in top-level .go files

vangent:
Updates #2548.

This updates most of the top-level `.go` files in each portable type.


docstore/doc.go
```diff
@@ -134,7 +134,7 @@
 // Times deserve special mention. Docstore can store and retrieve values of type
 // time.Time, with two caveats. First, the timezone will not be preserved. Second,
 // Docstore guarantees only that time.Time values are represented to millisecond
-// precision. Many providers will do better, but if you need to be sure that times
+// precision. Many drivers will do better, but if you need to be sure that times
```
polinasok: services?

docstore/doc.go
```diff
@@ -231,8 +231,8 @@
 //   iter := coll.Query().Where("Author.Last", "=", "Bulgakov").Limit(3).Get(ctx)
 //
 // You can make multiple Where calls. In some cases, parts of a Where clause may be
-// processed on the client rather than natively by the provider, which may have
-// performance implications for large result sets. See the provider-specific package
+// processed in the driver rather than natively by the backend service, which may have
```
polinasok: I believe "backing service" is more common with cloud/apps.

docstore/doc.go
vangent: Sure, done.

docstore/doc.go
vangent: Done.

PR #2551, opened 2019-07-15
docs: update godoc uses of "provider" to "driver" or "service" in driver[test] packages

vangent:
Updates #2548.

This updates the .go files in each portable type's `driver/` and `drivertest/` sub-packages.


runtimevar/driver/driver.go
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package driver provides the interface for providers of runtimevar.  This serves as a contract
-// of how the runtimevar API uses a provider implementation.
+// Package driver defines an interface that the runtimevar package uses to
```
polinasok: "an interface" in conceptual sense vs "interfaces" or "set of interfaces", referring to the Go interfaces - database/sql uses the latter, should we?

runtimevar/driver/driver.go
```diff
@@ -12,8 +12,8 @@
...
+// interact with the underlying runtime configuration services.
```
polinasok: Similarly to https://golang.org/pkg/database/sql/driver/, suggest the following rewrite (here and in all other APIs' driver.go): 
"Package driver defines interfaces to be implemented by runtime configuration drivers and to be used by package runtimevar to interact with the underlying services. Most code should use package runtimevar."

runtimevar/driver/driver.go
vangent: Which `database/sql`? https://golang.org/pkg/database/sql/ says "Package sql provides a generic interface around SQL (or SQL-like) databases."

The wording in `driver` packages means `interface(s)` referring to Go interfaces, not in a conceptual sense. I used the singular here because I thought that this package only defined a single interface, but it actually defines 2, so I updated it to "a set of interfaces".

runtimevar/driver/driver.go
vangent: I updated them all to be consistent, with a slight rewording of your suggestion. PTAL.

runtimevar/driver/driver.go
polinasok: I was referring to https://golang.org/pkg/database/sql/driver/ since this was a comment for driver, not the main top-level package.

PR #2553, opened 2019-07-15
internal/gocdk: add CLI tests for commands that require stdin

vangent:
When I created these CLI tests, there was no way to run commands that required input from the user. @jba added support for doing so; thus, this PR is now fleshing out the CLI tests with a bunch of commands that require user input.

I made one change to `internal/testing/cmdtest`, which is to have its `echo` and `echof` commands (which write a string to stdout or to a file) replace a literal `\n` in the input with an actual `\n` character. This allows a test to write a file with contents something like "foo\nbar", which you can use as stdin in order to answer two prompts (the first with `foo`, the second with `bar`.

In a bunch of cases, the file piped to stdin is empty; since `echof` adds a `\n` to the input, this simulates the user hitting Enter once, which works for commands that require input but supply a default.

Note: I also changed the default AWS region to `us-west-1` to match our other defaults.

### DIFF
--- a/internal/cmd/gocdk/testdata/biome.ct
+++ b/internal/cmd/gocdk/testdata/biome.ct
@@ -19,21 +19,76 @@ $ cd myproj
 $ gocdk biome list
 dev
 
-$ gocdk biome add mybiome --launcher local
-gocdk: Adding biome "mybiome"...
+# "biome add" prompts for a biome name.
+$ echof stdin
+$ gocdk biome add localbiomeprompt < stdin
+gocdk: Adding biome "localbiomeprompt"...
+
+Please enter a launcher to use (local, cloudrun, ecs)
+  default: local
+Enter a value, or "cancel": gocdk: Success!
+
+# Local launcher; no inputs needed.
+$ gocdk biome add localbiome --launcher local
+gocdk: Adding biome "localbiome"...
 gocdk: Success!
+
+# Cloudrun launcher; requires projectID and service name.
+$ echof stdin myprojectid\n
+$ gocdk biome add cloudrunbiome --launcher cloudrun < stdin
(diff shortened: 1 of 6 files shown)


internal/cmd/gocdk/testdata/biome.ct
```diff
@@ -19,21 +19,76 @@ $ cd myproj
 $ gocdk biome list
 dev
 
-$ gocdk biome add mybiome --launcher local
-gocdk: Adding biome "mybiome"...
+# "biome add" prompts for a biome name.
+$ echof stdin
```
vangent: This is an example of creating a `stdin` file with a single `\n` in it, which simulates the user hitting enter once, accepting the default of `local`.

internal/cmd/gocdk/testdata/resource.ct
```diff
@@ -44,28 +44,154 @@ $ gocdk resource add dev blob/foo --> FAIL
...
+gocdk: Adding "blob/s3blob" to "dev"...
+
+Please enter an AWS region.
+  See https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html for more information.
+  default: us-west-1
+Enter a value, or "cancel": gocdk:   added a Terraform provider "aws" to "main.tf"
+gocdk:   added an output variable "BLOB_BUCKET_URL" to "outputs.tf"
+gocdk:   added a new file "s3blob.tf"
+gocdk: Success!
+
```
vangent: Note that many of the resources below would require prompts if done initially (e.g., `docstore/awsdynamodb` requires an AWS region. However, since we've already asked for AWS region earlier (for `blob/s3blob`), the CLI saves that answer and doesn't ask again.

Issue #2410, opened 2019-06-26
blob/azureblob: support using AAD credentials for default URLOpener


omerlh:
Please use a title starting with the name of the affected package, or \"all\",
followed by a colon, followed by a short summary of the feature request.
Example: `blob/gcsblob: add support for more blobbing`.

### Is your feature request related to a problem? Please describe.
Azure storage now supports authentication using [Azure Active Directory](https://docs.microsoft.com/en-us/azure/storage/common/storage-auth-aad) user, which is easier and more secure compared to storage account key or SAS tokens.

### Describe the solution you'd like

Be able to use active directory credentials instead of SAS or access key to login to blob.

vangent:
I believe this should be possible already, using [`azblob.NewTokenCredential`](https://godoc.org/github.com/Azure/azure-storage-blob-go/azblob#NewTokenCredential). The `azureblob` documentation is hard to follow, but something like this:

```
import "github.com/Azure/azure-storage-blob-go/azblob"
import "gocloud.dev/blob/azureblob"

...

// First, get an oauth2 token from AAD.
// There are examples here:
// https://docs.microsoft.com/en-us/go/azure/azure-sdk-go-auth-aad-samples.

// Use it to create an azblob.Credentials, which can be used to make a
// pipeline for opening a bucket.
cred := azblob.NewTokenCredential(token, nil /* or add a refresher */)
pipeline := azureblob.NewPipeline(cred, azblob.PipelineOptions{})
bucket, err := azureblob.OpenBucket(ctx, pipeline, accountName, containerName, new(azureblob.Options)) 

// ... use your bucket!
```

I don't think it's feasible to support this kind of authentication using an `URLOpener`. Please advise if you think otherwise.

omerlh:
Well, it's relying on environment variables today for authentication (which is kinda not the best practices anyway). It's possible to add AAD authentication based on env var - maybe even with Azure managed identity.

vangent:
OK. I'm going to retitle this issue to include "in URL openers" since it is supported using constructors, and put it to Unplanned.

vangent:
Also: note that you can create a custom [`URLOpener`](https://godoc.org/gocloud.dev/blob/azureblob#URLOpener) and set the `Pipeline` field after you've authenticated, and you can use that `URLOpener` to open URLs. The environment variable approach is used in the default URLOpener.

dadgar:
[...]

PR #2554, opened 2019-07-15
blob/azureblob: Azure Blob example using AAD credentials

dadgar:
Requested by @vangent: https://github.com/google/go-cloud/issues/2410#issuecomment-511062585


dadgar:
I signed it!

blob/azureblob/example_test.go
```diff
@@ -109,3 +111,56 @@ func Example_openBucketFromURL() {
...
+		clientSecret = "456"
+		tenantID     = "789"
+
+		// The storage container to access.
+		containerName = "my-container"
+	)
+
+	// Get an Oauth2 token for the account for use with Azure Storage
+	ccc := auth.NewClientCredentialsConfig(clientID, clientSecret, tenantID)
+	ccc.Resource = azure.PublicCloud.ResourceIdentifiers.Storage
```
vangent: This constant doesn't appear to exist in the version we're using. Maybe just replace it with the constant string? (`https://storage.azure.com/`).

blob/azureblob/example_test.go
```diff
@@ -109,3 +111,56 @@ func Example_openBucketFromURL() {
 	}
 	defer bucket.Close()
 }
+
+func ExampleOpenBucket_usingAADCredentials() {
+	const (
+		// Your Azure Storage Account.
+		accountName = azureblob.AccountName("my-account")
+
+		// Your Azure AAD Service Principal with access to the storage account
```
vangent: Nit: Please end comments with a period, here and below.

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences.

blob/azureblob/example_test.go
```diff
@@ -109,3 +111,56 @@ func Example_openBucketFromURL() {
...
+	// Note that we're not supplying azureblob.Options.Credential, so SignedURL
+	// won't work. To use SignedURL, you need a real credential (see the other
+	// example).
+	ctx := context.Background()
+	b, err := azureblob.OpenBucket(ctx, pipeline, accountName, containerName, new(azureblob.Options))
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer b.Close()
+
```
vangent: Nit: I think you can leave the `ReadAll` part off; there are plenty of other examples showing how to use a `blob.Bucket`.

go.mod
```diff
@@ -24,10 +24,12 @@ require (
 	github.com/Azure/azure-sdk-for-go v30.0.0+incompatible
 	github.com/Azure/azure-service-bus-go v0.8.0
 	github.com/Azure/azure-storage-blob-go v0.6.0
-	github.com/Azure/go-autorest v12.0.0+incompatible
+	github.com/Azure/go-autorest/autorest v0.5.0
```
vangent: Please revert all of the `go.mod` and `go.sum` changes, they shouldn't be needed.

dadgar: The changes are to be able to reference `azure.PublicCloud.ResourceIdentifiers.Storage`. Will revert if you want

vangent: Yes, please revert; we are doing a release soon and don't want to upgrade dependencies right now. Thanks!

dadgar:
@vangent Thanks for the review. Addressed your comments, PTAL.

vangent:
@googlebot ping

PR #2557, opened 2019-07-15
docs: update godoc uses of "provider" to "driver" or "service" in top-level documentation

vangent:
Updates #2548.

This updates two top-level doc files; I pulled it out separately since they are higher visibility.


doc.go
```diff
@@ -21,15 +21,15 @@ It does this by providing stable, idiomatic interfaces for common uses like
 storage and databases. Think `database/sql` for cloud products.
 
 At the core of the Go CDK are common "portable types" implemented by cloud
-providers. For example, objects of the blob.Bucket portable type can be created
-using gcsblob.OpenBucket, s3blob.OpenBucket, or any other provider. Then, the
-blob.Bucket can be used throughout your application without worrying about
-the underlying implementation.
+services. For example, objects of the blob.Bucket portable type can be created
```
polinasok: "portable types". These are service-independent abstractions for interacting with any supported cloud services that have a corresponding driver implementation.
OR
"portable types" implemented for supported cloud services by service-dependent drivers.

doc.go
```diff
@@ -21,15 +21,15 @@ It does this by providing stable, idiomatic interfaces for common uses like
...
+using gcsblob.OpenBucket, s3blob.OpenBucket, or any other service with a
+Go CDK driver implementation. Then, the blob.Bucket can be used throughout
```
polinasok: And with the change above, this can be simplified to just "or any other driver".

doc.go
vangent: Reworded a bit, PTAL.

doc.go
vangent: Done.

doc.go
```diff
@@ -20,11 +20,12 @@ seamlessly deploy cloud applications on any combination of cloud providers.
 It does this by providing stable, idiomatic interfaces for common uses like
 storage and databases. Think `database/sql` for cloud products.
 
-At the core of the Go CDK are common "portable types" implemented by cloud
-services. For example, objects of the blob.Bucket portable type can be created
-using gcsblob.OpenBucket, s3blob.OpenBucket, or any other service with a
-Go CDK driver implementation. Then, the blob.Bucket can be used throughout
-your application without worrying about the underlying implementation.
+At the core of the Go CDK are common "portable types", implemented on top of
+service-specific drivers for supported supported cloud services. For example,
```
polinasok: remove 2nd "supported"

vangent: Oops, done.

PR #2558, opened 2019-07-15
docs: update godoc uses of "provider" to "driver" or "service" in a bunch of misc files

vangent:
Updates #2548.

This updates all of the remaining uses that I think should be updated.


blob/example_test.go
```diff
@@ -148,7 +148,7 @@ func Example() {
 
 	// We now have a *blob.Bucket! We can write our application using the
 	// *blob.Bucket type, and have the freedom to change the initialization code
-	// above to choose a different provider later.
+	// above to choose a different driver later.
```
polinasok: service-specific driver
(since this is an educational example, it might help to be more descriptive)

internal/testing/prereleasechecks.sh
```diff
@@ -16,7 +16,7 @@
 # This script runs expensive checks that we don't normally run on Travis, but
 # that should run periodically, before each release.
 # For example, tests that can't use record/replay, so must be performed live
-# against the service provider.
+# against the backend service.
```
polinasok: backing service?

vangent: Done.

blob/example_test.go
vangent: Done.

PR #2559, opened 2019-07-15
docstore: document decoder does case-sensitive matching

### LINKED ISSUE #2552: docstore: document why we chose case-sensitive field matching
`encoding/json` uses case-insensitive field matching; we chose to be base-sensitive. We should clearly document that this is the case, and explain why.

shantuo:
The change clarifies the order in which docstore's decoder find for a struct field.

Fixes #2552


docstore/doc.go
```diff
@@ -101,13 +101,12 @@
 // documents. Using structs is preferred because it enforces some structure on your
 // data.
 //
-// Docstore mimics the encoding/json package in its treatment of
-// structs: by default, a struct's exported fields are the fields of the document.
-// You can alter this default mapping by using a struct tag beginning with
-// "docstore:". Docstore struct tags support renaming, omitting fields
+// Docstore treats a struct's exported fields as the fields of the document by
```
vangent: Nit: I think "By default, docstore treats...." reads better.

vangent:
The test failure looks like a flake that should be fixed (or at least, made less flaky) by #2555.

docstore/doc.go
shantuo: Done.

docstore/doc.go
```diff
@@ -101,13 +101,12 @@
...
+// default. You can alter this default mapping by using a struct tag beginning
+// with "docstore:". Docstore struct tags support renaming, omitting fields
 // unconditionally, or omitting them only when they are empty, exactly like
-// encoding/json. Docstore also honors a "json" struct tag if there is no "docstore"
```
polinasok: So does docstore not honor `json:"..."`?

shantuo: Not any more, that was changed in https://github.com/google/go-cloud/commit/384e3e9eb5ccdf4a539e6bbf60e797b8795c40d3

PR #2565, opened 2019-07-16
blob: include key in error messages

### LINKED ISSUE #2564: blob: include blob key in errors
I noticed this from an [issue filed against Pulumi](https://github.com/pulumi/pulumi/pull/2455#issuecomment-511277129). It doesn't appear to be a Go CDK issue, but the error message wasn't very clear ("The specified blob does not exist") and the user noted that it "doesnt tell which blob it would expect and what content that should have."

Services appear to be inconsistent about whether they include the key name in "not found" errors (see below for details).

Since the portable type (`gocloud.dev/blob`) knows the key, we should just make sure it's included in the error message for all errors that apply to a key. We already include a prefix, so it's shouldn't be hard to add it there.

GCS:
```
blob (code=NotFound): storage: object doesn't exist
```

S3:
```
blob (code=NotFound): NoSuchKey: The specified key does not exist.
	status code: 404, request id: ...
```

Azure:
```
Read after failed write: blob (code=NotFound): -> github.com/Azure/azure-storage-blob-go/azblob.newStorageError, /usr/local/google/home/rvangent/go/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.6.0/azblob/zc_storage_error.go:42
===== RESPONSE ERROR (ServiceCode=BlobNotFound) =====
Description=The specified blob does not exist.
RequestId:d68d0b9f-e01e-0107-36b4-092564000000
Time:2019-05-13T17:52:46.5463275Z, Details: 
   Code: BlobNotFound
   GET https://gocloudblobtests.blob.core.windows.net/go-cloud-bucket/blob-for-reading
   Authorization: REDACTED
   Content-Type: []
   User-Agent: [go-cloud/blob/0.1.0 Azure-Storage/0.6 (go1.12.5; linux)]
   X-Cloud-Trace-Context: [e2790d62e0779aaa44a51111597b5f38/16106888398616378301;o=0]
   X-Ms-Date: [Tue, 16 Jul 2019 15:41:08 GMT]
   X-Ms-Version: [2018-11-09]
   --------------------------------------------------------------------------------
   RESPONSE Status: 404 Not Found
   Content-Length: [215]
   Content-Type: [application/xml]
   Date: [Mon, 13 May 2019 17:52:45 GMT]
   Server: [Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0]
   X-Ms-Error-Code: [BlobNotFound]
   X-Ms-Request-Id: [d68d0b9f-e01e-0107-36b4-092564000000]
   X-Ms-Version: [2018-11-09]

[Other errors looked similar]
```

File:
```
blob (code=NotFound): stat /tmp/go-cloud-fileblob/blob-for-reading: no such file or directory
```

vangent:
Fixes #2564.

See the bug for context. TL;DR, services don't usually include the key in error messages (e.g., "The specified key does not exist" without saying what the key was). That lack of info can make debugging difficult. This PR adds the key (where possible) to error messages from `blob`, via the portable type.

For example, the GCS error for a missing blob was:

```blob (code=NotFound): storage: object doesn't exist```

and after this change it is:

```blob (key not-found) (code=NotFound): storage: object doesn't exist```

I also added a conformance test for `Bucket.Attributes` for a nonexistent key.

### CODE REVIEW

File: blob/blob.go [hunk 1]
```diff
@@ -768,8 +772,6 @@ func (b *Bucket) NewWriter(ctx context.Context, key string, opts *WriterOptions)
 		end:        end,
 		cancel:     cancel,
 		key:        key,
-		opts:       dopts,
```
vangent: These two fields (along with `ctx`) are only supposed to be set when we need to call `NewTypedWriter` later; they are explicitly set in the `else` a few lines down. I'm not sure why we were setting them here, they are not needed and it made the code unclear because the comment in the struct says they are only set when `w` is `nil`.

I removed them and clarified the comments.

blob/blob.go
```diff
@@ -1089,14 +1090,18 @@ func OpenBucket(ctx context.Context, urlstr string) (*Bucket, error) {
...
 	if err == nil {
 		return nil
 	}
 	if gcerr.DoNotWrap(err) {
 		return err
 	}
-	return gcerr.New(b.ErrorCode(err), err, 2, "blob")
+	msg := "blob"
+	if key != "" {
+		msg += fmt.Sprintf(" (key %s)", key)
```
shantuo: nit: use `%q` to quote `key` since it might be something like key1 -> key2 in the copy scenario.

PR #2567, opened 2019-07-16
internal/gocdk: reduce duplicate code in "deploy" for generate image refs

vangent:
We had more or less the same code to generate a tag in both `build` and `deploy`. This PR changes the `build` function to return the actual tags it uses, so that `deploy` can use them.

I changed the order of the default image refs so that the most specific one (the snapshot tag) would first and be used by `deploy`.

### DIFF
--- a/internal/cmd/gocdk/deploy.go
+++ b/internal/cmd/gocdk/deploy.go
@@ -57,13 +57,14 @@ func deploy(ctx context.Context, pctx *processContext, biome, dockerImage string
 		return xerrors.Errorf("gocdk deploy: %w", err)
 	}
 
-	// If no image was specified, compute a snapshot tag and build it.
+	// If no image was specified, do a build.
 	if dockerImage == "" {
-		var err error
-		dockerImage, err = buildForDeploy(ctx, pctx, moduleRoot)
+		refs, err := build(ctx, pctx, nil)
 		if err != nil {
 			return xerrors.Errorf("gocdk deploy: %w", err)
 		}
+		dockerImage = refs[0]
+		pctx.Logf("Deploying Docker image %q...", dockerImage)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/cmd/gocdk/deploy.go
```diff
@@ -57,13 +57,14 @@ func deploy(ctx context.Context, pctx *processContext, biome, dockerImage string
...
-	// If no image was specified, compute a snapshot tag and build it.
+	// If no image was specified, do a build.
 	if dockerImage == "" {
-		var err error
-		dockerImage, err = buildForDeploy(ctx, pctx, moduleRoot)
+		refs, err := build(ctx, pctx, nil)
 		if err != nil {
 			return xerrors.Errorf("gocdk deploy: %w", err)
 		}
+		dockerImage = refs[0]
```
zombiezen: This subscript (`refs[0]`) feels too magical to me, which I think is why I had it the way it currently is. WDYT about making the return value of `build()` be a struct with the two refs split out as named fields?

vangent: The problem is that there may be more than 2.

WDYT about just returning the first one? So the function would return a single string instead of a `[]string`. It's still a bit magic in the sense that the default has to make sure to put the snapshot tag first in the slice, and the return value would be somewhat unuseful in the case where the `refs` are user-supplied (but we don't use it in that case), but maybe better since it's more localized?

zombiezen: Oh right, I had forgotten that the user could also specify the tags.

Yes, just returning the first one SGTM. Please add a comment in the code where that happens that the ordering is significant to the function, because there is some action-at-a-distance going on here.

vangent: Done.

Issue #2304, opened 2019-06-11
gocdk: fill in TODOs for help text


vangent:
A bunch of help text was left with TODOs that need to be filled in, e.g.,

```
apply.go:		Short: "TODO Apply Terraform for BIOME",
apply.go:		Long:  "TODO more about apply",
```

PR #2569, opened 2019-07-16
internal/gocdk: fill in Help text for the biome command

vangent:
Updates #2304.


internal/cmd/gocdk/testdata/biome-help.ct
```diff
@@ -1,61 +1,73 @@
...
 
 Usage:
   gocdk biome [command]
 
 Available Commands:
-  add         TODO Add BIOME_NAME
-  apply       TODO Apply Terraform for BIOME
-  list        TODO list
+  add         Add a new biome
+  apply       Apply any changes required for the biome's resource configuration
```
clausti: I'm concerned not everyone will understand that applying configuration changes will mean provisioning cloud resources, can we add something like "(e.g. creating a bucket)"

vangent: Sure! Done.

PR #2573, opened 2019-07-16
internal/gocdk: don't use 'go list' to find the project root dir

### LINKED ISSUE #2544: internal/gocdk: see if we can avoid "go mod list" for most commands
Almost all of our commands currently call `pctx.ModuleRoot`, which execs `go mod list` to find the directory the root module is in.

The first time for a project, `go mod list` is quite slow (5-10 seconds for me); this is quite painful in tests, which generally create a new project. It's also a bit annoying for users.

Subsequent calls are significantly faster, but exec'ing a process is always going to incur some overhead.

Perhaps we should look for the root module another way? E.g., check the current directory for `go.mod` and `biomes/`, check up a level if not found, repeat?

vangent:
Fixes #2544.

Instead of exec'ing `go list` to find the root module, just start at the current directory and look for a `go.mod` file and a `biomes` directory. If not found, move up one.

This reduces the time needed for CLI tests from ~1 minute to ~2 seconds.

### DIFF
--- a/internal/cmd/gocdk/main.go
+++ b/internal/cmd/gocdk/main.go
@@ -15,7 +15,6 @@
 package main
 
 import (
-	"bytes"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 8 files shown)


internal/cmd/gocdk/main.go
```diff
@@ -161,25 +160,23 @@ func (pctx *processContext) Println(a ...interface{}) {
...
-		return "", xerrors.Errorf("couldn't find a Go module root at or above %s", pctx.workdir)
-	}
-	output = bytes.TrimSuffix(output, []byte("\n"))
-	if len(output) == 0 {
-		return "", xerrors.Errorf("couldn't find a Go module root at or above %s", pctx.workdir)
-	}
-	root := string(output)
-	if _, err := os.Stat(biomesRootDir(root)); err != nil {
-		return "", xerrors.Errorf("Go module root %s doesn't look like a Go CDK project (no biomes/ directory)", root)
+	var prevDir string
```
zombiezen: Optional: you could move this declaration to the loop clause:

```go
for dir, prevDir := pctx.workdir, ""; ; dir = filepath.Dir(dir) {
```

PR #2574, opened 2019-07-16
internal/gocdk: remove explicit wire support from serve

### LINKED ISSUE #1954: internal/cmd/gocdk: build errors in "serve" are reported as being from "wire"
Ignore the actual build errors, the bug is that "gocdk serve" is telling me the errors are from `wire` instead of from `go build`.

```
gocdk: 2019/04/29 16:53:31 Building server...
/usr/local/google/home/rvangent/go/src/gocdk/demo_blob_bucket.go:72:3: expected ';', found '('
/usr/local/google/home/rvangent/go/src/gocdk/demo_blob_bucket.go:96:2: function must be invoked in defer statement
wire: generate failed
gocdk: 2019/04/29 16:53:34 Build: build server: wire: exit status 1
```

### LINKED ISSUE #2572: internal/gocdk: consider removing explicit call to wire in "gocdk serve"
`gocdk serve` tries to call `wire` if it exists; we could consider removing this and let the user do it on their own if they want to use `wire` (they will need to do so anyway for deploy).

vangent:
Fixes #2572.
Fixes #1954.

Don't call `wire` as part of building for "gocdk build".

Rationale: It seems pretty unlikely that Go CDK users will be using `wire`. Even if they do, they will need to make sure to call `wire` before they `gocdk build` or `gocdk deploy`; doing it for them as part of `gocdk serve` isn't necessarily saving much.

Also, IMHO it might be surprising to have generated code being modified on the fly by a background process.

### DIFF
--- a/internal/cmd/gocdk/serve_test.go
+++ b/internal/cmd/gocdk/serve_test.go
@@ -24,8 +24,6 @@ import (
 )
 
 func TestBuildForServe(t *testing.T) {
-	// TODO(light): This test is not hermetic because it brings
-	// in an external module.
 	const content = `package main
 import "fmt"
 func main() { fmt.Println("Hello, World!") }`
--- a/internal/cmd/gocdk/serve_test.go
+++ b/internal/cmd/gocdk/serve_test.go
@@ -24,47 +24,22 @@ import (
 )
 
 func TestBuildForServe(t *testing.T) {
-	t.Run("NoWire", func(t *testing.T) {
-		testBuildForServe(t, map[string]string{
-			"main.go": `package main
+	// TODO(light): This test is not hermetic because it brings
+	// in an external module.
+	const content = `package main
 import "fmt"
-func main() { fmt.Println("Hello, World!") }`,
-		})
-	})
-	t.Run("Wire", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


internal/cmd/gocdk/serve_test.go
```diff
@@ -24,47 +24,22 @@ import (
 )
 
 func TestBuildForServe(t *testing.T) {
-	t.Run("NoWire", func(t *testing.T) {
-		testBuildForServe(t, map[string]string{
-			"main.go": `package main
+	// TODO(light): This test is not hermetic because it brings
```
zombiezen: This TODO is not accurate: it only brought in an external module for the Wire case. Please remove.

zombiezen:
> Even if they do, they will need to make sure to call `wire` before they `gocdk build` or `gocdk deploy`; doing it for them as part of `gocdk serve` isn't necessarily saving much.

The fact that `gocdk build` hasn't been calling Wire is an oversight IMO, but I agree with the overall change. Less moving parts is better.

internal/cmd/gocdk/serve_test.go
vangent: Ah, right -- done.

PR #2577, opened 2019-07-16
internal/gocdk: some minor renames and comment fixes

vangent:


internal/cmd/gocdk/biome_test.go
```diff
@@ -88,47 +88,45 @@ func TestBiomeApply(t *testing.T) {
...
 	}
 
-	t.Run("TerraformNotInitialized", func(t *testing.T) {
-		ctx := context.Background()
-		pctx, cleanup, err := newTestProject(ctx)
-		if err != nil {
-			t.Fatal(err)
-		}
-		defer cleanup()
+	ctx := context.Background()
```
vangent: This change is just removing the subtest from a test that only has one test.

I'm planning on adding the other tests via TestCLI.

PR #2578, opened 2019-07-16
internal/gocdk: update Help for the root command

### LINKED ISSUE #2304: gocdk: fill in TODOs for help text

vangent:
Fixes #2304.

Last one! This is for the top-level / root command, which is shown for `gocdk` or `gocdk help`.

### DIFF
--- a/internal/cmd/gocdk/main.go
+++ b/internal/cmd/gocdk/main.go
@@ -58,9 +58,12 @@ func doit() int {
 
 func run(ctx context.Context, pctx *processContext, args []string) error {
 	var rootCmd = &cobra.Command{
-		Use:   "gocdk",
-		Short: "TODO gocdk",
-		Long:  "TODO more about gocdk",
+		Use: "gocdk",
+		// Short is not used since this is the root command.
+		Long: `gocdk can initialize a new project, add deployment targets (known as "biomes"),
+add demos of Go CDK portable types, provision Cloud resources and use them in the demos,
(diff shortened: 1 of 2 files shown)


internal/cmd/gocdk/main.go
```diff
@@ -58,9 +58,12 @@ func doit() int {
 
 func run(ctx context.Context, pctx *processContext, args []string) error {
 	var rootCmd = &cobra.Command{
-		Use:   "gocdk",
-		Short: "TODO gocdk",
-		Long:  "TODO more about gocdk",
+		Use: "gocdk",
+		// Short is not used since this is the root command.
+		Long: `gocdk can initialize a new project, add deployment targets (known as "biomes"),
+add demos of Go CDK portable types, provision Cloud resources and use them in the demos,
```
zombiezen: I don't think "Cloud" is a proper noun. I think like "internet", it has become conventionally lowercased, especially when discussing multiple clouds in the generic sense. Some cursory Googling seems to confirm this. [[1]] [[2]]

[1]: https://www.quora.com/Should-the-word-Cloud-be-capitalized-It-is-often-referred-to-as-The-Cloud-in-terms-of-cloud-computing-etc
[2]: https://english.stackexchange.com/a/112518

PR #2581, opened 2019-07-17
internal/gocdk: capture output from Terraform during "biome apply"

### LINKED ISSUE #1821: internal/cmd/gocdk: deal with Terraform output

vangent:
Fixes #1821.

Previous to this PR, during "biome apply" we would run `terraform init`, then `terraform apply`, and dump all of the stdout/stderr output from those commands to the terminal. That's a lot of confusing text.

This PR changes "biome apply" to:

* Run "terraform init", then "terraform plan" (which generates a "plan" file with all the needed changes), then "terraform apply" on the generated plan. The result is basically the same as before, but splits the plan + apply into separate steps so that we can prompt for approval in between, and so that we can be more judicious about what to output for the user to see. Terraform won't prompt in `plan`, and also won't prompt in `apply` if you give it an explicit plan.

* Capture the output from each call to `terraform`, and only sometimes emit it to the terminal. When we do emit it, it's surrounded by line separators to try to distinguish it (see example below).
  * We always emit it if there's an error.
  * We always emit proposed changes from `terraform plan` if there are any (unless the user specifies `--quiet`).
  * We always emit output if the user specifies `--verbose`.

* If there are changes, we prompt for approval (unless the user specifies `--auto-approve`).

Sample output (you can see more in the `biome-apply.ct` file, but it uses `--quiet` to avoid Terraform output in the log, since Terraform output differs by OS):

```
$ gocdk biome apply dev
gocdk: Checking to see if resource updates are needed...
gocdk: Resources updates are needed.

#########################################
Output from "terraform plan -detailed-exitcode -out=/usr/local/google/home/rvangent/go/src/foo/biomes/dev/tf.plan -input=true"
#########################################
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

local_file.samplefile: Refreshing state... [id=d3486ae9136e7856bc42212385ea797094475802]

------------------------------------------------------------------------

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
-/+ destroy and then create replacement

Terraform will perform the following actions:

  # local_file.samplefile must be replaced
-/+ resource "local_file" "samplefile" {
      ~ content  = "Hello world!" -> "xHello world!" # forces replacement
        filename = "/usr/local/google/home/rvangent/go/src/foo/biomes/dev/fileblob_scratchdir/samplefile.txt"
      ~ id       = "d3486ae9136e7856bc42212385ea797094475802" -> (known after apply)
    }

Plan: 1 to add, 0 to change, 1 to destroy.
#########################################


Do you want to perform these actions (only 'yes' will be accepted to approve)?
Enter a value, or "cancel": yes
gocdk: Success!


$ gocdk biome apply dev
gocdk: Checking to see if resource updates are needed...
gocdk: No resources updates are needed.
gocdk: Success!```

### DIFF
--- a/internal/cmd/gocdk/biome.go
+++ b/internal/cmd/gocdk/biome.go
@@ -331,10 +331,14 @@ func biomeApply(ctx context.Context, pctx *processContext, biome string, opts *b
 		// Exit code 2 --> there's a diff.
 		// Print out a summary of the changes (unless Verbose).
 		if !opts.Verbose {
-			if idx := bytes.LastIndex(out, []byte("Plan:")); idx == -1 {
-				pctx.Logf("Resources changes are needed; re-run with --verbose for more details.")
-			} else {
-				pctx.Logf(string(out[idx:]))
+			pctx.Logf("Resources changes are needed; re-run with --verbose for more details.")
+			// Best-effort try to print the summary "Plan: " line.
+			if idx := bytes.LastIndex(out, []byte("Plan:")); idx != -1 {
+				// Move to start of that line.
+				for idx >= 0 && out[idx] != '\n' {
+					idx--
+				}
+				pctx.Logf(string(out[idx+1:]))
 			}
 		}
 		// Prompt for approval (unless AutoApprove).
--- a/internal/cmd/gocdk/biome.go
+++ b/internal/cmd/gocdk/biome.go
@@ -17,10 +17,15 @@ package main
 import (
 	"bufio"
 	"context"
+	"errors"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
 	"sort"
 	"strconv"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"gocloud.dev/internal/cmd/gocdk/internal/prompt"
@@ -66,7 +71,7 @@ information needed to deploy.`,
 	}
 	biomeCmd.AddCommand(listCmd)
 
-	var input bool
+	var applyOpts biomeApplyOptions
 	applyCmd := &cobra.Command{
 		Use:   "apply <biome name>",
 		Short: "Apply any changes required for the biome's resource configuration (e.g., creating Cloud resources)",
@@ -76,10 +81,13 @@ configuration (e.g., creating or updating Cloud resources).
 Runs "terraform init" followed by "terraform apply".`,
 		Args: cobra.ExactArgs(1),
 		RunE: func(_ *cobra.Command, args []string) error {
-			return biomeApply(ctx, pctx, args[0], input)
+			return biomeApply(ctx, pctx, args[0], &applyOpts)
 		},
 	}
-	applyCmd.Flags().BoolVar(&input, "input", true, "ask for input for Terraform variables if not directly set")
+	applyCmd.Flags().BoolVar(&applyOpts.Input, "input", true, "ask for input for Terraform variables if not directly set")
+	applyCmd.Flags().BoolVar(&applyOpts.Verbose, "verbose", false, "true to print all output from Terraform")
+	applyCmd.Flags().BoolVar(&applyOpts.Quiet, "quiet", false, "true to suppress all output from Terraform unless there's an error")
+	applyCmd.Flags().BoolVar(&applyOpts.AutoApprove, "auto-approve", false, "true to auto-approve resource changes")
 	biomeCmd.AddCommand(applyCmd)
 
 	rootCmd.AddCommand(biomeCmd)
@@ -232,10 +240,19 @@ func biomeList(ctx context.Context, pctx *processContext) error {
 	return nil
 }
 
+// biomeApply holds options for "biome apply".
+type biomeApplyOptions struct {
+	Quiet       bool // suppress all Terraform output unless there's an error
+	Verbose     bool // show all Terraform output
+	AutoApprove bool // auto-approve Terraform changes
+	Input       bool // allow Terraform prompts for input
+}
+
 // biomeApply implements the "biome apply" subcommand.
 //
-// It runs "terraform init" and "terraform apply" for the named biome.
-func biomeApply(ctx context.Context, pctx *processContext, biome string, input bool) error {
+// It runs "terraform init", "terraform plan", and "terraform apply" for the
+// named biome.
+func biomeApply(ctx context.Context, pctx *processContext, biome string, opts *biomeApplyOptions) error {
 	moduleRoot, err := pctx.ModuleRoot(ctx)
 	if err != nil {
 		return xerrors.Errorf("biome apply %s: %w", biome, err)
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: internal/cmd/gocdk/biome.go [hunk 1]
```diff
@@ -246,17 +263,91 @@ func biomeApply(ctx context.Context, pctx *processContext, biome string, input b
 		return xerrors.Errorf("biome apply %s: %w", biome, err)
 	}
 
-	c := pctx.NewCommand(ctx, biomePath, "terraform", "init", "-input="+strconv.FormatBool(input))
-	if err := c.Run(); err != nil {
+	writeTerraformOutput := func(cmd *exec.Cmd, msg []byte) {
+		const separator = "#########################################\n"
+		pctx.stderr.Write([]byte("\n"))
```
vangent: Feedback on including this? I thought it would be useful to include the exact command line so the user can reproduce it manually, but maybe only with `--verbose`?

internal/cmd/gocdk/biome.go
```diff
@@ -246,17 +263,91 @@ func biomeApply(ctx context.Context, pctx *processContext, biome string, input b
...
+		writeTerraformOutput(c, out)
 		return xerrors.Errorf("biome apply %s: %w", biome, err)
+	} else if err != nil {
+		// Exit code 2 --> there's a diff.
+		// Print them out and prompt for approval.
+		pctx.Logf("Resources updates are needed.")
+		if opts.Quiet {
+			pctx.Logf("Not showing update details due to --quiet.")
+		} else {
+			writeTerraformOutput(c, out)
```
vangent: This prints all of the output from `terraform plan` -- as you can see from the example it is still quite chatty. We could try to parse it to reduce it.... Ick.

internal/cmd/gocdk/biome_test.go
```diff
@@ -76,14 +76,6 @@ func TestBiomeAdd(t *testing.T) {
 }
 
 func TestBiomeApply(t *testing.T) {
-	// TODO(#1809): test cases
```
vangent: I believe these tests are all covered by the CLI test now, although a couple aren't needed anymore because we just always run `terraform init`.

internal/cmd/gocdk/deploy.go
```diff
@@ -72,7 +72,7 @@ func deploy(ctx context.Context, pctx *processContext, biome, dockerImage string
 
 	// Run "biome apply".
 	if apply {
-		if err := biomeApply(ctx, pctx, biome, true); err != nil {
+		if err := biomeApply(ctx, pctx, biome, nil); err != nil {
```
vangent: Note that this changes the value for `Input` from `true` to `false` -- this means that if you run `gocdk deploy` and Terraform has questions for you, it will fail; I think that's OK but let me know what you think.

Our normal workflow in the tool doesn't use Terraform variables, so it's unlikely to happen (the user would have to manually add a variable).

We could also turn this into a flag.

internal/cmd/gocdk/main_test.go
```diff
@@ -106,8 +106,17 @@ func TestCLI(t *testing.T) {
 		t.Fatal(err)
 	}
 	goroot := strings.TrimSpace(string(gorootOut))
-	os.Setenv("PATH", fmt.Sprintf("%s/bin", goroot)) // gocdk runs the go command
-	os.Setenv("GO111MODULE", "on")                   // gocdk requires module behavior even when under GOPATH
+	tfPath, err := exec.LookPath("terraform")
```
vangent: This change modifies `PATH` for TestCLI so that it can find `terraform`.

internal/cmd/gocdk/biome.go
```diff
@@ -246,17 +263,98 @@ func biomeApply(ctx context.Context, pctx *processContext, biome string, input b
...
+	c.Stdout, c.Stderr = nil, nil
+	if out, err := c.CombinedOutput(); err != nil {
+		writeTerraformOutput(c, out)
+		return xerrors.Errorf("biome apply %s: %w", biome, err)
+	} else if opts.Verbose {
+		writeTerraformOutput(c, out)
+	}
+
+	// Next, run "terraform plan", writing the plan to a temp file.
+	planFile, err := ioutil.TempFile(pctx.workdir, "tfplan")
```
zombiezen: Just checking: is this intentionally writing the file in the current directory rather than the system temporary directory? I don't see a good reason to expose the file to the user unless the idea is to permit the user to re-execute the plan in case of failure. However, that seems to not be the case, since below there is a `defer os.Remove(...)`.

internal/cmd/gocdk/biome.go
```diff
@@ -246,17 +263,98 @@ func biomeApply(ctx context.Context, pctx *processContext, biome string, input b
...
+	if err != nil {
+		return xerrors.Errorf("biome apply %s: %w", biome, err)
+	}
+	// Note: using just the base filename instead of passing the file path to --out
+	// because Terraform appears to not work with full Windows paths.
+	planFileName := filepath.Base(planFile.Name())
+	defer planFile.Close()
+	defer os.Remove(planFile.Name())
+	c = pctx.NewCommand(ctx, biomePath, "terraform", "plan", "-detailed-exitcode", "-out="+planFileName, inputArg)
```
zombiezen: Wait though: this is expecting the plan file to be in the biome directory, which may be different from the working directory. This seems like a bug.

internal/cmd/gocdk/biome.go
```diff
@@ -246,17 +263,98 @@ func biomeApply(ctx context.Context, pctx *processContext, biome string, input b
...
+	// Next, run "terraform plan", writing the plan to a temp file.
+	planFile, err := ioutil.TempFile(pctx.workdir, "tfplan")
+	if err != nil {
+		return xerrors.Errorf("biome apply %s: %w", biome, err)
+	}
+	// Note: using just the base filename instead of passing the file path to --out
+	// because Terraform appears to not work with full Windows paths.
+	planFileName := filepath.Base(planFile.Name())
+	defer planFile.Close()
+	defer os.Remove(planFile.Name())
```
zombiezen: A failure to remove the file should at the least be logged. (Ignoring the `Close` error above is fine, since we're not directly writing to the file.)

However, I think the ordering of this is off: given the `defer` order, this will remove the file then close it. On Windows, I believe this will cause the remove to fail since the file is in use.

internal/cmd/gocdk/biome.go
```diff
@@ -246,17 +263,98 @@ func biomeApply(ctx context.Context, pctx *processContext, biome string, input b
...
+	c = pctx.NewCommand(ctx, biomePath, "terraform", "plan", "-detailed-exitcode", "-out="+planFileName, inputArg)
+	c.Env = overrideEnv(c.Env, "TF_IN_AUTOMATION=1")
+	c.Stdout, c.Stderr = nil, nil
+
+	// No error means success and no diffs.
+	// Exit code 1 means the command failed.
+	// Exit code 2 means the command succeeded, but there's a diff.
+	if out, err := c.CombinedOutput(); err != nil && !strings.Contains(err.Error(), "exit status 2") {
```
zombiezen: Using a `strings.Contains` for this is brittle. Instead, use `xerrors.As` to extract the [`*exec.ExitError`](https://godoc.org/os/exec#ExitError) and use [`os.ProcessState.ExitCode()`](https://godoc.org/os#ProcessState.ExitCode) to test for this condition.

internal/cmd/gocdk/deploy.go
zombiezen: This change seems okay, and it would be easy to relax it later by adding a flag, as you say.

internal/cmd/gocdk/biome.go
zombiezen: As discussed offline, I think a better behavior would be to nix the `--quiet` flag, print out the summary of how many resources added, removed, changed by default, and then make it so that full Terraform output is shown when passing `--verbose`.

internal/cmd/gocdk/biome.go
zombiezen: I think the command line is helpful.

On a different note, I think I would rather that the Terraform command is streamed to stdout in the case where `--verbose` is passed, so that the user can see the progress.

vangent: Done. I created a local helper function to exec the command and return the buffered output (if any; in Verbose mode it is `nil`) and the error.

internal/cmd/gocdk/biome.go
vangent: Done.

internal/cmd/gocdk/biome.go
vangent: Yes, it is intentional, although not for a great reason. See the comment a couple of lines down (maybe I should move it up?).

```
	// Note: using just the base filename instead of passing the file path to --out
	// because Terraform appears to not work with full Windows paths.
```

Basically, I can't get `terraform apply` to work on Windows when passed a full path to a plan file. It just exits with status code 1 and no message. It works fine when the plan file is in the current directory and you pass it just the filename.

internal/cmd/gocdk/biome.go
vangent: Yes, the plan file is generated in the biome directory. All of the Terraform commands are also executed using that same directory as their working dir (the 2nd argument to `pctx.NewCommand`.

vangent: Ah, never mind, you are right -- updated to make the temp file in `biomeDir` as well.

internal/cmd/gocdk/biome.go
vangent: Fixed to always close the file right away; we don't actually need the `os.File`, just the name.

internal/cmd/gocdk/biome.go
vangent: Tried to improve the comments for this.

internal/cmd/gocdk/biome.go
vangent: Ooh, that would be great... but it doesn't work in Go 1.11. I can leave a `TODO` to fix it after 1.11 support is dropped...?

vangent:
After the changes to the default mode, it now looks like this:

```
$ gocdk biome apply dev
gocdk: Checking to see if resource updates are needed...
gocdk: Plan: 1 to add, 0 to change, 1 to destroy.

Do you want to perform these actions (only 'yes' will be accepted to approve)?
Enter a value, or "cancel": yes
gocdk: Success!
```

PR #2583, opened 2019-07-18
samples/guestbook: fix links for external documentation

LeniM:
I think it is quite self explanatory :)

### DIFF
--- a/samples/guestbook/README.md
+++ b/samples/guestbook/README.md
@@ -6,7 +6,7 @@ in a cloud-agnostic manner using MySQL, the generic blob API, and the generic
 runtimevar API. All platform-specific code is set up by
 [Wire](https://github.com/google/wire).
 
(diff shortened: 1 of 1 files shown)


LeniM:
I signed it!

samples/guestbook/README.md
```diff
@@ -9,9 +9,9 @@ runtimevar API. All platform-specific code is set up by
 You can see how to build, run, and deploy this sample by following one of the
 following tutorials based on the cloud platform of your choosing:
```
vangent: Thanks for catching this! I missed these references when combining the 3 tutorial pages into 1.

Given that there are no longer three separate tutorial pages, but rather a single page with local, AWS, GCP, and Azure, how about we drop this list entirely? I suggest something like:

```
The [online tutorial][https://gocloud.dev/tutorials/guestbook] will walk you through how to build, run, and deploy this sample locally, on Google Cloud Platform (GCP), on Amazon Web Servicess (AWS), or on Microsoft Azure.
```

LeniM: no worries

PR #2586, opened 2019-07-18
internal/gocdk: rename "destroy" -> cleanup; add --delete flag; make it work like "biome apply"

vangent:
One last review...

This PR:

* Renames the command from "destroy" to "cleanup".
* Adds a flag to actually delete the biome directory.
* Removes the Go test in favor of cmdline tests that have better coverage.
* Updates the command to use the same framework as `biome apply`, so that it has `--verbose`, `--auto-approve`, etc.


.travis.yml
```diff
@@ -49,15 +49,15 @@ jobs:
   include:
     - go: "1.11.x"
       os: linux
-      if: branch = master AND type = push
```
vangent: This is temporary, will revert before checkin.

Issue #2589, opened 2019-07-19
docstore: all checks for keys should be checking for empty value


shantuo:
It currently checks for `nil` only, for example
```go
doc := map[string]interface{}{"key": ""}
```
should be treated as missing key as well.

We currently only does empty value check in firestore since it requires string type for name field, we should do this in the portable API ideally.

jba:
We should also document that you can't use zero-valued keys.

I'm a little concerned that we can't support an empty key in a map. It's possible that someone needs that. But I think it's probably better that we start by returning an error, and see if anyone complains.

Issue #2422, opened 2019-06-26
docstore: move tests to conformance tests, or document why they are provider-specific


vangent:
E.g., `memdocstore` has tests called `TestUpdateEncodesValues` and `TestUpdateAtomic`. These sound like they should be conformance tests for all providers, not `memdocstore` specific. If I'm wrong, they should explain why they are provider-specific.

jba:
For `TestUpdateEncodesValues`, I'm not sure that every driver will blind-decode an int to an int64, so I don't know how general that test is.

For `TestUpdateAtomic`, note that the doc for update says "It is undefined whether updating a sub-field of a non-map field will succeed.". This test assumes it will not succeed. I suppose we could add a boolean function to the harness so we can run this test only on drivers where it fails.

polinasok:
memtest still has 4 tests: `TestUpdateEncodesValues`, `TestUpdateAtomic`, `TestSortDocs`, `TestSaveAndLoad`. Tuo will continue with `TestUpdateEncodesValues`, and I will take on the other 3, moving them to conformance test or adding comments if they cannot/should not be moved.

shantuo:
There are dynamo-specific tests around query plans and I will add comments to them.

jba:
@polinasok Tuo and I already had conversations about those tests. I think they were on a PR.

The atomic one is hard to move because it depends on undefined behavior (that _is_ defined for memdocstore).

SaveAndLoad tests a feature specific to memdocstore.

PR #2590, opened 2019-07-19
docstore: merge driver specific tests into conformance tests

shantuo:
Updates #2422

This is part I of the change, which remove some of the action tests, if the idea is accepted, I'll continue with part II which will contain query tests.

### docstore
* TestToDriverMods: merged into `TestActionsDo` as the correctionness of the mods should be covered in conformance tests.

### gcpfirestore

* TestNewGetRequest: no need to test the implementation of how batch get requests are constructed, conformance tests should cover various cases.

### memdocstore

* TestUpdateEncodesValues: no need to test `update` encodes values, this should be tested in the codec tests.
* TestMissingKeyCreateFailsWithKeyFunc: this is testing docstore.ActionList.Do and should be replaced by portable API and conformance tests (Done in #2589).
* TestExampleInDoc: should be covered by conformance tests.

### docstore

* Added struct test cases.
* TestToDriverActionsErrors: changed to `TestActionsDo` to use the exported method `Do`.

### conformance tests

* Merged `MultipleActions` and `UnorderedActions` since we only have unordered mode now.
* Added `testExampleInDoc`.


jba:
> * TestToDriverMods: merged into `TestActionsDo` as the correctionness of the mods should be covered in conformance tests.

I'm OK with this change, but it has nothing to do with the conformance tests. 

> * TestResourceIDRegexp: no need to test the resource ID regex.

Why not? Regexes are hard to get right. It's important to test the edge cases.

> * TestUpdateEncodesValues: no need to test `update` encodes values, this should be tested in the codec tests.

But the codec tests check that various things are encoded correctly, not that encoding is performed in the first place.

Please do the following: 
1. Remove the logic from memdocstore that encodes updates.
2. Run all the tests and verify that some test breaks, *and* that it breaks in a way that clearly points to this bug.

The second part is important: if a test breaks but it's hard to see why, then it can be useful to have a redundant test that targets the exact problem.


> * TestMissingKeyCreateFailsWithKeyFunc: this is testing docstore.ActionList.Do and should be replaced by portable API and conformance tests

Please do the same exercise as above.

In general, don't remove a test unless you can prove that it's covered elsewhere.

> * TestExampleInDoc: should be covered by conformance tests.

Yes, it's probably a good idea to make sure this example works for all providers. But don't remove the test until then.

> Changed portable API and conformance tests

You haven't changed any conformance tests. The file `drivertest.go` is unmodified.

shantuo:
> > * TestToDriverMods: merged into `TestActionsDo` as the correctionness of the mods should be covered in conformance tests.
> 
> I'm OK with this change, but it has nothing to do with the conformance tests.
> 
The conformance part is talking about testing the results of updates.
> > * TestResourceIDRegexp: no need to test the resource ID regex.
> 
> Why not? Regexes are hard to get right. It's important to test the edge cases.
> 
I was thinking that would fail anyway if the user passes in an invalid ID. But I guess it makes sense to keep the test since we don't want to reject good ones?
> > * TestUpdateEncodesValues: no need to test `update` encodes values, this should be tested in the codec tests.
> 
> But the codec tests check that various things are encoded correctly, not that encoding is performed in the first place.
> 
> Please do the following:
> 
> 1. Remove the logic from memdocstore that encodes updates.
> 2. Run all the tests and verify that some test breaks, _and_ that it breaks in a way that clearly points to this bug.
> 
> The second part is important: if a test breaks but it's hard to see why, then it can be useful to have a redundant test that targets the exact problem.
> 
Yes I noticed that, the tests will break if the mods are not encoded, it either returns invalidArgument (for bad increment value), or has type mismatch (for directly updating numbers). I've also added test cases to testUpdate for that, but in a different PR #2594.
> > * TestMissingKeyCreateFailsWithKeyFunc: this is testing docstore.ActionList.Do and should be replaced by portable API and conformance tests
> 
> Please do the same exercise as above.
> 
Yes I do have tests for that but blocked by #2589, can I make a separate PR to address that together? The test for this case is not working as I expected now.
> In general, don't remove a test unless you can prove that it's covered elsewhere.
> 
> > * TestExampleInDoc: should be covered by conformance tests.
> 
> Yes, it's probably a good idea to make sure this example works for all providers. But don't remove the test until then.
> 
Yes, I added this example type to the portable API test, and I plan to also test it in the conformance tests.
> > Changed portable API and conformance tests
> 
> You haven't changed any conformance tests. The file `drivertest.go` is unmodified.
I wasn't include that file since some tests are failing due to #2589 , can I address that in a separate PR?

shantuo:
PTAL

jba:
It sounds like #2589 should be addressed before continuing with this PR.

shantuo:
PTAL

docstore/drivertest/drivertest.go
```diff
@@ -1819,6 +1764,76 @@ func testActionsOnStructWithoutRevision(t *testing.T, coll *ds.Collection) {
...
+			t.Fatal(err)
+		}
+	}
+	checkFieldEqual := func(got, want interface{}, field string) {
+		t.Helper()
+		dg, err := driver.NewDocument(got)
+		must(err)
+		dw, err := driver.NewDocument(want)
+		must(err)
+		fvg, err := dg.GetField(field)
```
jba:
```
fvg, err := MustDocument(got).GetField(field)
```
etc.

PR #2595, opened 2019-07-19
docstore: add a test for Query.Update without a Where clause

jba:
Also, fix a bug in the Firestore native codec test. (I believe it got
tickled by a mistaken update I did to the latest cloud.google.com/go
release.  So even though it's not bothering us now, it will when we
update.)


docstore/drivertest/drivertest.go
```diff
@@ -1433,28 +1430,41 @@ func testUpdateQuery(t *testing.T, coll *ds.Collection) {
...
-		g.DocstoreRevision = nil
+	check := func(q *ds.Query, filter func(*HighScore) bool) {
+		t.Helper()
+		err := q.Update(ctx, docstore.Mods{"Score": 13, "Time": nil})
+		if err != nil {
+			t.Fatal(err)
+		}
+		got := mustCollectHighScores(ctx, t, coll.Query().Get(ctx))
+		for _, g := range got {
+			if g.DocstoreRevision == nil {
```
shantuo: You can use `checkHasRevisionField` here.

jba: This seems simpler and clearer, since I know the type and field name.

PR #2597, opened 2019-07-22
docstore: support Query.Update on documents with no revision field

### LINKED ISSUE #2458: docstore: Query.Update should not do blind writes
The MongoDB implementation of `Query.Update` is a single RPC. It cannot check revisions. So `Query.Update` must always do blind writes. We need to document that, and make other drivers conform.

Other drivers run a query and update each returned document. A driver may retrieve the revision field in the query and then use it to construct the update, which would perform optimistic locking. They shouldn't do that.

jba:
- Add a conformance test for Query.Update on a struct with no revision field.

- Change drivers as needed to pass the test.

- Since this required yet another kind of collection for testing, simplify the code
  that creates harnesses and test collections.

Updates #2458.


jba:
PTAL.

docstore/drivertest/drivertest.go
```diff
@@ -183,67 +193,54 @@ func RunConformanceTests(t *testing.T, newHarness HarnessMaker, ct CodecTester,
...
 	if err != nil {
 		t.Fatal(err)
 	}
 	defer h.Close()
 
-	dc, err := h.MakeCollection(ctx)
+	withColl(t, h, kind, f)
+}
+
+func withColl(t *testing.T, h Harness, kind CollectionKind, f func(*testing.T, Harness, *ds.Collection)) {
```
shantuo: nit: move this above or below the other with...Collection methods and maybe add a comment to it.

docstore/gcpfirestore/fs.go
```diff
@@ -165,6 +165,10 @@ func OpenCollection(client *vkit.Client, collResourceID, nameField string, opts
 // Providing a function to construct the primary key is useful in two situations: if
 // your desired primary key field is not a string, or if there is more than one field
 // you want to use as a primary key.
+//
+// For the collection to be usable with Query.Delete and Query.Update,
+// nameFunc must work with map[string]interface{} as well as whatever
```
shantuo: I find this comment a bit vague if the user doesn't know about the implementation. Maybe add something like "nameFunc must work with both map and struct types representing the same underlying data structure", an example can be helpful as well, or direct the user to our test struct.

docstore/mongodocstore/mongo.go
```diff
@@ -125,6 +125,11 @@ type Options struct {
 	// The name of the field holding the document revision.
 	// Defaults to docstore.DefaultRevisionField.
 	RevisionField string
+	// Whether Query.Update writes a new revision into the updated documents.
+	// The default is false, meaning that a revision will be written to all
+	// documents that satisfy the query's conditions.. Set to true if the
```
shantuo: nit: there is one extra ".".

docstore/mongodocstore/mongo.go
```diff
@@ -125,6 +125,11 @@ type Options struct {
 	// The name of the field holding the document revision.
 	// Defaults to docstore.DefaultRevisionField.
 	RevisionField string
+	// Whether Query.Update writes a new revision into the updated documents.
```
shantuo: We should remind the user that this option should be used on the collections where there is no revision at all. It doesn't make sense to have this option on with revision field in the data because the user will end up with staled revision after doing an update query.

docstore/query.go
```diff
@@ -196,20 +197,100 @@ func (q *Query) Delete(ctx context.Context) error {
...
+	if d, ok := q.coll.driver.(rdq); ok {
+		return d.RunDeleteQuery(ctx, q.dq)
+	}
+
+	return q.runAction(ctx, func(al *ActionList, doc Document) {
+		al.Delete(doc)
+	})
+}
+
+func (q *Query) runAction(ctx context.Context, addAction func(*ActionList, Document)) error {
```
shantuo: `runQueryActions`?

docstore/query.go
```diff
@@ -196,20 +197,100 @@ func (q *Query) Delete(ctx context.Context) error {
 	if err := q.validateWrite("delete"); err != nil {
 		return err
 	}
-	return q.coll.driver.RunDeleteQuery(ctx, q.dq)
+	type rdq interface {
```
shantuo: This interface should be defined in the driver package.

docstore/query.go
```diff
@@ -196,20 +197,100 @@ func (q *Query) Delete(ctx context.Context) error {
...
+		for _, e := range alerr {
+			retries[keys[e.Index]] = true
+		}
+		// Re-run the query.
+	}
 }
 
 // Update updates all the documents specified by the query.
 // It is an error if the query has a limit.
+// TODO: export after fixing #2611.
```
shantuo: Should this be un-exported now?

docstore/query.go
polinasok: runAndRetryActions?

docstore/drivertest/drivertest.go
jba: Done.

I agree about the separate PR. You are quite right.

docstore/gcpfirestore/fs.go
jba: Done.

docstore/mongodocstore/mongo.go
jba: Done

docstore/mongodocstore/mongo.go
jba: Changed the "if" to an "if and only if". Is that what you mean? Maybe suggest the doc you want to see?

docstore/query.go
jba: Went with a variation on polinasok's. Didn't want to use the word "Query" since this is a method on Query.

docstore/query.go
jba: @shantuo done.

@polinasok It's a common Go idiom that lets you dynamically ask the question, does some value implement an interface? It lets you override behavior in a general way. Some examples:
- https://godoc.org/encoding/json#Marshal, from "If an encountered value...".
- https://godoc.org/io#Copy, from "If dst implements..."

docstore/query.go
jba: Left for a follow-up CL (by you :). The changes are extensive (rewriting the How-To, for example) and I don't have the bandwidth for them.

docstore/awsdynamodb/dynamo_test.go
```diff
@@ -83,20 +83,21 @@ func (h *harness) Close() {
...
-func (h *harness) MakeAlternateRevisionFieldCollection(context.Context) (driver.Collection, error) {
-	return newCollection(dyn.New(h.sess), collectionName1, drivertest.KeyField, "",
-		&Options{AllowScans: true, RevisionField: drivertest.AlternateRevisionField})
+func (h *harness) MakeCollection(_ context.Context, kind drivertest.CollectionKind) (driver.Collection, error) {
+	switch kind {
+	case drivertest.SingleKey, drivertest.NoRev:
+		return newCollection(dyn.New(h.sess), collectionName1, drivertest.KeyField, "", &Options{AllowScans: true})
+	case drivertest.TwoKey:
+		return newCollection(dyn.New(h.sess), collectionName2, "Game", "Player", &Options{
+			AllowScans:       true,
```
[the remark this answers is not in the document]
jba: I don't fully understand your question. This doesn't return a revision. It returns a collection that has two keys, "Game" and "Player", which is what the tests that use drivertest.TwoKey want. Whether the values stored in the collection have revisions or not is up to the user of the collection; it's not part of creating a collection.

docstore/drivertest/drivertest.go
```diff
@@ -35,24 +35,32 @@ import (
...
-// Harness descibes the functionality test harnesses must provide to run
-// conformance tests.
-type Harness interface {
-	// MakeCollection makes a driver.Collection for testing.
-	// The collection should have a single primary key field of type string named
+// CollectionKind describes the kind of testing collection to create.
+type CollectionKind int
+
+const (
+	// A collection with a single primary key field of type string named
```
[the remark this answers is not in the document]
jba: See above. Except for the `RevisionField` option used by `AltRev`, the driver doesn't need to know whether the documents have revisions or not.

docstore/drivertest/drivertest.go
```diff
@@ -155,23 +163,25 @@ func RunConformanceTests(t *testing.T, newHarness HarnessMaker, ct CodecTester,
...
+	t.Run("UnorderedActions", func(t *testing.T) { withRevCollections(t, newHarness, testUnorderedActions) })
+	t.Run("GetQueryKeyField", func(t *testing.T) { withRevCollections(t, newHarness, testGetQueryKeyField) })
+	t.Run("SerializeRevision", func(t *testing.T) { withCollection(t, newHarness, SingleKey, testSerializeRevision) })
+	t.Run("ActionsOnStructNoRev", func(t *testing.T) {
+		withCollection(t, newHarness, NoRev, testActionsOnStructNoRev)
+	})
+	t.Run("ActionsWithCompositeID", func(t *testing.T) { withCollection(t, newHarness, TwoKey, testActionsWithCompositeID) })
+	t.Run("GetQuery", func(t *testing.T) { withCollection(t, newHarness, TwoKey, testGetQuery) })
+	t.Run("DeleteQuery", func(t *testing.T) { withCollection(t, newHarness, TwoKey, testDeleteQuery) })
+	t.Run("UpdateQuery", func(t *testing.T) { withCollection(t, newHarness, TwoKey, testUpdateQuery) })
```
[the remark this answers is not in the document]
jba: See above.

docstore/drivertest/drivertest.go
```diff
@@ -183,67 +193,54 @@ func RunConformanceTests(t *testing.T, newHarness HarnessMaker, ct CodecTester,
...
-		coll = ds.NewCollection(dc)
-		defer coll.Close()
-		clearCollection(t, coll)
-		t.Run("AltRev", func(t *testing.T) { f(t, coll, AlternateRevisionField) })
-	})
+	f(t, h, coll)
 }
 
-func withTwoKeyCollection(t *testing.T, newHarness HarnessMaker, f func(*testing.T, *ds.Collection)) {
+func withRevCollections(t *testing.T, newHarness HarnessMaker, f func(*testing.T, *ds.Collection, string)) {
```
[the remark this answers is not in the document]
jba: Yes, these functions are very confusing. I wrote some doc comments for them.

Feel free to rename them as you wish in another PR.

docstore/drivertest/drivertest.go
```diff
@@ -1187,7 +1185,16 @@ func newHighScore() interface{} { return &HighScore{} }
 
 // HighScoreKey constructs a single primary key from a HighScore struct
 // by concatenating the Game and Player fields.
-func HighScoreKey(doc docstore.Document) interface{} { return doc.(*HighScore).key() }
+func HighScoreKey(doc docstore.Document) interface{} {
+	switch d := doc.(type) {
+	case *HighScore:
+		return d.key()
+	case map[string]interface{}:
```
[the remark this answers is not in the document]
jba: done

docstore/drivertest/drivertest.go
```diff
@@ -1187,7 +1185,16 @@ func newHighScore() interface{} { return &HighScore{} }
...
+		return fmt.Sprintf("%v|%v", d["Game"], d["Player"])
```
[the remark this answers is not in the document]
jba: done

docstore/drivertest/drivertest.go
```diff
@@ -1558,6 +1565,50 @@ func testUpdateQuery(t *testing.T, coll *ds.Collection) {
 	})
 }
 
+func testUpdateQueryNoRev(t *testing.T, _ Harness, coll *ds.Collection) {
```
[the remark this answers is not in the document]
jba: Added some comments.

They will work the same way; we're not testing whether the updates are atomic here. (That test would be hard to write, I think.) This test just verifies that all the machinery works when there's no revision field.

docstore/drivertest/drivertest.go
```diff
@@ -1558,6 +1565,50 @@ func testUpdateQuery(t *testing.T, coll *ds.Collection) {
...
+		{"A", 90},
+		{"B", 95},
+		{"C", 88},
+	}
+	al := coll.Actions()
+	for _, g := range grades {
+		al.Put(g)
+	}
+	if err := al.Do(ctx); err != nil {
+		t.Fatal(err)
```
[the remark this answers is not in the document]
jba: I don't agree with you in general, but I do in this case; if you look elsewhere in this file, you'll see locally defined functions called `must`. I was prevented from writing `must` as a top-level function in this file, or I would have done so and used it here. In fact, it might be worth defining locally, since we could use it two or three times.

jba:
PTAL.

docstore/drivertest/drivertest.go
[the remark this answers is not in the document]
jba: I thought I replied to this, but I don't see it....just in case it got lost:

Except for the name of the revision fields, collections don't know whether they're used with revisions or not.  A Two-key collection is just one with two keys, "Game" and "Player".

docstore/mongodocstore/mongo.go
shantuo: Yes, I think this is good enough.

docstore/query.go
jba: Sorry, I didn't fully answer this. I told you in general about it but not why I'm using it here.

We want the mongo driver to do its optimized thing for these kinds of queries, but we want to default to a more general algorithm. This trick lets us do that without hard-coding mongo into the portable part of the code.

docstore/memdocstore/mem.go
```diff
@@ -326,9 +332,6 @@ func (c *collection) update(doc storedDoc, mods []driver.Mod, revOn bool) error
 			m.parentMap[m.key] = m.encodedValue
 		}
 	}
-	if revOn {
-		c.changeRevision(doc)
```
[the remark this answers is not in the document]
jba: Query.Update for memdocstore will now use the general algorithm in docstore/query.go. That algorithm reads each document individually and then updates it. So revision checks and changes are handled by the ordinary update logic (the driver.Update case of `collection.runAction` in mem.go).

PR #2598, opened 2019-07-22
docstore: move time and number comparison to driver

jba:
Factor out common code from dynamo, firestore and mem drivers.

### DIFF
--- a/docstore/gcpfirestore/query.go
+++ b/docstore/gcpfirestore/query.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"io"
 	"math"
-	"math/big"
 	"path"
 	"reflect"
 	"strings"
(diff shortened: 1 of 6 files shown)


docstore/gcpfirestore/query.go
```diff
@@ -139,7 +138,7 @@ func evaluateFilter(f driver.Filter, doc driver.Document) bool {
 	// Compare times.
 	if t1, ok := val.(time.Time); ok {
 		if t2, ok := f.Value.(time.Time); ok {
-			return applyComparison(f.Op, compareTimes(t1, t2))
+			return applyComparison(f.Op, driver.CompareTimes(t1, t2))
```
shantuo: It might be an edge case, but I wonder whether the behavior is consistent with the backend service, I mean, would the result be different when local filters and query filters got swapped due to change of order of the `Where` clauses?

jba: I agree that is something we should test. I haven't done that. I opened #2601.

Meanwhile, this PR preserves existing behavior; it's just a refactor.

PR #2599, opened 2019-07-22
internal/testing/setup: use google/go-replayers for gRPC

jba:
I moved both the gRPC and HTTP record/replay proxies to a new repo,
github.com/google/go-replayers.

I also changed their API somewhat, to make it cleaner and reduce
the dependence on Google-specific API packages.

This PR switches the Go CDK to use the gRPC record-replay proxy.

I had to rewrite the setup code a bit to accommodate the new API.

I also unexported a couple of functions that were never called outside
the package.

### DIFF
--- a/internal/testing/setup/setup.go
+++ b/internal/testing/setup/setup.go
@@ -185,14 +185,13 @@ func NewGCPgRPCConn(ctx context.Context, t *testing.T, endPoint, api string) (*g
 			t.Fatal(err)
 		}
 		return conn, done
-	} else {
-		rep, done := newGCPReplayer(t, filename)
-		conn, err := rep.Connection()
-		if err != nil {
-			t.Fatal(err)
-		}
-		return conn, done
 	}
+	rep, done := newGCPReplayer(t, filename)
+	conn, err := rep.Connection()
+	if err != nil {
+		t.Fatal(err)
+	}
+	return conn, done
 }
 
 // contentTypeInjectPolicy and contentTypeInjector are somewhat of a hack to
@@ -288,11 +287,9 @@ func FakeGCPDefaultCredentials(t *testing.T) func() {
 	}
 }
 
-// newGCPDialOptions return grpc.DialOptions that are to be appended to a GRPC
-// dial request. These options allow a recorder/replayer to intercept RPCs and
-// save RPCs to the file at filename, or read the RPCs from the file and return
-// them. When recording is set to true, we're in recording mode; otherwise we're
-// in replaying mode.
+// newGCPRecordDialOptions return grpc.DialOptions that are to be appended to a
+// GRPC dial request. These options allow a recorder to intercept RPCs and save
+// RPCs to the file at filename, or read the RPCs from the file and return them.
 func newGCPRecordDialOptions(t *testing.T, filename string) (opts []grpc.DialOption, done func()) {
 	path := filepath.Join("testdata", filename)
 	t.Logf("Recording into golden file %s", path)
--- a/internal/testing/setup/setup.go
+++ b/internal/testing/setup/setup.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"flag"
 	"io/ioutil"
-	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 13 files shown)


internal/testing/setup/setup.go
```diff
@@ -170,39 +169,30 @@ func NewGCPClient(ctx context.Context, t *testing.T) (client *gcp.HTTPClient, rt
...
 			t.Fatal(err)
 		}
 		opts = append(opts, grpc.WithTransportCredentials(grpccreds.NewClientTLSFromCert(nil, "")))
 		opts = append(opts, grpc.WithPerRPCCredentials(oauth.TokenSource{TokenSource: gcp.CredentialsTokenSource(creds)}))
+		conn, err := grpc.DialContext(ctx, endPoint, opts...)
+		if err != nil {
+			t.Fatal(err)
+		}
+		return conn, done
 	} else {
```
vangent: Drop the `else` here since the `if` ends with a `return` now?

internal/testing/setup/setup.go
```diff
@@ -298,34 +288,18 @@ func FakeGCPDefaultCredentials(t *testing.T) func() {
 	}
 }
 
-// NewGCPDialOptions return grpc.DialOptions that are to be appended to a GRPC
+// newGCPDialOptions return grpc.DialOptions that are to be appended to a GRPC
```
vangent: Nit: newGCPRecordDialOptions.

internal/testing/setup/setup.go
```diff
@@ -298,34 +288,18 @@ func FakeGCPDefaultCredentials(t *testing.T) func() {
...
 // dial request. These options allow a recorder/replayer to intercept RPCs and
```
vangent: recorder/replayer -> recorder. The last sentence of the comment is also stale.

pubsub/kafkapubsub/go.sum
```diff
@@ -65,6 +65,8 @@ github.com/google/btree v0.0.0-20180813153112-4030bb1f1f0c/go.mod h1:lNA+9X1NB3Z
 github.com/google/go-cmp v0.2.0/go.mod h1:oXzfMopK8JAjlY9xF4vHSVASa0yLyX7SntLO5aqRK0M=
 github.com/google/go-cmp v0.3.0 h1:crn/baboCvb5fXaQ0IJ1SGTsTVrWpDsCWC8EGETZijY=
 github.com/google/go-cmp v0.3.0/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
+github.com/google/go-replayers/grpcreplay v0.1.0 h1:eNb1y9rZFmY4ax45uEEECSa8fsxGRU+8Bil52ASAwic=
```
vangent: Why does `kafkapubsub` need `grpcreplay`? I guess this is just `go.sum`, but is this really needed? Same for `mongodocstore` and `natspubsub` and `gocdk` and `contributebot` and others.

jba: I'm not sure how to answer that. `go mod tidy` doesn't change anything.

For that matter, why does `docstore/mongodocstore` have aws entries in its `go.sum`?

I guess `go.sum` just accumulates stuff and never gets cleaned up. 

What action would you like me to take?

jba:
PTAL

internal/testing/setup/setup.go
vangent: Ping on this one.

pubsub/kafkapubsub/go.sum
vangent: I guess you are right, there is nothing that cleans `go.sum` up. Maybe we should delete them and recreate them from scratch periodically?

At any rate, maybe try deleting the new lines, run `go build && go test`; if they come back, fine, if they don't, let's avoid cluttering the file unnecessarily.

jba: I tried this with samples and runtimevar/etcd. In both cases the lines came back. I stopped experimenting at that point.

PR #2600, opened 2019-07-23
docstore: disallow zero-value key

### LINKED ISSUE #2589: docstore: all checks for keys should be checking for empty value

### LINKED ISSUE #2593: docstore: conformance tests should test actions on collections with IDFunc

shantuo:
Fixes #2589 
Fixes #2593 

This PR checks whether the key of a document is nil or an empty value.
- When the user names a field as key, the portable API checks whether the key is empty and decide whether it should return an error (missing key is not an error for create)
- When the user provides us a nameFunc to construct a key for some driver, the driver will check whether the key is missing since that will always be an invalid argument error.


docstore/awsdynamodb/dynamo.go
```diff
@@ -131,12 +132,11 @@ func newCollection(db *dyn.DynamoDB, tableName, partitionKey, sortKey string, op
 
 // Key returns a two-element array with the partition key and sort key, if any.
 func (c *collection) Key(doc driver.Document) (interface{}, error) {
-	var keys [2]interface{}
-	var err error
-	keys[0], err = doc.GetField(c.partitionKey)
-	if err != nil {
+	pkey, err := doc.GetField(c.partitionKey)
+	if err != nil || pkey == nil || driver.IsEmptyValue(reflect.ValueOf(pkey)) {
```
shantuo: I've tried in `IsEmptyValue` to loop through an array to check whether each element is also empty, but that didn't work for a case like `[2]interface{}{"", nil}`, since both elements are `reflect.Interface` kind and the first one is not a "zero interface". Another reason is that only the partition key is actually required.

docstore/drivertest/drivertest.go
```diff
@@ -1254,17 +1275,49 @@ func testGetQueryKeyField(t *testing.T, coll *ds.Collection, revField string) {
 
 func sortByKeyField(d1, d2 docmap) bool { return d1[KeyField].(string) < d2[KeyField].(string) }
 
+// TODO(shantuo): consider add this test to all action tests, like the AltRev
```
shantuo: I didn't integrate this test into all other action tests since there are a few PR's touching the conformance test currently and this can wait I think.

docstore/awsdynamodb/dynamo.go
jba: What you have seems fine.

docstore/doc.go
```diff
@@ -151,14 +151,15 @@
...
 // Keys never appear alone in the docstore API, only as part of a document. For
 // instance, to retrieve a document by key, you pass the Collection.Get method
 // a document as a struct pointer or map with the key field populated, and docstore
-// populates the rest of that argument with the stored contents.
+// populates the rest of that argument with the stored contents. Docstore
+// doesn't take zero-value key.
 //
-// When you open a collection using the OpenCollection method of the service-specific driver or
-// a URL, you specify how to extract the key from a document.
+// When you open a collection using the constructor of the service-specific
```
jba: "Constructor" isn't a standard term in Go or this project, although we use it informally amongst ourselves. And I think it will confuse people coming from languages like Java. So can we leave this as it was?

docstore/drivertest/drivertest.go
```diff
@@ -1154,7 +1168,8 @@ type nativeMinimal struct {
 	LS []string
 }
 
-// The following is the schema for the collection used for query testing.
+// The following is the schema for the collection that the ID is composed from
```
jba: s/that/where/

docstore/drivertest/drivertest.go
jba: OK.

docstore/drivertest/drivertest.go
shantuo: Done.

shantuo:
In `doc.go` it currently says zero-value key is not allowed, I also updated comments in `docstore.Create`.

PTAL

docstore/drivertest/drivertest.go
```diff
@@ -1168,7 +1168,7 @@ type nativeMinimal struct {
 	LS []string
 }
 
-// The following is the schema for the collection that the ID is composed from
+// The following is the schema for the collection what the ID is composed from
```
jba: I'm sorry, this should be "where", not "what".

PR #2609, opened 2019-07-24
internal/gocdk: add verbose options to serve+deploy

vangent:
This is needed because both `serve` and `deploy` can call `biome apply`, which normally hides the Terraform output if there are changes and says "re-run with --verbose to see details" or something like that.

### DIFF
--- a/internal/cmd/gocdk/deploy.go
+++ b/internal/cmd/gocdk/deploy.go
@@ -34,6 +34,7 @@ import (
 func registerDeployCmd(ctx context.Context, pctx *processContext, rootCmd *cobra.Command) {
 	var dockerImage string
 	var apply bool
+	var verbose bool
 	deployCmd := &cobra.Command{
 		Use:   "deploy <biome name>",
 		Short: "Deploy the application to the biome's deployment target",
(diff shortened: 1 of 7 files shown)


internal/cmd/gocdk/deploy.go
```diff
@@ -43,18 +44,19 @@ By default, a new Docker image is built and deployed; use --image to skip the
...
+	deployCmd.Flags().BoolVar(&apply, "apply", true, "whether to run \"biome apply\" before deploying")
+	deployCmd.Flags().BoolVar(&verbose, "verbose", false, "true to print all output from Terraform during \"biome apply\"")
 	rootCmd.AddCommand(deployCmd)
 }
 
 // deploy implements the "deploy" command.
 //
 // If dockerImage is not provided, it does a build for a generated tag.
-func deploy(ctx context.Context, pctx *processContext, biome, dockerImage string, apply bool) error {
+func deploy(ctx context.Context, pctx *processContext, biome, dockerImage string, apply, verbose bool) error {
```
shantuo: Both `apply` and `verbose` could be put into an options struct since they are not required.

internal/cmd/gocdk/testdata/deploy-help.ct
```diff
@@ -8,9 +8,10 @@ Usage:
   gocdk deploy <biome name> [flags]
 
 Flags:
-      --apply biome apply    whether to run biome apply before deploying (default true)
+      --apply                whether to run "biome apply" before deploying (default true)
   -h, --help                 help for deploy
       --image [name][:tag]   Docker image to deploy in the form [name][:tag]; name defaults to image name from Dockerfile, tag defaults to latest; empty string builds a new image
+      --verbose              true to print all output from Terraform during "biome apply"
```
shantuo: `-v` as well? It seems pretty common. Unless we are going to have other verbosity.

internal/cmd/gocdk/deploy.go
vangent: Done. I re-used `biomeApplyOptions` in case we want to add flag for the other fields.

internal/cmd/gocdk/testdata/deploy-help.ct
vangent: Done, here and elsewhere.

PR #2617, opened 2019-07-30
pubsub/awssnssqs: send up to 10 messages in a batch for SQSTopic

### LINKED ISSUE #2615: pubsub/sqs: support multiple messages in SendBatch
# Is your feature request related to a problem? Please describe.
Currently SendBatch in SQS implementation only supports sending one message, but SQS supports up to 10 messages in one batch.

Support up to 10 messages

vangent:
Fixes #2615.

When I added `SQSTopic`, I didn't notice that SQS (as opposed to SNS) has a batch API for sending messages. This PR updates SQSTopic to ask the portable type batcher for up to 10 messages at a time, and to send them with the AWS `SendMessageBatch` API.

The code is somewhat complicated by the need to keep backwards compatibility for the previous type for `As` (`sqs.SendMessageInput`, now replaced by `sqs.SendMessageBatchRequestEntry`). I handled this by copying the fields that we set from the latter to the former on demand, and then copying **all** of the fields back after the callback (in case the user modified any of them).


pubsub/awssnssqs/awssnssqs.go
```diff
@@ -452,46 +457,78 @@ func openSQSTopic(ctx context.Context, sess client.ConfigProvider, qURL string,
...
+					*p = entry
+					return true
+				}
+				return false
+			}
+			if err := dm.BeforeSend(asFunc); err != nil {
+				return err
+			}
+			if smi != nil {
+				// Copy all of the fields that may have been modified back to the entry.
```
shantuo: Add a comment to mention this is for backward-compatibility and the two types between `entry` and `smi` have matching fields so that you can copy them back instead of sending those single inputs one by one.

pubsub/awssnssqs/awssnssqs.go
```diff
@@ -98,8 +98,13 @@ const (
 	noMessagesPollDuration = 250 * time.Millisecond
 )
 
-var sendBatcherOpts = &batcher.Options{
-	MaxBatchSize: 1,   // Both SNS and SQS SendBatch only support one message at a time
+var sendBatcherOptsSNS = &batcher.Options{
+	MaxBatchSize: 1,   // SNS SendBatch only supports one message at a time
```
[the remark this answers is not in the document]
vangent: Yes. The SNS API only supports sending a single message.

pubsub/awssnssqs/awssnssqs.go
vangent: I already have that comment, up on line 488. I don't think it's useful to copy it here...?

I am not sure what you mean by "instead of sending those single inputs one by one".

shantuo: I mean if they are not exactly matching with each other, you would need to call the single RPC's on `smi`s for the user calling `as` on the old type. I think the comment above could be sufficient.

Issue #2387, opened 2019-06-20
all: highlight URL openers over constructors in package descriptions


zombiezen:
Thanks @polinasok for the find!

Current documentation for the portable API packages (e.g. https://godoc.org/gocloud.dev/blob) start with a recommendation like:

> use its exported function(s) to create a *Bucket; do not use the NewBucket function in this package.

and then further down:

> Alternatively, you can construct a *Bucket via a URL and OpenBucket. See https://gocloud.dev/concepts/urls/ for more information.

New users should use URL openers, which is what we are recommending in our how-to guides. To fix this issue, we should rewrite the package documentation introductions to either:

a) Highlight the URL openers and deemphasize the constructors, or
b) Just point to the how-to guides, which already do this.

eliben:
:+1: 

Just ran into this with docstore. https://godoc.org/gocloud.dev/docstore starts by saying:

> Your application should import one of these provider-specific subpackages and use its exported functions to create a *Collection; do not use the NewCollection function in this package.

Which is most definitely *not* what we should be recommending to users to try initially. Reducing duplication with the how-to would also be good

PR #2622, opened 2019-07-31
blob/docs: Migrate content from godoc to how-to.

polinasok:
Updates #2387
Add more details to "Opening" how-to section.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -16,7 +16,7 @@
 // within a storage location. Subpackages contain driver implementations of
 // blob for supported services.
 //
-// See https://gocloud.dev/howto/blob/ for detailed how-to guides.
+// See https://gocloud.dev/howto/blob/ for a detailed how-to guide.
 //
 //
 // Errors
@@ -453,7 +453,7 @@ var (
 		})
 )
 
-// NewBucket is intended for use by drivers.
+// NewBucket is intended for use by drivers only. Do not use in application code.
 var NewBucket = newBucket
 
 // newBucket creates a new *Bucket based on a specific driver implementation.
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -13,34 +13,10 @@
 // limitations under the License.
 
 // Package blob provides an easy and portable way to interact with blobs
-// within a storage location, hereafter called a "bucket". See
-// https://gocloud.dev/howto/blob/ for how-to guides.
+// within a storage location. Subpackages contain driver implementations of
+// blob for supported services.
 //
-// It supports operations like reading and writing blobs (using standard
-// interfaces from the io package), deleting blobs, and listing blobs in a
-// bucket.
-//
-// Subpackages contain distinct implementations of blob for various services,
-// including Cloud and on-prem solutions. For example, "fileblob" supports
-// blobs backed by a filesystem. Your application should import one of these
-// driver subpackages and use its exported function(s) to create a
-// *Bucket; do not use the NewBucket function in this package. For example:
-//
-//  bucket, err := fileblob.OpenBucket("path/to/dir", nil)
-//  if err != nil {
-//      return fmt.Errorf("could not open bucket: %v", err)
-//  }
-//  buf, err := bucket.ReadAll(context.Background(), "myfile.txt")
-//  ...
-//
-// Then, write your application code using the *Bucket type. You can easily
-// reconfigure your initialization code to choose a different driver.
-// You can develop your application locally using fileblob, or deploy it to
-// multiple Cloud providers. You may find http://github.com/google/wire useful
-// for managing your initialization code.
-//
-// Alternatively, you can construct a *Bucket via a URL and OpenBucket.
-// See https://gocloud.dev/concepts/urls/ for more information.
+// See https://gocloud.dev/howto/blob/ for detailed how-to guides.
 //
 //
 // Errors
--- a/internal/website/content/howto/blob/_index.md
+++ b/internal/website/content/howto/blob/_index.md
@@ -5,29 +5,64 @@ showInSidenav: true
 toc: true
 ---
 
-Blobs are a common abstraction for storing unstructured data on cloud storage
-providers and accessing them via HTTP. These guides show how to work with
+Blobs are a common abstraction for storing unstructured data on Cloud storage
+services and accessing them via HTTP. These guides show how to work with
 blobs in the Go CDK.
 
 <!--more-->
 
+The Go CDK supports operations like reading and writing blobs (using standard
+[io package][] interfaces), deleting blobs, and listing blobs in a bucket.
+
+Subpackages contain driver implementations of blob for various services,
+including Cloud and on-prem solutions. You can develop your application
+locally using [`fileblob`][], then deploy it to multiple Cloud providers with
+minimal reconfiguration of your initialization code.
+
+[io package]: https://golang.org/pkg/io/
+[`fileblob`]: https://godoc.org/gocloud.dev/blob/fileblob
+
 ## Opening a Bucket {#opening}
 
-The first step in interacting with unstructured storage is connecting to your
-storage provider. Every storage provider is a little different, but the Go CDK
-lets you interact with all of them using the [`*blob.Bucket` type][].
+The first step in interacting with unstructured storage is opening a connection
+to your storage service to instantiate a portable [`*blob.Bucket`][].
 
 The easiest way to open a blob is using [`blob.OpenBucket`][] and a URL
 pointing to the blob, making sure you ["blank import"][] the driver package to
-link it in. See [Concepts: URLs][] for more details. If you need
+link it in. Do not use `NewBucket` function, intended for use by the driver
+implementations only. See [Concepts: URLs][] for more details.
+
+```go
+import _ "gocloud.dev/blob/<driver>"
+
+bucket, err := blob.OpenBucket(context.Background(), "<driver-url>")
+if err != nil {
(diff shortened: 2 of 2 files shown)


blob/blob.go
```diff
@@ -13,34 +13,10 @@
...
-//
-// Then, write your application code using the *Bucket type. You can easily
-// reconfigure your initialization code to choose a different driver.
-// You can develop your application locally using fileblob, or deploy it to
-// multiple Cloud providers. You may find http://github.com/google/wire useful
-// for managing your initialization code.
-//
-// Alternatively, you can construct a *Bucket via a URL and OpenBucket.
-// See https://gocloud.dev/concepts/urls/ for more information.
+// See https://gocloud.dev/howto/blob/ for detailed how-to guides.
```
vangent: Nit: "a detailed how-to guide." ?

internal/website/content/howto/blob/_index.md
```diff
@@ -5,29 +5,64 @@ showInSidenav: true
...
 
 <!--more-->
 
+The Go CDK supports operations like reading and writing blobs (using standard
+[io package][] interfaces), deleting blobs, and listing blobs in a bucket.
+
+Subpackages contain driver implementations of blob for various services,
+including Cloud and on-prem solutions. You can develop your application
+locally using [`fileblob`][], then deploy it to multiple Cloud providers with
+minimal reconfiguration of your initialization code.
```
vangent: Consider dropping "of your initialization code". With URLs, you don't have to change any code at all.

internal/website/content/howto/blob/_index.md
```diff
@@ -5,29 +5,64 @@ showInSidenav: true
...
+
+[io package]: https://golang.org/pkg/io/
+[`fileblob`]: https://godoc.org/gocloud.dev/blob/fileblob
+
 ## Opening a Bucket {#opening}
 
-The first step in interacting with unstructured storage is connecting to your
-storage provider. Every storage provider is a little different, but the Go CDK
-lets you interact with all of them using the [`*blob.Bucket` type][].
+The first step in interacting with unstructured storage is opening a connection
```
vangent: I'm not sure that "opening a connection "is useful here. E.g., if you are using a URL opener, you don't explicitly open a connection. How about 

The first step in interacting with unstructured storage is instantiating a portable [`*blob.Bucket`][] for your storage service.

?

internal/website/content/howto/blob/_index.md
```diff
@@ -5,29 +5,64 @@ showInSidenav: true
...
+to your storage service to instantiate a portable [`*blob.Bucket`][].
 
 The easiest way to open a blob is using [`blob.OpenBucket`][] and a URL
```
vangent: I think it should be "open a Bucket", not "open a blob". Or just drop that part, e.g., "The easiest way is to use ..."

internal/website/content/howto/blob/_index.md
```diff
@@ -5,29 +5,64 @@ showInSidenav: true
...
 pointing to the blob, making sure you ["blank import"][] the driver package to
```
vangent: Similarly, I don't like "pointing to the blob." How about "a service-specific URL pointing to the bucket." ?

internal/website/content/howto/blob/_index.md
```diff
@@ -5,29 +5,64 @@ showInSidenav: true
...
-link it in. See [Concepts: URLs][] for more details. If you need
+link it in. Do not use `NewBucket` function, intended for use by the driver
```
vangent: I would leave the sentence about `NewBucket` in the godoc. We're telling them how to use it, if they are here they won't even see `NewBucket` :-).

internal/website/content/howto/blob/_index.md
```diff
@@ -5,29 +5,64 @@ showInSidenav: true
...
+implementations only. See [Concepts: URLs][] for more details.
```
vangent: Maybe "for more background on how URLs are used in the Go CDK"? Otherwise it sounds like that's where you should go for more details on what the service-specific URLs look like. Maybe move this sentence after the example? And/or, extend it to include a link to below, e.g.,

"See the [guide below][] for URL forms for supported services, and [Concepts: URLs][] for more background on how URLs are used in the Go CDK."

internal/website/content/howto/blob/_index.md
```diff
@@ -5,29 +5,64 @@ showInSidenav: true
...
+
+```go
+import _ "gocloud.dev/blob/<driver>"
```
vangent: Should probably add "gocloud.dev/blob" here too.

internal/website/content/howto/blob/_index.md
```diff
@@ -5,29 +5,64 @@ showInSidenav: true
...
+...
+``` 
+
+Alternatively, if you need
 fine-grained control over the connection settings, you can call the constructor
-function in the driver package directly (like `s3blob.OpenBucket`).
+function in the driver package directly.
+
+```go
+import "gocloud.dev/blob/<driver>"
```
vangent: same

internal/website/content/howto/blob/_index.md
```diff
@@ -5,29 +5,64 @@ showInSidenav: true
...
 
-See the [guide below][] for usage of both forms for each supported provider.
+bucket, err := <driver>.OpenBucket(...)
+...
+```
+
+You may find the [wire package][] useful for managing your initialization code
+when switching between different backing services.
+
+See the [guide below][] for usage of both connection forms for each supported service.
```
vangent: If you follow my suggestion above about including this right after the URL example, reword this to drop "both connection forms".

internal/website/content/howto/blob/_index.md
```diff
@@ -35,12 +35,12 @@ implementations only. See [Concepts: URLs][] for more details.
 ```go
 import _ "gocloud.dev/blob/<driver>"
 
+var bucket *blob.Bucket;
```
vangent: No semicolon in Go :-).

You'll also want to declare

```
var err error
bucket, err = ... // instead of :=
```

internal/website/content/howto/blob/_index.md
polinasok: Done

I think it would be a good idea to add a table of API/url schemas to the concepts doc (in a different PR).

internal/website/content/howto/blob/_index.md
polinasok: Done

internal/website/content/howto/blob/_index.md
polinasok: Is this a style violation? I tested all code using Go playground and it compiles. I only chose to declare the bucket explicitly to demonstrate the portable type. We don't really need that for error.

internal/website/content/howto/blob/_index.md
polinasok: Why? Nothing from blob is referenced in this code snippet. Besides, this was meant to show only the lines that are different. Or do you think we should include the full example with error checking and defer?

internal/website/content/howto/blob/_index.md
polinasok: Done

internal/website/content/howto/blob/_index.md
vangent: > Is the URL parameter part of initialization code?
> Or are you implying that because it can be a command-line parameter, the code doesn't change, just the way it is called.

It could be in code (e.g., as a constant), but it's more like to be some kind of configuration variable or command-line parameter. For both of the latter, the code doesn't change at all, just the input.

> I think this is useful to highlight that in most cases you will only need to change how you open the portable type and not how you use it. What about "minimal initialization reconfiguration"?

"minimal initialization reconfiguration", or just "minimal reconfiguration" SGTM.

internal/website/content/howto/blob/_index.md
vangent: Ah, you are right, this one doesn't use `blob`.

internal/website/content/howto/blob/_index.md
vangent: It doesn't look right to me to do:

```
var bucket *blob.Bucket
bucket, err := ...
```

because the bucket declared on line 1 is shadowed by the new variable (implicitly declared due to `:=`) on line 2. For example, https://play.golang.org/p/6OusvTRTD98. Although, you are right that it will give a compile error if the two variables have different types.

I'd either leave it as it was (e.g., a comment explaining that `bucket` has type `*blob.Bucket`), or declare both `bucket` and `error` and use `=`.

internal/website/content/howto/blob/_index.md
polinasok: Done

internal/website/content/howto/blob/_index.md
polinasok: If you add statements to your program to print the address of the variable, you will see that it is the same after :=, so there is no shadowing. In fact, https://golang.org/doc/effective_go.html#redeclaration says that this is legal usage, common with err. Does you still mind?

vangent: You are right; it only shadows if you do it in an a different block (i.e., as part of an `if`).

It's up to you; overall, I find the code odd-looking and nobody would ever write it that way, which probably isn't great for a how-to. Adding a line for `var err error` is also unusual looking, so I think I still vote for either nothing, or a comment like `// bucket is a *blob.Bucket; see below for how to use it.`

polinasok: Done

PR #2630, opened 2019-08-05
blob: add support for PUT and DELETE for SignedURLs.

### LINKED ISSUE #2625: blob: support PUT method for signed url

We need to return a signed url that allows end user to directly upload a file to cloud storage. AFAK, cloud storage and s3 support the feature (allowing end user to send PUT/DELETE requests). 

Allow `blob.SignedURLOptions`  to specify HTTP method.

vangent:
Fixes #2625.

This was mostly straightforward, with 2 wrinkles:

1. Azure PUTs aren't fully portable; they require an additional `x-ms-blob-type` header to set the type of the blob being created. I hacked around that in `azureblob_test.go` by intercepting requests and inserting the header where needed.

2. I had to create a new service account for our GCS project with more privileges (previously one only had "read"), which resulted in having to update a bunch of record/replay goldens.


blob/gcsblob/gcsblob_test.go
```diff
@@ -87,7 +87,7 @@ func newHarness(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
 }
 
 func (h *harness) HTTPClient() *http.Client {
-	return &h.client.Client
+	return &http.Client{Transport: h.rt}
```
vangent: Note: this was a bug previously -- `HTTPClient` is supposed to return an unauthorized HTTP client, but it was returning an authorized one. Fixed by updating `setup.go` (below) to return the `http.RoundTripper` transport before adding the authorization wrapper, and creating a new HTTPClient using that transport here.

blob/driver/driver.go
```diff
@@ -306,6 +306,8 @@ type Bucket interface {
 type SignedURLOptions struct {
 	// Expiry sets how long the returned URL is valid for. It is guaranteed to be > 0.
 	Expiry time.Duration
+	// Method is the HTTP method that can be used on the URL; one of "GET", "PUT", or "DELETE".
```
shantuo: nit: this should clarify to the drivers that only one of the three will be passed in, thus let the driver know that they must handle all of the three.

blob/azureblob/azureblob.go
```diff
@@ -15,6 +15,11 @@
 // Package azureblob provides a blob implementation that uses Azure Storage’s
 // BlockBlob. Use OpenBucket to construct a *blob.Bucket.
 //
+// NOTE: SignedURLs for PUT created with this package are not fully portable;
```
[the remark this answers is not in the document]
vangent: The result of calling `SignedURL` is a string URL; the will likely send it to a user somehow so that the user can issue the `PUT`. When they do so, they will need to make sure that that `PUT` request has this header in it. I don't think there's anything we can do in code.

blob/driver/driver.go
vangent: Done.

PR #2631, opened 2019-08-05
docstore/mongodocstore: add a conformance test for Amazon DocumentDB

### LINKED ISSUE #2294: docstore: script conformance testing against Amazon DocumentDB
You can't connect to DocumentDB over the internet, only over a VPC (virtual private cloud), which means from an EC2 instance.

Ideally, we'd have a script that you could run that would:
1. set up a DocumentDB instance if one wasn't already
2. provision and start an EC2 instance
3. run the mongo conformance tests
4. shut down the instance.

If some of those are impossible to script, we should fall back to detailed instructions.

shantuo:
Also added a `awsdocdb` directory containing terraform files to provision the resources needed for running the tests.

Fixes #2294

### DIFF
--- a/docstore/mongodocstore/awsdocdb/main.tf
+++ b/docstore/mongodocstore/awsdocdb/main.tf
@@ -1,0 +1,85 @@
+# Specify the provider and access details
+provider "aws" {
+  version = "~> 2.0"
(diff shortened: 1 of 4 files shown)


shantuo:
Done, PTAL.

PR #2632, opened 2019-08-05
docstore/mongodocstore: tests against both MongoDB v3 and v4

### LINKED ISSUE #2460: docstore/mongodocstore: test both mongo v3 and v4
We are testing v4 in Travis, we should also do v3, the reason is that both Azure Cosmos DB and Amazon DocumentDB are compatible with v3, while it is hard to run test against their backend service in Travis, having tests run against v3 would be helpful. cc @jba @vangent

shantuo:
Fixes #2460

### DIFF
--- a/docstore/mongodocstore/localmongo.sh
+++ b/docstore/mongodocstore/localmongo.sh
@@ -13,7 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Starts two local MongoDB instances (v3 and v4) via Docker.
+# Starts two local MongoDB instances (v3 and v4) via Docker listening on two
+# different ports.
 
 # https://coderwall.com/p/fkfaqq/safer-bash-scripts-with-set-euxo-pipefail
 set -euo pipefail
--- a/docstore/mongodocstore/localmongo.sh
+++ b/docstore/mongodocstore/localmongo.sh
@@ -13,14 +13,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Starts a local MongoDB instance via Docker.
+# Starts two local MongoDB instances (v3 and v4) via Docker.
 
 # https://coderwall.com/p/fkfaqq/safer-bash-scripts-with-set-euxo-pipefail
 set -euo pipefail
 
-echo "Starting MongoDB..."
-docker rm -f mongo &> /dev/null || :
-docker run -d --name mongo  -p 27017:27017 mongo:4 &> /dev/null
-echo "...done. Run \"docker rm -f mongo\" to clean up the container."
+echo "Starting MongoDB v4 listening on 27017..."
+docker rm -f mongo4 &> /dev/null || :
+docker run -d --name mongo4  -p 27017:27017 mongo:4 &> /dev/null
+echo "...done. Run \"docker rm -f mongo4\" to clean up the container."
+echo
+
+echo "Starting MongoDB v3 listening on 27018..."
+docker rm -f mongo3 &> /dev/null || :
+docker run -d --name mongo3  -p 27018:27017 mongo:3 &> /dev/null
+echo "...done. Run \"docker rm -f mongo3\" to clean up the container."
 echo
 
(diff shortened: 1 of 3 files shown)


docstore/mongodocstore/localmongo.sh
```diff
@@ -13,14 +13,20 @@
...
-echo "...done. Run \"docker rm -f mongo\" to clean up the container."
+echo "Starting MongoDB v4 listening on 27017..."
+docker rm -f mongo4 &> /dev/null || :
+docker run -d --name mongo4  -p 27017:27017 mongo:4 &> /dev/null
+echo "...done. Run \"docker rm -f mongo4\" to clean up the container."
+echo
+
+echo "Starting MongoDB v3 listening on 27018..."
+docker rm -f mongo3 &> /dev/null || :
+docker run -d --name mongo3  -p 27018:27017 mongo:3 &> /dev/null
```
polinasok: From docker docs, I see that this means port 27018 of the host is mapped to port 27017 of the container. So does this mean both mongos will map to the same container? Might be a good idea to have the container have a unique number other than the same as the port number for one of mongos.

docstore/mongodocstore/mongo_test.go
```diff
@@ -163,7 +164,17 @@ func TestConformance(t *testing.T) {
 	drivertest.RunConformanceTests(t, newHarness, codecTester{}, []drivertest.AsTest{verifyAs{}})
 }
 
-func newTestClient(t *testing.T) *mongo.Client {
+func TestConformanceV3(t *testing.T) {
+	client := newTestClient(t, serverURIV3)
```
polinasok: These two functions are identical except for the URI. Why not put the body code into a helper function that takes the URL as an arg or alternatively use a loop in a single function?

docstore/mongodocstore/mongo_test.go
```diff
@@ -182,12 +193,12 @@ func newTestClient(t *testing.T) *mongo.Client {
 func BenchmarkConformance(b *testing.B) {
 	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
 	defer cancel()
-	client, err := Dial(ctx, serverURI)
+	client, err := Dial(ctx, serverURIV4)
```
polinasok: Why only v4?

docstore/mongodocstore/localmongo.sh
shantuo: There are two containers differentiated by the tags: `mongo:3` vs `mongo:4`, so this maps the 27017 port of mongo:4 to the host's 27017 and the 27017 port of mongo:3 to the host's 27018.

docstore/mongodocstore/mongo_test.go
shantuo: This is a benchmark so there isn't compatibility concern.

docstore/mongodocstore/mongo_test.go
```diff
@@ -154,18 +154,17 @@ func (verifyAs) ErrorCheck(c *docstore.Collection, err error) error {
...
-		return &harness{client.Database(dbName)}, nil
-	}
-	drivertest.RunConformanceTests(t, newHarness, codecTester{}, []drivertest.AsTest{verifyAs{}})
+func TestConformance(t *testing.T) {
+	t.Run("V3", func(t *testing.T) {
+		testConformance(t, serverURIV4)
+	})
+	t.Run("V4", func(t *testing.T) {
+		testConformance(t, serverURIV3)
+	})
```
polinasok: Suggest inlining these very short function bodies.

docstore/mongodocstore/mongo_test.go
polinasok: Add a comment?

docstore/mongodocstore/localmongo.sh
polinasok: It looked like a copy and paste mistake to me because I did find the host ports in the files below, but not the mongo's to confirm. So maybe add a comment at least?

docstore/mongodocstore/mongo_test.go
shantuo: Done.

docstore/mongodocstore/mongo_test.go
shantuo: Done.

docstore/mongodocstore/localmongo.sh
shantuo: I think the `echo "Starting MongoDB ..."` above serves the purpose already. I clarified a bit at the top-level comment.

PR #2634, opened 2019-08-06
docstore/awsdynamodb: add a ConsistentRead mode to collection options

### LINKED ISSUE #1752: docstore: put clients in strong consistency mode for testing
Our tests write documents, then immediately read them back. This can be flaky for providers that use eventual consistency by default. Where possible, create clients using strong consistency for testing.

This isn't essential, because it only affects tests when we record, and we do that rarely. 

Currently this only applies to dynamo. Firestore is strongly consistent, and we test mongo through a single-process server.

shantuo:
Fixes #1752


docstore/awsdynamodb/dynamo_test.go
```diff
@@ -86,15 +86,24 @@ func (h *harness) Close() {
...
 	switch kind {
 	case drivertest.SingleKey, drivertest.NoRev:
-		return newCollection(dyn.New(h.sess), collectionName1, drivertest.KeyField, "", &Options{AllowScans: true})
+		return newCollection(dyn.New(h.sess), collectionName1, drivertest.KeyField, "", &Options{
+			AllowScans:     true,
+			ConsistentRead: true,
+		})
 	case drivertest.TwoKey:
+		// For query test we don't use strong consistency mode since some tests are
+		// running on global secondary index and it doesn't support ConsistentRead.
```
polinasok: What would happen if you do set it in those tests? Ignored? Error?

docstore/awsdynamodb/query.go
```diff
@@ -110,7 +111,7 @@ func (c *collection) planQuery(q *driver.Query) (*queryRunner, error) {
 			cb = cb.WithFilter(filtersToConditionBuilder(q.Filters))
 			cbUsed = true
 		}
-		in := &dyn.ScanInput{TableName: &c.table}
+		in := &dyn.ScanInput{TableName: &c.table, ConsistentRead: aws.Bool(c.opts.ConsistentRead)}
```
polinasok: this would be easier to read if you had a newline before ConsistentRead

docstore/awsdynamodb/query_test.go
```diff
@@ -58,7 +58,8 @@ func TestPlanQuery(t *testing.T) {
 		return m
 	}
 
-	cmpopt := cmp.AllowUnexported(dynamodb.ScanInput{}, dynamodb.QueryInput{}, dynamodb.AttributeValue{})
+	// Ignores the ConsistentRead field from both QueryInput and ScanInput.
```
polinasok: Why? What would happen if you did not?

docstore/awsdynamodb/dynamo.go
```diff
@@ -95,6 +95,18 @@ type Options struct {
...
+
+	// If true, a strongly consistent read is used whenever possible, including
+	// get, query, scan, etc.; default to false, where an eventually consistent
+	// read is used.
+	//
+	// Not all read operations support this mode however, please check the official
+	// DynamoDB documentation for more details.
+	//
+	// The native client for DynamoDB uses this option in a per-action basis, if
+	// you need the flexibility to run both modes on the same colleciton, create
```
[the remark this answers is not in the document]
shantuo: Done.

docstore/awsdynamodb/dynamo.go
```diff
@@ -95,6 +95,18 @@ type Options struct {
...
+	// two colletions with different mode.
```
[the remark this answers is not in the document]
shantuo: Done.

docstore/awsdynamodb/dynamo.go
```diff
@@ -200,7 +212,7 @@ func (c *collection) batchGet(ctx context.Context, gets []*driver.Action, errs [
 	}
 	ka := &dyn.KeysAndAttributes{
 		Keys:           keys,
-		ConsistentRead: aws.Bool(true),
+		ConsistentRead: aws.Bool(c.opts.ConsistentRead),
```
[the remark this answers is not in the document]
shantuo: I think the old one should be a bug (or we wanted to make sure the test be deterministic). This does change the behavior slightly but should be the correct thing to do.

docstore/awsdynamodb/dynamo_test.go
shantuo: It will result in error in those cases that queries the global secondary index. I think that is the correct behavior since we don't want to make it "work" silently by ignoring the option when the user expects a `ConsistentRead`. And the error message returned is clear enough to point to this issue.

docstore/awsdynamodb/query.go
shantuo: Done.

docstore/awsdynamodb/query_test.go
shantuo: This is a dynamo specific test that tests the `planQuery` function only that is only relevant to the expressions.

The `ConsistentRead` field was nil (we didn't set it), now it is a false `*Bool`. The other way is to add that field to every `want` field but we are not testing it and it just makes the test longer. WDYT?

polinasok: Let's try IgnoreFields like we discussed in person.

docstore/awsdynamodb/dynamo_test.go
shantuo: I think this comment here is fine. I explained that in the option comment.

docstore/awsdynamodb/query_test.go
shantuo: Done.

shantuo:
PTAL

Issue #2611, opened 2019-07-26
docstore: Query.Update should retry on revision check failure


jba:
Split off from #2458 

Non-mongo implementations of Query.Update should retry documents that fail a revision check. 

For example, say the call
```
Query().Where("Score", ">", 100).Update(ctx, docstore.Mods{"Score", 100})
```
returns 100 documents. During the update of those documents, three fail because their revisions have changed. The driver should retry those three.

That involves either repeating the query (narrowed, preferably, to the keys of the failed documents), or retrieving the documents by key and evaluating the filters on the client.

We also need to consider what happens if a document is deleted between the time it is retrieved and updated.

shantuo:
The current API is hidden by #2637 , add them back when this is done.

PR #2637, opened 2019-08-08
docstore: remove Query.Update and .Delete API's

shantuo:
Removes these two API's from the portable type in docstore alpha and wait for #2611 to add them back.


vangent:
Changes so far LGTM, but tests are failing -- you missed a few uses.

It's fine to delete the code, but when you file the issue for what the next steps would be to restore this functionality, please include a link to the existing code (i.e., to a commit before it is deleted).

shantuo:
Yes I mentioned it in the issue.

PR #2645, opened 2019-08-10
mysql: remove mysql:// prefix when passing DSN to driver

### LINKED ISSUE #2644: mysql: question regarding mysql.Open in a dockerized environment
### Describe the bug

I'm having trouble trying to connect with `mysql.Open(...)` in a development environment that runs MySQL via Docker Compose.

The documentation in `gocloud.dev/mysql` describes to open a local MySQL connection with the `mysql://` scheme/prefix. However, when running this in my sample environment, I cannot find any connection string to successfully connect to MySQL in a Docker container.

E.g. the following connection string should work, if I understood correctly: `mysql://user:password@tcp(localhost:3306)/testdb`. Notice that the `tcp(...)` is necessary, otherwise the driver will try to connect via a local socket, which won't connect to the dockerized MySQL (as described [in their README](https://github.com/go-sql-driver/mysql/blob/master/README.md#protocol)).

From what I see, the problem is that the underlying MySQL driver for Go at `github.com/go-sql-driver/mysql` expects a connection string in the form of `[user[:password]@][net[(addr)]]/dbname[?param1=value1&paramN=valueN]`, while you are passing the original URL to it, including the `mysql://` prefix [here](https://github.com/google/go-cloud/blob/master/mysql/mysql.go#L49).

### To Reproduce

[Here's an example program](https://play.golang.org/p/L3mHegeNR1M) that tries to connect to MySQL via TCP on `localhost:3306`. The MySQL server can be exectuted via Docker Compose with a `docker-compose.yml` like:

```
version: '3'
services:
  mysql:
    image: mysql:5.7
    hostname: mysql
    restart: always
    environment:
      - reschedule=on-node-failure
      # - MYSQL_ROOT_PASSWORD=${MYSQL_ROOT_PASSWORD}
      - MYSQL_ALLOW_EMPTY_PASSWORD=true
      - MYSQL_DATABASE=
    command: mysqld --max_allowed_packet=30M
    ports:
      - 3306:3306
```

When you run that, you'll end up with:

```
$ go run main.go -urn="mysql://root@tcp(localhost:3306)/"
2019/08/10 16:23:08 unable to ping: Error 1045: Access denied for user 'mysql'@'172.18.0.1' (using password: YES)
```

As you can see, MySQL sees `mysql` as the user name, not `root`. That's because, as far as I understand, the Go driver receives an invalid connection string `mysql://root@tcp(localhost:3306)/`. Instead, it should get `root@tcp(localhost:3306)/` (excluding the `mysql://` prefix).

## Expected behavior

I expect to be able to successfully connect:

```
$ go run main.go -urn="mysql://root@tcp(localhost:3306)/"
$
```

### Version

I'm using Go 1.12.7 with `gocloud.dev v0.16.0`.

Edit: Fixed the example program.

olivere:
This commit removes the `mysql://` prefix before passing the DSN to the
MySQL driver for Go. The latter requires the DSN in a format like:

```
[username[:password]@][protocol[(address)]]/dbname[?param1=value1&...&paramN=valueN]
```

as described in
https://github.com/go-sql-driver/mysql/blob/master/README.md#dsn-data-source-name.

It doesn't accept a `mysql://` prefix and, when used, will not parse the
DSN correctly.

Fixes #2644

### DIFF
--- a/mysql/mysql.go
+++ b/mysql/mysql.go
@@ -20,6 +20,7 @@ import (
 	"database/sql"
 	"database/sql/driver"
 	"net/url"
+	"strings"
 
 	"github.com/go-sql-driver/mysql"
 	"gocloud.dev/internal/openurl"
@@ -46,7 +47,7 @@ func (*URLOpener) OpenMySQLURL(ctx context.Context, u *url.URL) (*sql.DB, error)
 }
 
 func openWithURL(url *url.URL) (*sql.DB, error) {
-	return sql.OpenDB(connector{dsn: url.String()}), nil
+	return sql.OpenDB(connector{dsn: strings.TrimLeft(url.String(), Scheme+"://")}), nil
 }
 
 type connector struct {


vangent:
Thanks!

PR #2655, opened 2019-08-22
all: Update golang.org/x/xerrors for Go 1.13rc1

jayschwa:
The older version of xerrors does not build with Go 1.13rc1.

xerrors issue fixed in the new version: https://golang.org/issues/32246

I tested this change by running `go test ./...` with both Go 1.12.9 and
Go 1.13rc1.


jayschwa:
@googlebot I signed it!

PR #2661, opened 2019-09-05
internal: add go 1.13 build

### LINKED ISSUE #2662: all: support go 1.13

vangent:
Fixes #2662 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,15 +49,14 @@ jobs:
   include:
     - go: "1.11.x"
       os: linux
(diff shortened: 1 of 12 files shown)


.travis.yml
```diff
@@ -49,15 +49,14 @@ jobs:
   include:
     - go: "1.11.x"
```
eliben: Should we drop 1.11 now?

vangent: We could eventually. I don't think there's much to gain by dropping it now.

PR #2679, opened 2019-09-29
runtimevar: Latest must check for good snapshot before selecting on passed in ctx.

alexsn:
Selecting on both <-haveGood and <-ctx.Done() is non deterministic so on
ctx.Done() selection (say ctx is already done to prevent blocking) an empty
snapshot will be returned by Latest even though a good snapshot might be available.

### DIFF
--- a/runtimevar/runtimevar.go
+++ b/runtimevar/runtimevar.go
@@ -106,8 +106,8 @@ type Variable struct {
 	backgroundCancel context.CancelFunc
 	backgroundDone   chan struct{}
 
-	// haveGood is closed when we get the first good value for the variable.
-	haveGood chan struct{}
+	// haveGoodCh is closed when we get the first good value for the variable.
+	haveGoodCh chan struct{}
 	// A reference to changed at the last time Watch was called.
 	// Not protected by mu because it's only referenced in Watch, which is not
 	// supposed to be called from multiple goroutines.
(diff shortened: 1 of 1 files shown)


runtimevar/runtimevar.go
```diff
@@ -266,15 +276,10 @@ func (c *Variable) Latest(ctx context.Context) (Snapshot, error) {
 // CheckHealth returns an error unless Latest will return a good value
 // without blocking.
 func (c *Variable) CheckHealth() error {
-	haveGood := false
-	select {
-	case <-c.haveGood:
-		haveGood = true
-	default:
-	}
+	haveLatest := c.haveLatest()
```
vangent: Can you keep the variable/function named `haveGood` (here and above)? I think that's more accurate (and matches the channel) than `haveLatest`.

alexsn: I wanted to do that but *Variable holds a haveGood channel member so we can't add a method with the same name.

vangent: How about renaming the `haveGood` channel member `haveGoodCh`?

runtimevar/runtimevar.go
```diff
@@ -248,16 +257,17 @@ func (c *Variable) background(ctx context.Context) {
...
 // Latest returns ErrClosed if the Variable has been closed.
 func (c *Variable) Latest(ctx context.Context) (Snapshot, error) {
-	var haveGood bool
+	var haveLatest bool
 	select {
 	case <-c.haveGood:
-		haveGood = true
+		haveLatest = true
 	case <-ctx.Done():
+		haveLatest = c.haveLatest()
```
vangent: I see what you're doing here, but I think it might be more clear like this:

```
haveGood := c.haveGood()
if !haveGood {
  select {
    case <-c.haveGood:
      haveGood = true
    case <-ctx.Done():
      // We don't return ctx.Err().
  }
  ...
```

WDYT?

alexsn: Sure, note that we can't have `c.haveGood()` due to `<-c.haveGood`. You ok with `.haveLatest()`?

PR #2680, opened 2019-09-30
pubsub/natspubsub: add nats streaming support

### LINKED ISSUE #1559: pubsub: support NATS streaming
Now that we have support for NATS, it would be nice to support NATS Streaming with the same driver. It would give users a choice between at-most-once and at-least-once NATS implementations.

@derekcollison would that be a lot work? Is that something you want to take on? (At your convenience, of course.)

srikrsna:
**pubsub/natspubsub: add nats streaming support**

Fixes #1559 

Adds support for NATS Streaming Server.


pubsub/natspubsub/nats_streaming.go
```diff
@@ -0,0 +1,277 @@
...
+	"sync"
+
+	"github.com/nats-io/nats.go"
+	"github.com/nats-io/stan.go"
+
+	"gocloud.dev/gcerrors"
+	"gocloud.dev/pubsub"
+	"gocloud.dev/pubsub/driver"
+)
+
```
vangent: There's no URL openers defined here...?

Also, a fair amount of the code is already duplicated with `nats.go` (and adding the URL opener will increase it). It might be simpler to use the same file, the same struct, and do conditionals where needed?

E.g.,

```
type topic struct {
        // Exactly one of nc, snc will be non-nil.
	nc   *nats.Conn
        snc stan.Conn
	subj string
}
```

The URL opener could have a URL param to indicate to use streaming, or a separate environment variable (`NATS_STREAMING_SERVER_URL`).

pubsub/natspubsub/nats_streaming.go
```diff
@@ -0,0 +1,277 @@
...
+	sub      stan.Subscription
+	ack      bool
+	messages []*stan.Msg
+	sync.Mutex
+}
+
+// OpenStreamingSubscription returns a *pubsub.Subscription representing a NATS Streaming subscription.
+// The subject is the NATS Subject to subscribe to; for more info, see
+// https://nats.io/documentation/writing_applications/subjects.
+func OpenStreamingSubscription(nc stan.Conn, subject, queue string, oo ...stan.SubscriptionOption) (*pubsub.Subscription, error) {
```
vangent: Please create a separate `Options` struct, in case we need to add options that are not `stan.SubscriptionOptions`.

pubsub/natspubsub/nats_streaming.go
```diff
@@ -0,0 +1,277 @@
...
+	for _, o := range oo {
+		if err := o(&opt); err != nil {
+			return nil, err
+		}
+	}
+
+	s := &stmSub{nc: nc, ack: opt.ManualAcks}
+	handler := func(msg *stan.Msg) {
+		s.Lock()
+		s.messages = append(s.messages, msg)
```
vangent: Hmm, this seems dangerous. Is there no back pressure? I.e., what happens if a subscription is created, but never calls `Receive`? This seems like it will just keep reading messages and adding them to an in-memory queue without any bound. If `!ManualAcks`, and the subscription is eventually closed, those messages are lost.

I'm not sure what the required semantics for `handler` is supposed to be, but an alternative might be to send the message to a channel with buffersize=1, and read from that channel in `ReceiveBatch` below, so that `handler` is blocked until the message is actually needed.

pubsub/natspubsub/nats_streaming.go
```diff
@@ -0,0 +1,277 @@
...
+
+func streamingMessageAsFunc(msg *stan.Msg) func(interface{}) bool {
+	return func(i interface{}) bool {
+		p, ok := i.(**stan.Msg)
+		if !ok {
+			return false
+		}
+		*p = msg
+		return true
+	}
```
vangent: Please add a test that runs the conformance tests. See `nats_test.go` for an example.

pubsub/natspubsub/nats_streaming.go
srikrsna: I think a better way would be to always allow manual acknowledge and set the channel size based on the MaxInflight value. It will align better with pubsub semantics

pubsub/natspubsub/nats_streaming.go
srikrsna: Should it contain all the options of `stan.SubscriptionOptions` as fields or `stan.SubscriptionOptions` as a field?

vangent:
Closing as this is quite stale at this point. Feel free to resend if you want to continue.

PR #2683, opened 2019-10-01
blob/azureblob: Add support for User Delegation Credentials

### LINKED ISSUE #2681: blob/azureblob: Support for User Delegation Keys

The problem we're trying to solve is that we need to generate pre-signed URLs for storage accounts without using storage access keys. Storage access keys cannot be scoped to specific containers or blobs within an account and are long-lived.

To generate pre-signed URLs (SAS in the context of Azure), one currently needs to provide a `*azblob.SharedKeyCredential` in the options when opening a bucket.

Instead of the `*azblob.SharedKeyCredential`, I would like to be able to pass in a `azblob.StorageAccountCredential` which is an interface implemented by `azblob.SharedKeyCredential` and `azblob.UserDelegationCredential` in [`github.com/Azure/azure-storage-blob-go`](github.com/Azure/azure-storage-blob-go) since [this PR](https://github.com/Azure/azure-storage-blob-go/pull/131) which is part of the module since v0.7.0. The `azblob.UserDelegationCredential` could then be obtained using Azure AD OAuth tokens and would eliminate the need for storage access keys.

We experimented with using storage access keys, but that doesn't align well with our permission model at the moment. An alternative would be to generate pre-signed URLs using the Azure SDK directly which would be bad for portability and increase code complexity.

seiffert:
Fixes #2681 

This PR adds support for User Delegation credentials in the options passed to `OpenBucket`. By using the `azblob.StorageAccountCredential` interface, it allows the use of an `*azblob.UserDelegationCredential` instead of an `*azblob.SharedKeyCredential` for generating pre-signed blob URLs (shared access signatures or SAS in the context of Azure).

The usage of User Delegation credentials allows applications to sign such URLs without requiring access to the storage account access keys which are pretty privileged.   The creation of User Delegation keys is protected by Azure AD and only requires a sub-set of permissions.

[`github.com/Azure/azure-storage-blob-go`](github.com/Azure/azure-storage-blob-go) supports user delegation credentials since [this PR](https://github.com/Azure/azure-storage-blob-go/pull/131) which is part of the module since v0.7.0.

**UPDATE:**
The change to the type of the `Options.Credential` attribute was detected as breaking change by CI. Because the change only replaces the concrete implementation with an interface and all internal usages rely on the methods contain in the interface, I believe that this is fine and added `BREAKING_CHANGE_OK` to the commit message.

**Azure Documentation on User Delegation SAS:**
* [SAS Overview and comparison of SAS types](https://docs.microsoft.com/en-us/rest/api/storageservices/delegate-access-with-shared-access-signature)
* [Create a user delegation SAS (preview)](https://docs.microsoft.com/en-us/rest/api/storageservices/delegate-access-with-shared-access-signature)


seiffert:
Do you have plans for cutting a new release in the near future?

vangent:
Working on `v0.18.0` now.

PR #2685, opened 2019-10-03
mysql: use TrimPrefix instead of TrimLeft

vangent:
yueeong [pointed out](https://github.com/google/go-cloud/commit/a5f6c05207b2d477d31267b24edd23331ac6f228#commitcomment-35353748) that this use of `TrimLeft` should be `TrimPrefix`. `TrimLeft` trims all character in the "cutset" of its second argument, so `TrimLeft(s, "mysql://")` would remove any leading `m`, `y`, `s`, etc.. That's not what we want.


vangent:
> Could you please add more details in the PR description?

Sorry, yes -- done.

PR #2686, opened 2019-10-08
runtimevar/gcpruntimeconfig: Use non-deprecated gRPC APIs

bvwells:
Use non-deprecated gRPC APIs in runtimevar/gcpruntimeconfig package.


bvwells:
Thanks for th merge @vangent!

PR #2687, opened 2019-10-08
blob/gcsblob: Use non-deprecated gcs blob storage APIs

bvwells:
Use non-deprecated gcs blob storage APIs


bvwells:
Thanks @vangent - hope this helps!

Issue #14, opened 2018-05-24
server: add request metrics


zombiezen:
**EDIT 2019-07-09 Adding more context to this issue:**

### Is your feature request related to a problem? Please describe.

Right now, a server using the Go CDK server package will automatically capture OpenCensus traces, but will not capture request metrics like `opencensus.io/http/server/request_count`. Go CDK servers should send these metrics out of the box.

### Describe the solution you'd like

A new field on `*server.Options` that to initialize the metrics exporter (#13), and for the server's request handler to use [`ochttp.Handler`](https://godoc.org/go.opencensus.io/plugin/ochttp#Handler).

zombiezen:
@clausti, are you still working on this issue? It might be a good one for @jba to take since he's working on metrics.

clausti:
I have not worked on this, no. I don't mind if someone else takes it.

Tazer:
When looking around a little there seem to be a http plugin for open census **ochttp**
https://github.com/census-instrumentation/opencensus-go/blob/master/examples/http/helloworld_server/main.go#L89

Not sure if it even would be a possibility to use that plugin? Anyway they have some defined standard metrics 

Server:
https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ochttp/stats.go#L226
Client: 
https://github.com/census-instrumentation/opencensus-go/blob/master/plugin/ochttp/stats.go#L137

jba:
You should use that plugin. See https://code.googlesource.com/google-api-go-client/+/refs/heads/master/transport/http/dial.go.

Issue #13, opened 2018-05-24
server: install monitoring exporter


zombiezen:
**EDIT 2019-07-09 Adding more context to this issue:**

### Is your feature request related to a problem? Please describe.

Right now, a server using the Go CDK server package can configure the OpenCensus trace exporter, but not the [metric exporter](https://godoc.org/go.opencensus.io/metric/metricexport#Exporter).

### Describe the solution you'd like

A new field on `*server.Options` to initialize the metrics exporter.

zombiezen:
@clausti, are you still working on this issue? It might be a good one for @jba to take since he's working on metrics.

jba:
I'd rather we spread the knowledge, but I'll do it if no one else wants to.

clausti:
I haven't worked on this, no. I don't mind if someone else takes it.

renilJoseph:
hi. i would like to work on this.

jba:
Great! I believe the right starting point is [`ochttp.Handler`](https://godoc.org/go.opencensus.io/plugin/ochttp#Handler). Let us know if you have any questions or need more help.

bvwells:
@zombiezen, @jba - I've added an initial implementation of OpenCensus stats exporter. Hope this is okay and what you were after. 

At the moment, it is using the default http server views and no propagation is set. I can put some tests around this if needed.

eliben:
Thanks for the contribution, @bvwells 
We're not looking at adding features to the server at this time - I should have closed this issue earlier, sorry for the confusion.

bvwells:
@eliben, no problems. I guess it might be worth waiting for OpenTelemetry....

PR #2689, opened 2019-10-08
server: Add monitoring exporter to server

bvwells:
Add monitoring exporter to server to address issue #13, #14.


eliben:
Thanks! See comment in #13 for closure

PR #2691, opened 2019-10-15
blob: allow setting Content-Type for PUT requests

zombiezen:
This is a follow-on to #2630. In practice, most applications need to set the Content-Type on the blob they upload. The current mode of operation causes GCS to reject any PUT request that has a Content-Type header, which is hard to debug (just an HTTP 403).

I am confident in the GCS implementation, but I have not manually tested the S3 or Azure Blob implementations. I require assistance in regenerating the replay logs.

@shantuo and @vangent for visibility.


blob/drivertest/drivertest.go
```diff
@@ -2066,7 +2066,14 @@ func testSignedURL(t *testing.T, newHarness HarnessMaker) {
 	}
 	getURLNoParamsURL.RawQuery = ""
 	getURLNoParams := getURLNoParamsURL.String()
-	putURL, err := b.SignedURL(ctx, key, &blob.SignedURLOptions{Method: http.MethodPut})
+	const (
+		goodContentType = "text/plain"
```
vangent: The "good" and "bad" descriptions confused me a bit. Maybe `allowedContentType` and `disallowedContentType`? Not sure if that's better.

blob/drivertest/drivertest.go
```diff
@@ -2087,24 +2094,30 @@ func testSignedURL(t *testing.T, newHarness HarnessMaker) {
 	// PUT the blob. Try with all URLs, only putURL should work.
```
vangent: Can you extend this comment to say "only putURL should work, and only when the content type matches" or something?

Or alternatively, put comments by each testcase explaining why they fail.

blob/drivertest/drivertest.go
```diff
@@ -2087,24 +2094,30 @@ func testSignedURL(t *testing.T, newHarness HarnessMaker) {
...
 		wantSuccess bool
 	}{
-		{http.MethodGet, getURL, false},
-		{http.MethodDelete, deleteURL, false},
-		{http.MethodPut, putURL, true},
+		{http.MethodGet, "", getURL, false},
+		{http.MethodDelete, "", deleteURL, false},
+		{http.MethodPut, goodContentType, putURL, true},
+		{http.MethodPut, badContentType, putURL, false},
+		{http.MethodPut, "", putURL, false},
```
vangent: This only tests using a content type in the URL. Do you think it's worth adding a second putURL with ContentType="" and adding testcases with that?

blob/fileblob/fileblob.go
```diff
@@ -693,6 +693,7 @@ func (h *URLSignerHMAC) URLFromKey(ctx context.Context, key string, opts *driver
 	q.Set("obj", key)
 	q.Set("expiry", strconv.FormatInt(time.Now().Add(opts.Expiry).Unix(), 10))
 	q.Set("method", opts.Method)
+	q.Set("contentType", opts.ContentType)
```
vangent: The fileblob tests fail with this change, I think you need to verify the content type in the handler?

blob/azureblob/azureblob.go
```diff
@@ -661,6 +661,7 @@ func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedU
 		ContainerName: b.name,
 		BlobName:      srcBlobParts.BlobName,
 		Permissions:   perms.String(),
+		ContentType:   opts.ContentType,
```
vangent: azure tests are failing:

```
    --- FAIL: TestConformance/TestSignedURL (0.67s)
        setup.go:79: Recording into golden file testdata/TestConformance/TestSignedURL.replay
        drivertest.go:2120: PUT with PUT URL (content-type="application/octet-stream") got status code 201, want 2xx? false
        drivertest.go:2122: 
        drivertest.go:2120: PUT with PUT URL (content-type="") got status code 201, want 2xx? false
        drivertest.go:2122: 
```

vangent:
> Changes made. You'll need to regenerate the logs again. (I did a force-push to fix my commit metadata, so you may have to do some Git trickery on your end to re-fetch the branch.)

I don't see the changes you made yet ...?

blob/azureblob/azureblob.go
vangent: What do you mean by "they don't really conform to our contract for PUT URLs"? Is there something else other than the content-type?

We could just relax the contract, e.g.:

```
        // ContentType specifies the Content-Type HTTP header the user agent is
        // permitted to use in the PUT request. It must match exactly. Empty means
        // the Content-Type header is not permitted.
```
--

```
        // ContentType specifies the Content-Type HTTP header the user agent is
        // expected to use in the PUT request. Some providers enforce it, others do not.
        // When enforced, it must match exactly, and an empty string means the
        // Content-Type header is not permitted.
```

or something like that.

zombiezen: I'm referring to the issue you identified originally in #2630 about Azure requiring an additional `x-ms-blob-type` header which the others don't.

zombiezen:
> I don't see the changes you made yet ...?

Sorry. Forgot to run `git commit`.

blob/azureblob/azureblob.go
vangent: Ah yes, I forgot about that.

Let's:
-- return a `NotImplemented` error for `azblob` if the `ContentType` field is set
-- check for `NotImplemented` in the conformance test to skip the test
-- update the comment about `PUT` not being fully conformant at the top of `azblob.go` to also note that `opts.ContentType` can't be set.

WDYT?

zombiezen: It doesn't really fit the spirit of the contract: if the `ContentType` field isn't set, that's supposed to indicate that a `Content-Type` header **can't** be set on the `PUT` request (this is the behavior that S3 and GCS enforce). Thus, my recommendation to just forbid `PUT` signed URLs on Azure, because I think it's too far away from the contract the others align with.

vangent: That's a backward incompatible change though; there may be users using `PUT` signed URLs today.

zombiezen: Sure, but this is functionality that was only introduced in this last release IIUC.

The reason I'm pushing is because it significantly weakens what guarantees can be made of the `ContentType` field. Since Azure doesn't enforce the `ContentType` field's behavior at all, then the Azure implementation has to go out of its way to say "if you use `PUT` signed URLs at all, it's non-compliant and behaves in a totally different way than the other implementations".

If you really feel strongly that we should keep the behavior, I'll drop it to move this PR forward, since I'd very much like to see this fix for S3 and GCS. I bumped into the problem the instant I tried to use the new `PUT` URLs on GCS and the failure modes are really hard to debug from the GCS side. I fear other users will encounter the same issues and get frustrated with the CDK.

vangent: What do you think about making `ContentType` be a `*string` instead of a `string`, where:

* `nil` (the default) means "no assertion about ContentType"
* a non-`nil` value means an assertion ("" asserts no ContentType header)
* There's a comment that says that not all providers implement verification of non-nil `ContentType` values (which would be true for Azure).

zombiezen: Alright, I've added a new field `EnforceAbsentContentType`, WDYT? It's spiritually pretty close to what you suggested, but it has two advantages:

1. It doesn't require the API surface awkwardness of `new(string)` or a helper function.
2. It makes it easier to communicate that a `(ContentType, EnforceAbsentContentType)` of `("", false)` may still perform enforcement, depending on implementation (i.e. GCS and S3). `("", true)` will always perform enforcement or will raise an error (Azure). Most implementations will never need to care about the distinction and can just read from `ContentType`.

Thanks for the patience here. Work has been busy, and I wanted to wait until I had adequate time to ponder the correct API.

blob/azureblob/azureblob.go
```diff
@@ -661,6 +668,7 @@ func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedU
...
```
vangent: Should `opts.ContentType` be here? IIUC it can never be anything other than the empty string.

blob/azureblob/azureblob.go
```diff
@@ -636,7 +640,10 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
 // SignedURL implements driver.SignedURL.
 func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error) {
 	if b.opts.Credential == nil {
-		return "", errors.New("to use SignedURL, you must call OpenBucket with a non-nil Options.Credential")
+		return "", errors.New("azureblob: to use SignedURL, you must call OpenBucket with a non-nil Options.Credential")
+	}
+	if opts != nil && (opts.ContentType != "" || opts.EnforceAbsentContentType) {
```
vangent: Nit: you don't need to check for `opts == nil` here, the concrete type always passes a non-nil `driver.SignedURLOptions`.

blob/blob.go
```diff
@@ -904,9 +909,26 @@ type SignedURLOptions struct {
...
+
 	// Method is the HTTP method that can be used on the URL; one of "GET", "PUT",
 	// or "DELETE". Defaults to "GET".
 	Method string
+
+	// ContentType specifies the Content-Type HTTP header the user agent is
+	// permitted to use in the PUT request. It must match exactly. See
+	// EnforceAbsentContentType for behavior when ContentType is the empty string.
+	//
+	// Must be empty for non-PUT requests.
```
vangent: Maybe add something here to indicate that this is not supported by all drivers?

blob/blob.go
```diff
@@ -904,9 +909,26 @@ type SignedURLOptions struct {
...
+	ContentType string
+
+	// If EnforceAbsentContentType is true and ContentType is the empty string,
+	// then PUTing to the signed URL will fail if the Content-Type header is
+	// present. If EnforceAbsentContentType is false and ContentType is the empty
+	// string, then PUTing without a Content-Type header will succeed, but it is
+	// implementation-specific whether providing a Content-Type header will fail.
```
vangent: Why is it implementation-specific? Can't we make it so that the behavior is consistent across providers, but sometimes you get Unimplemented?

blob/blob.go
vangent: (i.e., IIUC azureblob will return an Unimplemented error if you put something here).

blob/blob.go
vangent: In particular, the backwards-compatible behavior of `("", false)` puts you in the "implementation-specific" bucket, which is unfortunate. It should instead not enforce anything, similar to the current release. 

How about this:

```
// ContentType specifies the Content-Type HTTP header the user agent is
// permitted to use in the PUT request. It must match exactly. See
// EnforceAbsentContentType for behavior when ContentType is the empty string.
// Some providers do not support this and will return an Unimplemented error.
//
// Must be empty for non-PUT requests.

and

// EnforceAbsenceContentType determines behavior when ContentType is the empty string.
//
// If true, PUTing to the signed URL will fail if any Content-Type header is provided.
// Some providers do not support this and will return an Unimplemented error.
//
// If false, no check of the `Content-Type` header is performed.
//
// Must be false for non-PUT requests.
```

IIUC, the Azure implementation matches this already, but the others would need to be updated to look at EnforceAbsentContentType.

blob/azureblob/azureblob.go
zombiezen: Done.

blob/azureblob/azureblob.go
zombiezen: Oops, good catch. A carryover from an earlier revision.

blob/blob.go
zombiezen: Added docs here and in driver to that effect.

blob/blob.go
zombiezen: > In particular, the backwards-compatible behavior of `("", false)` puts you in the "implementation-specific" bucket, which is unfortunate. It should instead not enforce anything, similar to the current release.

I understand the behavior is unfortunate, but that's actually how I encountered this problem in a real-world app. Your understanding is incorrect: GCS and S3 always enforce the Content-Type header. If you use v0.18.0 of CDK and create a signed PUT URL with no Content-Type and then try to upload with a Content-Type header set to something, it will always fail. There is no way AFAICT to disable the behavior, nor do I think it's actually desirable from a security perspective.

The end-user impact of this is really hard to debug, too, since GCS necessarily gives a generic HTTP 403 because it's supposed to be indistinguishable from any other invalid signed URL.

> IIUC, the Azure implementation matches this already, but the others would need to be updated to look at EnforceAbsentContentType.

You'll notice that I don't even check `EnforceAbsentContentType` in GCS and S3 drivers, since there's nothing meaningful they can do with the information.

vangent: Ah, got it -- I misunderstood the current behavior. This looks OK.

vangent:
Can you review my non-golden-file changes from the last commit? I had to make some changes to get tests to pass. In particular, it was using some possibly-empty URLs for some tests, and I dropped a case that checked for `Unimplemented` that should always be implemented.

zombiezen:
LGTM, thanks Robert!

PR #2696, opened 2019-10-28
pubsub: added capacity to slice creation, when capacity is known

AdityaHarindar:
Micro-optimization to reduce memory allocations, copying.


AdityaHarindar:
Ah, I didn't notice there was a specific convention being followed here, feel free to close this if you feel this optimization is unnecessary.

vangent:
Thanks for the contribution idea, but yes, I think that's the case here.

PR #2705, opened 2019-11-13
all: Update dependencies.

vangent:
I had to update the `update_deps.sh` script to handles the new column in `allmodules`.

And, I removed a test about bad subjects in NATS, apparently the underlying library behavior changed there.


vangent:
Note: This PR builds on #2704, I will rebase after that one is submitted.

PR #2707, opened 2019-11-16
blob/gcsblob: Remove line for duplicated function comment

bannzai:
Remove line for duplicated function comment.


bannzai:
@googlebot I signed it!

PR #2717, opened 2019-12-22
blob/azureblob: Add support for specifying Azure Cloud Environment 

### LINKED ISSUE #2716: blob/azureblob: Add support for specifying different Azure cloud environments (Government, China, Germany) to default URLOpener

Currently the azureblob package reads the Azure Storage Account name and a key or SAS token from the environment variables. However, it always attempts to open the URL as "<account_name>.blob.core.windows.net". Furthermore, this hardcoded URL is embedded in the non-exported openBucket function, which generates the azureblob.bucket that implements all of the driver functions. As a result, it doesn't seem possible to write my own URLOpener implementation to do this without rewriting a very large portion of the package, including the underlying bucket and driver implementations. 

I would like to be able to specify an Azure cloud environment such as "public", "usgovernment", "german", etc either as an environment variable or somewhere in the azblob:// URL. The blob opener would then attempt to access the account at the Azure storage endpoint for that environment, rather than the standard public endpoint. For example, "usgovernment" would target "<account_name>.blob.core.usgovcloudapi.net". 

If it is too difficult or intrusive to write this functionality into the package itself, it would be nice if the target URL were somehow exposed for developers to modify or extend it. 

Alternatively, if the azureblob bucket structure were exposed, it would be much easier to write a URLOpener that targets a different URL and then instantiates the bucket with the new ServiceURL.

Strife96:
Fixes #2716
Updated 12/24/19 to reflect changes in implementation:
The current go-cloud azureblob implementation always targets "blob.core.windows.net" when attempting to open a bucket, making USGovernment, China, and German cloud buckets inaccessible. 

With this addition, users can specify the Azure Cloud Environment they would like to target by setting  the AZURE_STORAGE_DOMAIN environment variable to a value like "blob.core.govcloudapi.net", "blob.core.chinacloudapi.cn", etc. allowing them to access the different Azure blob storage endpoints for each of these clouds. 

If users leave the AZURE_STORAGE_DOMAIN variable unset, the bucket opener will target "blob.core.windows.net" by default, which is the public domain that most users are likely already using. Any changes made by this request should be transparent to end users who have not been setting this environment variable. 

If you have any questions or feedback regarding this change, please let me know! Thank you! 🙂

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -182,8 +182,8 @@ func openerFromEnv(accountName AccountName, accountKey AccountKey, sasToken SAST
 		AccountName: accountName,
 		Pipeline:    NewPipeline(credential, azblob.PipelineOptions{}),
 		Options: Options{
-			Credential: storageAccountCredential,
-			SASToken:   sasToken,
+			Credential:       storageAccountCredential,
+			SASToken:         sasToken,
 			CloudEnvironment: cloudEnv,
 		},
 	}, nil
@@ -227,16 +227,17 @@ type AccountKey string
 type SASToken string
 
 // CloudEnvironment is an Azure Cloud Environment to target (i.e. AzureCloud, AzureUSGovernment, AzureChinaCloud).
-// It is read from the AZURE_CLOUD_ENVIRONMENT environment variable. 
+// It is read from the AZURE_CLOUD_ENVIRONMENT environment variable.
 type CloudEnvironment string
 
-// Mappings from the AZURE_CLOUD_ENVIRONMENT value to their respective blob storage domains. 
+// Mappings from the AZURE_CLOUD_ENVIRONMENT value to their respective blob storage domains.
 var cloudEnvMap = map[CloudEnvironment]string{
-	CloudEnvironment("AzureCloud"): "blob.core.windows.net",
+	CloudEnvironment("AzureCloud"):        "blob.core.windows.net",
 	CloudEnvironment("AzureUSGovernment"): "blob.core.usgovcloudapi.net",
-	CloudEnvironment("AzureChinaCloud"): "blob.core.chinacloudapi.cn",
-	CloudEnvironment("AzureGermanCloud"): "blob.core.cloudapi.de",
+	CloudEnvironment("AzureChinaCloud"):   "blob.core.chinacloudapi.cn",
+	CloudEnvironment("AzureGermanCloud"):  "blob.core.cloudapi.de",
 }
+
 // DefaultAccountName loads the Azure storage account name from the
 // AZURE_STORAGE_ACCOUNT environment variable.
 func DefaultAccountName() (AccountName, error) {
@@ -267,12 +268,12 @@ func DefaultSASToken() (SASToken, error) {
 	return SASToken(s), nil
 }
 
-// DefaultCloudEnvironment loads the desired Azure Cloud to target from 
-// the AZURE_CLOUD_ENVIRONMENT environment variable. 
+// DefaultCloudEnvironment loads the desired Azure Cloud to target from
+// the AZURE_CLOUD_ENVIRONMENT environment variable.
 func DefaultCloudEnvironment() (CloudEnvironment, error) {
 	s := os.Getenv("AZURE_CLOUD_ENVIRONMENT")
-	if s == "" {
-		return CloudEnvironment("AzureCloud"), nil
+	if _, exists := cloudEnvMap[CloudEnvironment(s)]; s != "" && !exists { // If AZURE_CLOUD_ENVIRONMENT is set but the value is not recognized
+		return "", errors.New("azureblob: environment variable AZURE_CLOUD_ENVIRONMENT has invalid value... Must be one of (AzureCloud, AzureUSGovernment, AzureChinaCloud, AzureGermanCloud)")
 	}
 	return CloudEnvironment(s), nil
 }
@@ -323,9 +324,9 @@ func openBucket(ctx context.Context, pipeline pipeline.Pipeline, accountName Acc
 	}
 	if opts == nil {
 		opts = &Options{}
-	} 
+	}
 	if opts.CloudEnvironment == "" {
-		// If opts waas nil or no Cloud Environment was specified, default to public Azure cloud endpoint.
+		// If opts was nil or no Cloud Environment was specified, default to public Azure cloud endpoint.
 		opts.CloudEnvironment = CloudEnvironment("AzureCloud")
 	}
 	blobURL, err := url.Parse(fmt.Sprintf("https://%s.%s", accountName, cloudEnvMap[opts.CloudEnvironment]))
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -26,6 +26,9 @@
 // The default URL opener will use credentials from the environment variables
 // AZURE_STORAGE_ACCOUNT, AZURE_STORAGE_KEY, and AZURE_STORAGE_SAS_TOKEN.
 // AZURE_STORAGE_ACCOUNT is required, along with one of the other two.
+// AZURE_BLOB_DOMAIN can optionally be used to provide an Azure Environment
+// blob storage domain to use. If no AZURE_BLOB_DOMAIN is provided, the
+// default Azure public domain "blob.core.windows.net" will be used.
 // To customize the URL opener, or for more details on the URL format,
 // see URLOpener.
 // See https://gocloud.dev/concepts/urls/ for background information.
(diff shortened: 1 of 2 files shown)


blob/azureblob/azureblob.go
```diff
@@ -96,6 +96,8 @@ type Options struct {
 	// delegated privileges.
 	// See https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1#shared-access-signature-parameters.
 	SASToken SASToken
+
+	CloudEnvironment CloudEnvironment
```
vangent: Can you add a comment here explaining what this option is, what the possible values are, and what it defaults to?

blob/azureblob/azureblob.go
```diff
@@ -223,6 +226,18 @@ type AccountKey string
 // https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1
 type SASToken string
 
+// CloudEnvironment is an Azure Cloud Environment to target (i.e. AzureCloud, AzureUSGovernment, AzureChinaCloud).
+// It is read from the AZURE_CLOUD_ENVIRONMENT environment variable.
+type CloudEnvironment string
+
+// Mappings from the AZURE_CLOUD_ENVIRONMENT value to their respective blob storage domains.
+var cloudEnvMap = map[CloudEnvironment]string{
```
vangent: Is there an advantage to having this mapping? I.e., why not have the CloudEnvironment variable just be the storage domain itself?

For example, with the mapping approach we'll need to update the code every time there's a new environment.

blob/azureblob/azureblob_test.go
```diff
@@ -53,6 +55,7 @@ import (
 const (
 	bucketName  = "go-cloud-bucket"
 	accountName = AccountName("gocloudblobtests")
+	cloudEnv    = CloudEnvironment("")
```
vangent: I think you can leave this out as well.

blob/azureblob/azureblob_test.go
```diff
@@ -32,15 +32,17 @@ import (
...
+// Prerequisites for -record mode
 // 1. Sign-in to your Azure Subscription at http://portal.azure.com.
 //
 // 2. Create a Storage Account.
 //
 // 3. Locate the Access Key (Primary or Secondary) under your Storage Account > Settings > Access Keys.
 //
-// 4. Set the environment variables AZURE_STORAGE_ACCOUNT, AZURE_STORAGE_KEY to
-//    the storage account name and your access key.
+// 4. Set the environment variables AZURE_STORAGE_ACCOUNT, AZURE_STORAGE_KEY, and AZURE_CLOUD_ENVIRONMENT to
```
vangent: I think you can leave this unchanged; we're not going to `-record` in anything other than AzureCloud so we can just leave it empty.

blob/azureblob/azureblob_test.go
```diff
@@ -113,7 +123,7 @@ func (h *harness) HTTPClient() *http.Client {
 }
 
 func (h *harness) MakeDriver(ctx context.Context) (driver.Bucket, error) {
-	return openBucket(ctx, h.pipeline, accountName, bucketName, &Options{Credential: h.credential})
+	return openBucket(ctx, h.pipeline, accountName, bucketName, &Options{Credential: h.credential, CloudEnvironment: cloudEnv})
```
vangent: Same.

blob/azureblob/azureblob_test.go
```diff
@@ -76,6 +79,13 @@ func newHarness(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
 		if err != nil {
 			t.Fatal(err)
 		}
+		env, err := DefaultCloudEnvironment()
```
vangent: Same.

blob/azureblob/azureblob.go
Strife96: The purpose of the mapping is to be able to use the same environment variable values ("AzureCloud", "AzureUSGovernment", etc.) that users may already be familiar with for this functionality (especially if they use Ansible, which is fairly popular). Furthermore, the values supplied mirror the Azure CLI and Azure Powershell names for these Cloud environments (when you do an "az cloud list" for example). It seemed a little more user-friendly in the end. 

However, if you think this setup is too hard to maintain into the future (I'm unsure how often Azure modifies or adds environments/endpoints), I can easily change the option to simply take in a specific URL suffix instead, it's certainly more flexible. The environment variable name should probably also be changed though (maybe something like GOCLOUD_AZURE_ENVIRONMENT?), so as not to collide with Ansible users who will be using AZURE_CLOUD_ENVIRONMENT with different values. Would you like me to do this instead?

blob/azureblob/azureblob_test.go
Strife96: Sure (also to all the below changes). Something made me squeamish about adding this option without any automated way to test it, but if you're okay with it, I can remove all of these changes to the testing file.

Strife96:
PTAL: I went ahead and made the changes to address the comments from the first iteration. The test harness modifications have been reverted, and the environment variable is now used as a domain name rather than a mapped value. Because this now differs from other existing usages of "AZURE_CLOUD_ENVIRONMENT", the environment variable is now renamed to "AZURE_BLOB_DOMAIN". The default target remains "blob.core.windows.net". Some documentation was added on what values are expected and how to find them.

Over the course of these commits 3 test builds were run. Travis build 7400 succeeded, then build 7401 passed on all environments but OSX (https://travis-ci.com/google/go-cloud/builds/142083379). The failed tests were in the runtimevar/httpvar module. Between these two builds, only whitespace removals were made from running gofmt -w -s on azureblob.go and azureblob_test.go. After changing only documentation, Travis build 7402 then succeeded on all environments again. The error appears transient, but I am unsure of the specifics of runtimevar/httpvar and whether or not it is related to the changes in these commits.

blob/azureblob/azureblob.go
```diff
@@ -253,6 +270,16 @@ func DefaultSASToken() (SASToken, error) {
 	return SASToken(s), nil
 }
 
+// DefaultCloudEnvironment loads the desired Azure Cloud to target from
+// the AZURE_BLOB_DOMAIN environment variable.
+func DefaultCloudEnvironment() (CloudEnvironment, error) {
+	s := os.Getenv("AZURE_BLOB_DOMAIN")
+	if s == "" {
+		return CloudEnvironment("blob.core.windows.net"), nil // If unspecified, use default public domain.
```
vangent: Nit: this would read simpler with `s = "blob.core.windows.net"` instead of a separate `return`.

vangent: Or, could just leave it the empty string, since `openBucket` handles the default.

blob/azureblob/azureblob.go
```diff
@@ -223,6 +236,10 @@ type AccountKey string
 // https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1
 type SASToken string
 
+// CloudEnvironment is an Azure Cloud Environment domain name to target (i.e. AzureCloud, AzureUSGovernment, AzureChinaCloud).
```
vangent: Nit: the examples are stale.

blob/azureblob/azureblob.go
```diff
@@ -223,6 +236,10 @@ type AccountKey string
...
+// It is read from the AZURE_BLOB_DOMAIN environment variable.
+type CloudEnvironment string
```
vangent: Naming: it's a bit weird to have the type and variable called `Environment` but the docstring and env variable called `Domain`. How about using `CloudEnvironmentDomain` everywhere? (e.g., `AZURE_BLOB_ENVIRONMENT_DOMAIN`, `CloudEnvironmentDomain` type, etc.).

blob/azureblob/azureblob.go
```diff
@@ -26,6 +26,11 @@
 // The default URL opener will use credentials from the environment variables
 // AZURE_STORAGE_ACCOUNT, AZURE_STORAGE_KEY, and AZURE_STORAGE_SAS_TOKEN.
 // AZURE_STORAGE_ACCOUNT is required, along with one of the other two.
+// AZURE_BLOB_DOMAIN can optionally be used to provide an Azure Environment
```
vangent: Naming nit: let's use `AZURE_STORAGE_` prefix for consistency with the other ones.

Strife96:
PTAL: Removed redundant default domain check, openBucket now exclusively handles the empty string case. Renamed the environment variable from AZURE_BLOB_DOMAIN to AZURE_STORAGE_DOMAIN and the internal type from CloudEnvironment to StorageDomain, as that is more accurate for its function now. Updated the documentation to reflect these changes and fix the outdated examples. 

Please let me know if I missed anything or if you would like any other changes.

blob/azureblob/azureblob_test.go
```diff
@@ -371,18 +375,24 @@ func TestOpenerFromEnv(t *testing.T) {
...
+				t.Errorf("Options.StorageDomain = %q; want %q", o.Options.StorageDomain, test.wantStorageDomain)
+			}
 		})
 	}
 }
 
 func TestOpenBucketFromURL(t *testing.T) {
 	prevAccount := os.Getenv("AZURE_STORAGE_ACCOUNT")
 	prevKey := os.Getenv("AZURE_STORAGE_KEY")
+	prevEnv := os.Getenv("AZURE_CLOUD_ENVIRONMENT")
```
vangent: One last thing -- please fix these env references.

Strife96:
Oops, that's silly of me. Fixed!

PR #2719, opened 2019-12-26
blob/gcsblob: use STORAGE_EMULATOR_HOST if present

### LINKED ISSUE #2718: blob/gcsblob: does not honor STORAGE_EMULATOR_HOST
[STORAGE_EMULATOR_HOST](https://github.com/googleapis/google-cloud-go/blob/9f47821ebd7e01e6ff3fa65f57dcd8afe59e2e57/storage/storage.go#L99) is used by `cloud.google.com/go/storage.NewClient` to enable testing against a GCS fake server.

In `gcsblob`, we always call [storage.Newclient](https://github.com/google/go-cloud/blob/ca79fc73f4d3639ab1d849991ec2b02ad3042edd/blob/gcsblob/gcsblob.go#L193) before [processing credentials](https://github.com/google/go-cloud/blob/ca79fc73f4d3639ab1d849991ec2b02ad3042edd/blob/gcsblob/gcsblob.go#L97). So we can't use a GCS fake server for testing with `gcsblob`.

It would be better if we also adopt `STORAGE_EMULATOR_HOST` logic before checking for credentials.

cuonglm:
Fixes #2718

### DIFF
--- a/blob/gcsblob/gcsblob.go
+++ b/blob/gcsblob/gcsblob.go
@@ -55,6 +55,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"os"
 	"sort"
 	"strings"
 	"sync"
@@ -62,15 +63,17 @@ import (
 
 	"cloud.google.com/go/storage"
 	"github.com/google/wire"
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/googleapi"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+
 	"gocloud.dev/blob"
 	"gocloud.dev/blob/driver"
 	"gocloud.dev/gcerrors"
 	"gocloud.dev/gcp"
 	"gocloud.dev/internal/escape"
 	"gocloud.dev/internal/useragent"
-	"google.golang.org/api/googleapi"
-	"google.golang.org/api/iterator"
-	"google.golang.org/api/option"
 )
 
 const defaultPageSize = 1000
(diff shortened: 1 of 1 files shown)


blob/gcsblob/gcsblob.go
```diff
@@ -94,11 +96,16 @@ type lazyCredsOpener struct {
 
 func (o *lazyCredsOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket, error) {
 	o.init.Do(func() {
-		creds, err := gcp.DefaultCredentials(ctx)
-		if err != nil {
-			o.err = err
-			return
+		creds, _ := google.CredentialsFromJSON(ctx, []byte(`{"type": "service_account", "project_id": "my-project-id"}`))
```
vangent: Let's do this is an `if/else` to be more clear, something like:

```
var creds google.Credentials // or whatever the type is
if os.Getenv("STORAGE_EMULATOR_HOST") == "" {
  var err error
  creds, err = gcp.DefaultCredentials(ctx)
  if err != nil {
    o.err = err
    return
  }
} else {
  // When using storage emulator, use dummy creds.
  creds, _ = google.CredentialsFromJSON(ctx, []byte(`{"type": "service_account", "project_id": "my-project-id"}`))
}
```

blob/gcsblob/gcsblob.go
```diff
@@ -189,8 +196,16 @@ func openBucket(ctx context.Context, client *gcp.HTTPClient, bucketName string,
 	if bucketName == "" {
 		return nil, errors.New("gcsblob.OpenBucket: bucketName is required")
 	}
+	clientOpts := []option.ClientOption{option.WithHTTPClient(useragent.HTTPClient(&client.Client, "blob"))}
```
vangent: Same here, use an `if/else`. Right now you have `option.WithHTTPClient` twice, the behavior is probably undefined.

blob/gcsblob/gcsblob.go
cuonglm: Fixed

blob/gcsblob/gcsblob.go
cuonglm: The later will override the former, but let just re-assign to `clientOpts`.

PR #2721, opened 2020-01-01
blob: Add ReadFrom to blob.Writer and WriteTo to blob.Reader

### LINKED ISSUE #2720: blob: implement ReaderFrom and WriterTo interfaces
An extremely welcome addition would be for blobs to implement the ReaderFrom/WriterTo interfaces:

```
type ReaderFrom interface {
    ReadFrom(r Reader) (n int64, err error)
}
type WriterTo interface {
    WriteTo(w Writer) (n int64, err error)
}
```

Right now we can use a TeeReader to "proxy-chain" and implement (for example) an upload progress bar:

```
dstBlob, err := bucket.NewWriter(ctx, dstKey, &blob.WriterOptions{BeforeWrite: before})
// [...more code...more stuff...]
barReader := bar.NewProxyReader(statReader)
teer := io.TeeReader(barReader, dstBlob)
_, err := ioutil.ReadAll(teer)
```

But the above is (in my humble opinion) not as simple, linear, and elegant as it could be. Ideally I would love to have the option to write something like the following (which I can do, for example, with Golang's built-in SFTP client):

```
dstBlob, err := bucket.NewWriter(ctx, dstKey, &blob.WriterOptions{BeforeWrite: before})
// [...more code...more stuff...]
barReader := bar.NewProxyReader(statReader)
byteCount, err := dstBlob.ReadFrom(barReader)
```

I would see a tremendous amount of value in this. I think the above solution would allow very short and elegant code to be written. It's not just for proxying readers/writers, if ReadFrom/WriteTo were supported, we could literally upload a file like this:

```
dstBlob, err := bucket.NewWriter(ctx, dstKey, &blob.WriterOptions{BeforeWrite: before})
srcFile, err := os.Open(srcFileName)
byteCount, err := dstBlob.ReadFrom(srcFile)
```

vangent:
Fixes #2720.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -155,21 +155,23 @@ func (r *Reader) WriteTo(w io.Writer) (int64, error) {
 // written to w.
 func readFromWriteTo(r io.Reader, w io.Writer) (int64, int64, error) {
 	buf := make([]byte, 1024)
-	var nr, nw int64
+	var totalRead, totalWritten int64
 	for {
-		nb, err := r.Read(buf)
-		nr += int64(nb)
-		if err == io.EOF {
+		numRead, rerr := r.Read(buf)
+		if numRead > 0 {
+			totalRead += int64(numRead)
+			numWritten, werr := w.Write(buf[0:numRead])
+			totalWritten += int64(numWritten)
+			if werr != nil {
+				return totalRead, totalWritten, fmt.Errorf("failed to write: %v", werr)
+			}
+		}
+		if rerr == io.EOF {
 			// Done!
-			return nr, nw, nil
+			return totalRead, totalWritten, nil
 		}
-		if err != nil {
-			return nr, nw, fmt.Errorf("failed to read: %v", err)
-		}
-		nb, err = w.Write(buf[0:nb])
-		nw += int64(nb)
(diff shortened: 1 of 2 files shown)


blob/blob.go
```diff
@@ -139,6 +139,41 @@ func (r *Reader) As(i interface{}) bool {
...
+// It reads data from r and writes to w, until EOF or a read/write error.
+// It returns the number of bytes read from r and the number of bytes
+// written to w.
+func readFromWriteTo(r io.Reader, w io.Writer) (int64, int64, error) {
+	buf := make([]byte, 1024)
+	var nr, nw int64
+	for {
+		nb, err := r.Read(buf)
+		nr += int64(nb)
+		if err == io.EOF {
```
eliben: I think it's OK for `Read` to return `nb>0` and `err == io.EOF`. Are you sure this is handled correctly here?

From the docs:

"Callers should always process the n > 0 bytes returned before considering the error err. Doing so correctly handles I/O errors that happen after reading some bytes and also both of the allowed EOF behaviors. "

vangent: Good catch. A conformance test actually caught this as well. Fixed.

PR #2723, opened 2020-01-08
server: use ochttp to create spans

zombiezen:
Updates #14


vangent:
Looks fine but tests failed on one image, PTAL.

zombiezen:
Done

PR #2726, opened 2020-01-17
docstore: fix batch Creates

### LINKED ISSUE #2725: docstore/driver: batch Create would only create single document across all drivers
### Describe the bug

Multiple documents cannot be created within single ActionList, only one of the documents would be created. 

### To Reproduce

Steps to reproduce the behavior.

```golang
package docstorecreate

import (
	"context"
	"testing"

	"github.com/stretchr/testify/assert"

	"gocloud.dev/docstore/memdocstore"
)

type Doc struct {
	ID      string
	Payload string
}

func TestMemDocstore(t *testing.T) {
	ctx := context.Background()

	coll, err := memdocstore.OpenCollection("ID", nil)
	assert.NoError(t, err)

	in1 := Doc{Payload: "one"}
	in2 := Doc{Payload: "two"}

	err = coll.Actions().Create(&in1).Create(&in2).Do(ctx)
	assert.NoError(t, err)
	assert.NotEmpty(t, in1.ID)
	assert.NotEmpty(t, in2.ID)
}

```

## Expected behavior

All documents would be created. 

### Version

v0.18.0

### Additional context

The root cause is implementation of [`GroupActions`](https://github.com/google/go-cloud/blob/master/docstore/driver/util.go#L58) function that maps write ops under a key. When documents are provided for `Create` operation, all keys are empty, and therefore only one of the documents would be created. 

This is generic bug that happens for all docstore drivers.

vangent:
Fixes #2725.

As described in the bug, a batched set of actions with multiple Create actions would drop all but one of the Creates. This is because an internal grouping operation used to order actions so that Gets and Writes for the same key are in an appropriate order uses `Key` as a map key. Create actions may have a `nil` Key, so only one survived.

To fix this, keep a separate slice of actions with `nil` Keys during the grouping, and append them to the "write" group at the end.


docstore/driver/util.go
```diff
@@ -61,8 +61,12 @@ func GroupActions(actions []*Action) (beforeGets, getList, writeList, afterGets
 	agets := map[interface{}]*Action{}
 	cgets := map[interface{}]*Action{}
 	writes := map[interface{}]*Action{}
+	var nilkeys []*Action
 	for _, a := range actions {
-		if a.Kind == Get {
+		if a.Key == nil {
```
deniswg: I believe Key is not necessarily nil, it could be empty string. Probably better check a.Kind == Create

vangent:
@eliben  or @jba ping?

Issue #2732, opened 2020-01-29
blob/gcsblob: incorrect metadata when iterating over objects


rusenask:
### Describe the bug

When listing objects the keys are correct but the obj attributes are the same for all objects. 

### To Reproduce

Create more than 1 object in GCS with custom attributes, try to list them and observe the attributes:

```golang
for {
	obj, err := iter.Next(ctx)
	if err == io.EOF {
		break
	}
	if err != nil {
		return nil, err
	}
       var oa storage.ObjectAttrs
       if obj.As(&oa) {
           fmt.Println(oa.Metadata)
       }
}

```

### Workaround

Currently I just get the attributes separately for each object from the list, although it would be nice to stop doing this :) 

```golang
attr, err := s.bucket.Attributes(context.Background(), obj.Key)
// use attr.Metadata
```

## Expected behavior

I would expect that if I get a list of objects I would see attributes of object 1 that belong to object 1, attributes of object 2 that belong to object 2 (and not 1).

### Version

v0.18.0

PR #2733, opened 2020-01-29
blob/gcsblob: fix obj attributes when listing bucket contents

rusenask:
Fixes https://github.com/google/go-cloud/issues/2732

### DIFF
--- a/blob/gcsblob/gcsblob.go
+++ b/blob/gcsblob/gcsblob.go
@@ -323,12 +323,13 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
 	if len(objects) > 0 {
 		page.Objects = make([]*driver.ListObject, len(objects))
 		for i, obj := range objects {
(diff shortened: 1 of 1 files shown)


rusenask:
@googlebot I signed it!

blob/gcsblob/gcsblob.go
```diff
@@ -323,12 +323,13 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
 	if len(objects) > 0 {
 		page.Objects = make([]*driver.ListObject, len(objects))
 		for i, obj := range objects {
-			asFunc := func(i interface{}) bool {
-				p, ok := i.(*storage.ObjectAttrs)
+			toCopy := objects[i]
```
vangent: Nit: I believe this can just be `obj`, no?

rusenask: yes, updated

PR #2735, opened 2020-01-31
blob: Let readFromWriteTo return the original err

petr-tichy:
Allows the caller to detect the error type instead of returning a
formatted string


petr-tichy:
@googlebot I signed it!

PR #2736, opened 2020-01-31
Add Header and Body to Entry struct to allow logging Request details

### LINKED ISSUE #2734: /server/requestlog: add http.Request to Entry struct
# Is your feature request related to a problem? Please describe.

Currently, the Entry struct has the RequestURL field, which stores the entire http.URL string. I would like to easily log only certain pieces of the URL (e.g. path, etc.). 

Is it possible to just add the request URL to the Entry struct so we can easily access and log whichever field from the URL we like or pull out the individual elements of the url and expose them as separate fields?

I've thought through other ways of doing this, and it's of course possible outside of this interface, but I'm trying to stick within the parameters of go-cloud whenever possible.

### Additional context

I'd also like to log the request body in some cases, but I can imagine that's a bridge too far. I can have a separate handler to do that...

gilcrest:
Fixes #2734 

This commit brings the `http.Request` into the `requestlog.Entry` struct so developers have access to it when logging using the `Log` method of the `requestlog.Logger` interface.

Thanks so much for your consideration of this!

Best,

Dan

### DIFF
--- a/server/requestlog/requestlog.go
+++ b/server/requestlog/requestlog.go
@@ -67,6 +67,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		RemoteIP:          ipFromHostPort(r.RemoteAddr),
 		TraceID:           sc.TraceID,
 		SpanID:            sc.SpanID,
+		Request:           r,
 	}
 	if addr, ok := r.Context().Value(http.LocalAddrContextKey).(net.Addr); ok {
 		ent.ServerIP = ipFromHostPort(addr.String())
@@ -114,6 +115,7 @@ type Entry struct {
 	Latency            time.Duration
 	TraceID            trace.TraceID
 	SpanID             trace.SpanID
+	Request            *http.Request
 }
 
 func ipFromHostPort(hp string) string {
--- a/server/requestlog/requestlog.go
+++ b/server/requestlog/requestlog.go
@@ -67,7 +67,8 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		RemoteIP:          ipFromHostPort(r.RemoteAddr),
 		TraceID:           sc.TraceID,
 		SpanID:            sc.SpanID,
-		Request:           r,
+		Header:            r.Header,
+		Body:              r.Body,
 	}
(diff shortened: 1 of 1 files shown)


gilcrest:
@googlebot I signed it!

server/requestlog/requestlog.go
```diff
@@ -114,6 +115,7 @@ type Entry struct {
 	Latency            time.Duration
 	TraceID            trace.TraceID
 	SpanID             trace.SpanID
+	Request            *http.Request
```
vangent: @zombiezen thoughts?

This makes me nervous because Request is potentially mutable by the caller, all of the other info in this struct is immutable.

If it's just URL info you want, we could consider cloning the net.URL, but it's possible that the easiest way to do that is to re-parse the URL string (e.g., that's the first result for https://www.google.com/search?q=golang+clone+url for me anyway), which you can already do.

zombiezen: I'm in agreement with @vangent here: the current struct is simple and immutable now. The `*http.Request` struct is large and contains many fields not relevant to an `Entry`, so I don't think it's appropriate to add one.

I could be convinced that this struct should include the HTTP headers, but that's beyond the use case outlined in #2734 which only needed the URL.

gilcrest: Totally fair. Would you be game if I just add the Header and Body? I can live without a clone of net.URL and will just parse the string back myself.

What I'm trying to get at is - as part of an API error response, I'm adding the opencensus TraceID that you've so nicely packaged into Entry. For customer support issues, I'd like to be able to find that TraceID in logs and inspect the request headers and body for debugging. It can be helpful. If this is outside the scope of what you've intended for Entry, I can understand that and will create a a separate log to get these details. I know the Body being an `io.Closer` makes it tricky, but I'd just leave it as is and separately dump it similar to how it's done in `httputil.DumpRequest`.

Thank you both again for your time and consideration!

Dan

gilcrest: Added a 2nd commit with these changes, if it helps!

gilcrest:
Weird. The build is now failing on older versions of Go, but not new ones... sigh..  I'll try again later some time.

PR #2737, opened 2020-02-05
server/requestlog: responseStats should implement http.Hijacker

alexsn:
responseStats should support connection hijacking if underlying response
writer does.

Signed-off-by: Alex Snast <alexsn@fb.com>


alexsn:
Currently websocket upgrades fails when using requestlog middleware as the wrapped response writer doesn't implement `http.Hijacker`.

PR #2747, opened 2020-02-18
GCP: Add NewAnonymousHTTPClient

alvaroaleman:
Currently, an anonymous GCP client has to be constructed manually via
```
client := &gcp.HTTPClient{Client: http.Client{Transport: gcp.DefaultTransport()}}
```

While this works, it is not very intuitive. This PR adds a simple
constructor for an anonymous client, making sure this capability is
visible in the godocs and easy to discover.

### DIFF
--- a/gcp/gcp.go
+++ b/gcp/gcp.go
@@ -45,6 +45,15 @@ type HTTPClient struct {
 	http.Client
 }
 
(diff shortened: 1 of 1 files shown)


alvaroaleman:
@googlebot I signed it!

PR #2752, opened 2020-02-25
all: drop go 1.11 support and make go 1.14 the current version.

vangent:
I had to drop some stuff from the `gocdk-runtimevar` sample; it was racy and got more racy with go 1.14.

I also had to upgrade `go-cmp` due to https://github.com/google/go-cmp/issues/167.

### DIFF
--- a/samples/gocdk-runtimevar/main_test.go
+++ b/samples/gocdk-runtimevar/main_test.go
@@ -16,10 +16,8 @@ package main
 
 import (
 	"context"
-	"errors"
 	"flag"
 	"testing"
-	"time"
 
 	"github.com/google/go-cmdtest"
 )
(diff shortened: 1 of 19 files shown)


samples/gocdk-runtimevar/main_test.go
```diff
@@ -37,61 +35,11 @@ func Test(t *testing.T) {
...
-		cancel func()
-		outc   chan result
-	)
+	ctx := context.Background()
 
 	runtimevar := cmdtest.InProcessProgram("gocdk-runtimevar", func() int {
 		return run(ctx)
 	})
 	ts.Commands["gocdk-runtimevar"] = runtimevar
-
```
eliben: Why all this?

vangent: See the PR description: I had to drop some stuff from the gocdk-runtimevar sample; it was racy and got 100% racy with go 1.14.

PR #2757, opened 2020-03-11
Blob/blob: Reduce allocations in bucket.Read() and bucket.Write()

richardartoul:
In my read-heavy workloads the stats allocations completely dominated all of the allocations in my application. This amortizes the cost of those allocations so they only have to be paid once for write/read stream instead of once per write/read call.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -92,18 +92,21 @@ import (
 // It implements io.ReadCloser, and must be closed after
 // reads are finished.
 type Reader struct {
-	b        driver.Bucket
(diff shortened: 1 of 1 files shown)


richardartoul:
@googlebot I signed it!

richardartoul:
Alternatively it would be better to just emit the stat once when the reader/writer is closed becuase `RecordWithTags` still allocates internally a lot I think, but I dont know if that would be acceptable sicne you'd only see the metric at the end

Issue #2745, opened 2020-02-14
gocloud.dev/mysql/gcpmysql: gcp mySQL options are not being handle in Open


g8rswimmer:
#1348 # Describe the bug

When calling the `mysql.Open` with the DNS string including options.  The database connection is opened without any of the options applied.

### To Reproduce

```go
db, err := mysql.Open(ctx,
		"gcpmysql://user:password@example-project/region/my-instance01/testdb?parseTime=true")
```
After this line of code successfully completes, the database connection will be opened.  However, the `parseTime` option will not have been applied.

## Expected behavior

When the `mysql.Open` connection string is passed, all of the options will be applied to the database connection.

### Version

gocloud.dev v0.19.0

### Additional context

After some investigating, I think that the problem could be in the use of the `go-sql-driver/mysql` configuration structure.  When calling `FormatDSN` method, it looks like the options are taken from the configuration structure members and not the string.

vangent:
Could you send a PR and test for this?

g8rswimmer:
@vangent Sure, I will get something up and add you as a reviewer (or anyone else that would need to approve)

g8rswimmer:
@vangent 
I have a branch for PR, but I do not have write access to push a branch.

vangent:
Hi @g8rswimmer -- you will need to create a fork and then a branch in your fork:
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork

Also see https://github.com/google/go-cloud/blob/master/CONTRIBUTING.md#making-a-pull-request.

g8rswimmer:
@vangent Should I close this as the new version addresses the issue?  I have tested on multiple mySQL gcp instances.

vangent:
Sure, if the issue is fixed let's close it.

PR #2759, opened 2020-03-18
mysql/gcpmysql: Parse Options in DSN

g8rswimmer:
This PR addresses Issue #2745 

### Background
When opening a `mysql` database instance, options may be needed.  For example, if you have a date/time column, the `parseTime` option will need to be set `true`.  In the `gcpmysql` package, the DSN options are not being handled when opening the instance connection.

### Root Cause
The differences of the URLs needed for the gcp database instance prevents the usage of the drivers (github.com/go-sql-driver/mysql) parsing of the DSN options.  
GCP:
```
user:password@example-project/region/my-instance01/testdb
```
 Vs
```
username:password@tcp(127.0.0.1:3306)/test
```
The code in the `OpenMySQLURL` only parses the connect parameters and none of the options that may be included when calling the driver's `OpenDB` method.

### Possible Solution
The approach taken was to minimize the code changes to add in the options to the configuration.  A method was created to generate a base configuration from a DSN string with a default database name and the passed queries.  After the base configuration is created, the original values where added to the configuration structure (net, addr, etc.).

This approach uses the drivers logic to properly set the configuration options rather than to have option parsing to two different locations.

#### Limitations
* The new configuration will have default values in the configuration structure (I did check to see what the default values where and how it would effect.  From what I am seeing, it should not have any negative side effects).  The original code set the configuration structure explicitly.

### DIFF
--- a/mysql/gcpmysql/gcpmysql.go
+++ b/mysql/gcpmysql/gcpmysql.go
@@ -113,16 +113,33 @@ func (uo *URLOpener) OpenMySQLURL(ctx context.Context, u *url.URL) (*sql.DB, err
 	mysql.RegisterDial(dialerName, client.Dial)
 
 	password, _ := u.User.Password()
-	cfg := &mysql.Config{
-		AllowNativePasswords: true,
-		Net:                  dialerName,
-		Addr:                 instance,
-		User:                 u.User.Username(),
-		Passwd:               password,
-		DBName:               dbName,
+
+	cfg, err := configWithOptions(u)
+	if err != nil {
+		return nil, fmt.Errorf("gcpmysql: open config %v", err)
 	}
+
+	cfg.AllowNativePasswords = true
+	cfg.Net = dialerName
+	cfg.Addr = instance
+	cfg.User = u.User.Username()
+	cfg.Passwd = password
+	cfg.DBName = dbName
+
 	db := sql.OpenDB(connector{cfg.FormatDSN(), uo.TraceOpts})
 	return db, nil
+}
+
+func configWithOptions(u *url.URL) (*mysql.Config, error) {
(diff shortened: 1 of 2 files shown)


g8rswimmer:
@googlebot I signed it!

mysql/gcpmysql/gcpmysql.go
```diff
@@ -113,18 +113,35 @@ func (uo *URLOpener) OpenMySQLURL(ctx context.Context, u *url.URL) (*sql.DB, err
...
-		User:                 u.User.Username(),
-		Passwd:               password,
-		DBName:               dbName,
+
+	cfg, err := configWithOptions(u)
+	if err != nil {
+		return nil, fmt.Errorf("gcpmysql: open config %v", err)
 	}
+
+	cfg.AllowNativePasswords = true
```
vangent: I think the tests for this would be more clear and more end-to-end if this code moved into `configWithOptions` (which maybe then should be renamed `configFromURL` or something).

mysql/gcpmysql/gcpmysql_test.go
```diff
@@ -130,3 +132,80 @@ func TestInstanceFromURL(t *testing.T) {
...
+		},
+		{
+			name: "single option",
+			args: args{
+				urlStr: "gcpmysql://my-project-id/us-central1/my-instance-id/my-db?parseTime=true",
+			},
+			want: func() *drvr.Config {
+				cfg := drvr.NewConfig()
+				cfg.Net = "tcp"
+				cfg.Addr = "127.0.0.1:3306"
```
vangent: For example, we won't see an `Addr` like this anymore, but rather the real one from the URL.

PR #2766, opened 2020-04-01
runtimevar/awsparamstore: Use WithDecryption=true to decrypt encrypted parameters

### LINKED ISSUE #2764: runtimevar/awsparameter: add support to automatically decrypt SecureString value
In AWS SSM Parameter Store, values can be stored in a _encrypted_ form; an SDK option allows to easily and transparently decrypt the value at retrieval. For example, this looks like  this using aws CLI tool:

```
# aws ssm delete-parameter --name myvar
aws ssm put-parameter --name myvar --value secretvalue --type SecureString  # encrypted
aws ssm get-parameter --with-decryption --name myvar  # transparently decrypted
secretvalue
```
Note the options: `--type SecureString` and `--with-decryption`.

With the current runtimevar/awsparameter implementation such SecureString parameters are not decrypted automatically and their values are not directly usable.

### Proposal 1

Change the current behaviour and always automatically decrypt SecureString values. This probably correspond to what most users expect. Minimal changes for this may look as follows.

```diff
[PATCH] runtimevar/awsparamstore: automatically decrypt encrypted (SecureString) value

---
 runtimevar/awsparamstore/awsparamstore.go | 8 ++++++--
 1 file changed, 6 insertions(+), 2 deletions(-)

diff --git a/runtimevar/awsparamstore/awsparamstore.go b/runtimevar/awsparamstore/awsparamstore.go
index c17e091d..250da207 100644
--- a/runtimevar/awsparamstore/awsparamstore.go
+++ b/runtimevar/awsparamstore/awsparamstore.go
@@ -249,9 +249,13 @@ func (w *watcher) WatchVariable(ctx context.Context, prev driver.State) (driver.
 	if prev != nil {
 		lastVersion = prev.(*state).version
 	}
-	// GetParameter from S3 to get the current value and version.
+	// GetParameter from SSM to get the current value and
+	// version. Automatically decrypt SecureString value.
 	svc := ssm.New(w.sess)
-	getResp, err := svc.GetParameter(&ssm.GetParameterInput{Name: aws.String(w.name)})
+	getResp, err := svc.GetParameter(&ssm.GetParameterInput{
+		Name:           aws.String(w.name),
+		WithDecryption: aws.Bool(true),
+	})
 	if err != nil {
 		return errorState(err, prev), w.wait
 	}
-- 
2.26.0
```

**NOTE.** This will work even for default (unencrypted) String as well since `WithDecryption: aws.Bool(true)` is ignored (and documented as this) for non SecureString parameters.

### Proposal 2

Add a way to specifiy if encrypted (SecureString) parameter should be automatically decrypted.

vangent:
`WithDecryption=true` has no effect if the parameter is not encrypted, so it's safe to just always set it.

Fixes #2764.

### DIFF
--- a/runtimevar/awsparamstore/awsparamstore.go
+++ b/runtimevar/awsparamstore/awsparamstore.go
@@ -251,7 +251,11 @@ func (w *watcher) WatchVariable(ctx context.Context, prev driver.State) (driver.
 	}
 	// GetParameter from S3 to get the current value and version.
 	svc := ssm.New(w.sess)
-	getResp, err := svc.GetParameter(&ssm.GetParameterInput{Name: aws.String(w.name)})
+	getResp, err := svc.GetParameter(&ssm.GetParameterInput{
+		Name: aws.String(w.name),
+		// Ignored if the parameter is not encrypted.
+		WithDecryption: aws.Bool(true),
+	})
 	if err != nil {
 		return errorState(err, prev), w.wait
 	}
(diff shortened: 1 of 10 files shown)


chumbert2:
Thanks!

PR #2768, opened 2020-04-07
server/all: add support for ListenAndServeTLS

### LINKED ISSUE #2767: server: support ListenAndServeTLS

I am using the `go-cloud` library for creating a HTTP server. I looked for documentation and examples about how to handle HTTPS but couldn't find anything.

We could add a method to a server for supporting HTTS, like `ListenAndServeTLS` for example so it is consistent with the standard library.

Or maybe am I missing something?

vangent:
Fixes #2767.

@zombiezen

### DIFF
--- a/server/driver/driver.go
+++ b/server/driver/driver.go
@@ -32,10 +32,20 @@ type Server interface {
 	// requests to the given http.Handler, and be interruptable by Shutdown.
 	// Drivers should use the given address if they serve using TCP directly.
 	ListenAndServe(addr string, h http.Handler) error
+
 	// Shutdown gracefully shuts down the server without interrupting
 	// any active connections. If the provided context expires before
 	// the shutdown is complete, Shutdown returns the context's error,
 	// otherwise it returns any error returned from closing the Server's
 	// underlying Listener(s).
 	Shutdown(ctx context.Context) error
+}
+
+// TLSServer is an optional interface for Server drivers, that adds support
+// for serving TLS.
+type TLSServer interface {
+	// ListenAndServeTLS is similar to Server.ListenAndServe, but should
+	// serve using TLS.
+	// See http://go/godoc/net/http/#Server.ListenAndServeTLS.
+	ListenAndServeTLS(addr, certFile, keyFile string, h http.Handler) error
 }
--- a/server/server.go
+++ b/server/server.go
@@ -17,6 +17,7 @@ package server // import "gocloud.dev/server"
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"sync"
 	"time"
@@ -42,13 +43,14 @@ var Set = wire.NewSet(
 // Server is a preconfigured HTTP server with diagnostic hooks.
 // The zero value is a server with the default options.
 type Server struct {
-	reqlog        requestlog.Logger
-	handler       http.Handler
-	healthHandler health.Handler
-	te            trace.Exporter
-	sampler       trace.Sampler
-	once          sync.Once
-	driver        driver.Server
+	reqlog         requestlog.Logger
+	handler        http.Handler
+	wrappedHandler http.Handler
+	healthHandler  health.Handler
+	te             trace.Exporter
+	sampler        trace.Sampler
+	once           sync.Once
+	driver         driver.Server
 }
 
 // Options is the set of optional parameters.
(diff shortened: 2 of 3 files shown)


server/driver/driver.go
```diff
@@ -26,12 +26,15 @@ type Server interface {
...
 	// The addr argument will be a non-empty string specifying "host:port".
+	// If certFile or keyFile are non-empty, the Server should listen using
+	// TLS.
 	// The http.Handler will always be non-nil.
 	// Drivers must block until serving is done (or
 	// return an error if serving can't occur for some reason), serve
 	// requests to the given http.Handler, and be interruptable by Shutdown.
 	// Drivers should use the given address if they serve using TCP directly.
-	ListenAndServe(addr string, h http.Handler) error
+	ListenAndServe(addr, certFile, keyFile string, h http.Handler) error
```
eliben, line 36: Why not a separate method to avoid breaking the API and to be more consistent with stdlib?

vangent: Could do that; I considered it. I don't think it avoids breaking though; anyone who is implementing this interface will be broken whether we change an existing method or add a new one. The logic was largely the same so I went this way, but happy to change it if you prefer. LMK.

zombiezen: (I haven't looked too closely at the rest of this PR, but...) If you add a different interface for `ListenAndServeTLS`, you can introduce this as a non-breaking change, and then have the `server` return an error for drivers that don't implement this optional interface.

server/server.go
```diff
@@ -158,10 +167,13 @@ func NewDefaultDriver() *DefaultDriver {
 
 // ListenAndServe sets the address and handler on DefaultDriver's http.Server,
 // then calls ListenAndServe on it.
-func (dd *DefaultDriver) ListenAndServe(addr string, h http.Handler) error {
+func (dd *DefaultDriver) ListenAndServe(addr, certFile, keyFile string, h http.Handler) error {
```
zombiezen, line 170: Consider making the drivers take in a [`*tls.Config`](https://pkg.go.dev/crypto/tls?tab=doc#Config) and then having the portable type provide a shortcut glue to use files. It is not uncommon to need to customize the `*tls.Config`, especially in scenarios with dynamically created (Let's Encrypt) certificates.

server/driver/driver.go
shantuo: Also, separating the method saves two `""`s for user that just wants to use `ListenAndServe`.

vangent: > If you add a different interface for ListenAndServeTLS, you can introduce this as a non-breaking change

True; I will give that a try. Although I still don't think it's likely that anyone is actually implementing this interface, it's good practice.

> separating the method saves two ""s for user that just wants to use ListenAndServe

The only user of the interface is the concrete type, so I don't think this is a big concern.

vangent: Now using a separate / optional interface with a separate method. Thanks for the suggestion @zombiezen !

The PR is no longer marked as a breaking change.

server/server.go
vangent: This turned out not to be necessary; the `*tls.Config` can be given to drivers at construction time. For example, the `server.DefaultDriver` already has a `http.Server` in it that the user can access to set the `TLSConfig` field.

vangent:
PTAL.

server/server.go
```diff
@@ -101,35 +103,48 @@ func (srv *Server) init() {
...
-	h = &ochttp.Handler{
-		Handler:          h,
-		IsPublicEndpoint: true,
+// ListenAndServeTLS is a wrapper to use wherever http.ListenAndServeTLS is used.
+// It wraps the http.Handler provided to New with a handler that handles tracing and
+// request logging. If the handler is nil, then http.DefaultServeMux will be used.
+// A configured Requestlogger will log all requests except HealthChecks.
+func (srv *Server) ListenAndServeTLS(addr, certFile, keyFile string) error {
+	// Check if the driver implements the optional interface.
+	tlsDriver, ok := srv.driver.(driver.TLSServer)
```
shantuo, line 142: Is this check necessary? Would the compiler give a similar error message?

achew22:
Out of curiosity, so I can learn more about the thought processes of `go-cloud`, why did you decide to implement `ListenAndServeTLS(...)` instead of `Serve(net.Listener)`? The `Serve` method also exists on most of the `Driver`s by default, and is probably just as backward compatible. If you want to use TLS you can use [`tls.Listen`](https://golang.org/pkg/crypto/tls/#Listen), but it would also open up listening on a UNIX socket, regular file, smoke signal, or some other non-standard listening scheme.

server/server.go
vangent: No, because one can instantiate a `Server` with a `driver.Server`; `driver.TLSServer` is an additional interface that drivers can choose whether to implement. For example, the test has two drivers, one that supports TLS and one that doesn't.

vangent:
> Out of curiosity, so I can learn more about the thought processes of `go-cloud`, why did you decide to implement `ListenAndServeTLS(...)` instead of `Serve(net.Listener)`? The `Serve` method also exists on most of the `Driver`s by default, and is probably just as backward compatible. If you want to use TLS you can use [`tls.Listen`](https://golang.org/pkg/crypto/tls/#Listen), but it would also open up listening on a UNIX socket, regular file, smoke signal, or some other non-standard listening scheme.


I suppose we could also expose `server.Serve` and/or `server.ServeTLS` in the same way, but haven't had any demand for that.

achew22:
> > Out of curiosity, so I can learn more about the thought processes of `go-cloud`, why did you decide to implement `ListenAndServeTLS(...)` instead of `Serve(net.Listener)`? The `Serve` method also exists on most of the `Driver`s by default, and is probably just as backward compatible. If you want to use TLS you can use [`tls.Listen`](https://golang.org/pkg/crypto/tls/#Listen), but it would also open up listening on a UNIX socket, regular file, smoke signal, or some other non-standard listening scheme.
> 
> I suppose we could also expose `server.Serve` and/or `server.ServeTLS` in the same way, but haven't had any demand for that.

If that's is a thing that you're considering, I opted not to use go-cloud because you didn't have a way to invoke `Serve` and it I couldn't find a transitional path for preexisting code to work with go-cloud without it. Our code uses [`cmux`](https://github.com/soheilhy/cmux) to server a single port with multiple protocols attached and we couldn't figure out a way to use that with go-cloud so we have it on our "to revisit in 6 months" list.

zombiezen:
> If that's is a thing that you're considering, I opted not to use go-cloud because you didn't have a way to invoke `Serve` and it I couldn't find a transitional path for preexisting code to work with go-cloud without it. Our code uses [`cmux`](https://github.com/soheilhy/cmux) to server a single port with multiple protocols attached [...]

This package's API was designed to give the driver control over how to surface an `http.Handler` to the world. Implementing the [`server/driver.Driver`](https://pkg.go.dev/gocloud.dev/server/driver?tab=doc#Server) interface would likely be the best way to accomplish your goal.

If you encounter issues with that, please open a new issue so we can better track this discussion rather than being buried in a PR.

> and we couldn't figure out a way to use that with go-cloud so we have it on our "to revisit in 6 months" list.

Hm. We may need to do more to communicate this, but the Go CDK packages can be used independently of the `server` package. The `server` package contains a prefabbed diagnostics wrapper for an application's `http.Handler` to help onboard folks who are new to writing cloud-native services, but it sounds like you already have a stack that works for your environment.

achew22:
> Implementing the server/driver.Driver interface would likely be the best way to accomplish your goal.

but `server/driver.Driver` doesn't have a `Serve(net.Listener)` method on it either? Is there another interface that I should be implementing?

zombiezen:
Would it be an issue to take in the `net.Listener` as a field of the type you're using to implement `Driver`?

achew22:
Would I then ignore the parameters passed into `ListenAndServe`?

zombiezen:
Yes.

achew22:
I guess I could do that. It feels pretty counter intuitive to do that though. If that's the official recommendation, would you be willing to put together some docs describing it?

PR #2773, opened 2020-04-14
blob/s3blob: update aws-sdk-go to support IRSA

### LINKED ISSUE #2772: blob/s3blob: use newer aws-sdk-go to use IAM Roles for Service Accounts
### Describe the bug

aws-sdk-go needs to be at least 1.23.13 to support IAM Roles for Service Accounts. see https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html

### To Reproduce

1. Use go-cloud in an application
1. Use relevant IRSA annotations in the SA for the application pods

## Expected behavior

The application should be able to access s3

### Version

v0.17.0

### Additional context

Application in which I faced the issue: https://github.com/jenkins-x/bucketrepo

gazal-k:
fix #2772

### DIFF
--- a/docstore/mongodocstore/go.mod
+++ b/docstore/mongodocstore/go.mod
@@ -21,13 +21,11 @@ require (
 	github.com/google/go-cmp v0.4.0
 	github.com/google/wire v0.4.0
 	github.com/klauspost/compress v1.9.8 // indirect
-	github.com/pkg/errors v0.9.1 // indirect
 	github.com/xdg/stringprep v1.0.0 // indirect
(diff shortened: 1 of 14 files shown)


gazal-k:
> Thanks!

can u pls let me know how often releases are created for go-cloud? The reason I'm asking is, I'd like to reference the version of go-cloud with this fix in another repo

vangent:
We do releases every few months, but you can use [Pseudo-versions](https://golang.org/cmd/go/#hdr-Pseudo_versions) to refer to the current commit until then.

PR #2776, opened 2020-04-19
blob/azureblob: allow specifying credentials in URL query params

stepanbujnak:
The credentials from the environment variables can now optionally be overriden by specifing query parameters on the URL (account_name, account_key, sas_token and storage_domain).

Signed-off-by: Stepan Bujnak <stepan.bujnak@gmail.com>

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -28,7 +28,10 @@
 // AZURE_STORAGE_ACCOUNT is required, along with one of the other two.
 // AZURE_STORAGE_DOMAIN can optionally be used to provide an Azure Environment
 // blob storage domain to use. If no AZURE_STORAGE_DOMAIN is provided, the
-// default Azure public domain "blob.core.windows.net" will be used. Check
+// default Azure public domain "blob.core.windows.net" will be used.
+// The credentials from the environment variables can optionally be
+// overriden by specifing query parameters on the URL
+// (account_name, account_key, sas_token and storage_domain). Check
 // the Azure Developer Guide for your particular cloud environment to see
 // the proper blob storage domain name to provide.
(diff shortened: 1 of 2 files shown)


vangent:
Hmm, IIRC we decided not to support these parameters in the URL because we didn't want to encourage users to put sensitive data like the SAS token into a plain-string URL.

stepanbujnak:
That's a shame because having to specify them as an environment variable makes it difficult to use multiple accounts at the same time. Besides, this argument is rather weak, you can stop users from specifying sensitive data in the URL, but nothing prevents them from setting those parameters via `os.Setenv()` from a plain-string.

Are you specifically against having the SAS token as a param or are you against all query parameters?

vangent:
Of course we can't prevent users from leaking sensitive data, but the intent is to not encourage it.

The design principle we documented is here:
https://github.com/google/go-cloud/blob/master/internal/docs/design.md#urls

```
Driver packages should include their own URLOpener struct type which implements all the relevant WidgetURLOpener methods. The URL should only serve to identify which resource to open. Any credentials or other complex values should be taken in as struct fields, not as input from URL. If the driver package registers its URLOpener with the DefaultURLMux, then it should populate these complex fields from environment variables. If doing so is undesirable or expensive, then it should not register with the DefaultURLMux and instead rely on users to create their own mux. If there already exists a well-established URI format for the backend (like S3 URLs or database connection URIs), then drivers should honor them where possible.
```

If you want to specify the credentials in another way, you can create your own `URLOpener` +  mux as described here:
https://gocloud.dev/concepts/urls/#muxes

The `azureblob.URLOpener` lets you set the `Pipeline` etc., see `openerFromEnv` and replace it how you'd like.

Or, don't use the URL-based openers instead, use the constructors.

PR #2778, opened 2020-04-21
pubsub: move batcher out of internal

### LINKED ISSUE #2771: Move `gocloud.dev/pubsub/internal/batcher` out of internal
`gocloud.dev/pubsub/internal/batcher` can only be imported by `gocloud.dev/pubsub/`.  I'm making a new Redis driver using gocloud.dev so that we can migrate easily to Google Pubsub.  

However, I'm unable to set batcher opts for at-most-once delivery in the way that the nats package does - because batcher is internal.

Is this _truly_ meant to be private if people can bring their own drivers?

vangent:
Other people are trying to develop their own `pubsub` drivers, and the existing ones depend on `batcher`. This means they can't do so unless it's in a fork, which is restrictive.

Fixes #2771.


vangent:
FYI: Similar request for `internal/gcerr`: #2779.

Issue #2753, opened 2020-02-26
Implement nats queue groups


tonyhb:
It would be good to add support for NATS queue groups for single-subscriber delivery.  There's a comment to this effect (https://github.com/google/go-cloud/blob/master/pubsub/natspubsub/nats.go#L284).

tonyhb:
Semi-related to https://github.com/google/go-cloud/issues/2141.

mathantunes:
Hi,

Any progress on this topic so far ? I'd like to work on it.

vangent:
Nobody is actively working on it. Could you summarize what kind of a change you think this is? I.e., if it's a new constructor and a few lines of code, vs a brand new driver.

vangent:
@mathantunes Nobody is actively working on it. Could you summarize what kind of a change you think this is? I.e., if it's a new constructor and a few lines of code, vs a brand new driver.

mathantunes:
Hi @vangent,

The [NATS QueueSubscribeSync](https://github.com/nats-io/nats.go/blob/76f36eea7e4c57cc5325bb80de40ecede69ffc48/nats.go#L2885) implementation is just a subscribe with a specified queue group and NATS takes care of the message delivery to only one queue subscriber.

* [NATS Docs on QueueSubscriber](https://docs.nats.io/nats-concepts/queue)

On go-cloud, I thing we can **modify the current NATS driver** either by utilizing the  *SubscriptionOptions present [here](https://github.com/google/go-cloud/blob/bb59d59f26d596dcd5eb970cc06572e9da7de7bd/pubsub/natspubsub/nats.go#L285)

Or by creating a new Subscription constructor, as sketched below:
```go
func OpenQueueSubscription(nc *nats.Conn, subject, queue string) (*pubsub.Subscription, error) {
	sub, err = nc.QueueSubscribeSync(subject, queue)
        ds := &subscription{nc, sub}
	return pubsub.NewSubscription(ds, recvBatcherOpts, nil), nil
}
```

vangent:
OK, this sounds like a minor change then; feel free to send a PR. Adding an option to `SubscriberOptions` SGTM.

mathantunes:
Alright, here it goes [pubsub/natspubsub: implement nats queue subscription #2782](https://github.com/google/go-cloud/pull/2782). 

It seems like I can't link the PR to the Issue.

Thank you

PR #2782, opened 2020-04-30
pubsub/natspubsub: implement nats queue subscription

mathantunes:
Implementation of Nats queue subscription through OpenSubscription function on natspubsub package.

Unit tests and example code were also added

### DIFF
--- a/pubsub/natspubsub/nats.go
+++ b/pubsub/natspubsub/nats.go
@@ -52,6 +52,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pubsub/natspubsub/nats.go
```diff
@@ -169,7 +169,11 @@ type TopicOptions struct{}
 
 // SubscriptionOptions sets options for constructing a *pubsub.Subscription
 // backed by NATS.
-type SubscriptionOptions struct{}
+type SubscriptionOptions struct {
+	//Queue sets the subscription as a QueueSubcription
```
vangent, line 173: Nit: For consistency, please include a space after `//` and end comments with a `.`.

mathantunes: Done!

pubsub/natspubsub/nats.go
```diff
@@ -169,7 +169,11 @@ type TopicOptions struct{}
 
 // SubscriptionOptions sets options for constructing a *pubsub.Subscription
 // backed by NATS.
-type SubscriptionOptions struct{}
+type SubscriptionOptions struct {
```
vangent, line 172: This allows you to set the Queue if you're using the regular constructor, or if you're using a custom URLOpener. How about adding some code near line 160 to support a "queue" URL parameter so that it can be done for the default URLOpener?

Something like this:

opts := o.SubscriptionOptions
for param, values := range u.Query() {
  if param == "queue" {
    opts.Queue = values[0]
  } else {
    return nil, fmt.Errorf("open subscription %v: invalid query parameter %s", u, param)
  }
}
...

vangent: And add a test for this near line 381.

mathantunes: Good thinking.

I also added a ```strings.ToLower(param)``` and a check for ```values != nil```, just in case anything goes wrong.

The URL to be decoded is ```"nats://mytopic?queue=queue1"``` and used on unit test ```TestOpenSubscriptionFromURL```

Also added an example ```Example_openQueueSubscriptionFromURL```

vangent:
Fixes #2753.

Issue #2783, opened 2020-05-04
blob/s3blob: ListObject.As() promotes wrong object


wmark:
Using the s3-driver, promoting a `blob.ListObject` to `s3.Object` will give the wrong item.

Specifically, this will throw a *mismatch* (fill the bucket with a few files):
```go
func main() {
	ctx := context.Background()

	bucketURL := "s3://example"
	bucket, _ := blob.OpenBucket(ctx, bucketURL)
	defer bucket.Close()

	iter := bucket.List(nil)
	for {
		obj, err := iter.Next(ctx)
		if err != nil {
			break
		}
		if obj == nil || obj.IsDir {
			continue
		}

		var s3o s3.Object
		if obj.As(&s3o) {
			if obj.Key != *s3o.Key { // Check for nil omitted.
				log.Fatalf("Mismatch with s3:\n -- %v\n -- %v\n",
					obj, s3o)
				break
			}
			// Count bytes per storage class, or whatever else.
		}
	}
	return
}
```

I expected this to halt without indicating any “mismatch”, i. e. the keys (names) being in agreement. Yet the unexpected result is something like this:
```console
2020/05/04 20:22:25 Mismatch with s3:
 -- &{foo.ext 2020-03-21 19:09:33 +0000 UTC 333996350 [] false 0xad1890}
 -- {
  ETag: "\"ccd2d4dc9b1217aa1a7f4190512ae3ca\"",
  Key: "bar.ext",
  LastModified: 2020-04-09 16:26:28 +0000 UTC,
  Size: 1890083,
  StorageClass: "STANDARD"
}

```

### Version

```
go 1.14

require (
	cloud.google.com/go v0.44.3
	github.com/aws/aws-sdk-go v1.19.45
	gocloud.dev v0.19.0
)
```

PR #2784, opened 2020-05-04
blob/s3blob: fix ListObject.As promoting the wrong object (#2783)

wmark:


vangent:
The basic change LGTM, but the tests are failing with:

```
blob/s3blob/s3blob_test.go:202:9: cannot convert nil to type s3.Object
```

wmark:
Thanks for the heads-up. Fixed. Apologies.

PR #2785, opened 2020-05-06
mysql/awsmysql: Parse options from URL

ymt2:
Because of the required URL representation differences between "awsmysql" package and github.com/go-sql-driver/mysql, we drop some options like the "parseTime".
So, this work change to respect the given URL parameter.

### DIFF
--- a/mysql/awsmysql/awsmysql.go
+++ b/mysql/awsmysql/awsmysql.go
@@ -33,7 +33,6 @@ import (
 	"database/sql/driver"
 	"fmt"
 	"net/url"
-	"strings"
 	"sync"
 
 	"contrib.go.opencensus.io/integrations/ocsql"
@@ -78,27 +77,21 @@ func (uo *URLOpener) OpenMySQLURL(ctx context.Context, u *url.URL) (*sql.DB, err
 		return nil, fmt.Errorf("open RDS: empty endpoint")
 	}
 	password, _ := u.User.Password()
+	dsn := fmt.Sprintf("%s:%s@tcp(%s)%s?%s", u.User.Username(), password, u.Host, u.Path, u.RawQuery)
 	c := &connector{
-		addr:     u.Host,
-		user:     u.User.Username(),
-		password: password,
-		dbName:   strings.TrimPrefix(u.Path, "/"),
 		// Make a copy of TraceOpts to avoid caller modifying.
 		traceOpts: append([]ocsql.TraceOption(nil), uo.TraceOpts...),
 		provider:  source,
 
 		sem:   make(chan struct{}, 1),
 		ready: make(chan struct{}),
+		dsn:   dsn,
 	}
 	c.sem <- struct{}{}
 	return sql.OpenDB(c), nil
 }
 
 type connector struct {
-	addr      string
-	user      string
-	password  string
-	dbName    string
 	traceOpts []ocsql.TraceOption
 
 	sem      chan struct{} // receive to acquire, send to release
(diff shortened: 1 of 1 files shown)


ymt2:
@googlebot I signed it!

vangent:
@zombiezen

mysql/awsmysql/awsmysql.go
```diff
@@ -78,27 +77,21 @@ func (uo *URLOpener) OpenMySQLURL(ctx context.Context, u *url.URL) (*sql.DB, err
 		return nil, fmt.Errorf("open RDS: empty endpoint")
 	}
 	password, _ := u.User.Password()
+	dsn := fmt.Sprintf("%s:%s@tcp(%s)%s?%s", u.User.Username(), password, u.Host, u.Path, u.RawQuery)
```
zombiezen, line 80: Rather than using `fmt.Sprintf`, use `*url.URL` to construct the DSN to ensure the proper escaping.

mysql/awsmysql/awsmysql.go
```diff
@@ -129,16 +122,14 @@ func (c *connector) Connect(ctx context.Context) (driver.Conn, error) {
...
-		cfg := &mysql.Config{
-			Net:                     "tcp",
-			Addr:                    c.addr,
-			User:                    c.user,
-			Passwd:                  c.password,
-			TLSConfig:               tlsConfigName,
-			AllowCleartextPasswords: true,
-			AllowNativePasswords:    true,
-			DBName:                  c.dbName,
+		cfg, err := mysql.ParseDSN(c.dsn)
```
zombiezen, line 125: This requires parsing the same string on every new connection. Better to parse the DSN during `OpenMySQLURL` and then clone it during `Connect`.

mysql/awsmysql/awsmysql.go
ymt2: MySQL allows to use special characters like a `*` in password or username.
And package `url` encodes such as special characters.
So, I think we cannot use package `url` in this usecase.

mysql/awsmysql/awsmysql.go
ymt2: Agree. But we have to call [mysql Config.normalize](https://github.com/go-sql-driver/mysql/blob/6844171c7b652e0a8271a909fc48f6a32492d04f/dsn.go#L98) here to call [mysql.getTLSConfigClone](https://github.com/go-sql-driver/mysql/blob/d9aa6d3abe0f1072adb4aa71e1eb1d3eb3d3921c/utils.go#L82) to use SSL. 
So, I think it should be called here even if it is redundant.

vangent:
@zombiezen can you finish your review? I'd like a LGTM from you before moving forward. Thanks!

zombiezen:
Yes, sorry about the delay here. I'll try to review this week.

mysql/awsmysql/awsmysql.go
zombiezen: Right, but otherwise a `:` in a password will be interpreted incorrectly. The escaping issue isn't being handled here, so this seems like a step backward.

mysql/awsmysql/awsmysql.go
zombiezen: I don't understand this statement. What's the end effect of this unexported field? Does this mean the same `*mysql.Config` can't be used for multiple connections? That seems like a pretty unfortunate flaw in the underlying MySQL library and seems like it should be fixed upstream.

zombiezen:
Taking another look and I stand by my prior objections to this PR. I'm onboard with the direction, but I think the implementation needs more work before accepting.

mysql/awsmysql/awsmysql.go
g8rswimmer: @ymt2 Did you look at the https://github.com/google/go-cloud/blob/master/mysql/gcpmysql/gcpmysql.go#L121 where this has also been done.  I have used it in several projects and it is work well for me.

vangent:
Closing in favor of #2831.

PR #2786, opened 2020-05-14
secrets/azurekeyvault: add support for other azure keyvault regions

jaxxstorm:
Azure keyvault has multiple regions, as described [here](https://docs.microsoft.com/en-us/azure/key-vault/general/secure-your-key-vault#resource-endpoints)

This PR updates the regex currently used to detect a valid azure
keyvault region to include:

- Azure US Government: https://<vault-name>.vault.usgovcloudapi.net
- Azure China 21Vianet: https://<vault-name>.vault.azure.cn:443
- Azure Germany: https://.vault.microsoftazure.de:443

See

### DIFF
--- a/secrets/azurekeyvault/akv.go
+++ b/secrets/azurekeyvault/akv.go
@@ -208,7 +208,7 @@ func dial(useCLI bool) (*keyvault.BaseClient, error) {
 
 var (
 	// Note that the last binding may be just a key, or key/version.
-	keyIDRE = regexp.MustCompile("^(https://.+\\.vault\\.azure\\.net/)keys/(.+)$")
+	keyIDRE = regexp.MustCompile(`^(https://.+\.vault\.(?:azure\.net|azure\.cn|usgovcloudapi\.net|microsoftazure\.de)/)keys/(.+)$`)
 )
 
 // OpenKeeper returns a *secrets.Keeper that uses Azure keyVault.


secrets/azurekeyvault/akv.go
```diff
@@ -208,7 +208,7 @@ func dial(useCLI bool) (*keyvault.BaseClient, error) {
 
 var (
 	// Note that the last binding may be just a key, or key/version.
-	keyIDRE = regexp.MustCompile("^(https://.+\\.vault\\.azure\\.net/)keys/(.+)$")
+	keyIDRE = regexp.MustCompile("^(https://.+\\.vault\\.(?:azure\\.net|azure\\.cn|usgovcloudapi\\.net|microsoftazure\\.de)/)keys/(.+)$")
```
wmark, line 211: Use *raw strings* for regular expressions to avoid excessive quoting.

```suggestion
	keyIDRE = regexp.MustCompile(`^(https://.+\.vault\.(?:azure\.net|azure\.cn|usgovcloudapi\.net|microsoftazure\.de)/)keys/(.+)$`)
```

And you might want to prefix your commit with `secrets/azurekeyvault:` (the PR might get discarded if the commit is merged with `ff`).

jaxxstorm: @wmark I didn't use the raw string to stick with the current convention, would you prefer me to switch it to the in this case?

jaxxstorm: As suggested, I've updated this to use a raw string and squashed the commit messages

PR #2788, opened 2020-05-14
pubsub/awssnssqs/example_test.go:remove http:// protocol from URL in Example_openSQSTopicFromURL

DallanQ:
remove "http://" from queueURL so we get the correct awssqs://sqs.us-east-... url in the OpenTopic function call.


DallanQ:
@googlebot I signed it!

vangent:
Looks like the tests are failing, you need to run:

```
internal/website/gatherexamples/run.sh > internal/website/data/examples.json
```

to update the JSON file of examples that gets included in the website.

PR #2796, opened 2020-05-21
blob/fileblob: Support relative paths in URL openers via a host of "."

### LINKED ISSUE #2795: blob/fileblob: Support relative paths in bucket URL

In application configuration the bucket URL constructor comes really handy. Buckets can be configured with a single URL.

Unfortunately, using bucket URLs we cannot achieve the same results: using file blobs, a bucket URL cannot reference a relative path. Relative paths are often sane defaults in development environments.

Find a way to represent relative paths in URLs. For example:

- `file://./path/to/dir` (host would be `.`)
- `file://path/to/dir` (host would `path`. non empty host could mean relative path)

(Thinking it through, the above two solutions are essentially the same)

Use a query parameter to make the path relative:

`file:///path/to/dir?relative=true`

vangent:
Fixes #2795.


vangent:
gentle ping eliben@ ?

PR #2800, opened 2020-05-26
blob/gcsblob: set defaults for SignURL

### LINKED ISSUE #2653: blob/gcsblob: Make GoogleAccessID and PrivateKey optional

I'm creating a microservice on cloud-run and recently ran into the 32MB response limit. So I thought, I'll just use the SignedURL functionality and redirect to it.

But I can't seem to access the private-key anywhere on cloud-run. The credentials seem to be fetched from the metadata service

I came across this issue on the Go SDK: https://github.com/googleapis/google-cloud-go/issues/1130

It feels like a really weird abstraction, that I need to tell OpenBucket about encryption keys when everything else is automatically detected.

The developer experience is *otherwise* excellent.

`bucket.SignedURL()` should work out-of-the-box on GCP in my opinion without further configuration.

I've tried:

- Sending service-account.json file as an ENV var
- Retrieving the private-key from ENV
- Retrieving the private-key from known location.

I haven't tried hard-coding the credentials or shipping them through Git or Docker

wmark:
This enables to use blob.SignURL out-of-the-box by setting the required values, primarily the GoogleAccessID and a PrivateKey—where available—, and a default signing behaviour.

By moving the IAM Credentials API client into the bucket, it's henceforth possible to use it in `bucket.Options.SignBytes()` without any external wrapper.

Previously users ran into `SignURL` working properly with the likes of S3 without doing anything special, and GCS has been the outlier that needed additional work degrading productivity and developer experience.

closes #2653

### DIFF
--- a/blob/gcsblob/gcsblob.go
+++ b/blob/gcsblob/gcsblob.go
@@ -18,9 +18,16 @@
 // URLs
 //
 // For blob.OpenBucket, gcsblob registers for the scheme "gs".
-// The default URL opener will creating a connection using use default
-// credentials from the environment, as described in
+// The default URL opener will set up a connection using default credentials
+// from the environment, as described in
 // https://cloud.google.com/docs/authentication/production.
+// Some environments, such as GCE, come without a private key. In such cases
+// the IAM Credentials API will be configured for use in Options.MakeSignBytes,
+// which will introduce latency to any and all calls to bucket.SignedURL
+// that you can avoid by installing a service account credentials file or
+// obtaining and configuring a private key:
+// https://cloud.google.com/iam/docs/creating-managing-service-account-keys
+//
 // To customize the URL opener, or for more details on the URL format,
 // see URLOpener.
 // See https://gocloud.dev/concepts/urls/ for background information.
@@ -156,6 +163,14 @@ func (o *lazyCredsOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.
 			opts.GoogleAccessID, opts.PrivateKey = readDefaultCredentials(creds.JSON)
 		}
 
+		// Provide a default factory for SignBytes for environments without a private key.
+		if len(opts.PrivateKey) <= 0 && opts.GoogleAccessID != "" {
+			iam := new(credentialsClient)
+			// We cannot hold onto the first context: it might've been cancelled already.
+			ctx := context.Background()
+			opts.MakeSignBytes = iam.CreateMakeSignBytesWith(ctx, opts.GoogleAccessID)
+		}
+
 		client, err := gcp.NewHTTPClient(gcp.DefaultTransport(), creds.TokenSource)
 		if err != nil {
 			o.err = err
@@ -212,6 +227,9 @@ func (o *URLOpener) forParams(ctx context.Context, q url.Values) (*Options, erro
 	if accessID := q.Get("access_id"); accessID != "" && accessID != opts.GoogleAccessID {
 		opts.GoogleAccessID = accessID
 		opts.PrivateKey = nil // Clear any previous key unrelated to the new accessID.
+
+		// Clear this as well to prevent calls with the old and mismatched accessID.
+		opts.MakeSignBytes = nil
 	}
 	if keyPath := q.Get("private_key_path"); keyPath != "" {
 		pk, err := ioutil.ReadFile(keyPath)
@@ -219,6 +237,11 @@ func (o *URLOpener) forParams(ctx context.Context, q url.Values) (*Options, erro
 			return nil, err
 		}
 		opts.PrivateKey = pk
+	} else if _, exists := q["private_key_path"]; exists {
+		// A possible default value has been cleared by setting this to an empty value:
+		// The private key might have expired, or falling back to SignBytes/MakeSignBytes
+		// is intentional such as for tests or involving a key stored in a HSM/TPM.
+		opts.PrivateKey = nil
 	}
 	return opts, nil
 }
@@ -236,10 +259,17 @@ type Options struct {
 	PrivateKey []byte
 
 	// SignBytes is a function for implementing custom signing.
-	// Exactly one of PrivateKey or SignBytes must be non-nil to use SignedURL.
+	// Exactly one of PrivateKey, SignBytes, or MakeSignBytes must be non-nil to use SignedURL.
 	// See https://godoc.org/cloud.google.com/go/storage#SignedURLOptions.
 	SignBytes func([]byte) ([]byte, error)
+
+	// MakeSignBytes is a factory for functions that are being used in place of an empty SignBytes.
+	// If your implementation of 'SignBytes' needs a request context, set this instead.
+	MakeSignBytes func(requestCtx context.Context) SignBytesFunc
 }
+
+// SignBytesFunc is shorthand for the signature of Options.SignBytes.
+type SignBytesFunc func([]byte) ([]byte, error)
 
 // openBucket returns a GCS Bucket that communicates using the given HTTP client.
 func openBucket(ctx context.Context, client *gcp.HTTPClient, bucketName string, opts *Options) (*bucket, error) {
@@ -649,9 +679,20 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
 }
 
 func (b *bucket) SignedURL(ctx context.Context, key string, dopts *driver.SignedURLOptions) (string, error) {
-	if b.opts.GoogleAccessID == "" || (b.opts.PrivateKey == nil && b.opts.SignBytes == nil) {
-		return "", errors.New("to use SignedURL, you must call OpenBucket with a valid Options.GoogleAccessID and exactly one of Options.PrivateKey or Options.SignBytes")
+	numSigners := 0
+	if b.opts.PrivateKey != nil {
+		numSigners++
 	}
+	if b.opts.SignBytes != nil {
+		numSigners++
+	}
+	if b.opts.MakeSignBytes != nil {
+		numSigners++
+	}
+	if b.opts.GoogleAccessID == "" || numSigners != 1 {
+		return "", errors.New("to use SignedURL, you must call OpenBucket with a valid Options.GoogleAccessID and exactly one of Options.PrivateKey, Options.SignBytes, or Options.MakeSignBytes")
+	}
+
 	key = escapeKey(key)
 	opts := &storage.SignedURLOptions{
 		Expires:        time.Now().Add(dopts.Expiry),
@@ -660,6 +701,9 @@ func (b *bucket) SignedURL(ctx context.Context, key string, dopts *driver.Signed
 		GoogleAccessID: b.opts.GoogleAccessID,
 		PrivateKey:     b.opts.PrivateKey,
 		SignBytes:      b.opts.SignBytes,
+	}
+	if b.opts.MakeSignBytes != nil {
+		opts.SignBytes = b.opts.MakeSignBytes(ctx)
 	}
 	return storage.SignedURL(b.name, key, opts)
 }
--- a/blob/gcsblob/gcsblob.go
+++ b/blob/gcsblob/gcsblob.go
@@ -49,6 +49,7 @@ package gcsblob // import "gocloud.dev/blob/gcsblob"
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -87,6 +88,48 @@ var Set = wire.NewSet(
 	wire.Struct(new(URLOpener), "Client"),
 )
 
+// readDefaultCredentials gets the field values from the supplied JSON data.
+// For its possible formats please see
+// https://cloud.google.com/iam/docs/creating-managing-service-account-keys#iam-service-account-keys-create-go
+//
+// Use "golang.org/x/oauth2/google".DefaultCredentials.JSON to get
+// the contents of the preferred credential file.
+//
+// Returns null-values for fields that have not been obtained.
+func readDefaultCredentials(credFileAsJSON []byte) (AccessID string, PrivateKey []byte) {
+	// For example, a credentials file as generated for service accounts through the web console.
+	var contentVariantA struct {
+		ClientEmail string `json:"client_email"`
+		PrivateKey  string `json:"private_key"`
+	}
+	if err := json.Unmarshal(credFileAsJSON, &contentVariantA); err == nil {
+		AccessID = contentVariantA.ClientEmail
+		PrivateKey = []byte(contentVariantA.PrivateKey)
+	}
+	if AccessID != "" {
+		return
+	}
+
+	// If obtained through the REST API.
+	var contentVariantB struct {
+		Name           string `json:"name"`
+		PrivateKeyData string `json:"privateKeyData"`
+	}
+	if err := json.Unmarshal(credFileAsJSON, &contentVariantB); err == nil {
+		nextFieldIsAccessID := false
+		for _, s := range strings.Split(contentVariantB.Name, "/") {
+			if nextFieldIsAccessID {
+				AccessID = s
+				break
+			}
+			nextFieldIsAccessID = s == "serviceAccounts"
+		}
+		PrivateKey = []byte(contentVariantB.PrivateKeyData)
+	}
+
+	return
+}
+
 // lazyCredsOpener obtains Application Default Credentials on the first call
 // to OpenBucketURL.
 type lazyCredsOpener struct {
@@ -97,6 +140,7 @@ type lazyCredsOpener struct {
 
 func (o *lazyCredsOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket, error) {
 	o.init.Do(func() {
+		var opts Options
 		var creds *google.Credentials
 		if os.Getenv("STORAGE_EMULATOR_HOST") != "" {
 			creds, _ = google.CredentialsFromJSON(ctx, []byte(`{"type": "service_account", "project_id": "my-project-id"}`))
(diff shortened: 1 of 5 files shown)


wmark:
I'd love to get some guidance on how to properly test this, as well as someone from the team to implement them given the elaborate infrastructure this uses.

wmark:
I've dropped detection of the *service account* on GCE (as `GoogleAccessID`) from this PR and will file it separately once this is in. It required upgrading to `cloud.google.com/go v0.57.0`

blob/gcsblob/gcsblob.go
```diff
@@ -107,14 +111,24 @@ func (o *lazyCredsOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.
 				o.err = err
 				return
 			}
+
+			// Populate default values from any such file, if and where possible.
```
vangent, line 115: Can you add a link to the documentation where you found that this JSON blob might contain these fields? Maybe https://cloud.google.com/iam/docs/creating-managing-service-account-keys#creating_service_account_keys ?

blob/gcsblob/gcsblob.go
```diff
@@ -134,6 +148,12 @@ const Scheme = "gs"
 //
 //   - access_id: sets Options.GoogleAccessID
 //   - private_key_path: path to read for Options.PrivateKey
+//
+// Currently their use is limited to SignedURL. If left empty they will be guessed
+// which can result in inconsistencies and/or silently fail in environments such as Compute Engine.
```
vangent, line 153: Please explain how they are guessed. IIUC, this will only work with a credentials file found on the default paths that was created using Cloud Console or `gcloud`...?

blob/gcsblob/gcsblob.go
```diff
@@ -134,6 +148,12 @@ const Scheme = "gs"
 //
 //   - access_id: sets Options.GoogleAccessID
 //   - private_key_path: path to read for Options.PrivateKey
+//
+// Currently their use is limited to SignedURL. If left empty they will be guessed
```
vangent, line 152: IIUC, they are only guessed when using the default URLOpener. So, I think that the comments about guessing should be up near line 23, not here.

blob/gcsblob/gcsblob.go
```diff
@@ -134,6 +148,12 @@ const Scheme = "gs"
...
+// which can result in inconsistencies and/or silently fail in environments such as Compute Engine.
+//
+// Absent any other signer and without a private key, access_id will be used
```
vangent, line 155: When would this happen?

IIUC, before this PR, either neither of these fields would be set (using the default URLOpener via lazyCredsOpener, or using URLOpener explicitly but not setting the fields), or they would both be required to be set (only when using URLOpener explicitly).

Your "guessing" code tries to guess both fields. Is there any reason to support only having `access_id` set? I.e., is there any reason to believe that we'll successfully guess one of them but not the other?

If the answer is "no", then I think the "guessing" code is all you need to add; all of the changes below here aren't necessary. Am I missing something?

blob/gcsblob/gcsblob.go
```diff
@@ -160,15 +180,19 @@ func (o *URLOpener) forParams(ctx context.Context, q url.Values) (*Options, erro
...
 		opts.GoogleAccessID = accessID
+		opts.PrivateKey = nil // Clear any previous key unrelated to the new accessID.
 	}
 	if keyPath := q.Get("private_key_path"); keyPath != "" {
 		pk, err := ioutil.ReadFile(keyPath)
 		if err != nil {
 			return nil, err
 		}
 		opts.PrivateKey = pk
+	} else if _, exists := q["private_key_path"]; exists {
```
vangent, line 193: What's this for?

blob/gcsblob/gcsblob.go
```diff
@@ -227,15 +251,80 @@ func OpenBucket(ctx context.Context, client *gcp.HTTPClient, bucketName string,
 	if err != nil {
 		return nil, err
 	}
+	otherSignerAvailable := len(drv.opts.PrivateKey) >= 1 || drv.opts.SignBytes != nil
```
vangent, line 254: Assuming the above comments about whether this is necessary are resolved as "yes":

Could this be implemented instead by setting `Options.SignBytes` during initialization? Then again, you don't need to modify any of the code here or below.

blob/gcsblob/gcsblob.go
wmark: Appreciate your feedback to this PR so we can arrive at the best possible solution. I'll push a change asap, but reply to your points now so we're on the same page before the push will hide everything.

----
I'll add the link. I'll extract the “guessing” to its own function and cover the second example from aforementioned site.

I've asked around and received files without a private key (they have a ‘google’ sub-structure in them), so there's instances where only the `GoogleAccessID` is available.

blob/gcsblob/gcsblob.go
wmark: Commenting on this would add clarity, but giving a stronger guarantee than GCP does might be unwise, as this scheme might change in future (and we might still read `creds.JSON`).

One path I anticipate is GCP switching the scheme (today ECDSA is popular for this), implementing key rotation (ECDSA are to be retired every so many signings), and using something the likes of the upcoming *runtime config* to distribute them.

blob/gcsblob/gcsblob.go
wmark: That's for three use cases:
1. On GCE (but not limited to) the private key is not available.
2. The private key has expired (rotation).
3. Tests. (Needs *Service Account Token Creation* privileges.)

blob/gcsblob/gcsblob.go
wmark: Tests and to force going through the *Credentials API*. I'll add that to the comment.

blob/gcsblob/gcsblob.go
wmark: This would move the required fields into `gcsblob.Bucket` (going with the default style to prefer struct members over closures), and/or make the lazy initialization no longer a lazy one (the *Credentials API client* connects right away).

blob/gcsblob/gcsblob.go
vangent: OK, please add comments to this effect (explaining that in some cases, both fields are available, and in others, only the access_id. Thanks!

blob/gcsblob/gcsblob.go
vangent: Could you make a new struct with whatever fields you need, including a lazily initialized client, and pass a method of that struct as the `Options.SignBytes` func? I think it's OK to not close the client on shutdown, so `Bucket` and `bucket` wouldn't need to know about the struct at all.

vangent: The motivation for this suggestion is to keep the number of ways we sign the same instead of adding a new one; it keeps the main code simpler. It also provides a demonstration of how to use `SignBytes`; if a user wants to modify how the default `URLOpener` does it (e.g., different guessing, etc.), they can copy what the default URLOpener does and modify it how they like, and everything will work from there.

blob/gcsblob/gcsblob.go
wmark: `lazyCredsOpener` is supposed to set any ‘Application Default Credentials’, therefore I've skipped this comment altogether.

blob/gcsblob/gcsblob.go
wmark: I see the merits of this, not expanding struct `bucket` at all and setting `SignBytes` accordingly. Unfortunately the client needs to be closed or its connection remains lingering. I've considered starting a goroutine to close it once the *background context* concludes but found that it might be too late, i. e. the connection lingered still (the goroutine won't block shutdown without anything in place).

wmark: Further, `SignBytes` does not accept a context, specifically no *request context*, therefore it's not possible to properly express the IAM Client API call properly as merely `SignBytes([]byte)…`.

In extension, the user won't be able to use anything with a *request context* as `SignBytes` without any prior callback or the like to get that context in.

wmark: If you want I can add `context.Context` as parameter to `SignBytes` and implement your vision for this—that'll be a breaking change, though.

vangent: I think it's important that the functionality you're providing a default for in the default URLOpener is functionality that's customizable without making changes to the module. I see your point about a `SignBytes` function not being able to access the `context.Context` passed in to `SignedURL`. To resolve that, how about adding a `MakeSignBytes` field to `Options` with the signature `MakeSignBytes(ctx context.Context) func([]byte) ([]byte, error)`? If provided, `SignedURL` would call that function with the request `ctx` to get a `SignBytes` func to pass to `storage.SignedURL` via the `storage.SignedURLOptions`.

Then, your code in the default `URLOpener` would create a `credentialsSigner` client struct, with a lazily initialized client field and a `MakeSignBytes` method, and pass that `MakeSignBytes` method in `Options`.

It's true that the lazily initialized client will never be cleaned up, but that's OK; there's only one of them (not one per bucket), and it will live as long as the application. We've used that pattern in other default URLOpeners (check out `pubsub/rabbitpubsub` for example).

wmark: I've added a `MakeSignBytes` and expanded the comments accordingly.

Consequently we no longer properly close the client and have lost the ability to impersonate other *service accounts*, but at least the latter can be retrofitted by the user if need be. We've gained the ability to see HSM or TPM modules, and “keyless schemes” (a different process holds the key), be utilized for storing a signing key.

wmark:
Not sure what's outstanding – if there's something, please do kindly remind me.

blob/gcsblob/gcsblob.go
```diff
@@ -230,6 +306,56 @@ func OpenBucket(ctx context.Context, client *gcp.HTTPClient, bucketName string,
...
+	return nil, c.err
+}
+
+// CreateMakeSignBytesWith produces a MakeSignBytes variant from an expanded parameter set.
+// It essentially adapts a remote call to the IAM Credentials API
+// to the function signature expected by storage.SignedURLOptions.SignBytes.
+func (c *credentialsClient) CreateMakeSignBytesWith(lifetimeCtx context.Context, googleAccessID string) func(context.Context) SignBytesFunc {
+	return func(requestCtx context.Context) SignBytesFunc {
+		c.init.Do(func() {
+			c.client, c.err = credentials.NewIamCredentialsClient(lifetimeCtx)
```
vangent, line 329: What's this context used for in NewIamCredentialsClient? I am pretty sure it can be cancelled already by this point since it's being called lazily... Maybe better to use a Background instead?

blob/gcsblob/gcsblob.go
```diff
@@ -230,6 +306,56 @@ func OpenBucket(ctx context.Context, client *gcp.HTTPClient, bucketName string,
...
+			if err != nil {
+				return nil, err
+			}
+			return resp.GetSignedBlob(), nil
+		}
+	}
+}
+
+// Close will sever any established connection to the IAM Credentials API.
+// Call this when the corresponding bucket gets closed.
```
vangent, line 352: Comment about calling it when the bucket gets closed is stale; this credentialsClient is created once per application for the initOnce of the default URLOpener.

vangent: (Can probably just delete this function entirely)

blob/gcsblob/gcsblob.go
```diff
@@ -611,6 +737,10 @@ func (b *bucket) SignedURL(ctx context.Context, key string, dopts *driver.Signed
 		PrivateKey:     b.opts.PrivateKey,
 		SignBytes:      b.opts.SignBytes,
 	}
+	otherSignerAvailable := len(opts.PrivateKey) >= 1 || opts.SignBytes != nil
```
vangent, line 740: This "if" is confusing me. Can we just check at bucket creation time that only one of SignBytes/MakeSignBytes is provided, and then simplify this to:

sb := opts.SignBytes
if opts.MakeSignBytes != nil {
  sb = opts.MakeSignBytes(ctx)
}
...

?

blob/gcsblob/gcsblob.go
```diff
@@ -186,11 +255,18 @@ type Options struct {
 	PrivateKey []byte
 
 	// SignBytes is a function for implementing custom signing.
-	// Exactly one of PrivateKey or SignBytes must be non-nil to use SignedURL.
+	// Exactly one of PrivateKey, SignBytes, or MakeSignBytes must be non-nil to use SignedURL.
 	// See https://godoc.org/cloud.google.com/go/storage#SignedURLOptions.
 	SignBytes func([]byte) ([]byte, error)
```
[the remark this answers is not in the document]
wmark: I'd love to, been there [master..1a3645ee](https://github.com/google/go-cloud/compare/527c34b3c659fca56c7862f6df9c0a580ee6a541..1a3645eedc2f61a36e9c91eaab17d88c683e9c39), but it fails the tests: [Travis-CI job #7663](https://travis-ci.com/github/google/go-cloud/jobs/341989060#L722)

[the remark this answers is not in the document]
vangent: OK, never mind.

blob/gcsblob/gcsblob.go
wmark: You're right, but I believe the Background context should be passed further up instead and this left as-is: To enable users to change the AccessID used I'm considering making the `credentialsClient` public (capital `CredentialsClient`; consequently we'd not remove the currently redundant `Close`). Users will be able to instantiate their own. Or at least move it to its own file to make extraction easier.

Options I'd move on: (1) keep it private, remove and simplify as currently suggested by you; (2) private, but in its own file with everything in place; (3) make it public. I'm for (3), but it's no module owned by me.

vangent: I like the idea of making it public so that users can do this on their own if they want to.

I don't think the lazy initialization is adding much value and it's adding significant complexity; so maybe we can just do a func something like:

```
func CredentialsSigner(client *credentials.IamCredentialsClient, googleAccessID string) func(context.Context) SignBytesFunc {
  ...
}
```

You can initialize the client in the default lazyCredsOpener, which already has a init.Once and mechanism for persistently returning errors, and it can use the passed-in context there to instantiate the client. 

Then there's no need for the credentialsClient struct at all.

Users who want to specify a googleAccessID can initialize (and Close) the client on their own.

wmark: My own use-case is, I have this on AppEngine as one part of many, and I'd like to defer anything expensive like establishing remote connections. Further, a user might override the defaults we are about to set and the connection would've been made needlessly.

I chose to believe you'll like the new implementation, especially as it adds value by being reusable (at the expense of the user managing it, i. e. closing it) and serving as example for how to employ a HSM in an on-prem environment.

As to not further scope-creep of this patch, I'll prepare this now but petition to make it public/export in a separate PR with the method we currently won't use (`Close`).

blob/gcsblob/gcsblob.go
wmark: 1. Both need to be checked, `SignBytes` and `PrivateKey`, so we don't involve a `MakeSignBytes` that will initialize something.
2. You've probably meant this exemplary, but the contract is that `MakeSignBytes` won't overwrite a `SignBytes` – the latter is prefered as it's supposedly without context, and apparently without calling any remote API.

wmark:
Thanks for your patience negotiating this. Tests are in, and I believe we've arrived at an expanded functionality that opens up this to use-cases with improved security.

blob/gcsblob/iam.go
```diff
@@ -0,0 +1,81 @@
...
+	return func(requestCtx context.Context) SignBytesFunc {
+		c.init.Do(func() {
+			if c.client != nil {
+				// Set previously, likely to a mock implementation for tests.
+				return
+			}
+			c.client, c.err = credentials.NewIamCredentialsClient(lifetimeCtx)
+		})
+
+		if c.err != nil {
```
vangent, line 64: Seems like it would be simpler to check for `c.err != nil` in the func below...?

E.g.:

return func(p []byte) ([]byte, error) {
  if c.err != nil {
    return nil, c.err
  }
  resp, err := c.client.SignBlob(...)
  ...
}
```

and then delete errPassThrough.

File: blob/gcsblob/gcsblob.go [hunk 12]
```diff
@@ -611,6 +691,10 @@ func (b *bucket) SignedURL(ctx context.Context, key string, dopts *driver.Signed
 		PrivateKey:     b.opts.PrivateKey,
 		SignBytes:      b.opts.SignBytes,
 	}
+	// Avoid (needlessly) calling MakeSignBytes to not trigger any lazy initialization.
```
vangent, line 694: If you do the above, then this can be simplified to `if b.ops.MakeSignBytes != nil`.

blob/gcsblob/iam.go
```diff
@@ -0,0 +1,81 @@
...
+
+	credentials "cloud.google.com/go/iam/credentials/apiv1"
+	gax "github.com/googleapis/gax-go/v2"
+	credentialspb "google.golang.org/genproto/googleapis/iam/credentials/v1"
+)
+
+// credentialsClient wraps the IAM Credentials API client for a lazy initialization
+// and expresses it in the reduced format expected by SignBytes.
+// See https://cloud.google.com/iam/docs/reference/credentials/rest
+type credentialsClient struct {
```
vangent, line 33: I am fine with this for now, but when we make it public I don't think it should be a struct like this, it should just be a function as I suggested in my previous comment.

* Creating a new client is not expensive, I don't think it even blocks.
* The lazy URLOpener is only called if someone is instantiating a URL using it; they can't override the defaults after that.
* The lazy URLOpener already has code supporting the `sync.Once` stuff, it's unnecessary to implement it again here.

vangent:
Thanks again for your patience and for the contribution!

Issue #2803, opened 2020-05-30
External implementation of docstore not possible


sbulman:
It's currently not possible to create an external implementation of the docstore driver interface because the ErrorAs method returns a gcerr.ErrorCode which is from an internal package. The blob driver interface returns a gcerrors.ErrorCode which is not internal.

Can the error return types be aligned so that external driver implementations can be created and forking this repo can be avoided.

PR #2806, opened 2020-05-31
doctore: use gcerrors in driver interface

### LINKED ISSUE #2779: errors: constructing errors with codes is not possible for external implementers
I've been building an external implementation of pubsub and have not been able to get the tests to
pass because I can't construct the appropriate errors. 

The package to construct the errors is https://godoc.org/gocloud.dev/internal/gcerr

Relates to #2778

vangent:
Using the internal type in the interface makes it impossible to write external drivers.

Fixes #2779.

### DIFF
--- a/docstore/awsdynamodb/dynamo.go
+++ b/docstore/awsdynamodb/dynamo.go
@@ -698,30 +698,30 @@ func (c *collection) ErrorAs(err error, i interface{}) bool {
 	return true
 }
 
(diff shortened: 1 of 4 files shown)


vangent:
@sbulman it would be great if you could confirm that this would resolve the issue you reported in #2803.

sbulman:
@vangent I can confirm this resolves my issue

PR #2810, opened 2020-06-05
all: update dependencies

vangent:


vangent:
I had to back out the runtimevar/etcd updates, as they don't appear to be compatible with recent grpc versions.

vangent:
... and, had to back out secrets/hashicorp updates as well, as they depend on go 1.14.

PR #2814, opened 2020-06-09
all: update dependencies

vangent:
My last attempt didn't update a bunch of things, not sure why.

I re-recorded the golden files to make sure things still work. I had to fix a couple of things that no longer did, notably the `Timestamp` proto has some new hidden fields that have to be ignored for tests, and Azure Service Bus returns a new kind of error when a subscription/topic doesn't exist.

I also had to drop the Travis check for `go 1.12` as `go build` wants to modify the `.mod` files. I think this should be OK at this point but let me know if you disagree.


wmark:
(Please don't use Github's web UI to merge, squash, and rebase. It messes up the history.)

vangent:
> (Please don't use Github's web UI to merge, squash, and rebase. It messes up the history.)

Sorry, this is how we roll, per https://github.com/google/go-cloud/blob/master/CONTRIBUTING.md#making-a-pull-request.

PR #2823, opened 2020-06-21
mysql: add FromScheme allowing us to get opener my scheme

alexsn:
Add the ability to fetch `MySQLURLOpener` by scheme. This can be used to patch underlying URLOpeners with additional configs.

For example say we want to customize the trace options awsmysql driver uses:

```go
import (
  "gocloud.dev/mysql"
  "gocloud.dev/mysql/awsmysql"
  "contrib.go.opencensus.io/integrations/ocsql"
)

func init() {
  opener, _ := mysql.DefaultURLMux().FromScheme(awsmysql.Scheme)
  opener.(*awsmysql.URLOpener).TraceOpts = []ocsql.TraceOption{
    ocsql.WithAllowRoot(true),
  }
}
```

### DIFF
--- a/mysql/mysql.go
+++ b/mysql/mysql.go
@@ -106,6 +106,15 @@ func (mux *URLMux) OpenMySQLURL(ctx context.Context, u *url.URL) (*sql.DB, error
 	return opener.(MySQLURLOpener).OpenMySQLURL(ctx, u)
 }
 
+// FromScheme returns a opener for a given scheme.
+func (mux *URLMux) FromScheme(scheme string) (MySQLURLOpener, error) {
+	opener, err := mux.schemes.FromURL("DB", &url.URL{Scheme: scheme})
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


alexsn:
> I think the preferred way to do this is to re-register the mux.
> 
> E.g.,
> 
> ```
> o := &mysql.URLOpener{TraceOpts: ...}
> mysql.DefaultURLMux().RegisterMySQL(mysql.Scheme, o)
> ```

You can't as it [panics](https://github.com/google/go-cloud/blob/5212926ee47fade2a7786b7434c8b8955565cb00/internal/openurl/openurl.go#L50).

vangent:
Yes, you are right, sorry.

I don't think I'm in favor of this change because it feels fragile to me; you are mutating global state, and additionally mutating global state that might change from release to release, with potentially unpredictable results for your application.

Instead, can you create your own `URLMux`, install whatever opener(s) you want, and use that to open URLs instead of `DefaultURLMux`?

alexsn:
That'll work.

PR #2824, opened 2020-06-21
mysql: add trace options to default mysql opener

alexsn:
Add ocsql.TraceOpts to default opener as we do for cloud specific ones.


vangent:
This looks fine to me, thanks for the contribution!

Tests are failing because this is technically a breaking change (adding the new field makes the struct no longer comparable). This is OK, but you need to add a commit with `BREAKING_CHANGE_OK` in the first line of the message to get tests to pass.

```
FAIL. If this is expected and OK, you can pass this check by adding a commit with BREAKING_CHANGE_OK in the first line of the message.
```

alexsn:
@vangent - Can we please merge this?

vangent:
Yup! Thanks for the contribution.

PR #2831, opened 2020-06-28
mysql/awsmysql: Parse DSN options from URL

alexsn:
Same as #2785 but using similar approach as `gcpmysql.go`


vangent:
This change LGTM, @zombiezen can you take a look as well?

vangent:
Hi @alexsn, can you push an empty commit or something? Github is waiting for a Travis build that is borked and never going to finish, and I don't see a way I can retrigger it from here.

alexsn:
I've force pushed after rebasing on master.

PR #2832, opened 2020-06-29
blob/fileblob: fix outdated comment

MetalBlueberry:
Not much to say, just fix outdated comment modified by #2821


MetalBlueberry:
@googlebot I signed it!

PR #2834, opened 2020-07-27
runtimevar/httpvar: use context when watching values

alexsn:


runtimevar/httpvar/httpvar.go
```diff
@@ -193,14 +193,18 @@ func equivalentError(err1, err2 error) bool {
 // service.
 type watcher struct {
 	client   *http.Client
-	endpoint *url.URL
+	endpoint string
```
vangent, line 196: Can we leave this as a `*url.URL` please?

alexsn: Sure, didn't see the point of calling `.String()` on every GET.

Issue #2850, opened 2020-09-02
blob/azureblob: Add support for storage domain in URL opener


stanhu:
Go Cloud already supports an optional `StorageDomain` parameter. Just as AWS has the `endpoint` query parameter, we ought to be able to specify this via a query parameter in the URL opener.

### Describe the solution you'd like

Add a `domain` query parameter that allows setting of this value.

PR #2851, opened 2020-09-02
blob/azureblob: Support storage domain in URL opener (#2850)

stanhu:
This adds support for a `domain` query parameter to specify a different
host suffix for Azure China, Azure Germany, Azure Government, etc.


blob/azureblob/azureblob.go
```diff
@@ -208,12 +209,28 @@ func openerFromEnv(accountName AccountName, accountKey AccountKey, sasToken SAST
 
 // OpenBucketURL opens a blob.Bucket based on u.
 func (o *URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket, error) {
-	for k := range u.Query() {
-		return nil, fmt.Errorf("open bucket %v: invalid query parameter %q", u, k)
+	err := o.setOptionsFromURLParams(u.Query())
```
vangent, line 212: This has a side effect of changing the Option for all future URLs opened by the URLOpener. Please clone the o.Options first, similar to:

https://github.com/google/go-cloud/blob/master/blob/gcsblob/gcsblob.go#L232

You can clone it, and then pass it to setOptionsFromURLParams, which will no longer need to be a function of the struct.

vangent (changes requested):
Thanks for the PR!

vangent (approved):

PR #2855, opened 2020-09-04
mysql: fix DSN parsing from URL

alexsn:
Currently it's not possible to open a mysql instance via URL and when attempted the following error is shows:
URL: mysql://root:root@localhost/testdb
Error: default addr for network 'localhost' unknown

This occurs since mysql.OpenMySQLURL treats a mysql DSN as a URL prefixed with a schema which is incorrect.
In order to fix this I moved the `configFromURL` function I added in https://github.com/google/go-cloud/pull/2831 to mysql package and made it public. In addition I've added a terraform harnest for mysql driver using docker provider.

Fixed https://github.com/google/go-cloud/issues/2856.

### DIFF
--- a/mysql/main.tf
+++ b/mysql/main.tf
@@ -1,0 +1,86 @@
+# Copyright 2020 The Go Cloud Development Kit Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Harness for MySQL tests.
+
+terraform {
+  required_version = ">= 0.13"
+  required_providers {
+    docker = {
+      source  = "terraform-providers/docker"
+      version = "~> 2.0"
+    }
+    random = {
+      source  = "hashicorp/random"
+      version = "~> 2.0"
+    }
+  }
+}
+
+variable port {
+  type        = number
+  description = "Port exposed out of the MySQL container."
+  default     = 3306
+}
+
+resource random_pet mysql {}
+
+resource random_password db_password {
+  special = false
+  length  = 20
+}
+
+locals {
+  db_name = "testdb"
+}
+
+resource docker_container mysql {
+  name  = random_pet.mysql.id
+  image = docker_image.mysql.latest
+
+  env = [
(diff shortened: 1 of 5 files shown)


vangent (commented):
Sorry, this is a bit hard to review since it's a refactor and a bug fix in one (I think).

Could you perhaps separate those? I.e., one PR that's just fixing the bug, and another one that's the refactor to share code?

Consider filing an Issue to describe the problem as well, in case other people are running into it.

alexsn:
Hi @vangent, this PR has 2 commits. The first fixes the bug while the second adds/reactors testing code.

I think it should be easy to review if you do so one commit at a time.

I'll also file as issue linking to this to a bit later.

mysql/mysql.go
```diff
@@ -44,12 +45,37 @@ type URLOpener struct {
...
+		return nil, err
+	}
 	return sql.OpenDB(connector{
-		dsn:       strings.TrimPrefix(u.String(), Scheme+"://"),
+		dsn:       cfg.FormatDSN(),
 		traceOpts: append([]ocsql.TraceOption(nil), uo.TraceOpts...),
 	}), nil
 }
 
+// ConfigFromURL creates a mysql.Config from URL.
```
vangent, line 58: IIUC, this is a straight copy from the previous location, right? And same with the test below?

mysql/mysql.go
```diff
@@ -44,12 +45,37 @@ type URLOpener struct {
 
 // OpenMySQLURL opens a new database connection wrapped with OpenCensus instrumentation.
 func (uo *URLOpener) OpenMySQLURL(_ context.Context, u *url.URL) (*sql.DB, error) {
+	cfg, err := ConfigFromURL(u)
+	if err != nil {
+		return nil, err
+	}
 	return sql.OpenDB(connector{
-		dsn:       strings.TrimPrefix(u.String(), Scheme+"://"),
+		dsn:       cfg.FormatDSN(),
```
vangent, line 53: Is this the bug fix? What's an example `dsn` value before the change and after it?

mysql/main.tf
```diff
@@ -0,0 +1,86 @@
+# Copyright 2018 The Go Cloud Development Kit Authors
```
vangent, line 1: Nit: 2020 please

mysql/mysql.go
alexsn: Yes

mysql/mysql.go
alexsn: In tests `mysql://root@localhost/mysql` was used but didn't work if you tried to send any sql command over the created db (received `default addr for network 'localhost' unknown` error).

The new value that is now used in tests is generated by terraform (e.g. `mysql://root:password@localhost:3306/testdb`) while the test ensures that this works in both `.Open` and `.Ping` calls.

alexsn:
@vangent, any additional changes required? If not could you please merge this PR?

vangent (approved):

PR #2861, opened 2020-09-12
blob/all: adding BeforeSign support to SignedURL operation

alexsn:
`BeforeSign` is similar to `BeforeRead` and friends allowing users to hook into SignedURL operation. In my case it's needed to set `ContentDisposition` before signing s3blob get request.

### DIFF
--- a/blob/drivertest/drivertest.go
+++ b/blob/drivertest/drivertest.go
@@ -75,6 +75,8 @@ type HarnessMaker func(ctx context.Context, t *testing.T) (Harness, error)
 //    on the single blob entry returned.
 // 7. Tries to read a non-existent blob, and calls ErrorCheck with the error.
 // 8. Makes a copy of the blob, using BeforeCopy as a CopyOption.
+// 9. Calls SignedURL using BeforeSign as a SignedURLOption for each supported
+//    signing method (i.e. GET, PUT and DELETE).
 //
 // For example, an AsTest might set a driver-specific field to a custom
 // value in BeforeWrite, and then verify the custom value was returned in
@@ -98,6 +100,9 @@ type AsTest interface {
 	// BeforeList will be passed directly to ListOptions as part of listing the
 	// test blob.
 	BeforeList(as func(interface{}) bool) error
+	// BeforeSign will be passed directly to SignedURLOptions as part of
+	// generating a signed URL to the test blob.
+	BeforeSign(as func(interface{}) bool) error
 	// AttributesCheck will be called after fetching the test blob's attributes.
 	// It should call attrs.As and verify the results.
 	AttributesCheck(attrs *blob.Attributes) error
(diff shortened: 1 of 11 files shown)


vangent (changes requested):
Thanks for the contribution! Overall the change looks good to me, but it needs tests.

Can you extend drivertest's TestAs to cover this?
https://github.com/google/go-cloud/blob/master/blob/drivertest/drivertest.go#L67

Each of (azure, GCS, AWS) should provide the hook for testing their specific As tests; the existing BeforeWrite and BeforeCopy tests should give you an example to follow.

I *think* that Signing is done locally by all the providers, so that this won't need updates to the record/replay files. If that turns out not to be true, let me know and we can figure out what to do (it's hard for external contributors to do; I would need to figure out how to upload a commit to your pull request, which I have done in the past but can't remember how to do).

vangent (commented):
Also, this is technically a breaking change (because SignedURLOptions is changing from comparable to not comparable), so you'll need to add a commit with BREAKING_CHANGE_OK in the first line of the message.

blob/drivertest/drivertest.go
```diff
@@ -98,6 +100,9 @@ type AsTest interface {
 	// BeforeList will be passed directly to ListOptions as part of listing the
 	// test blob.
 	BeforeList(as func(interface{}) bool) error
+	// BeforeSign will be passed directly to SignedURLOptions as part of
+	// generating a singed URL to the test blob.
```
vangent, line 104: Nit: signed

blob/s3blob/s3blob.go
```diff
@@ -727,30 +732,66 @@ func (b *bucket) Delete(ctx context.Context, key string) error {
...
+			}
+			if err := opts.BeforeSign(asFunc); err != nil {
+				return "", err
+			}
+		}
+		req, _ = b.client.DeleteObjectRequest(in)
 	default:
 		return "", fmt.Errorf("unsupported Method %q", opts.Method)
 	}
+	req.SetContext(ctx)
```
vangent, line 795: This line looks new. What does it do? We don't want the `ctx` passed to `SIgnedURL` to be used after returning from this call.... My guess is that setting the context has no effect on `Presign`, but I haven't looked at the docs.

vangent (commented):
Thanks for adding the tests, looks great. One question remaining.

blob/s3blob/s3blob.go
alexsn: I've added it since the `SignedURL` signature accepts a context but it's ignored in the handler. I can remove it if you prefer as it's probably a nop and exists since the same input type is used for both sign / read.

vangent: Yea, let's remove it. The SignedURL `ctx` is intended to be used as a deadline for the signing (in case there's a driver that needs to make an RPC to sign). This driver doesn't need it since the signing is just a library call. The `ctx` is not intended to be used as part of the request that's being signed itself.

I agree that it's probably a no-op in the end, but it's a bit misleading to set it.

alexsn: Done.

vangent (approved):

PR #2863, opened 2020-09-28
blob/s3blob: Cache uploaders between calls to `NewTypedWriter`

### LINKED ISSUE #2807: blob/s3: High memory usage during multi-part uploads
### Describe the bug

High memory usage when doing many parallel multi-part uploads.

### To Reproduce

Spin up a bunch of goroutines doing muli-part uploads with a large part size.

## Expected behavior

Low memory usage.

### Version

Latest

### Additional context
 
The issue seems to be caused by the fact that the S3 implementation creates a new multi-part uploader for each writer: https://github.com/google/go-cloud/blob/master/blob/s3blob/s3blob.go#L626

The multipart uploader maintains a pool of byte slices and as a result when there are a lot of parallel uploads you end up with a pool of parts sitting in memory for each upload until the G.C has time to clean them up.

I think this issue could be solved by caching uploaders based on the uploader options and then reusing them (since they're already thread-safe). I can make a P.R for this change if its acceptable.

richardartoul:
Fixes #2807

### DIFF
--- a/blob/s3blob/s3blob.go
+++ b/blob/s3blob/s3blob.go
@@ -295,9 +295,44 @@ func (w *writer) Close() error {
 
 // bucket represents an S3 bucket and handles read, write and delete operations.
 type bucket struct {
+	sync.RWMutex
+
 	name          string
 	client        *s3.S3
 	useLegacyList bool
+
+	// cachedUploader stores a cached uploader along with a key specifying its
+	// configuration. Reusing the same upload over and over again (when possible)
+	// is important because the uploader maintains an internal buffer pool for
+	// multi-part uploads and creating a new uploader for each upload results
+	// in a large number of temporary []byte allocations.
+	cachedUploader *cachedUploader
+}
+
+// cachedUploader represent a cached uploader that can be reused as long as its
+// key matches the desired uploader configuration.
+type cachedUploader struct {
+	key      uploaderKey
+	uploader *s3manager.Uploader
+}
+
+func newCachedUploader(key uploaderKey, uploader *s3manager.Uploader) *cachedUploader {
+	return &cachedUploader{
+		key:      key,
+		uploader: uploader,
+	}
+}
+
+// uploaderKey contains all custom configuration for the uploader so that
+// it can be determined if a cached uploader can serve a callers request.
+type uploaderKey struct {
+	partSize int
+}
+
+func newUploaderKey(partSize int) uploaderKey {
+	return uploaderKey{
+		partSize: partSize,
+	}
 }
 
 func (b *bucket) Close() error {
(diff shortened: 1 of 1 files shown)


blob/s3blob/s3blob.go
```diff
@@ -295,9 +295,44 @@ func (w *writer) Close() error {
...
+func newCachedUploader(key uploaderKey, uploader *s3manager.Uploader) *cachedUploader {
+	return &cachedUploader{
+		key:      key,
+		uploader: uploader,
+	}
+}
+
+// uploaderKey contains all custom configuration for the uploader so that
+// it can be determined if a cached uploader can serve a callers request.
+type uploaderKey struct {
```
vangent, line 328: Why not use `driver.WriterOptions` directly as the key?

blob/s3blob/s3blob.go
```diff
@@ -628,11 +663,33 @@ func unescapeKey(key string) string {
 // NewTypedWriter implements driver.NewTypedWriter.
 func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType string, opts *driver.WriterOptions) (driver.Writer, error) {
 	key = escapeKey(key)
-	uploader := s3manager.NewUploaderWithClient(b.client, func(u *s3manager.Uploader) {
-		if opts.BufferSize != 0 {
-			u.PartSize = int64(opts.BufferSize)
-		}
-	})
+
+	// Check if a cached uploader already exists.
```
vangent, line 667: This would be more readable if it were factored out into a helper function, e.g. `func (b *bucket) uploader(opts *driver.WriterOptions) *s3manager.Uploader`.

blob/s3blob/s3blob.go
```diff
@@ -628,11 +663,33 @@ func unescapeKey(key string) string {
...
+	b.RLock()
+	cachedUploader := b.cachedUploader
+	b.RUnlock()
+
+	var (
+		uploader    *s3manager.Uploader
+		uploaderKey = newUploaderKey(opts.BufferSize)
+	)
+	if cachedUploader != nil && uploaderKey == cachedUploader.key {
+		// If a cached uploader already exists *and* it has the same key/configuration
```
vangent, line 677: So if some writes use one BufferSize, and other writes use another BufferSize, we'll never get to reuse the uploader? Maybe have a map instead of a single cached one?

blob/s3blob/s3blob.go
```diff
@@ -628,11 +663,33 @@ func unescapeKey(key string) string {
...
+		// as requested by the caller then reuse it.
+		uploader = cachedUploader.uploader
+	} else {
+		// Otherwise create a new uploader.
```
vangent, line 681: There is a race here; two goroutines could enter this function, both get a `nil` `cachedUploader`, and then both create a new one. One of them will win to be saved. Using a helper function that grabs the lock for the entire scope of the function should fix this.

blob/s3blob/s3blob.go
vangent: Please add a test for this new function as well.

vangent (changes requested):
Some comments on the implementation, but my main question is whether this is necessary at all. "increased memory until GC gathers it" does not sound like it's an actual problem.

segevfiner:
Actually, it might be possible to use the same uploader but override options per upload by virtue of the `Upload` methods taking `func(*Uploader)` that allows for it.

PR #2866, opened 2020-10-14
all: fix godoc URL

segevfiner:
It linked to the old import path so you got documentation for an old version

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ compile but are not supported.
 Documentation for the project lives primarily on https://gocloud.dev/, including
 [tutorials][].
 
-You can also browse Go package reference on [godoc.org][godoc].
+You can also browse Go package reference on [pkg.go.dev][PkgGoDev].
 
 [tutorials]: https://gocloud.dev/tutorials/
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,7 +49,7 @@ talk at Next 2018:
 [![Video: Building Go Applications for the Open Cloud (Cloud Next '18)](https://img.youtube.com/vi/_2ZwhvIkgek/0.jpg)][video]
 
 [announcement blog post]: https://blog.golang.org/go-cloud
-[godoc]: https://godoc.org/github.com/google/go-cloud
+[godoc]: https://pkg.go.dev/gocloud.dev
```
eliben, line 52: How about https://pkg.go.dev/mod/gocloud.dev ?

eliben (changes requested):

README.md
segevfiner: Also an option. Not sure what they recommend to use. There is also an official badge if you prefer https://pkg.go.dev/badge/.

eliben (approved):
Thanks!

PR #2869, opened 2020-10-14
blob: Return Unimplemented errors from SignedURL when the driver supports it but hasn't been configured correctly

### LINKED ISSUE #2867: blob/fileblob: SignedURL returns an errors.New instead of gcerr.Error with code Unimplemented when URLSigner is not set
### Describe the bug

When you call `bucket.SignedURL` with the fileblob driver without URLSigner set, you receive back an plain `errors.New` error:
https://github.com/google/go-cloud/blob/6be5a462804a1e9eb8bd63feb2d14f7aac068b0c/blob/fileblob/fileblob.go#L652

Instead of a `gcerr.Error` with code `Unimplemented` as the API is documented to return. This makes code that degrades gracefully when `SignedURL` is not supported to fail without explicitly checking for this error. Though just returning `Unimplemented` can make it difficult for someone to figure out that he can set `URLSigner` unless the error also includes a message to such an effect in addition to the error code.

### To Reproduce

```go
blob.OpenBucket("file:///some_path").SignedURL(context.Background(), "key", nil)
```

## Expected behavior

It returns a `gcerr.Error` with code `Unimplemented`.

### Version

v0.20.0

### Additional context
Go 1.15.2 using Go modules.

vangent:
Fixes #2867

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -589,9 +589,6 @@ func (b *bucket) ErrorAs(err error, i interface{}) bool {
 }
 
 func (b *bucket) ErrorCode(err error) gcerrors.ErrorCode {
-	if code := gcerrors.Code(err); code != gcerrors.Unknown {
-		return code
-	}
 	serr, ok := err.(azblob.StorageError)
 	switch {
 	case !ok:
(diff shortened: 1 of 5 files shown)


blob/azureblob/azureblob.go
```diff
@@ -589,9 +589,6 @@ func (b *bucket) ErrorAs(err error, i interface{}) bool {
 }
 
 func (b *bucket) ErrorCode(err error) gcerrors.ErrorCode {
-	if code := gcerrors.Code(err); code != gcerrors.Unknown {
```
vangent, line 592: Note: I moved this to the concrete class (blob.go). If the driver produces an error that already has an error code, we should always just use it.

PR #2874, opened 2020-10-26
`internal/website`: fix wrong document link

shibukawa:
Docstore's how to guide points to wrong web url (the label is memdocstore, but link destination is mongodocstore). This PR just fix this link issue.


eliben (approved):
Thanks!

PR #2873, opened 2020-10-23
blob/azblob: feat added in MSI based authentication for Azure buckets

chrismellard:
Fixes #2870

PR #2875, opened 2020-10-30
blob/azureblob: Update azure-storage-blob-go to v0.11.0

johejo:
refs https://github.com/Azure/azure-storage-blob-go/blob/master/ChangeLog.md#version-0110

Fixed because v0.11.0 included breaking change in API.


johejo:
Is there any suggestion to fix CI build? Help wanted.

vangent:
> Fixed because v0.11.0 included breaking change in API

Reference? The release notes link you provided doesn't note any breaking changes.

I suspect that I will have to take over this PR; the tests are probably breaking because the new version changed something internal such that the record/replay tests don't work anymore and we need to re-record them. Doing so depends on our Azure project etc.

johejo:
Sorry for the misunderstanding. 
Breaking change in my sense means that it contains changes to public methods and structs. 

It's a small change, so maintainers are free to take over this change on another branch, etc. 
Thank you.

vangent:
I am trying to take this over, but sadly it is non-trivial.

The newer versions of these Azure libraries appear to include some dynamic content (a "blockid") in the HTTP body of the `PUT` request. The record/replay library we use (https://github.com/google/go-replayers) allows clearing or ignoring request parameters and/or headers during matching, but does not support manipulating the Body itself.

I will probably have to add a feature to that library to support that, then proceed with this. It might take a while.

vangent:
Obsoleted by #2873 2876.

vangent:
Sorry, #2876.

PR #2876, opened 2020-11-05
blob/azureblob: Update azure-storage-blob-go to v0.11.0.

vangent:
Taking over #2875 since the change requires re-recording record/replay files.

This update from Azure also adds some dynamic content to the Body of some requests, which required a new feature for go-replayers.


vangent:
@johejo FYI

vangent:
Ping Eli?

eliben (approved):
Bah, sorry I missed the notification

PR #2888, opened 2020-11-18
blob: Add CreateTime and ETag to blob.Attributes

### LINKED ISSUE #2887: blob: Add ETag and Date to Attributes
I'm trying to implement support for conditional requests/caching using `blob`. It requires passing in and returning various HTTP headers.

As one part of this, `ETag` and `Date` can be added to `Attributes`.

See https://github.com/google/go-cloud/issues/2881#issuecomment-729888206

Having to retrieve them in a provider dependent way.

### Additional context
See https://github.com/google/go-cloud/issues/2881

vangent:
Fixes #2887 .

### DIFF
--- a/blob/fileblob/fileblob.go
+++ b/blob/fileblob/fileblob.go
@@ -484,7 +484,7 @@ func (b *bucket) Attributes(ctx context.Context, key string) (*driver.Attributes
 		ModTime: info.ModTime(),
 		Size:    info.Size(),
 		MD5:     xa.MD5,
-		ETag:    fmt.Sprintf("W/%d", info.ModTime().UnixNano()),
+		ETag:    fmt.Sprintf("%x-%x", info.ModTime().UnixNano(), info.Size()),
 		AsFunc: func(i interface{}) bool {
 			p, ok := i.(*os.FileInfo)
 			if !ok {
--- a/blob/fileblob/fileblob.go
+++ b/blob/fileblob/fileblob.go
@@ -480,9 +480,11 @@ func (b *bucket) Attributes(ctx context.Context, key string) (*driver.Attributes
 		ContentLanguage:    xa.ContentLanguage,
 		ContentType:        xa.ContentType,
 		Metadata:           xa.Metadata,
-		ModTime:            info.ModTime(),
-		Size:               info.Size(),
-		MD5:                xa.MD5,
+		// CreateTime left as the zero time.
+		ModTime: info.ModTime(),
+		Size:    info.Size(),
+		MD5:     xa.MD5,
+		ETag:    fmt.Sprintf("W/%d", info.ModTime().UnixNano()),
 		AsFunc: func(i interface{}) bool {
 			p, ok := i.(*os.FileInfo)
 			if !ok {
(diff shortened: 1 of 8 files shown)


vangent:
@segevfiner FYI

segevfiner:
@vangent See my two messages here https://github.com/google/go-cloud/issues/2887#issuecomment-730024650 https://github.com/google/go-cloud/issues/2887#issuecomment-730037902

(Just making sure, they won't somehow slip under your radar 😉, thanks for taking the time looking at all this)

blob/fileblob/fileblob.go
```diff
@@ -480,9 +480,11 @@ func (b *bucket) Attributes(ctx context.Context, key string) (*driver.Attributes
 		ContentLanguage:    xa.ContentLanguage,
 		ContentType:        xa.ContentType,
 		Metadata:           xa.Metadata,
+		CreateTime:         info.ModTime(),
 		ModTime:            info.ModTime(),
 		Size:               info.Size(),
 		MD5:                xa.MD5,
+		ETag:               fmt.Sprintf("W/%d", info.ModTime().UnixNano()),
```
segevfiner, line 487: Check out https://serverfault.com/a/690374 

Basically Nginx set the ETag based on the modification time and size and not as a weak etag. Easy to do here and should probably be better as it is done by a production grade server.

Also ETag has strange mandatory quotes for it to work as far as I remember, e.g.

```
ETag: "XYZ"
Or:
ETag: W/"XYZ"
```

(Probably applies to memblob as well)

blob/memblob/memblob.go
```diff
@@ -332,12 +333,17 @@ func (w *writer) Close() error {
 			ContentType:        w.contentType,
 			Metadata:           w.metadata,
 			Size:               int64(len(content)),
-			ModTime:            time.Now(),
+			CreateTime:         now,
```
segevfiner, line 336: Are we sure we don't just want to leave it at zero, so the user can explicitly know it is not supported?

(Also for the other drivers that don't support it)

vangent: Done.

blob/fileblob/fileblob.go
vangent: Thanks! I updated fileblob and memblob to include both mod time and size, and dropped the /W prefix.

I updated the concrete type (`blob`) to note that it strips surrounding quotes (providers are not consistent about whether they include them).

segevfiner:
Are you sure you want to strip the quotes rather than also include them in fileblob/memblob ETag? Because when later propagating it to an HTTP header, the quotes are required. And they are also needed when later passing it to `If-None-Match` (Including the `W/` prefix if any AFAIK).

vangent:
> Are you sure you want to strip the quotes rather than also include them in fileblob/memblob ETag? Because when later propagating it to an HTTP header, the quotes are required. And they are also needed when later passing it to `If-None-Match` (Including the `W/` prefix if any AFAIK).

Fixed to keep the quotes since they are mentioned explicitly in the RFC. I also added some basic syntax checking in `drivertest` to verify consistency (e.g., `gcsblob` seems to strip the quotes, so I updated the driver to add them).

eliben (approved):

PR #2889, opened 2020-11-19
blob: Add a ListPage function for listing blobs in pages

### LINKED ISSUE #2880: blob: Add/Expose ListPaged in blob.Bucket
When listing buckets contents in a web services, it is common/imperative to do so as a paginated API when the buckets can be big so as not to run out of memory or return a huge response all at once. To do so, you need an API that allows you to page through the results manually with the underlying pagination mechanism of the object/blob store. Often a keyset pagination where you set the size of the page and get a key to be passed to the next call to get the next one.

This exists in the underlying blob driver as `ListPaged` and is wrapped by `List` to provide an API that iterates through the pages for you, but `ListPaged` itself is not exposed in the `blob.Bucket` interface to be used in the scenario you want to paginate by yourself.

Expose a `ListPaged` method in `blob.Bucket` that allows the application to manage pagination by itself.

Not using the package for listing buckets. Trying to use the underlying blob driver interface directly (But the package purposefully makes this difficult)

### Additional context
https://pkg.go.dev/gocloud.dev/blob#Bucket.List
vs
`ListPaged` in https://pkg.go.dev/gocloud.dev@v0.20.0/blob/driver#Bucket

vangent:
Fixes #2880.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -39,6 +39,7 @@
 //  - Attributes
 //  - Copy
 //  - Delete
+//  - ListPage
 //  - NewRangeReader, from creation until the call to Close. (NewReader and ReadAll
 //    are included because they call NewRangeReader.)
 //  - NewWriter, from creation until the call to Close.
@@ -602,15 +603,23 @@ var FirstPageToken = []byte("first page")
 // ListPage returns a page of ListObject results for blobs in a bucket, in lexicographical
 // order of UTF-8 encoded keys.
 //
-// To fetch the first page, pass FirstPageToken as the pageToken. For subsequent pages, pass the pageToken
-// returned from a previous call to ListPage. It is not possible to "skip ahead" pages.
+// To fetch the first page, pass FirstPageToken as the pageToken. For subsequent pages, pass
+// the pageToken returned from a previous call to ListPage.
+// It is not possible to "skip ahead" pages.
 //
 // Each call will return pageSize results, unless there are not enough blobs to fill the
-// page. If there are no more blobs available, the returned pageToken will be nil. Calling List
-// Page with a nil pageToken will return io.EOF.
+// page, in which case it will return fewer results (possibly 0).
 //
-// The underlying implementation fetches results in pages, but one call to LastPage may
-// require multiple page fetches (and multiple calls to the BeforeList callback).
+// If there are no more blobs available, ListPage will return an empty pageToken. Note that
+// this may happen regardless of the number of returned results -- the last page might have
+// 0 results (i.e., if the last item was deleted), pageSize results, or anything in between.
+//
+// Calling ListPage with an empty pageToken will immediately return io.EOF. When looping
+// over pages, callers can either check for an empty pageToken, or they can make one more
+// call and check for io.EOF.
+//
+// The underlying implementation fetches results in pages, but one call to ListPage may
+// require multiple page fetches (and therefore, multiple calls to the BeforeList callback).
 //
 // A nil ListOptions is treated the same as the zero value.
 //
@@ -640,6 +649,10 @@ func (b *Bucket) ListPage(ctx context.Context, pageToken []byte, pageSize int, o
 	if b.closed {
 		return nil, nil, errClosed
 	}
+
+	ctx = b.tracer.Start(ctx, "ListPage")
+	defer func() { b.tracer.End(ctx, err) }()
+
 	dopts := &driver.ListOptions{
 		Prefix:     opts.Prefix,
 		Delimiter:  opts.Delimiter,
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -591,6 +591,88 @@ func (b *Bucket) List(opts *ListOptions) *ListIterator {
 	return &ListIterator{b: b, opts: dopts}
 }
 
+// FirstPageToken is the pageToken to pass to ListPage to retrieve the first page of results.
+var FirstPageToken = []byte("first page")
+
+// ListPage returns a page of ListObject results for blobs in a bucket, in lexicographical
+// order of UTF-8 encoded keys.
+//
+// To fetch the first page, pass FirstPageToken as the pageToken. For subsequent pages, pass the pageToken
+// returned from a previous call to ListPage. It is not possible to "skip ahead" pages.
+//
+// Each call will return pageSize results, unless there are not enough blobs to fill the
+// page. If there are no more blobs available, the returned pageToken will be nil. Calling List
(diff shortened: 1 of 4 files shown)


vangent:
@segevfiner

segevfiner:
I wonder if indeed calling listing with a decreasing max size to get exact result pages will work correctly... It really depends on what reasons the various object/blob stores had for returning a short result set to begin with. e.g. Is it because they query for the number of blobs requested and than transform them according to `Delimiter` into `prefixes` without querying for more results, will this work correctly than? I don't really know and totally guessing here.

segevfiner:
I wonder if indeed calling listing with a decreasing max size to get exact result pages will work correctly... It really depends on what reasons the various object/blob stores had for returning a short result set to begin with. e.g. Is it because they query for the number of blobs requested and than transform them according to `Delimiter` into `prefixes` without querying for more results, will this work correctly than? I don't really know and totally guessing here. Like it could be totally fine and better than just having the UI cope if it does work well.

vangent:
> I wonder if indeed calling listing with a decreasing max size to get exact result pages will work correctly... 

It should eventually work. Drivers are required to return an empty `NextPageToken` when there are more results, or else the existing `List` wouldn't work. Drivers are allowed to return fewer than pageSize results, including 0, but eventually they will have to return something. If it takes too long `ctx` could expire.

In practice, it should almost always work in a single call. The second call to retrieve more results should be rare; we have to handle it though.

I like that this change was possible purely in the concrete type :-).

@zombiezen

blob/blob.go
```diff
@@ -591,6 +591,88 @@ func (b *Bucket) List(opts *ListOptions) *ListIterator {
...
+var FirstPageToken = []byte("first page")
+
+// ListPage returns a page of ListObject results for blobs in a bucket, in lexicographical
+// order of UTF-8 encoded keys.
+//
+// To fetch the first page, pass FirstPageToken as the pageToken. For subsequent pages, pass the pageToken
+// returned from a previous call to ListPage. It is not possible to "skip ahead" pages.
+//
+// Each call will return pageSize results, unless there are not enough blobs to fill the
+// page. If there are no more blobs available, the returned pageToken will be nil. Calling List
```
eliben, line 604: micro-nit: should `ListPage` be a single word here?

[github PR review formatting may be fooling me]

blob/blob.go
```diff
@@ -591,6 +591,88 @@ func (b *Bucket) List(opts *ListOptions) *ListIterator {
...
+// Page with a nil pageToken will return io.EOF.
+//
+// The underlying implementation fetches results in pages, but one call to LastPage may
```
eliben, line 607: `LastPage` or `ListPage`?

blob/blob.go
```diff
@@ -591,6 +591,88 @@ func (b *Bucket) List(opts *ListOptions) *ListIterator {
...
+
+// ListPage returns a page of ListObject results for blobs in a bucket, in lexicographical
+// order of UTF-8 encoded keys.
+//
+// To fetch the first page, pass FirstPageToken as the pageToken. For subsequent pages, pass the pageToken
+// returned from a previous call to ListPage. It is not possible to "skip ahead" pages.
+//
+// Each call will return pageSize results, unless there are not enough blobs to fill the
+// page. If there are no more blobs available, the returned pageToken will be nil. Calling List
+// Page with a nil pageToken will return io.EOF.
```
eliben, line 605: Would it be clearer to talk about an "empty pageToken", given that it's a slice type? nil is a bit confusing here

blob/example_test.go
```diff
@@ -289,6 +289,80 @@ func ExampleBucket_List_withDelimiter() {
...
+	defer cleanup()
+
+	// Create the file-based bucket.
+	bucket, err := fileblob.OpenBucket(dir, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer bucket.Close()
+
+	// Create some blob objects for listing: "foo[0..4].txt".
```
eliben, line 305: bound in commend different from code.

blob/blob_test.go
```diff
@@ -132,27 +129,72 @@ func TestListIterator(t *testing.T) {
...
+	got := [][]string{}
+	for {
+		page, token, err := b.ListPage(ctx, nextToken, 2, nil)
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			t.Fatal(err)
+		}
+		if len(page) > 0 {
```
eliben, line 153: But if `len(page) == 0` we continue looping?

This makes me realize it's not clear how to treat all return values of `ListPage` consistently. `err == io.EOF` signals "done", but also when the page is empty we're done, and also when token is nil we're done.

`ListPage` should document which is true when a bit more precisely so there would be a very obvious way of writing this loop.

eliben (approved):

blob/blob.go
```diff
@@ -591,6 +591,88 @@ func (b *Bucket) List(opts *ListOptions) *ListIterator {
...
+	}
+
+	// FirstPageToken fetches the first page. Drivers use nil.
+	// The public API doesn't use nil for the first page because it would be too easy to
+	// keep fetching forever (since the last page return nil for the next pageToken).
+	if bytes.Equal(pageToken, FirstPageToken) {
+		pageToken = nil
+	}
+	b.mu.RLock()
+	defer b.mu.RUnlock()
```
Segflow, line 634: Wouldn't this defer impact performance, as the lock will only be released at the end? I see that it's used to only protect `b.closed`.

blob/blob.go
vangent: Yes, done.

blob/blob.go
vangent: ListPage, fixed.

blob/blob.go
vangent: Done.

blob/example_test.go
vangent: Fixed.

blob/blob_test.go
vangent: The conditional here was not necessary, removed.

The primary signal for ListPage being done is the empty token being returned. You can call it one more time if you want, but it will return an error (io.EOF). I tried to clarify the documentation a bit.

blob/blob.go
vangent: No, it shouldn't -- it's an `RLock` so it doesn't block any other `Bucket` calls other than the call to `Close` itself, and that is by design. A call to `Close` will block until outstanding calls are completed.

Issue #2890, opened 2020-11-20
pubsub/azuresb: return error on subscription not found


Segflow:
### Describe the bug

Opening a not found azure subscription will cause `.Receive` to never return. 

This happens because the return value of https://github.com/google/go-cloud/blob/master/pubsub/azuresb/azuresb.go#L458-L478 is not check and that's where the error can be detected. 

There is this comment in `openSubscription` 
```
// Initialize a link to the AMQP server, but save any errors to be
// returned in ReceiveBatch instead of returning them here, because we
// want "subscription not found" to be a Receive time error.
```

But on azure, if a subscription is not found, but the namespace exists, linkError will be `nil`. 

Here is what the error would look like if we capture and spew.Dump() it:

```
(*amqp.Error)(0xc000bae0c0)(*Error{Condition: amqp:not-found, Description: The messaging entity 'xxx:Topic:compute-events|subs1' could not be found. To know more visit https://aka.ms/sbResourceMgrExceptions.  TrackingId:988e3380-297f-45c1-81fb-afafa35eefcc_B6, SystemTracker:xxx:Topic:xxx-events|subs1, Timestamp:2020-11-20T09:40:45 TrackingId:4044f0ab55b74054bf86ae267091f316_G8, SystemTracker:gateway7, Timestamp:2020-11-20T09:40:45, Info: map[]})
```

### To Reproduce

Receive from any non existent azuresb subscription

## Expected behavior

`.Receive()` to return an error if the subscription is not found

### Version

v0.20

PR #2891, opened 2020-11-20
pubsub/azuresb: check for service bus ReceiveOne error

Segflow:
This PR checks the error value returned by azure's `ReceiveOne`. Previously the error was simply ignored. Which led to few things:

- If subscription is not found: A call to `ReceiveBatch` will never return.
- If subscription got deleted after initial connection (`NewSubscription`): Same as above, `ReceiveBatch` will block forever

With this PR, we check for the error.

Driver's ReceiveBatch should not handle retry, but in azuresb implemenation, we Receive messages one by one until we reach the desired message count. For each single message receive we use rctx,, so we should retry the call if the cause is rctx is done. Otherwise we return the error and let the concrete type decide.

Previously we loop until rctx is done, which I believe is wrong, we should loop over `ctx` is done, and retry when the error is caused by `rctx` being done.

Fixes https://github.com/google/go-cloud/issues/2890


Segflow:
@googlebot I fixed it.

pubsub/azuresb/azuresb.go
```diff
@@ -449,13 +449,14 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*dr
 	defer cancel()
 	var messages []*driver.Message
 
-	// Loop until rctx is Done, or until we've received maxMessages.
-	for len(messages) < maxMessages && rctx.Err() == nil {
+	// Loop until ctx is Done, or until we've received maxMessages.
+	for len(messages) < maxMessages && ctx.Err() == nil {
```
vangent, line 453: I don't think we want this change, as it changes the behavior of ReceiveBatch from waiting at most listenerTimeout (1 second) to waiting until ctx is Done.

pubsub/azuresb/azuresb.go
```diff
@@ -476,6 +477,15 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*dr
 			})
 			return nil
 		}))
+
+		if err != nil {
```
vangent, line 481: Per the driver contract for ReceiveBatch (https://github.com/google/go-cloud/blob/master/pubsub/driver/driver.go#L124),

```
	// If there is a transient failure, this method should not retry but
	// should return a nil slice and an error. The concrete API will take
	// care of retry logic.
```

vangent (commented):
Can you explain the error condition a bit more? There is definitely an existing test for trying to connect to a non-existent subscription (and IIRC it fails with a non-nil linkErr). How can we reproduce the error you're trying to fix?

Segflow:
To reproduce you just need to open a non found subscription like:
```
sub, err := pubsub.OpenSubscription(ctx, "azuresb://existant_opic?subscription=NOT_FOUND_SUBS")
if err != nil {
	panic(err)
}
```

Now 
```
sub.Receive(ctx)
```

Will hang forever. 

When printing the error returned here: https://github.com/google/go-cloud/blob/master/pubsub/azuresb/azuresb.go#L458-L478 (which is ignored right now) here is the output:
```
(*amqp.Error)(0xc000bae0c0)(*Error{Condition: amqp:not-found, Description: The messaging entity 'xxx:Topic:compute-events|subs1' could not be found. To know more visit https://aka.ms/sbResourceMgrExceptions.  TrackingId:988e3380-297f-45c1-81fb-afafa35eefcc_B6, SystemTracker:xxx:Topic:xxx-events|subs1, Timestamp:2020-11-20T09:40:45 TrackingId:4044f0ab55b74054bf86ae267091f316_G8, SystemTracker:gateway7, Timestamp:2020-11-20T09:40:45, Info: map[]})
```

So it did not fail when doing `OpenSubscription`, and also even though the subscription is not found, `s.linkErr` was still `nil` after `OpenSubscription` call

vangent:
> sub, err := pubsub.OpenSubscription(ctx, "azuresb://existant_opic?subscription=NOT_FOUND_SUBS")
> So it did not fail when doing `OpenSubscription`, and also even though the subscription is not found, `s.linkErr` was still `nil` after `OpenSubscription` call

So, we have this test:
https://github.com/google/go-cloud/blob/master/pubsub/azuresb/azuresb_test.go#L139

Is this difference here that the topic exists, but the subscription does not? (It looks like in the test we are using a nonexistent topic + nonexistent subscription).

vangent:
> (*amqp.Error)(0xc000bae0c0)(*Error{Condition: amqp:not-found, 

So could this PR just add a check for this specific error and return it?

Segflow:
Exactly yes, if the topic does not exists, it fails fine, if topic exist but subscription does not exists, it never ends.

> So could this PR just add a check for this specific error and return it?

What about other errors? 

(will revert ctx changes)

Segflow:
I think we should return all the error, and let the concrete type decide no?

vangent:
> I think we should return all the error, and let the concrete type decide no?

I think that should be fine. A common case will be a "context deadline exceeded" error due to the 1 second rctx expiring, but it is a "transient error" so I think that the concrete type should handle it.

Segflow:
@vangent reverted the ctx changes, and did as you suggested. Let me know if this looks good

vangent (approved):

Segflow:
Thanks @vangent for your quick review! This is the first PR to many to come. My team decided to adopt go-cloud early on

PR #2894, opened 2020-11-20
blob: Add a bucket.IsAccessible

### LINKED ISSUE #2892: blob: Add bucket method for "Does Bucket Exist? It it accessible?"

If a bucket doesn't exist at all, there is no way to use `gocloud.dev` to check if the bucket exists at all. My current solution is: 

```go
func canAccessBucket(ctx context.Context, bucket *blob.Bucket) bool {
	iter := bucket.List(&blob.ListOptions{Delimiter: "/"})
	_, err := iter.Next(ctx)
	if err != nil && !errors.Is(err, io.EOF) {
		return false
	}
	return true
}
```

But this has a few problems:
1. It's slower than necessary to list out a bucket just to see if it exists
2. For Azure, it takes 60s to timeout (i will file a different Issue for this)
3. Because Azure does not return a useful error code, we have to assume that *any and all* errors that are not `io.EOF` mean the bucket does not exist: this is dangerous! We could be swallowing legit errors, and instead should be checking for a specific "Bucket Doesn't Exist" error.

Please see https://github.com/google/go-cloud/issues/2712 for some additional context

I would like a method on bucket that returns a boolean (and an error too for exceptional errors) which is `true` if the bucket exists and is accessible and `false` otherwise:

```
bucketExists, err := bucket.IsAccessible(ctx)
```

vangent:
Fixes #2892.


vangent:
@bensussman FYI

eliben (approved):

PR #2898, opened 2020-11-21
pubsub/azuresb: Simplify ReceiveBatch to only return 1 message; fixes error handling

### LINKED ISSUE #2897: pubsub/azuresb: retry on internal context "rctx" expiration
### Describe the bug

Unlike gcp and aws' implemenation, azure's `ReceiveBatch` receive messages one by one. 

For each single message read we use a new context with 1 second timeout.
```
rctx, cancel := context.WithTimeout(ctx, listenerTimeout)
defer cancel()
```

If any request take more then 1 second to respond, `ReceiveOne` will return a `DeadlineExceeded` error to the concrete type.

The concrete type calls `ReceiveBatch` as follow:

```
err := retry.Call(ctx, gax.Backoff{}, s.driver.IsRetryable, func() error {
	var err error
	ctx2 := s.tracer.Start(ctx, "driver.Subscription.ReceiveBatch")
	defer func() { s.tracer.End(ctx2, err) }()
	msgs, err = s.driver.ReceiveBatch(ctx2, curMaxMessagesInBatch)
	return err
})
if err != nil {
	return wrapError(s.driver, err)
}
```

And azure's pubsub `IsRetryable` always return false. 

Thus a single slow request make `.Receive` to return immediately with no retry. And this is caused by an internal context (`rctx`) not even user context.
### To Reproduce

- Use azure's pubsub
- Reduce https://github.com/google/go-cloud/blob/master/pubsub/azuresb/azuresb.go#L85:L85 enough to make it fail sometimes
- Call `.Receive` with `context.Background()`

## Expected behavior

I think the expected behavior is that Receive should block until one message is available -- Not return with `DeadlineExceeded`.

In azure's ReceiveBatch we have:
```
	rctx, cancel := context.WithTimeout(ctx, listenerTimeout)
	defer cancel()
	var messages []*driver.Message

	// Loop until rctx is Done, or until we've received maxMessages.
	for len(messages) < maxMessages && rctx.Err() == nil {
           ...
        }
```

I think we should not return if `rctx` is done, and retry here! Because the concrete type is supposed to handle `ReceiveBatch` errors + external context,  not how internally `ReceiveBatch` is done. 

### Version

0.20

### LINKED ISSUE #2899: pubsub/azuresb: sub.Receive returned on transient error
### Describe the bug

I basically have this loop:
```
for {
	m, err := s.sub.Receive(ctx)
	if err != nil {
		select {
		case <-ctx.Done():
			return
		default:
			log.Fatalf("non retryable sub.Receive(): %v", err)
		}
	}

	fmt.Println(string(m.Body))

	m.Ack()
}
```

Left it running for a while, and kept using my laptop. Few moments later I got this error on my terminal.

```
pubsub (code=Unknown):\n    gocloud.dev/pubsub.(*Subscription).getNextBatch.func1\n        /Users/xxxx/go-cloud/pubsub/pubsub.go:655\n  - read tcp 192.168.1.5:49930->168.62.54.52:5671: i/o timeout
```

Which mean the concrete type returned instead of retrying.

### To Reproduce

- Run `Receive` in a loop
- Try to trigger an io/timeout error

## Expected behavior

Concrete type should automatically handle this type of transient errors

### Version

0.20

### Additional context

Using the changes introduced here https://github.com/google/go-cloud/pull/2898

I also have some more debug statement, here is the error returned by `ReceiveOne` call

```
(*net.OpError)(0xc000098640)(read tcp 192.168.1.5:49930->168.62.54.52:5671: i/o timeout)
```

### LINKED ISSUE #2900: pubsub/azuresb: receive blocks while it received at least one message
### Describe the bug

My code uses https://github.com/google/go-cloud/pull/2898

And it's just an infinite loop doing `sub.Receive`

```
ctx := context.Background()
for {
	m, err := s.sub.Receive(ctx)
	if err != nil {
		select {
		case <-ctx.Done():
			return
		default:
			log.With(zap.Error(err)).Fatal("non retryable sub.Receive() error")
		}
	}
	fmt.Printf("Got Message: %s", string(m.Body))

	m.Ack()
}
```

1. I published 1 message `m1` -> `.Receive` returned it (as expected)
2. I published second message `m2` -> Receive still blocked forever
3. I published a third message `m3` -> Receive returned `m2`, and then `m3`

I added some debug statement to `pubsub.go`, noticed that `batchSize` gets doubled every time, and `Receive` only return when there is `batchSize` message available. 

I added more debug statement to `azuresb.go`, and tried the same thing again, I noted that at step [2], `ReceiveOne` actually received message `m2`, but that message was not returned by `sub.Receive` because it was waiting for 2 goroutines to finish before returning. 

```
g, ctx := errgroup.WithContext(s.backgroundCtx)
for _, maxMessagesInBatch := range batches {
	// Make a copy of the loop variable since it will be used by a goroutine.
	curMaxMessagesInBatch := maxMessagesInBatch
	g.Go(func() error {
		var msgs []*driver.Message
		err := retry.Call(ctx, gax.Backoff{}, s.driver.IsRetryable, func() error {
			var err error
			ctx2 := s.tracer.Start(ctx, "driver.Subscription.ReceiveBatch")
			defer func() { s.tracer.End(ctx2, err) }()
			msgs, err = s.driver.ReceiveBatch(ctx2, curMaxMessagesInBatch)
			return err
		})
		if err != nil {
			return wrapError(s.driver, err)
		}
		mu.Lock()
		defer mu.Unlock()
		q = append(q, msgs...)
		return nil
	})
}
if err := g.Wait(); err != nil {
	return nil, err
}
```

In step 2, `batches` arrays is [1, 1], so it spawned 2 goroutines in an `errgroup` and kept waiting for both of them to finish using `g.Wait()` before returning. And since there is only 1 message available, one of those goroutine is blocked. 
In step 3, when I published another message, the second goroutine received the message, and `g.Wait()` returned.

I think this problem does not only affect `azuresb`, because in `getNextBatch` we start new goroutines to get batches, and only return when all of them finish, or one of them errors out. So if one them blocks forever, `getNextBatch` don't return.

## Expected behavior

When there is a message, `.Receive` should return it

### Version

v0.20 - branch https://github.com/google/go-cloud/pull/2898

## Additional Context

Here is a self contained example:

```
func main() {
	// Messages to send.
	n, _ := strconv.Atoi(os.Args[1])
	fmt.Printf("Sending %d messages\n", n)

	ctx := context.Background()
	// Create a topic.
	topic, err := openTopicAzure()
	if err != nil {
		panic(err)
	}

	// Create a subscription connected to that topic.
	subscription, err := openSubscriptionAzure()
	if err != nil {
		panic(err)
	}
	defer subscription.Shutdown(ctx)

	go func() {
		for i := 0; i < n; i++ {
			msg := fmt.Sprintf("message %d", i)
			err := topic.Send(ctx, &pubsub.Message{
				Body: []byte(msg),
			})
			if err != nil {
				panic(err)
			}
		}
	}()

	for {
		m, err := subscription.Receive(ctx)
		if err != nil {
			panic(err)
		}
		fmt.Printf("Received: %s\n", string(m.Body))
		m.Ack()
	}

}
```

```
$ ./reproduce 1
Sending 1 messages
Received: message 0
{hangs} [as expected]
```

```
$ ./reproduce 2
Sending 2 messages
Received: message 0
{hangs} [NOT expected]
```

```
./reproduce 3
Sending 3 messages
Received: message 0
Received: message 1
Received: message 2
{hangs} [as expected]
```

```
./reproduce 6
Sending 6 messages
Received: message 0
Received: message 1
Received: message 2
{hangs} [NOT expected]
```

vangent:
Fixes #2897 
Fixes #2899
Fixes #2900

This code is simpler than before; it relies on the concrete type behavior of retrying driver.ReceiveBatch calls as needed, instead of trying to batch up some messages artificially inside ReceiveBatch.

It also fixes a bug introduced in #2891, where a timeout in the call to `ReceiveOne` would cause failures.

I benchmarked this, and it's not noticeably slower than before; the overhead from letting the concrete type retry vs looping inside the driver is minimal. This also lets the concrete type react faster/better with more concurrency (by calling ReceiveBatch concurrently).

### DIFF
--- a/pubsub/awssnssqs/awssnssqs.go
+++ b/pubsub/awssnssqs/awssnssqs.go
@@ -647,6 +647,8 @@ type SubscriptionOptions struct {
 
 	// WaitTime passed to ReceiveMessage to enable long polling.
 	// https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-short-and-long-polling.html#sqs-long-polling.
+	// Note that a non-zero WaitTime can delay delivery of messages
+	// by up to that duration.
 	WaitTime time.Duration
 }
 
--- a/pubsub/azuresb/azuresb.go
+++ b/pubsub/azuresb/azuresb.go
@@ -62,6 +62,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"time"
 
 	common "github.com/Azure/azure-amqp-common-go/v3"
 	"github.com/Azure/azure-amqp-common-go/v3/cbs"
@@ -80,6 +81,8 @@ const (
 	// https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-amqp-request-response#update-disposition-status
 	dispositionForAck  = "completed"
 	dispositionForNack = "abandoned"
+
+	listenerTimeout = 1 * time.Second
 )
 
 var sendBatcherOpts = &batcher.Options{
@@ -449,12 +452,18 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*dr
 		return nil, s.linkErr
 	}
 
+	// ReceiveOne will block until ctx is Done; we want to return after
+	// a reasonably short delay even if there are no messages. So, create a
+	// sub context for the RPC.
+	rCtx, cancel := context.WithTimeout(ctx, listenerTimeout)
+	defer cancel()
+
 	// NOTE: there's also a Receive method, but it starts two goroutines
 	// that aren't necessarily finished when Receive returns, which causes
 	// data races if Receive is called again quickly. ReceiveOne is more
 	// straightforward.
 	var message *driver.Message
-	err := s.sbSub.ReceiveOne(ctx, servicebus.HandlerFunc(func(_ context.Context, sbmsg *servicebus.Message) error {
+	err := s.sbSub.ReceiveOne(rCtx, servicebus.HandlerFunc(func(_ context.Context, sbmsg *servicebus.Message) error {
 		metadata := map[string]string{}
 		for key, value := range sbmsg.GetKeyValues() {
 			if strVal, ok := value.(string); ok {
@@ -475,6 +484,10 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*dr
 		}
 		return nil
 	}))
+	// Mask rCtx timeouts, they are expected if no messages are available.
+	if err == rCtx.Err() {
+		err = nil
+	}
 	return []*driver.Message{message}, err
 }
 
--- a/pubsub/azuresb/azuresb.go
+++ b/pubsub/azuresb/azuresb.go
@@ -62,7 +62,6 @@ import (
 	"runtime"
 	"strings"
 	"sync"
-	"time"
 
 	common "github.com/Azure/azure-amqp-common-go/v3"
 	"github.com/Azure/azure-amqp-common-go/v3/cbs"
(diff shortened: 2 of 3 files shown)


vangent:
FYI @Segflow

eliben (approved):

PR #2904, opened 2020-11-25
pubsub/all: add message ID to pubssub messages

Segflow:
Add read only `ID` property to pubsub messages. 

Only Kafka and Nats don't have this property -- I could be wrong, please confirm

This property is needed if we want to guarantee at most once delivery. It's also useful for logging/tracing and tracking messages.

### DIFF
--- a/pubsub/awssnssqs/awssnssqs.go
+++ b/pubsub/awssnssqs/awssnssqs.go
@@ -710,6 +710,7 @@ func (s *subscription) ReceiveBatch(ctx context.Context, maxMessages int) ([]*dr
 		}
 
 		m2 := &driver.Message{
+			ID:       *m.MessageId,
 			Body:     b,
 			Metadata: attrs,
 			AckID:    m.ReceiptHandle,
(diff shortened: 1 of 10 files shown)


vangent:
(background: https://godoc.org/gocloud.dev/pubsub#hdr-At_most_once_and_At_least_once_Delivery)

Because of the differing semantics, we have to be very careful to state what the semantics of `Message` fields are. For example, is it guaranteed that `ID` will be the same if a message is Nacked and then re-delivered (or never acked and redelivered)? If so, we should add drivertests to verify that.

This also doesn't seem to be fully portable given that NATS and Kafka don't have it.

> This property is needed if we want to guarantee at most once delivery.

How would you do that? Semantics vary across providers, but I'm not sure doing so client-side will work well.
An easy alternative for this would be for applications to add a `Metadata` key/value pair with a unique ID.

> It's also useful for logging/tracing and tracking messages.

I buy this though. Perhaps we could call it a DebugID, be clear that it has no semantics beyond being useful for logging, and fill it in for all providers (even Kafka and NATS have some metadata that would be useful for that).

PR #2905, opened 2020-11-26
pubsub: fix data race accessing topic error on shutdown

alexsn:
Similar to https://github.com/google/go-cloud/pull/2901 but for topic shutdown.


vangent (approved):
Thanks!

PR #2906, opened 2020-11-27
blob/azureblob: Ensure container exist on write

MetalBlueberry:
I've found that go-cloud doesn't create the container if it doesn't exist nor gives an option to create it. 

Adding this line will create it if it doesn't exist and do nothing if it already exists.

May is better to add this functionality at the driver level, but for now, this solves my issue.


MetalBlueberry:
@googlebot I fixed it.

vangent:
Sorry, this is an explicit design decision. See https://github.com/google/go-cloud/blob/master/internal/docs/design.md#developers-and-operators. We generally do not want to get in the business of provisioning, which includes creating a container.

PR #2908, opened 2020-12-02
blob/azureblob: Obtain Credentials from MSI for URL Signing

toddself:
In order to sign URLs we require credentials to be generated.  The previous pull #2873 added support for authentication via MSI in Azure, this makes it so MSI-enabled environments are able to obtain credentials correctly for URL signing.

We tried to add this to the openerFromMSI, but we ran into a chicken-and-egg issue where in order to create the Credentials we needed a service URL, which wasn't being created until the openBucket function.

Happy to make any changes necessary to get this approved!

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -836,16 +836,7 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
 // SignedURL implements driver.SignedURL.
 func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error) {
 	if b.opts.Credential == nil {
-		if opts.Credential == nil {
-			return "", gcerr.New(gcerr.Unimplemented, nil, 1, "azureblob: to use SignedURL, you must call OpenBucket with a non-nil Options.Credential")
-		} else {
-			switch t := opts.Credential.(type) {
-			case azblob.StorageAccountCredential:
-				b.opts.Credential = opts.Credential.(azblob.StorageAccountCredential)
-			default:
-				return "", gcerr.New(gcerr.Unimplemented, nil, 1, fmt.Sprintf("azureblob: to use SignedURL, you must call OpenBucket with a non-nil Options.Credential got %+v", t))
-			}
-		}
+		return "", gcerr.New(gcerr.Unimplemented, nil, 1, "azureblob: to use SignedURL, you must call OpenBucket with a non-nil Options.Credential")
 	}
 	if opts.ContentType != "" || opts.EnforceAbsentContentType {
 		return "", gcerr.New(gcerr.Unimplemented, nil, 1, "azureblob: does not enforce Content-Type on PUT")
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -862,6 +862,8 @@ func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedU
 		return "", gcerr.New(gcerr.Unimplemented, nil, 1, "azureblob: does not enforce Content-Type on PUT")
 	}
 
+	fmt.Printf("\n\n\n\ntoken: %+v\n", b.opts.Credential)
+
 	key = escapeKey(key, false)
 	blockBlobURL := b.containerURL.NewBlockBlobURL(key)
 	srcBlobParts := azblob.NewBlobURLParts(blockBlobURL.URL())
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -420,14 +420,15 @@ func NewPipeline(credential azblob.Credential, opts azblob.PipelineOptions) pipe
 // write and delete operations on objects within it.
 // See https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction.
 type bucket struct {
-	name                  string
-	pageMarkers           map[string]azblob.Marker
-	serviceURL            *azblob.ServiceURL
-	containerURL          azblob.ContainerURL
-	opts                  *Options
-	credentialExpiration  *time.Time
+	name         string
+	pageMarkers  map[string]azblob.Marker
+	serviceURL   *azblob.ServiceURL
+	containerURL azblob.ContainerURL
+	opts         *Options
+
+	mu                    sync.Mutex // protect the fields below
+	credentialExpiration  time.Time
 	delegationCredentials azblob.StorageAccountCredential
-	mtx                   sync.Mutex
 }
 
 // OpenBucket returns a *blob.Bucket backed by Azure Storage Account. See the package
@@ -485,13 +486,13 @@ func openBucket(ctx context.Context, pipeline pipeline.Pipeline, accountName Acc
 		blobURL.RawQuery = strings.TrimPrefix(string(opts.SASToken), "?")
 	}
 	serviceURL := azblob.NewServiceURL(*blobURL, pipeline)
-
 	return &bucket{
-		name:         containerName,
-		pageMarkers:  map[string]azblob.Marker{},
-		serviceURL:   &serviceURL,
-		containerURL: serviceURL.NewContainerURL(containerName),
-		opts:         opts,
+		name:                 containerName,
+		pageMarkers:          map[string]azblob.Marker{},
+		serviceURL:           &serviceURL,
+		containerURL:         serviceURL.NewContainerURL(containerName),
+		opts:                 opts,
+		credentialExpiration: time.Time{},
 	}, nil
 }
 
@@ -825,9 +826,9 @@ func (b *bucket) getDelegationCredentials(ctx context.Context) (azblob.StorageAc
 	validPeriod := 48 * time.Hour
 	currentTime := time.Now().UTC()
 	expires := currentTime.Add(validPeriod)
-	b.credentialExpiration = &expires
+	b.credentialExpiration = expires
 	keyInfo := azblob.NewKeyInfo(currentTime, expires)
-	delegationCredentials, err := b.serviceURL.GetUserDelegationCredential(ctx, keyInfo, nil, nil)
+	delegationCredentials, err := b.serviceURL.GetUserDelegationCredential(ctx, keyInfo, nil /* default timeout */, nil /* no request id */)
 
 	if err != nil {
 		return nil, fmt.Errorf("error while retrieving user delegation credential: %s", err)
@@ -838,11 +839,11 @@ func (b *bucket) getDelegationCredentials(ctx context.Context) (azblob.StorageAc
 
 func (b *bucket) refreshDelegationCredentials(ctx context.Context) (azblob.StorageAccountCredential, error) {
 	var err error
-	b.mtx.Lock()
-	defer b.mtx.Unlock()
+	b.mu.Lock()
+	defer b.mu.Unlock()
 	creds := b.delegationCredentials
 
-	if b.credentialExpiration != nil && time.Now().UTC().After(*b.credentialExpiration) {
+	if time.Now().UTC().After(b.credentialExpiration) {
 		creds, err = b.getDelegationCredentials(ctx)
 		b.delegationCredentials = creds
 	}
@@ -852,10 +853,9 @@ func (b *bucket) refreshDelegationCredentials(ctx context.Context) (azblob.Stora
 // SignedURL implements driver.SignedURL.
 func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error) {
 	var credential azblob.StorageAccountCredential
-	isMSIEnvironment := adal.MSIAvailable(ctx, adal.CreateSender())
 	if b.opts.Credential != nil {
 		credential = b.opts.Credential
-	} else if isMSIEnvironment {
+	} else if isMSIEnvironment := adal.MSIAvailable(ctx, adal.CreateSender()); isMSIEnvironment {
 		var err error
 		credential, err = b.refreshDelegationCredentials(ctx)
 		if err != nil {
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -420,14 +420,11 @@ func NewPipeline(credential azblob.Credential, opts azblob.PipelineOptions) pipe
 // write and delete operations on objects within it.
 // See https://docs.microsoft.com/en-us/azure/storage/blobs/storage-blobs-introduction.
 type bucket struct {
-	name         string
-	pageMarkers  map[string]azblob.Marker
-	serviceURL   *azblob.ServiceURL
-	containerURL azblob.ContainerURL
-	opts         *Options
-
-	// CredentialExpiration is when the current MSI-obtained signing credential
-	// expires so that we can refresh it on demand.
+	name                 string
+	pageMarkers          map[string]azblob.Marker
+	serviceURL           *azblob.ServiceURL
+	containerURL         azblob.ContainerURL
+	opts                 *Options
 	credentialExpiration time.Time
 }
 
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -487,12 +487,11 @@ func openBucket(ctx context.Context, pipeline pipeline.Pipeline, accountName Acc
 	}
 	serviceURL := azblob.NewServiceURL(*blobURL, pipeline)
 	return &bucket{
-		name:                 containerName,
-		pageMarkers:          map[string]azblob.Marker{},
-		serviceURL:           &serviceURL,
-		containerURL:         serviceURL.NewContainerURL(containerName),
-		opts:                 opts,
-		credentialExpiration: time.Time{},
+		name:         containerName,
+		pageMarkers:  map[string]azblob.Marker{},
+		serviceURL:   &serviceURL,
+		containerURL: serviceURL.NewContainerURL(containerName),
+		opts:         opts,
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


toddself:
@googlebot I fixed it.

blob/azureblob/azureblob.go
```diff
@@ -481,6 +481,23 @@ func openBucket(ctx context.Context, pipeline pipeline.Pipeline, accountName Acc
 		blobURL.RawQuery = strings.TrimPrefix(string(opts.SASToken), "?")
 	}
 	serviceURL := azblob.NewServiceURL(*blobURL, pipeline)
+
+	if opts.Credential == nil {
```
vangent, line 485: Could you get credentials earlier using https://godoc.org/github.com/Azure/azure-storage-blob-go/azblob#NewUserDelegationCredential ?

Alternatively, could you do this lazily in SignURL?

Either way, please update the docstring for Options.Credential.

vangent (changes requested):
This is causing tests to fail which is concerning.

blob/azureblob/azureblob.go
```diff
@@ -101,7 +101,8 @@ const (
 // Options sets options for constructing a *blob.Bucket backed by Azure Block Blob.
 type Options struct {
 	// Credential represents the authorizer for SignedURL.
-	// Required to use SignedURL.
+	// Required to use SignedURL. If you're using MSI for authentication, this will
+	// attempt to be loaded lazily the first time you call SignedURL
```
vangent, line 105: Nit: missing closing ".".

blob/azureblob/azureblob.go
```diff
@@ -816,10 +818,31 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
...
+		keyInfo := azblob.NewKeyInfo(currentTime, currentTime.Add(48*time.Hour))
+		delegationCredentials, err := b.serviceURL.GetUserDelegationCredential(ctx, keyInfo, nil, nil)
+
+		if err != nil {
+			return nil, fmt.Errorf("azureblob.OpenBucket: error while retrieving user delegation credential: %s", err)
+		}
+
+		return &delegationCredentials, nil
+	}
+	return nil, errors.New("Cannot generate credentials in a non MSI environment")
```
vangent: Nit: Cannot -> cannot

blob/azureblob/azureblob.go
```diff
@@ -816,10 +818,31 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
...
+func (b *bucket) getDelegationCredentials(ctx context.Context) (*azblob.UserDelegationCredential, error) {
+	isMSIEnvironment := adal.MSIAvailable(ctx, adal.CreateSender())
+
+	if isMSIEnvironment {
+		currentTime := time.Now().UTC()
+		keyInfo := azblob.NewKeyInfo(currentTime, currentTime.Add(48*time.Hour))
+		delegationCredentials, err := b.serviceURL.GetUserDelegationCredential(ctx, keyInfo, nil, nil)
+
+		if err != nil {
+			return nil, fmt.Errorf("azureblob.OpenBucket: error while retrieving user delegation credential: %s", err)
```
vangent: Nit: drop the "azureblob.OpenBucket: " prefix.

blob/azureblob/azureblob.go
```diff
@@ -816,10 +818,31 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
...
+}
+
 // SignedURL implements driver.SignedURL.
 func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error) {
+
 	if b.opts.Credential == nil {
-		return "", gcerr.New(gcerr.Unimplemented, nil, 1, "azureblob: to use SignedURL, you must call OpenBucket with a non-nil Options.Credential")
+		var err error
+		if b.opts.Credential, err = b.getDelegationCredentials(ctx); err != nil {
+			return "", gcerr.New(gcerr.Unimplemented, nil, 1, "azureblob: to use SignedURL, you must call OpenBucket with a non-nil Options.Credential")
```
vangent, line 851: How about including err in the returned error so the caller can distinguish between "cannot generate credentials in a non MSI environment" and "error while retrieving..."?

blob/azureblob/azureblob.go
```diff
@@ -101,9 +101,14 @@ const (
 // Options sets options for constructing a *blob.Bucket backed by Azure Block Blob.
 type Options struct {
 	// Credential represents the authorizer for SignedURL.
-	// Required to use SignedURL.
+	// Required to use SignedURL. If you're using MSI for authentication, this will
+	// attempt to be loaded lazily the first time you call SignedURL
 	Credential azblob.StorageAccountCredential
 
+	// CredentialExpiration is when the current MSI-obtained signing credential
+	// expires so that we can refresh it on demand
```
vangent, line 109: Same.

blob/azureblob/azureblob.go
```diff
@@ -816,11 +822,43 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
...
+func (b *bucket) getDelegationCredentials(ctx context.Context) (*azblob.UserDelegationCredential, error) {
+	validity := 48 * time.Hour
+	currentTime := time.Now().UTC()
+	expires := currentTime.Add(validity)
+	b.opts.CredentialExpiration = expires
+	keyInfo := azblob.NewKeyInfo(currentTime, expires)
+	delegationCredentials, err := b.serviceURL.GetUserDelegationCredential(ctx, keyInfo, nil, nil)
+
+	if err != nil {
+		return nil, fmt.Errorf("azureblob.OpenBucket: error while retrieving user delegation credential: %s", err)
```
vangent, line 834: Nit: drop "azureblob.OpenBucket: " prefix.

blob/azureblob/azureblob.go
```diff
@@ -816,11 +822,43 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
...
+
 // SignedURL implements driver.SignedURL.
 func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error) {
+
 	if b.opts.Credential == nil {
-		return "", gcerr.New(gcerr.Unimplemented, nil, 1, "azureblob: to use SignedURL, you must call OpenBucket with a non-nil Options.Credential")
+		var err error
+		isMSIEnvironment := adal.MSIAvailable(ctx, adal.CreateSender())
+		if isMSIEnvironment {
+			if b.opts.Credential, err = b.getDelegationCredentials(ctx); err != nil {
```
vangent, line 847: Let's store the delegation credentials (and expiration) on `bucket` instead of in `Options`.

blob/azureblob/azureblob.go
```diff
@@ -816,11 +822,43 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
...
+	if err != nil {
+		return nil, fmt.Errorf("azureblob.OpenBucket: error while retrieving user delegation credential: %s", err)
+	}
+
+	return &delegationCredentials, nil
+}
+
 // SignedURL implements driver.SignedURL.
 func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error) {
+
```
r-medina, line 842: get rid of this blank line

blob/azureblob/azureblob.go
```diff
@@ -816,11 +822,43 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
...
 	if b.opts.Credential == nil {
-		return "", gcerr.New(gcerr.Unimplemented, nil, 1, "azureblob: to use SignedURL, you must call OpenBucket with a non-nil Options.Credential")
+		var err error
+		isMSIEnvironment := adal.MSIAvailable(ctx, adal.CreateSender())
+		if isMSIEnvironment {
+			if b.opts.Credential, err = b.getDelegationCredentials(ctx); err != nil {
+				return "", err
+			}
+		} else {
```
r-medina, line 850: you can avoid this else clause by just having a `if !isMSIEnvironment` above

blob/azureblob/azureblob.go
```diff
@@ -816,11 +822,43 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
 	return page, nil
 }
 
+func (b *bucket) getDelegationCredentials(ctx context.Context) (*azblob.UserDelegationCredential, error) {
+	validity := 48 * time.Hour
```
r-medina, line 826: `validity` is not the best name here `currentTime.Add(validity)` isn't as obvious as it could be (validity implies a boolean to me)

blob/azureblob/azureblob.go
vangent: Ping? `Options` should be immutable.

toddself: oh lord i missed the first part of this!

sorry!

blob/azureblob/azureblob.go
```diff
@@ -101,9 +101,14 @@ const (
 // Options sets options for constructing a *blob.Bucket backed by Azure Block Blob.
 type Options struct {
 	// Credential represents the authorizer for SignedURL.
-	// Required to use SignedURL.
+	// Required to use SignedURL. If you're using MSI for authentication, this will
+	// attempt to be loaded lazily the first time you call SignedURL.
 	Credential azblob.StorageAccountCredential
 
+	// IsMSIEnvironment represents if the blob storage is being authenticated via
```
vangent, line 108: Is this really an Option? IIUC, you compute it automatically for the URLOpener case, why would we require a user to fill it in when using the constructor?

I.e., can't we drop this and use `adal.MSIAvailable` instead? If that call is expensive, cache it on `bucket`, not `Options`.

blob/azureblob/azureblob.go
```diff
@@ -816,14 +828,48 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
...
+
+	return &delegationCredentials, nil
+}
+
 // SignedURL implements driver.SignedURL.
 func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error) {
+	var credential azblob.StorageAccountCredential
 	if b.opts.Credential == nil {
-		return "", gcerr.New(gcerr.Unimplemented, nil, 1, "azureblob: to use SignedURL, you must call OpenBucket with a non-nil Options.Credential")
+		var err error
```
vangent, line 850: I think this could be simplified to something like:

```
var credential azblob.StorageAccountCredential
if b.opts.Credential != nil {
  credential = b.opts.Credential
} else if isMSIEnvironment {
  var err error
  credential, err = b.refreshDelegationCredentials(ctx)
  if err != nil { ...}
} else {
  // return error
}
// ... use credential
```

where `refreshDelegationCredentials` checks the expiry and refreshes if needed.

`SignedURL` can be called by multiple goroutines, so b.delegationCredentials and b.credentialExpiration need to be protected by a mutex; doing all reading/writing of those variables inside a single function makes that easy.

vangent (changes requested):

blob/azureblob/azureblob.go
toddself: So the problem I was trying to solve here is that just because MSI is available, we don't want to always use it.

MSI can be available, but you might have not used it to create your credentials.  If this is the case, refreshing the credentials fails.  And then since we don't have access to `bucket` when we use `openerFromMSI` I needed to stick the data somewhere that was available from both contexts.

I've now changed this to only set `credentialExpiration` if we are using delegation credentials, and then we only refresh the token if we have MSI available AND the delegation credentials are being used.

blob/azureblob/azureblob.go
```diff
@@ -419,11 +420,14 @@ func NewPipeline(credential azblob.Credential, opts azblob.PipelineOptions) pipe
...
-	pageMarkers  map[string]azblob.Marker
-	serviceURL   *azblob.ServiceURL
-	containerURL azblob.ContainerURL
-	opts         *Options
+	name                  string
+	pageMarkers           map[string]azblob.Marker
+	serviceURL            *azblob.ServiceURL
+	containerURL          azblob.ContainerURL
+	opts                  *Options
+	credentialExpiration  *time.Time
```
vangent, line 428: Formatting nits:

* Can you insert a blank line before the 3 new fields?
* Rename mtx -> mu and add an inline comment "// mu protect the fields below."

(This is consistent with how mutexes are used in this repo, and make it clear
what the mutex is used for).

Thanks!

blob/azureblob/azureblob.go
```diff
@@ -481,6 +485,7 @@ func openBucket(ctx context.Context, pipeline pipeline.Pipeline, accountName Acc
 		blobURL.RawQuery = strings.TrimPrefix(string(opts.SASToken), "?")
 	}
 	serviceURL := azblob.NewServiceURL(*blobURL, pipeline)
+
```
vangent, line 488: Nit: remove any random blank lines you inserted.

blob/azureblob/azureblob.go
```diff
@@ -816,14 +821,54 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
...
 }
 
+func (b *bucket) getDelegationCredentials(ctx context.Context) (azblob.StorageAccountCredential, error) {
+	validPeriod := 48 * time.Hour
+	currentTime := time.Now().UTC()
+	expires := currentTime.Add(validPeriod)
+	b.credentialExpiration = &expires
+	keyInfo := azblob.NewKeyInfo(currentTime, expires)
+	delegationCredentials, err := b.serviceURL.GetUserDelegationCredential(ctx, keyInfo, nil, nil)
+
```
vangent, line 831: Consider adding comments about what these `nil` parameters are? E.g., `keyInfo, nil /* no options */, nil /* whatever */) ?

blob/azureblob/azureblob.go
vangent, line 428: I think it's more standard to use `time.Time` and check for `ZeroTime` instead of `nil`. In this case, since "now" will be after the zero time, you don't even need that special check (on line 845).

blob/azureblob/azureblob.go
```diff
@@ -816,14 +821,54 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
...
+		b.delegationCredentials = creds
+	}
+	return creds, err
+}
+
 // SignedURL implements driver.SignedURL.
 func (b *bucket) SignedURL(ctx context.Context, key string, opts *driver.SignedURLOptions) (string, error) {
-	if b.opts.Credential == nil {
+	var credential azblob.StorageAccountCredential
+	isMSIEnvironment := adal.MSIAvailable(ctx, adal.CreateSender())
```
vangent, line 855: Nit: might as well do this inline on line 858 to avoid doing it if b.opts.Credential is provided.

vangent (commented):
Thanks for the contribution and for your patience, getting close!

vangent (commented):
Is there any way to add a test for this?

toddself:
@vangent i've been trying to figure that out.  if i mock the underlying stuff i think i should be able to?

toddself:
@vangent: My colleague @r-medina and I looked into writing tests for this, but we noticed there are open questions around soup-to-nuts Azure testing. For example nothing tests writing blobs, or the basic scenarios for connecting via MSI or even signing a non-MSI-based URL. The MSI environment apparently requires mocking out underlying Azure Active Directory REST interfaces.

Given the lack of test coverage for the azure blob interface, would it be better to create a separate issue and start chipping away at the test debt there meanwhile merging this code in?

(We have verified that this works in an Azure Manged Service Interface environment.)

vangent:
> nothing tests writing blobs

Actually, there are pretty comprehensive tests for all of the `blob` drivers through the `drivertests` package, including reading, writing, listing, etc.

The different connection/authentication/authorization mechanisms are not well tested (the tests basically use only one way).
If you have tested this locally and it's working for you I'm fine with that.

blob/azureblob/azureblob.go
```diff
@@ -482,11 +487,12 @@ func openBucket(ctx context.Context, pipeline pipeline.Pipeline, accountName Acc
...
-		pageMarkers:  map[string]azblob.Marker{},
-		serviceURL:   &serviceURL,
-		containerURL: serviceURL.NewContainerURL(containerName),
-		opts:         opts,
+		name:                 containerName,
+		pageMarkers:          map[string]azblob.Marker{},
+		serviceURL:           &serviceURL,
+		containerURL:         serviceURL.NewContainerURL(containerName),
+		opts:                 opts,
+		credentialExpiration: time.Time{},
```
vangent, line 495: Nit: this shouldn't be necessary, time.Time{} is the default value for time.Time.

blob/azureblob/azureblob.go
```diff
@@ -816,14 +822,53 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
 	return page, nil
 }
 
+func (b *bucket) getDelegationCredentials(ctx context.Context) (azblob.StorageAccountCredential, error) {
+	validPeriod := 48 * time.Hour
```
vangent, line 826: Sorry, I thought I left this comment last time but it looks like it got lost. Can you delete this function and do it inline in refreshDelegationCredentials? It makes the locking more clear (e.g., this function sets `b.credentialExpiration`, which is a mutation that's a bit unexpected from a "get" function, and also requires that whoever calls this has the mutex).

blob/azureblob/azureblob.go
```diff
@@ -816,14 +822,53 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
...
+	currentTime := time.Now().UTC()
+	expires := currentTime.Add(validPeriod)
+	b.credentialExpiration = expires
```
vangent, line 829: `b.credentialExpiration` should only be updated if the new credentials are successfully updated (i.e., update it at the same time as `b.credentials`.

vangent (approved):

vangent:
@r-medina if you sign the CLA I can merge this.

r-medina:
@vangent i signed something on https://cla.developers.google.com/clas - is there another step i need to do?

vangent:
I'm not sure. See https://github.com/google/go-cloud/pull/2908#issuecomment-742095656 -- maybe try a "@googlebot I fixed it" comment?

vangent:
@googlebot I fixed it

vangent:
@toddself it looks like you need to add a comment "@googlebot I consent", see above.

toddself:
@googlebot I consent

r-medina:
@googlebot I consent

Issue #2907, opened 2020-12-01
runtimevar: add drivers for AWS Secrets Manager and GCP Secret Manager


THE108:
AWS Secrets Manager ([https://aws.amazon.com/secrets-manager](https://aws.amazon.com/secrets-manager)) and GCP Secret Manager ([https://cloud.google.com/secret-manager](https://cloud.google.com/secret-manager)) seem to be better alternatives for storing secrets than using config services like AWS SSM Parameter Store / GCP Cloud Runtime Configurator + KMS.

### Describe the solution you'd like

Available runtimevar drivers for AWS Secrets Manager and GCP Secret Manager.

### Describe alternatives you've considered

AWS SSM Parameter Store / GCP Cloud Runtime Configurator + KMS.

THE108:
P.S. I can help with a PR.

vangent:
PRs for this are welcome. Hopefully a new driver is fairly straightforward by copying one of the existing ones. We will likely have to work together because we'll need to record test files with our test projects (GCP and AWS). Let me know what support you need.

THE108:
@vangent 

I've submitted a [PR for AWS Secrets Manager](https://github.com/google/go-cloud/pull/2909). Tests obviously fail due to missing `.replay` files so I'll need some help to generate them.

vangent:
Thanks for the contribution! Are you planning on sending a PR for GCP Secret Manager as well?

THE108:
> Thanks for the contribution! Are you planning on sending a PR for GCP Secret Manager as well?

Yes, I'll submit a separate PR probably next week.

THE108:
@vangent 

Could you point me out to the documentation repo so I can a relevant section for AWS Secrets Manager?

vangent:
@THE108 it's the same repo, for example you probably want to edit:
https://github.com/google/go-cloud/tree/master/internal/website/content/howto/runtimevar

THE108:
@vangent 

I created a [PR](https://github.com/google/go-cloud/pull/2938) for GCP Secret Manager.

PR #2909, opened 2020-12-03
runtimevar: add driver for AWS Secrets Manager

THE108:
Add driver for AWS Secrets Manager.

Updates #2907.

A couple of things to consider:

1. AWS Secrets Manager API is quite similar to AWS SSM Parameter Store, so it looks a bit repetitive. Please let me know if you think it's worth it to refactor the common parts out into a common AWS specific package for `runtimevar`.

2. AWS Secrets Manager does actual removing of secrets in an async way. Thus, it's possible that the creation of a secret with the same name right after deletion will fail due to the SecretAlreadyExists error until it will be actually removed. AWS docs recommend implementing retry logic for this use case which makes `go test -record` a bit slower.


runtimevar/awssecretsmanager/awssecretsmanager.go
```diff
@@ -0,0 +1,359 @@
+// Copyright 2018 The Go Cloud Development Kit Authors
```
vangent, line 1: Nit: 2020

runtimevar/awssecretsmanager/awssecretsmanager_test.go
```diff
@@ -0,0 +1,236 @@
+// Copyright 2018 The Go Cloud Development Kit Authors
```
vangent, line 1: Same.

runtimevar/awssecretsmanager/awssecretsmanager_test.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func (h *harness) DeleteVariable(ctx context.Context, name string) error {
+	svc := secretsmanager.New(h.session)
+
+	_, err := svc.DeleteSecretWithContext(ctx, &secretsmanager.DeleteSecretInput{
+		SecretId:                   aws.String(name),
+		ForceDeleteWithoutRecovery: aws.Bool(true),
+	})
+
```
vangent, line 119: Is there a reasonable RPC to put here so that we could poll until the Delete is complete, instead of retrying in Create? It seems like that would make each tests more hermetic.

runtimevar/awssecretsmanager/example_test.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2018 The Go Cloud Development Kit Authors
```
vangent, line 1: Same.

runtimevar/awssecretsmanager/example_test.go
```diff
@@ -0,0 +1,57 @@
...
+	// Construct a *runtimevar.Variable that watches the variable.
+	v, err := awssecretsmanager.OpenVariable(sess, "cfg-variable-name", runtimevar.StringDecoder, nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer v.Close()
+}
+
+func Example_openVariableFromURL() {
+	// PRAGMA: This example is used on gocloud.dev; PRAGMA comments adjust how it is shown and can be ignored.
```
vangent, line 45: You can delete these PRAGMAs, as this example is not in fact used on gocloud.dev. I made add it in a subsequent PR.

runtimevar/awssecretsmanager/awssecretsmanager.go
```diff
@@ -0,0 +1,359 @@
...
+	// WaitDuration controls the rate at which AWS Secrets Manager is polled.
+	// Defaults to 30 seconds.
+	WaitDuration time.Duration
+}
+
+// OpenVariable constructs a *runtimevar.Variable backed by the variable secretID in AWS Secrets Manager.
+// Secrets Manager returns raw bytes; provide a decoder to decode the raw bytes
+// into the appropriate type for runtimevar.Snapshot.Value.
+// See the runtimevar package documentation for examples of decoders.
+func OpenVariable(sess client.ConfigProvider, name string, decoder *runtimevar.Decoder, opts *Options) (*runtimevar.Variable, error) {
```
vangent, line 148: name -> secretID

runtimevar/awssecretsmanager/awssecretsmanager.go
```diff
@@ -0,0 +1,359 @@
...
+func init() {
+	runtimevar.DefaultURLMux().RegisterVariable(Scheme, new(lazySessionOpener))
+}
+
+// Set holds Wire providers for this package.
+var Set = wire.NewSet(
+	wire.Struct(new(URLOpener), "ConfigProvider"),
+)
+
+// URLOpener opens AWS Secrets Manager URLs like "awssecretsmanager://myvar".
```
vangent, line 64: Should "myvar" be "secretID" here?

runtimevar/awssecretsmanager/example_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// Establish an AWS session.
+	// See https://docs.aws.amazon.com/sdk-for-go/api/aws/session/ for more info.
+	sess, err := session.NewSession(nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Construct a *runtimevar.Variable that watches the variable.
+	v, err := awssecretsmanager.OpenVariable(sess, "cfg-variable-name", runtimevar.StringDecoder, nil)
```
vangent, line 37: cfg-variable-name should probably be "my-secret-id" or something here.

runtimevar/awssecretsmanager/example_test.go
```diff
@@ -0,0 +1,57 @@
...
+
+func Example_openVariableFromURL() {
+	// PRAGMA: This example is used on gocloud.dev; PRAGMA comments adjust how it is shown and can be ignored.
+	// PRAGMA: On gocloud.dev, add a blank import: _ "gocloud.dev/runtimevar/awssecretsmanager"
+	// PRAGMA: On gocloud.dev, hide lines until the next blank line.
+	ctx := context.Background()
+
+	// runtimevar.OpenVariable creates a *runtimevar.Variable from a URL.
+
+	v, err := runtimevar.OpenVariable(ctx, "awssecretsmanager://myvar?region=us-east-2&decoder=string")
```
vangent, line 52: Similar here.

vangent (changes requested):
Overall looks great.

Do the tests pass with -record for you? If so, the easiest thing to do is probably for you to comment out the conformance tests for now so that the tests pass; once the review is complete then I'll approve and create another PR uncommenting the tests and adding the .replay files.

Are there any provisioning steps required to begin with, or will the tests just run out of the box with an existing AWS project?

vangent:
> Please let me know if you think it's worth it to refactor the common parts out into a common AWS specific package for runtimevar

No, this is fine.

> AWS Secrets Manager does actual removing of secrets in an async way

Ack. I left a comment suggesting we do polling to make sure the Delete succeeded instead of retry during Create.

vangent (commented):
You will need to create a directory + file under here:
https://github.com/google/go-cloud/tree/master/internal/website/content/runtimevar

to address this test failure:

Ensuring that any new packages have the corresponding entries in Hugo...
FAIL: missing package meta tags for:
runtimevar/awssecretsmanager

runtimevar/awssecretsmanager/awssecretsmanager.go
THE108: AWS Secrets Manager doc states:

> 	// Specifies the secret containing the version that you want to retrieve. You
	// can specify either the Amazon Resource Name (ARN) or the friendly name of
	// the secret.

Thus, I think it would be better to keep it as `name` for better clarity and consistency with other drivers.

runtimevar/awssecretsmanager/awssecretsmanager.go
THE108: See my comment above.

runtimevar/awssecretsmanager/example_test.go
THE108: See my comment above.

runtimevar/awssecretsmanager/example_test.go
THE108: See my comment above.

runtimevar/awssecretsmanager/awssecretsmanager_test.go
THE108: Yes, the [DescribeSecret](https://aws.amazon.com/premiumsupport/knowledge-center/delete-secrets-manager-secret/) API call could be used to verify that the secret is permanently deleted. Let me do the change.

THE108: I tried that approach out and it seems I can't make tests reproducible - some tests are failing randomly.

THE108:
@vangent 

> Do the tests pass with -record for you?

Yes, but only after making AWS Secrets Manager's testing harness immutable and secret variable names unique per test case (hence, `.replay` files need to be regenerated for `runtimevar` package).

Tests were failing randomly with previous approaches. I suspect it could happen because of some AWS Secrets Manager's APIs behavior.

> If so, the easiest thing to do is probably for you to comment out the conformance tests for now so that the tests pass; once the review is complete then I'll approve and create another PR uncommenting the tests and adding the .replay files.

Done.

> Are there any provisioning steps required to begin with, or will the tests just run out of the box with an existing AWS project?

I did not do anything to set up my AWS account before running the tests so it should work as-is.

Some useful commands for AWS CLI to manage secrets:
```
# List secrets
aws secretsmanager list-secrets

# Describe secret
aws secretsmanager describe-secret --secret-id <name>

# Delete a secret
aws secretsmanager delete-secret --secret-id <name> --force-delete-without-recovery

# Delete all secrets
aws secretsmanager list-secrets | jq ".SecretList[].Name" | xargs -I {} aws secretsmanager delete-secret --secret-id {} --force-delete-without-recovery 2>&1 > /dev/null
```

runtimevar/awssecretsmanager/awssecretsmanager.go
```diff
@@ -0,0 +1,359 @@
...
```
vangent, line 64: Can you add a reference to the "AWS Secrets Manager doc" that describes what "myvar" is here, and also in the docstring for OpenVariable? ("name" is not very descriptive).

runtimevar/awssecretsmanager/awssecretsmanager_test.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func (h *harness) Close() {
+	h.closer()
+}
+
+func (h *harness) Mutable() bool { return false }
+
+func TestConformance(t *testing.T) {
+	drivertest.RunConformanceTests(t, newHarness, []drivertest.AsTest{verifyAs{}})
```
vangent, line 135: Comment out here instead.

runtimevar/awssecretsmanager/example_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+func ExampleOpenVariable() {
+	// Establish an AWS session.
+	// See https://docs.aws.amazon.com/sdk-for-go/api/aws/session/ for more info.
+	sess, err := session.NewSession(nil)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Construct a *runtimevar.Variable that watches the variable.
```
vangent, line 34: Maybe add a comment here as well about what cfg-variable-name could be, and/or make the example more realistic.

runtimevar/drivertest/drivertest.go
```diff
@@ -95,38 +96,38 @@ func (verifyAsFailsOnNil) ErrorCheck(v *runtimevar.Variable, err error) (ret err
...
-		for _, st := range asTests {
-			if st.Name() == "" {
-				t.Fatalf("AsTest.Name is required")
-			}
-			t.Run(st.Name(), func(t *testing.T) {
-				testAs(t, newHarness, st)
-			})
-		}
-	})
+	//t.Run("TestNonExistentVariable", func(t *testing.T) {
```
vangent, line 99: Please don't comment these out, as it disables tests for all drivers not just this new one. See other comment above for where to comment it out for just the new one.

runtimevar/drivertest/drivertest.go
```diff
@@ -176,7 +177,7 @@ func testNonExistentVariable(t *testing.T, newHarness HarnessMaker) {
 
 func testString(t *testing.T, newHarness HarnessMaker) {
 	const (
-		name    = "test-config-variable"
+		name    = "test-config-variable-string"
```
vangent, line 180: Please revert this for now.

runtimevar/awssecretsmanager/awssecretsmanager_test.go
vangent: This is a bit concerning. Can you dig in a bit more to see why they are not reproducible? The drivertests are enforcing semantic requirements for the driver, and having to make it immutable and using different variable names across tests seem to indicate that something is fishy.

vangent: I have grabbed the PR and am trying to get the -record tests to pass. It seems like AWS is quite flaky, for example I am able to reproducibly do something like:

* create secret x
* delete secret x
* wait for DescribeSecret to fail
* create secret x
* get secret x
* describe secret x
* get secret x again --> fails!

I'm guessing that these mutating operations are only eventually consistent on the server side, which makes it really hard to write reproducible tests.

vangent: https://docs.aws.amazon.com/secretsmanager/latest/userguide/troubleshoot_general.html#troubleshoot_general_eventual-consistency

vangent: OK, I've been able to get the tests to pass now, by modifying each of Create/Update/Delete in the test harness to apply the change, sleep for some time, and then verify the change has been applied (and if not sleep again). I will send a PR to make those changes and add the .replay files after merging this.

vangent (approved):

vangent:
Thanks for the contribution!

PR #2912, opened 2020-12-03
runtimevar/drivertest: add variable cleanup in testUpdateWithErrors

THE108:
Add variable cleanup in testUpdateWithErrors.

TestUpdateWithErrors fails since the new delete variable request is missing in TestUpdateWithErrors.replay file.

Thus, added a TODO item: Uncomment after TestUpdateWithErrors.replay will be regenerated.


vangent:
Sorry for the confusion, this was fixed in #2911.

Issue #2914, opened 2020-12-03
runtimevar/etcdvar: etcd has all sorts of go.mod problems


vangent:
https://github.com/etcd-io/etcd/issues/12124
https://github.com/etcd-io/etcd/issues/11931
https://github.com/etcd-io/etcd/issues/11749

This means we can't update etcd or a bunch of other unrelated modules (like cloud.google.com).

Plan is to drop support for etcd going forward until these are fixed. Users can continue using the current tagged version of go-cloud.

saheljalal:
All the indicated issues appear closed now. Could we evaluate if support will be added back here?

PR #2915, opened 2020-12-03
runtimevar/etcdvar: drop support for runtimevar/etcd going forward until they fix their go module problems

vangent:
Updates #2914


vangent:
FYI my plan after this is to update all dependencies again (after dropping etcd, everything upgrades cleanly!), then do a release. Possibly wait for the 2 other pending PRs from contributors.

eliben:
Is there a chance etcd are just going to fix their module snafu soon? (it's not the fist time...)

vangent:
> Is there a chance etcd are just going to fix their module snafu soon? (it's not the fist time...)

Not really. It's been this way since July, and there is no plan or ETA yet.

eliben:
OK, and OOC - do we have to upgrade etcd at all?

eliben (approved):
LGTM

The questions in the comments are just informational

vangent:
> do we have to upgrade etcd at all

No, and that's what I've done in the past -- just refrained from updating it if the newer versions have problems. However, in this case, etcd's problems prevent us from upgrading other modules (gorpc, and then everything that depends on that, which is all of the GCP modules). That's not OK.

PR #2919, opened 2020-12-04
runtimevar/awssecretsmanager: make drivertests repeatable and add .replay files

vangent:
Updates #2907.

@THE108 please review as well.

### DIFF
--- a/runtimevar/awssecretsmanager/awssecretsmanager_test.go
+++ b/runtimevar/awssecretsmanager/awssecretsmanager_test.go
@@ -15,6 +15,7 @@
 package awssecretsmanager
 
 import (
+	"bytes"
 	"context"
 	"crypto/sha1"
 	"errors"
@@ -28,8 +29,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/secretsmanager"
-	"github.com/googleapis/gax-go"
-	"gocloud.dev/internal/retry"
 	"gocloud.dev/internal/testing/setup"
 	"gocloud.dev/runtimevar"
 	"gocloud.dev/runtimevar/driver"
@@ -47,7 +46,10 @@ type harness struct {
 	closer  func()
 }
 
-const maxClientRequestTokenLen = 64
+const (
+	consistencyDelay         = 30 * time.Second
+	maxClientRequestTokenLen = 64
+)
 
 // AWS Secrets Manager requires unique token for Create and Update requests to ensure idempotency.
 // From the other side, request data must be deterministic in order to make tests reproducible.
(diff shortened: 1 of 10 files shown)


runtimevar/awssecretsmanager/awssecretsmanager_test.go
```diff
@@ -81,60 +83,92 @@ func (h *harness) MakeWatcher(_ context.Context, name string, decoder *runtimeva
...
+		ClientRequestToken: aws.String(generateClientRequestToken(name, val)),
+		SecretBinary:       val,
+	})
+	if err != nil {
+		return err
+	}
+	// Secret Manager is only eventually consistent, so we insert a delay
+	// and verification that the mutation was applied. This is still not a guarantee
+	// but in practice seems to work well enough to make tests repeatable.
+	for {
```
THE108, line 97: I think we could use `retry.Call` method from `gocloud.dev/internal/retry` package instead of retrying indefinitely.

runtimevar/awssecretsmanager/awssecretsmanager_test.go
```diff
@@ -81,60 +83,92 @@ func (h *harness) MakeWatcher(_ context.Context, name string, decoder *runtimeva
...
+		if *setup.Record {
+			time.Sleep(consistencyDelay)
+		}
+		_, err := svc.GetSecretValueWithContext(ctx, &secretsmanager.GetSecretValueInput{
+			SecretId: aws.String(name),
+		})
+		if err == nil {
```
THE108, line 104: I think we should retry only on the particular error types (`resource not found` in this case) and return connection or authentication-related errors immediately.

runtimevar/awssecretsmanager/awssecretsmanager_test.go
```diff
@@ -81,60 +83,92 @@ func (h *harness) MakeWatcher(_ context.Context, name string, decoder *runtimeva
...
+	// and verification that the mutation was applied. This is still not a guarantee
+	// but in practice seems to work well enough to make tests repeatable.
+	for {
+		if *setup.Record {
+			time.Sleep(consistencyDelay)
+		}
+		getResp, err := svc.GetSecretValueWithContext(ctx, &secretsmanager.GetSecretValueInput{
+			SecretId: aws.String(name),
+		})
+		if err == nil && bytes.Equal(getResp.SecretBinary, val) {
```
THE108, line 131: Same here: I think we should retry only on the `resource not found` error and non-equal secret value.

runtimevar/awssecretsmanager/awssecretsmanager_test.go
```diff
@@ -81,60 +83,92 @@ func (h *harness) MakeWatcher(_ context.Context, name string, decoder *runtimeva
...
+	// and verification that the mutation was applied. This is still not a guarantee
+	// but in practice seems to work well enough to make tests repeatable.
+	for {
+		if *setup.Record {
+			time.Sleep(consistencyDelay)
+		}
+		_, err := svc.DescribeSecretWithContext(ctx, &secretsmanager.DescribeSecretInput{
+			SecretId: aws.String(name),
+		})
+		if err != nil {
```
THE108, line 157: Same here: I think we should retry only on the `resource not found` error.

THE108 (changes requested):

runtimevar/awssecretsmanager/awssecretsmanager_test.go
vangent: Done.

runtimevar/awssecretsmanager/awssecretsmanager_test.go
vangent: Done.

runtimevar/awssecretsmanager/awssecretsmanager_test.go
vangent: Done.

runtimevar/awssecretsmanager/awssecretsmanager_test.go
vangent: Done.

THE108 (approved):

runtimevar/awssecretsmanager/awssecretsmanager_test.go
```diff
@@ -47,12 +50,37 @@ type harness struct {
 	closer  func()
 }
 
-const maxClientRequestTokenLen = 64
+// waitForMutation uses check to wait until a mutation has taken effect.
```
eliben, line 53: I'd document briefly here how we know from check that it needs to be retried.

eliben (approved):

runtimevar/awssecretsmanager/awssecretsmanager_test.go
vangent: Good point, done.

PR #2924, opened 2020-12-11
mysql: parse url host in the form of 'protocol(addr)' (#2923)

### LINKED ISSUE #2923: sql/mysql protocol in connectionstring parsed wrong 0.21.0 regression
### Describe the bug
Hey all,
we are using a connection string like `mysql://user:password@tcp(dbtest-01)/database` to connect to a mysql instance. 
Up until 0.21.0 this worked fine but now it fails with `no such host`.
### To Reproduce
```golang
db, err := mysql.Open(ctx, "mysql://user:password@tcp(dbtest-01)/database")
if err != nil {
	log.Fatalf("open db failed: %s", err)
}
rows, err := db.Query("show variables like 'max_allowed_packet'")
if err != nil {
	log.Fatal(err)
}
```
Since 0.21.0 the query fails with:
```
dial tcp: lookup tcp(dbtest-01): no such host
```
Using `mysql://user:password@dbtest-01/database` without protocol works fine.
At version 0.20.0 using the connection string without the protocol returned the error:
```
default addr for network 'dbtest-01' unknown
```

## Expected behavior
Working sql.DB* instance.

### Version
gocloud.dev v0.21.0
$ go version
go version go1.15.6 linux/amd64

### Additional context
I guess the regression was caused by #2855 .
It interacts with the way the go-sql-driver formats the connection string. https://github.com/go-sql-driver/mysql#dsn-data-source-name.

I am not sure if we use the connection string in a special way or if this may impact other users.

alexsn:
Allow url host to take the form of `protocol(addr)` (similar to [mysql.ParseDSN](https://github.com/go-sql-driver/mysql/blob/master/dsn.go#L323-L338)).

Fixes #2923.

### DIFF
--- a/mysql/mysql.go
+++ b/mysql/mysql.go
@@ -21,6 +21,7 @@ import (
 	"database/sql/driver"
 	"fmt"
 	"net/url"
+	"regexp"
 	"strings"
 
 	"github.com/go-sql-driver/mysql"
@@ -55,6 +56,8 @@ func (uo *URLOpener) OpenMySQLURL(_ context.Context, u *url.URL) (*sql.DB, error
 	}), nil
 }
 
+var netAddrRE = regexp.MustCompile(`^(.+)\((.+)\)$`)
+
 // ConfigFromURL creates a mysql.Config from URL.
 func ConfigFromURL(u *url.URL) (cfg *mysql.Config, err error) {
 	dbName := strings.TrimPrefix(u.Path, "/")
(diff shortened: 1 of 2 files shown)


mysql/mysql.go
```diff
@@ -66,8 +69,13 @@ func ConfigFromURL(u *url.URL) (cfg *mysql.Config, err error) {
 	} else {
 		cfg = mysql.NewConfig()
 	}
-	cfg.Net = "tcp"
-	cfg.Addr = u.Host
+	if matches := netAddrRE.FindStringSubmatch(u.Host); len(matches) == 3 {
```
vangent, line 72: Can we include this stuff from u.Host in the string we pass to muysql.ParseDSN above instead of doing our own parsing?

mysql/mysql.go
```diff
@@ -66,8 +69,13 @@ func ConfigFromURL(u *url.URL) (cfg *mysql.Config, err error) {
...
+		cfg.Net = matches[1]
+		cfg.Addr = matches[2]
+	} else {
+		cfg.Net = "tcp"
+		cfg.Addr = u.Host
+	}
```
vangent, line 78: Unrelated to this PR, but maybe you can fix it anyway: There's a reference to "pq" down below in a docstring, which is PostGRES not MySQL, isn't that wrong?

mysql/mysql.go
alexsn: `mysql.ParseDSN` expects the host to be in the form of `protocol(address)` so simply passing `mysql://localhost/db` won't work so we'll need to wrap the host with `'('  + u.Host  +')'`. In any case we'll need to add our own logic.

mysql/mysql.go
alexsn: sure

vangent (approved):

PR #2926, opened 2020-12-11
docstore: fixes to improve handling of byte arrays and non-pointer protocol buffers

### LINKED ISSUE #2791: docstore/mongo protobuf timestamp panics on get
### Describe the bug

A struct generated from a .proto schema with a timestamp is written to mongodb but panics when reading it.

### To Reproduce
<pre>
$ go version
go version go1.14.3 linux/amd64
</pre>

Using a protobuf specification like
```
import "google/protobuf/timestamp.proto";

message Test{
  ...
  google.protobuf.Timestamp start_date = 4;
}
```
Will result in a struct like 
```golang
type Test struct{
  ...
  StartTime *timestamp.Timestamp `protobuf:"bytes,5,opt,name=start_time,json=startTime,proto3" json:"start_time,omitempty"`
}
```

Writing this to a mongodb instance will work and checking via mongo shell shows binary data:
```
{
	"_id" : "3d964883c5e60d36",
	"DocstoreRevision" : "2f1954b9-f038-440c-b0e0-c0ed88f2d04a",
	...
	"StartDate" : BinData(0,"CIDCr/AF"),
}
```
Which causes a panic when read via the go cloud cdk:
```
panic: reflect.Set: value of type []uint8 is not assignable to type *timestamppb.Timestamp

goroutine 28 [running]:
reflect.Value.assignTo(0xca0040, 0xc0000a8900, 0x97, 0xe166df, 0xb, 0xd81da0, 0x0, 0xc0000a8900, 0x0, 0x0)
	/snap/go/5759/src/reflect/value.go:2403 +0x426
reflect.Value.Set(0xd81da0, 0xc0000ae670, 0x196, 0xca0040, 0xc0000a8900, 0x97)
	/snap/go/5759/src/reflect/value.go:1532 +0xbd
gocloud.dev/docstore/driver.decode(0xd81da0, 0xc0000ae670, 0x196, 0x1013c00, 0xc0000a88e0, 0x2, 0xd81da0)
	/home/carsten/go/pkg/mod/gocloud.dev@v0.19.0/docstore/driver/codec.go:391 +0x193
gocloud.dev/docstore/driver.decodeStruct.func1(0xc0000aa970, 0x9, 0x1013c00, 0xc0000a88e0, 0x40dd01, 0xc000144d40)
	/home/carsten/go/pkg/mod/gocloud.dev@v0.19.0/docstore/driver/codec.go:679 +0x15b
gocloud.dev/docstore/mongodocstore.decoder.DecodeMap(0xcfdd00, 0xc000462630, 0x0, 0xc000144d40)
	/home/carsten/go/pkg/mod/gocloud.dev/docstore/mongodocstore@v0.19.0/codec.go:242 +0x12d
gocloud.dev/docstore/driver.decodeStruct(0xd6aae0, 0xc0000ae660, 0x199, 0x1013c00, 0xc0000a88c0, 0xc0005b1d18, 0xc0000a88c0)
	/home/carsten/go/pkg/mod/gocloud.dev@v0.19.0/docstore/driver/codec.go:658 +0x171
gocloud.dev/docstore/driver.Document.Decode(0xdbd640, 0xc0000ae660, 0x0, 0xd6aae0, 0xc0000ae660, 0x199, 0xc000121500, 0xa, 0x10, 0x1013c00, ...)
	/home/carsten/go/pkg/mod/gocloud.dev@v0.19.0/docstore/driver/document.go:184 +0x142
gocloud.dev/docstore/mongodocstore.decodeDoc(0xc000462630, 0xdbd640, 0xc0000ae660, 0x0, 0xd6aae0, 0xc0000ae660, 0x199, 0xc000121500, 0xa, 0x10, ...)
	/home/carsten/go/pkg/mod/gocloud.dev/docstore/mongodocstore@v0.19.0/codec.go:125 +0x115
gocloud.dev/docstore/mongodocstore.(*docIterator).Next(0xc0004623f0, 0x1005ca0, 0xc000481080, 0xdbd640, 0xc0000ae660, 0x0, 0xd6aae0, 0xc0000ae660, 0x199, 0xc000121500, ...)
	/home/carsten/go/pkg/mod/gocloud.dev/docstore/mongodocstore@v0.19.0/query.go:127 +0xe6
gocloud.dev/docstore.(*DocumentIterator).Next(0xc000462420, 0x1005ca0, 0xc000481080, 0xdbd640, 0xc0000ae660, 0xc0003426b0, 0x1)
	/home/carsten/go/pkg/mod/gocloud.dev@v0.19.0/docstore/query.go:234 +0x204
main.(*ServerJob).GetJobs(0xc00048a400, 0x1005ca0, 0xc000481080, 0xc0002b6508, 0x0, 0x0, 0x0)
...
```
## Expected behavior

Using the in memory docstore i can see the expected behavior of the struct being filled with data.
```
input: {StartDate:seconds:1589872418 nanos:319671987 ...
output: {StartDate:seconds:1589872418 nanos:319671987 ... 
```
### Version
	gocloud.dev v0.19.0
	gocloud.dev/docstore/mongodocstore v0.19.0

### LINKED ISSUE #2920: docstore/mongodb: []uint8 is not assignable to type uuid.UUID
### Describe the bug

Using a uuid.UUID field works well with the in memory docstore but not with a mongodb instance.

### To Reproduce

Startup a local mongodb instance. E.g.:
```
docker run -it --rm --net=host mongo 
```
And run the following short test programm
```go
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/google/uuid"
	"gocloud.dev/docstore"
	_ "gocloud.dev/docstore/memdocstore"
	_ "gocloud.dev/docstore/mongodocstore"
)

type Item struct {
	Id uuid.UUID

	DocstoreRevision interface{}
}

func main() {
	memCollection, err := docstore.OpenCollection(context.Background(), "mem://Items/Id")
	if err != nil {
		panic(err)
	}
	createAndQuery(memCollection)

	os.Setenv("MONGO_SERVER_URL", "mongodb://localhost/db")
	mongoCollection, err := docstore.OpenCollection(context.Background(), "mongo://db/Items?id_field=Id")
	if err != nil {
		panic(err)
	}
	createAndQuery(mongoCollection)
}

func createAndQuery(collection *docstore.Collection) {
	ctx := context.Background()
	item := &Item{Id: uuid.New()}
	result := &Item{}

	err := collection.Create(ctx, item)
	if err != nil {
		panic(err)
	}

	err = collection.Query().Get(ctx).Next(ctx, result)
	if err != nil {
		panic(err)
	}
	fmt.Println(result.Id)
}
```
On my machine the programm panics when reading from the mongodb instance:
```
$ go run .
405eb3eb-a0c6-4233-8554-736e479ef2c5
panic: reflect.Set: value of type []uint8 is not assignable to type uuid.UUID

goroutine 1 [running]:
reflect.Value.assignTo(0xb0f3e0, 0xc00000e440, 0x97, 0xc1f60b, 0xb, 0xbda5a0, 0x0, 0xc00000e440, 0x0, 0x0)
        /snap/go/current/src/reflect/value.go:2435 +0x405
reflect.Value.Set(0xbda5a0, 0xc0003a8200, 0x191, 0xb0f3e0, 0xc00000e440, 0x97)
        /snap/go/current/src/reflect/value.go:1564 +0xbd
gocloud.dev/docstore/driver.decode(0xbda5a0, 0xc0003a8200, 0x191, 0xd95d40, 0xc00000e420, 0x1, 0xbda5a0)
        /home/carsten/go/pkg/mod/gocloud.dev@v0.20.0/docstore/driver/codec.go:391 +0x193
gocloud.dev/docstore/driver.decodeStruct.func1(0xc2a843, 0x2, 0xd95d40, 0xc00000e420, 0x40f201, 0xc00006a301)
        /home/carsten/go/pkg/mod/gocloud.dev@v0.20.0/docstore/driver/codec.go:679 +0x15d
gocloud.dev/docstore/mongodocstore.decoder.DecodeMap(0xb54500, 0xc000078780, 0x0, 0xc00006a3c0)
        /home/carsten/go/pkg/mod/gocloud.dev/docstore/mongodocstore@v0.20.0/codec.go:242 +0x13c
gocloud.dev/docstore/driver.decodeStruct(0xb8a260, 0xc0003a8200, 0x199, 0xd95d40, 0xc00000e3e0, 0xc00035fcf8, 0xc00000e3e0)
        /home/carsten/go/pkg/mod/gocloud.dev@v0.20.0/docstore/driver/codec.go:658 +0x17d
gocloud.dev/docstore/driver.Document.Decode(0xaffae0, 0xc0003a8200, 0x0, 0xb8a260, 0xc0003a8200, 0x199, 0xc0002a02a0, 0x2, 0x2, 0xd95d40, ...)
        /home/carsten/go/pkg/mod/gocloud.dev@v0.20.0/docstore/driver/document.go:184 +0x145
gocloud.dev/docstore/mongodocstore.decodeDoc(0xc000078780, 0xaffae0, 0xc0003a8200, 0x0, 0xb8a260, 0xc0003a8200, 0x199, 0xc0002a02a0, 0x2, 0x2, ...)
        /home/carsten/go/pkg/mod/gocloud.dev/docstore/mongodocstore@v0.20.0/codec.go:125 +0x11e
gocloud.dev/docstore/mongodocstore.(*docIterator).Next(0xc0000786f0, 0xd8ce60, 0xc0000a2000, 0xaffae0, 0xc0003a8200, 0x0, 0xb8a260, 0xc0003a8200, 0x199, 0xc0002a02a0, ...)
        /home/carsten/go/pkg/mod/gocloud.dev/docstore/mongodocstore@v0.20.0/query.go:127 +0xeb
gocloud.dev/docstore.(*DocumentIterator).Next(0xc000078720, 0xd8ce60, 0xc0000a2000, 0xaffae0, 0xc0003a8200, 0x0, 0x0)
        /home/carsten/go/pkg/mod/gocloud.dev@v0.20.0/docstore/query.go:234 +0x1f8
main.createAndQuery(0xc0002b2d80)
        /home/carsten/projects/test/main.go:46 +0x205
main.main()
        /home/carsten/projects/test/main.go:33 +0x118
exit status 2
```

## Expected behavior
This should print the uuid of the items stored in the in memory instance and mongodb like:
010ce414-2d8e-4d21-9dc3-aabf6c55b317
adac3c1e-2b96-4274-a630-e590693da78e

### Version

Which version(s) of the Go Cloud modules are you seeing the bug with?

<pre>
$ go version
go version go1.15.6 linux/amd64

$ cat go.mod
module test

go 1.15

require (
        github.com/google/uuid v1.1.1
        gocloud.dev v0.20.0
        gocloud.dev/docstore/mongodocstore v0.20.0
)
</pre>

vangent:
Fixes #2920
Fixes #2791 (I think)

@jba I'm not familiar with those code and am not sure this is correct. Please advise.

The newly added tests panic without the changes in `codec.go`.


docstore/driver/codec.go
```diff
@@ -388,8 +388,10 @@ func decode(v reflect.Value, d Decoder) error {
 		if err != nil {
 			return err
 		}
-		v.Set(reflect.ValueOf(val))
-		return nil
+		if v.Type().AssignableTo(reflect.TypeOf(val)) {
+			v.Set(reflect.ValueOf(val))
+			return nil
+		}
```
vangent, line 394: Is it OK to fall through if it's not assignable? I was kind of surprised when the tests passed. Maybe one of the options below ends up working?

For the new tests, this panics because `**not assignable uuid.UUID to []uint8**`. `uuid.UUID` is an array, not a slice.

docstore/driver/codec.go
```diff
@@ -510,7 +512,12 @@ func decodeList(v reflect.Value, d Decoder) error {
 	// supports that, then do the decoding.
 	if v.Type().Elem().Kind() == reflect.Uint8 {
 		if b, ok := d.AsBytes(); ok {
-			v.SetBytes(b)
+			if v.Kind() == reflect.Slice {
+				v.SetBytes(b)
```
vangent, line 516: This would previously panic with `reflect: call of reflect.Value.SetBytes on array Value`.

docstore/driver/codec.go
```diff
@@ -144,6 +144,14 @@ func encode(v reflect.Value, enc Encoder) error {
 		}
 		return nil
 	}
+	if reflect.PtrTo(v.Type()).Implements(protoMessageType) {
```
vangent, line 147: Using a struct with a protocol buffer field (as opposed to a pointer to a protocol buffer field) failed before this change; see #2791.

docstore/drivertest/drivertest.go
```diff
@@ -1163,13 +1177,46 @@ type nativeMinimal struct {
...
 	M  map[string]bool
 	P  *string
 	T  time.Time
 	LF []float64
 	LS []string
 }
 
+// testProto tests encoding/decoding of a document with protocol buffer
+// and pointer-to-protocol-buffer fields.
+func testProto(t *testing.T, _ Harness, coll *ds.Collection) {
```
vangent, line 1192: This test would fail before the new `reflect.PtrTo(v.Type()).Implements(protoMessageType)` condition in codec.go's `encode`, because the `Proto` field would not match, and fall through to being encoded as a `map` below. That would fail when decoding since the `map` couldn't be decoded to the `Proto`.

docstore/driver/codec.go
jba: The check for slice is good, but if false I think you want to fall through to the general list-handling code on line 536/551. It should handle copying into arrays of various size, not just the same size. Currently you will not assign if the destination array is a different size, which is a bug.

docstore/drivertest/drivertest.go
jba: I'm fine with this change since it seems you made the user happy, but no one should ever be writing a proto struct directly (as opposed to by pointer).

docstore/driver/codec.go
jba: Here is what's happening, and why falling through works for mongo.

First, throughout this code, `v` is the destination and the value being decoded is inside `d`. We repeatedly ask `d` for the value in different forms. `AsSpecial` asks `d` to deal with any special cases, and `AsBytes` asks `d` for the value as a byte slice, if it can do that.

If `AsSpecial` returns true as its first return value (`done` here), then `val` is now the value to set into `v`, not whatever is in `d`. At that point, `val` must be used; falling through is not in general correct.
I thought about falling through but using `val` as the new value instead of what is inside `d`, but that is infeasible.

There is a mongo type called `primitive.Binary` which wraps a byte slice. The `AsSpecial` method unwraps it, but _so does `AsBytes`_. So in that particular case, falling through works because later, when we call `AsBytes`, we get the right value.

I think the surprising fix here is to remove the `primitive.Binary` case from the mongo driver's `AsSpecial`, and return an error if this assignability check fails. In general, if a driver can convert a decoder value using `AsBytes` or any similar method, it should not handle it in `AsSpecial`.
 
Let me know if that causes other problems.

jba (changes requested):

jba:
Thanks for dealing with this, Robert. I missed my mention on the original issue.

docstore/driver/codec.go
vangent: Thanks for the info. I tried that (removing the `primitive.Binary` case from the mongo `AsSpecial`, and returning an error here if `IsAssignable` return false).

The tests fail with:

```
ABOUT TO RETURN ERROR val 2019-03-26 17:00:00.005 -0700 PDT
v.Type is interface {}
TypeOf(val) is time.Time

decoding time.Time: cannot set type interface {} to 1553644800005 (code=InvalidArgument)
```

The code looks like this:

```
                if v.Type().AssignableTo(reflect.TypeOf(val)) {
                        v.Set(reflect.ValueOf(val))
                        return nil
                }
                fmt.Printf("ABOUT TO RETURN ERROR val %v\n", val)
                fmt.Printf("v.Type is %v\n", v.Type())
                fmt.Printf("TypeOf(val) is %v\n", reflect.TypeOf(val))
                return decodingError(v, d)
```

Interestingly, if I just revert this change entirely and always call `Set`, it seems to work -- so `Set` does something (or at least, doesn't panic) if the `AssignableTo` check returns false.....

docstore/driver/codec.go
vangent: Hmm, that does not work. If I let it fall through, the call to `d.ListLen()` below fails.

docstore/driver/codec.go
vangent: @jba this change looks OK?

jba: Yes.

docstore/driver/codec.go
jba: Oh, I think you have it reversed. You want to know if val is assignable to v.

jba: The order of the args to AssignableTo is the reverse of that of an assignment statement, hence the confusion.

docstore/driver/codec.go
jba: OK, keep your code and return an error if the byte array is a different size, maybe with a TODO that it should work like the general array-copying code below.

docstore/driver/codec.go
vangent: Arg. :-)
Fixed. Works now.

docstore/driver/codec.go
vangent: I modified it to call `prepareLen` it if it's an array, I think this does the right thing but PTAL.

docstore/driver/codec.go
```diff
@@ -704,11 +724,11 @@ func fieldByIndexCreate(v reflect.Value, index []int) (reflect.Value, bool) {
 }
 
 func decodingError(v reflect.Value, d Decoder) error {
```
vangent, line 726: This is unrelated to this PR but helps debugging significantly, otherwise all the stack traces for errors point here.

vangent:
Thanks for the review @jba, PTAL.

docstore/driver/codec.go
jba: LGTM.

jba (approved):

PR #2934, opened 2020-12-18
blob/azblob: Fix SAS token not being used to open a container

### LINKED ISSUE #2933: blob/azureblob: Using a SAS token no longer works
## Describe the bug

Before https://github.com/google/go-cloud/pull/2873, setting the environment variable `AZURE_STORAGE_SAS_TOKEN` would work fine because the default behavior was to open the bucket via the environment: https://github.com/google/go-cloud/pull/2873/files#diff-9ee8a1ad2dadc385fec784185cfeacacb8796a02b4da3b576dfe5e3a9346d753L154-R169

Now, an account key MUST be provided for that to work, but that defeats the purpose of the SAS token in the first place.
 
### To Reproduce

Attempt to download or upload a blob with `AZURE_STORAGE_SAS_TOKEN` and `AZURE_STORAGE_ACCOUNT`. Azure will attempt to access the file publicly:

```
FATAL: blob (key "myblob") (code=Unknown): write error: -> github.com/Azure/azure-storage-blob-go/azblob.newStorageError, /Users/stanhu/.gvm/pkgsets/go1.15.3/global/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.11.0/azblob/zc_storage_error.go:42
===== RESPONSE ERROR (ServiceCode=PublicAccessNotPermitted) =====
Description=Public access is not permitted on this storage account.
```

## Expected behavior

The upload works.

### Version

v0.21.0

### Additional context

I think the fix here should be to tweak that `if` statement to look at `sasToken`.

stanhu:
Prior to https://github.com/google/go-cloud/pull/2873, setting the
environment variable `AZURE_STORAGE_SAS_TOKEN` would work because the
default behavior was to open the bucket via the environment. However,
that pull request changed the behavior to only use the environment if an
account key were provided, but the SAS token can be thought of as a
temporary key.

To fix this, we check whether an account key or a SAS token is
provided. If one of them is available, open the container with the
environment.

Closes #2933

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -161,7 +161,7 @@ func (o *lazyCredsOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.
 
 		isMSIEnvironment := adal.MSIAvailable(ctx, adal.CreateSender())
 
-		if accountKey != "" {
+		if accountKey != "" || sasToken != "" {
 			o.opener, o.err = openerFromEnv(accountName, accountKey, sasToken, storageDomain, protocol)
 		} else if isMSIEnvironment {
 			o.opener, o.err = openerFromMSI(accountName, storageDomain, protocol)


blob/azureblob/azureblob.go
```diff
@@ -161,7 +161,7 @@ func (o *lazyCredsOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.
 
 		isMSIEnvironment := adal.MSIAvailable(ctx, adal.CreateSender())
 
-		if accountKey != "" {
+		if accountKey != "" || sasToken != "" {
```
stanhu, line 164: I didn't see existing tests for this `OpenBucketURL`. Not sure if it's easy to do without mocking these `opener` calls.

go.sum
```diff
@@ -64,7 +64,7 @@ github.com/Azure/azure-sdk-for-go v49.0.0+incompatible h1:rvYYNgKNBwoxUaBFmd/+Tp
 github.com/Azure/azure-sdk-for-go v49.0.0+incompatible/go.mod h1:9XXNKU+eRnpl9moKnB4QOLf1HestfXbmab5FXxiDBjc=
 github.com/Azure/azure-service-bus-go v0.10.7 h1:fB1FeQYkDJN0m1FF3WWta1zUf0pOCN93UXsGnKtdqWQ=
 github.com/Azure/azure-service-bus-go v0.10.7/go.mod h1:o5z/3lDG1iT/T/G7vgIwIqVDTx9Qa2wndf5OdzSzpF8=
-github.com/Azure/azure-storage-blob-go v0.12.0 h1:7bFXA1QB+lOK2/ASWHhp6/vnxjaeeZq6t8w1Jyp0Iaw=
+github.com/Azure/azure-storage-blob-go v0.12.0 h1:LEu0WO1uxAkeJFuNTZR+xxUpIoRUwOXgA2ipS798Ycc=
```
vangent, line 67: Please revert this change as it is unrelated.

blob/azureblob/azureblob.go
vangent: There are tests for the URL parsing etc. (https://github.com/google/go-cloud/blob/master/blob/azureblob/azureblob_test.go#L527) but we don't have good tests for the various authentication mechanisms unfortunately.

vangent (changes requested):

chrismellard (approved):

vangent (approved):

PR #2935, opened 2020-12-18
blob/azblob: Fix azure-storage-blob-go checksum

stanhu:
The tag for this module got yanked and now checksums are failing. Pin
the version to the pseudo-version as described in
https://github.com/Azure/azure-storage-blob-go/issues/236#issuecomment-747063565.


stanhu:
@vangent Would you mind merging this? I'm blocked on upgrading go-cloud due to this dependency issue.

vangent (approved):

PR #2943, opened 2021-01-07
blob/gcsblob: mark Options.SignBytes and .PrivateKey as deprecated

### LINKED ISSUE #2815: blob/gcsblob: deprecate Options.{PrivateKey,SignBytes}
With upcoming v0.20 of this CDK a `Options.MakeSignBytes(Context…)` gets introduced that can be used to express both, signing by `Options.PrivateKey` and `Options.SignBytes()`.

1. We arrive at a reduced complexity.
2. We enable a later *private key* rotation and other security best-practices.
3. `PrivateKey` is a wrapper to `SignBytes` [anyway](https://github.com/googleapis/google-cloud-go/blob/v0.57.0/storage/storage.go#L639-L654).

For the legacy `PrivateKey` we can provide a `MakeSignBytes` until they arrive at a better solution, then drop it altogether.

With `gocloud.dev/blob/gcsblob` now shipping with defaults for signing, setting a signer will be rather rare outside on-premise installations. A depreciation will shift the responsibility from legacy code to this module, eliminating a possible future blocker.

vangent:
Via comment only.

Fixes #2815.

### DIFF
--- a/blob/gcsblob/gcsblob.go
+++ b/blob/gcsblob/gcsblob.go
@@ -265,11 +265,13 @@ type Options struct {
 	// PrivateKey is the Google service account private key.
 	// Exactly one of PrivateKey or SignBytes must be non-nil to use SignedURL.
 	// See https://godoc.org/cloud.google.com/go/storage#SignedURLOptions.
+	// Deprecated: Use MakeSignBytes instead.
 	PrivateKey []byte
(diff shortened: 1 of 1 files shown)


wmark:
+1

eliben (approved):

PR #2951, opened 2021-02-02
all: remove incompatible gax-go

johejo:
Remove incopmatible gax-go from dependencies.


vangent (approved):
Thanks for the cleanup!

PR #2954, opened 2021-02-05
blob/fileblob: new param "metadata=skip" to not write sidecar files

wmark:
An option needed to use fileblob as gracefully degraded drop-in in
legacy software, to approximate behaviour of a “dumb” file system
where sidecar files (with metadata) are not wanted.

A use-case is an image hoster who currently interfaces on-prem storage.

### DIFF
--- a/blob/fileblob/fileblob.go
+++ b/blob/fileblob/fileblob.go
@@ -15,6 +15,12 @@
 // Package fileblob provides a blob implementation that uses the filesystem.
 // Use OpenBucket to construct a *blob.Bucket.
 //
+// By default fileblob stores blob metadata in 'sidecar files' with the original
+// filename but an additional ".attrs" suffix.
+// That behaviour can be changed via Options.Metadata;
+// writing of those metadata files suppressed by setting it to "skip".
+// Absent any stored metadata many blob.Attributes fields will be set to default values.
+//
 // URLs
 //
 // For blob.OpenBucket, fileblob registers for the scheme "file".
@@ -97,6 +103,8 @@ const Scheme = "file"
 //   - base_url: the base URL to use to construct signed URLs; see URLSignerHMAC
 //   - secret_key_path: path to read for the secret key used to construct signed URLs;
 //     see URLSignerHMAC
+//   - metadata: if set to "skip", won't write metadata such as blob.Attributes
+//     as per the package docstring
 //
 // If either of base_url / secret_key_path are provided, both must be.
 //
@@ -134,15 +142,38 @@ func (o *URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket
 	return OpenBucket(filepath.FromSlash(path), opts)
 }
 
+var recognizedParams = map[string]bool{
+	"create_dir":      true,
+	"base_url":        true,
+	"secret_key_path": true,
+	"metadata":        true,
+}
+
+type metadataOption string // Not exported as subject to change.
+
+// Settings for Options.Metadata.
+const (
+	// Metadata gets written to a separate file.
+	MetadataInSidecar metadataOption = ""
+	// Writes won't carry metadata, as per the package docstring.
+	MetadataDontWrite metadataOption = "skip"
+)
+
 func (o *URLOpener) forParams(ctx context.Context, q url.Values) (*Options, error) {
 	for k := range q {
-		if k != "create_dir" && k != "base_url" && k != "secret_key_path" {
+		if _, ok := recognizedParams[k]; !ok {
 			return nil, fmt.Errorf("invalid query parameter %q", k)
 		}
 	}
 	opts := new(Options)
 	*opts = o.Options
 
+	switch metadataOption(q.Get("metadata")) {
+	case MetadataDontWrite:
+		opts.Metadata = MetadataDontWrite
+	default:
+		opts.Metadata = MetadataInSidecar
+	}
 	if q.Get("create_dir") != "" {
 		opts.CreateDir = true
 	}
@@ -176,6 +207,9 @@ type Options struct {
 	// If true, create the directory backing the Bucket if it does not exist
 	// (using os.MkdirAll).
 	CreateDir bool
+
+	// Refers to the strategy for how to deal with metadata, such as blob.Attributes.
+	Metadata metadataOption
 }
 
 type bucket struct {
@@ -596,6 +630,16 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
 			return nil, err
 		}
 	}
+
+	if b.opts.Metadata == MetadataDontWrite {
+		w := &writer{
+			ctx:  ctx,
+			File: f,
+			path: path,
+		}
+		return w, nil
+	}
+
 	var metadata map[string]string
 	if len(opts.Metadata) > 0 {
 		metadata = opts.Metadata
@@ -608,7 +652,7 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
 		ContentType:        contentType,
 		Metadata:           metadata,
 	}
-	w := &writer{
+	w := &writerWithSidecar{
 		ctx:        ctx,
 		f:          f,
 		path:       path,
@@ -619,7 +663,8 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
 	return w, nil
 }
 
-type writer struct {
+// writerWithSidecar implements the strategy of storing metadata in a distinct file.
+type writerWithSidecar struct {
 	ctx        context.Context
 	f          *os.File
 	path       string
@@ -630,7 +675,7 @@ type writer struct {
 	md5hash hash.Hash
 }
 
-func (w *writer) Write(p []byte) (n int, err error) {
+func (w *writerWithSidecar) Write(p []byte) (n int, err error) {
 	n, err = w.f.Write(p)
 	if err != nil {
 		// Don't hash the unwritten tail twice when writing is resumed.
@@ -643,7 +688,7 @@ func (w *writer) Write(p []byte) (n int, err error) {
 	return n, nil
 }
 
-func (w *writer) Close() error {
+func (w *writerWithSidecar) Close() error {
 	err := w.f.Close()
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


blob/fileblob/fileblob.go
```diff
@@ -143,6 +143,7 @@ func (o *URLOpener) forParams(ctx context.Context, q url.Values) (*Options, erro
 	opts := new(Options)
 	*opts = o.Options
 
+	opts.SkipMetadata = (q.Get("skip_metadata") != "")
 	if q.Get("create_dir") != "" {
```
vangent, line 147: Can you update this one to use the same one-line style for consistency?

blob/fileblob/fileblob.go
```diff
@@ -143,6 +143,7 @@ func (o *URLOpener) forParams(ctx context.Context, q url.Values) (*Options, erro
 	opts := new(Options)
 	*opts = o.Options
 
+	opts.SkipMetadata = (q.Get("skip_metadata") != "")
```
vangent, line 146: Let's improve the documentation for this whole "sidecar" thing. Can you add something above near line 17, something like:

```
//
// fileblob stores blob metadata in sidecar files with an ".attrs" suffix, including
// blob.Attributes fields other than the ones like CreateTime and ModTime
// that are provided by the filesystem.
// This implies that many blob.Attributes fields will set to default values for
// blobs read via fileblob that were not originally written using fileblob.
// Writing of these sidecar files can be suppressed via Options.SkipMetadata.
//
// URLs
```

blob/fileblob/fileblob.go
```diff
@@ -176,6 +177,11 @@ type Options struct {
 	// If true, create the directory backing the Bucket if it does not exist
 	// (using os.MkdirAll).
 	CreateDir bool
+
+	// SkipMetadata will skip the writing of any metadata files (sidecar files).
```
vangent, line 181: After adding the, this should be OK as:
```
// SkipMetadata will skip the writing of the metadata sidecar files as described
// in the package docstring.
```

blob/fileblob/fileblob.go
```diff
@@ -600,13 +606,16 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
...
 		metadata = opts.Metadata
 	}
-	attrs := xattrs{
-		CacheControl:       opts.CacheControl,
-		ContentDisposition: opts.ContentDisposition,
-		ContentEncoding:    opts.ContentEncoding,
-		ContentLanguage:    opts.ContentLanguage,
-		ContentType:        contentType,
-		Metadata:           metadata,
+	var attrs *xattrs
```
vangent, line 609: I think instead of using `nil` attrs as a signal that we're not writing the metadata, it might be simpler and more clear to leave this as-is, and instead add a  `SkipMetadata` field to `writer` (initialized via `opts.SkipMetadata`).

Then just the `setAttrs` call and the `os.Remove` call need to check it.

blob/fileblob/fileblob.go
```diff
@@ -176,6 +177,11 @@ type Options struct {
 	// If true, create the directory backing the Bucket if it does not exist
 	// (using os.MkdirAll).
 	CreateDir bool
+
+	// SkipMetadata will skip the writing of any metadata files (sidecar files).
+	// Metadata read will be incomplete at best and should not be relied upon
+	// more than that of a ordinary file system's.
+	SkipMetadata bool
```
vangent, line 184: This is only skipping metadata write, so perhaps it would be more clear to name it `SkipMetadataWrites`? Or maybe it's clear enough from the docstring.

blob/fileblob/fileblob_test.go
```diff
@@ -36,14 +36,23 @@ import (
...
+	skipMetadata bool
+	server       *httptest.Server
+	urlSigner    URLSigner
+	closer       func()
 }
 
-func newHarness(ctx context.Context, t *testing.T, prefix string) (drivertest.Harness, error) {
+func newHarness(ctx context.Context, t *testing.T, prefix string, skipMetadata bool) (drivertest.Harness, error) {
+	switch name := t.Name(); {
+	case strings.HasSuffix(name, "TestAttributes"), strings.Contains(name, "TestMetadata/"):
```
vangent, line 49: Nice, but can you add a comment here explaining that we're skipping testcases that verify metadata? I'm a bit surprised it's only these two :-|.

blob/fileblob/fileblob_test.go
```diff
@@ -159,19 +169,26 @@ func (h *harness) Close() {
...
 func TestConformanceWithPrefix(t *testing.T) {
 	const prefix = "some/prefix/dir/"
 	newHarnessWithPrefix := func(ctx context.Context, t *testing.T) (drivertest.Harness, error) {
-		return newHarness(ctx, t, prefix)
+		return newHarness(ctx, t, prefix, false)
 	}
 	drivertest.RunConformanceTests(t, newHarnessWithPrefix, []drivertest.AsTest{verifyAs{prefix: prefix}})
 }
 
+func TestConformanceNoAttrs(t *testing.T) {
```
vangent, line 185: Nit: TestConformanceSkipMetadata ?

blob/fileblob/fileblob_test.go
```diff
@@ -440,3 +457,38 @@ func TestListAtRoot(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Got error creating temp dir: %#v", err)
+	}
+	defer os.RemoveAll(dir)
+	b := &bucket{dir: dir, opts: &Options{}}
+	bucket := blob.NewBucket(driver.Bucket(b))
+
+	for _, skipMetadata := range [...]bool{false, true} {
+		b.opts.SkipMetadata = skipMetadata
+		writer, err := bucket.NewWriter(ctx, "key", &blob.WriterOptions{
```
vangent, line 475: You can use `bucket.WriteAll` to simplify this.

blob/fileblob/fileblob_test.go
```diff
@@ -440,3 +457,38 @@ func TestListAtRoot(t *testing.T) {
...
+		}
+		if _, err := writer.Write([]byte("hello world")); err != nil {
+			t.Fatal(err)
+		}
+		if err := writer.Close(); err != nil {
+			t.Fatal(err)
+		}
+
+		_, err = os.Stat(filepath.Join(dir, "key"+attrsExt))
+		if !skipMetadata && os.IsNotExist(err) {
```
vangent, line 489: I think this is backwards -- it's only verifying that a metadata file *is* written in the `false` case. It is not verifying anything in the `true` case.

I think you want something like:

```
if skipMetadata != os.IsNotExist(err) {
   ...
}
```

blob/fileblob/fileblob_test.go
```diff
@@ -440,3 +457,38 @@ func TestListAtRoot(t *testing.T) {
...
+	dir, err := ioutil.TempDir("", "fileblob*")
+	if err != nil {
+		t.Fatalf("Got error creating temp dir: %#v", err)
+	}
+	defer os.RemoveAll(dir)
+	b := &bucket{dir: dir, opts: &Options{}}
+	bucket := blob.NewBucket(driver.Bucket(b))
+
+	for _, skipMetadata := range [...]bool{false, true} {
+		b.opts.SkipMetadata = skipMetadata
```
vangent, line 474: Modifying the options using the private field is a bit weird (options are supposed to be immutable once you've created the bucket).

How about moving the loop so it's around the whole thing, so you create a new `bucket` for each of the 2 cases, using  the public `fileblob.OpenBucket` (with the `SkipMetadata` option set appropriately).

vangent (changes requested):

blob/fileblob/fileblob.go
wmark: We can always go with a stronger guarantee (in place of “metadata read will be inconsistent and should not be relied upon”) later.

The idea is to retain the option to override or set metadata if need be, at the expense of futile reads—if this were not wanted, `http.FileServe` would fill that niche.

wmark:
Thanks for entertaining this. With v0.22 out and no need to rush this, I'll apply a broader stroke to `blob/fileblob`.

vangent:
> Thanks for entertaining this. With v0.22 out and no need to rush this, I'll apply a broader stroke to `blob/fileblob`.

Sure, thanks for the contribution! Just note, I'd rather get a series of smaller PRs than one giant one.

wmark:
PTAL

blob/fileblob/fileblob.go
```diff
@@ -15,6 +15,12 @@
 // Package fileblob provides a blob implementation that uses the filesystem.
 // Use OpenBucket to construct a *blob.Bucket.
 //
+// By default fileblob stores blob metadata in 'sidecar files' with the original
+// filename but an additional ".attrs" suffix.
+// That behaviour can be changed via Options.Metadata;
+// writing of those metadata files suppressed by setting it to "skip".
```
vangent, line 21: Nit: "can be suppressed"

blob/fileblob/fileblob.go
```diff
@@ -134,15 +142,38 @@ func (o *URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket
...
 			return nil, fmt.Errorf("invalid query parameter %q", k)
 		}
 	}
 	opts := new(Options)
 	*opts = o.Options
 
+	switch metadataOption(q.Get("metadata")) {
+	case MetadataDontWrite:
+		opts.Metadata = MetadataDontWrite
+	default:
```
vangent, line 174: Let's give an explicit error for unsupported values instead of always defaulting to Sidecar. You can test for both valid and invalid values in TestOpenBucketFromURL.

blob/fileblob/fileblob.go
```diff
@@ -176,6 +207,9 @@ type Options struct {
 	// If true, create the directory backing the Bucket if it does not exist
 	// (using os.MkdirAll).
 	CreateDir bool
+
+	// Refers to the strategy for how to deal with metadata, such as blob.Attributes.
```
vangent, line 211: Maybe add something like "Supported values are in the Metadata* constants."?

vangent, line 211: Any reason why you changed this to a string from a bool Option? Do you anticipate other values, or ....?

vangent (changes requested):

blob/fileblob/fileblob.go
wmark: Yes, I've this for *xattr* (“extended attributes”) implemented, but need to get to a better understanding of its limits, and this PR as base to actually extend both `writer*`, before filing a draft.

vangent (approved):

wmark:
Nice! Thanks for working with me on this.

PR #2960, opened 2021-02-12
blob/azureblog: Add support for CDN domains

### LINKED ISSUE #2959: blob/azureblob: Storage URL is always prefixed with account name
### Describe the bug

I was to configure Azure CDN for my blob storage container using [these guides](https://docs.microsoft.com/en-us/azure/cdn/cdn-create-a-storage-account-with-cdn)

The Azure CDN gives you a URL that you can ping directly which will proxy back to the storage account without the client having to care or know about it. 

The StorageDomain option seemed like the right place to configure the CDN URL. However, it always ends up being prefixed with the AccountName which is unfortunately not correct: https://github.com/google/go-cloud/blob/d1f7069afcebb8169000b4f61d1f095eafd46268/blob/azureblob/azureblob.go#L477

### To Reproduce

Create an Azure CDN for a blob account and use this library to access it. 

### Version

v0.21.0

### Additional Context

It's worth noting the underlying storage client also has an issue detailed here: https://github.com/Azure/azure-storage-blob-go/issues/255

vangent:
Fixes #2959

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -115,12 +115,22 @@ type Options struct {
 	// The default value is "blob.core.windows.net". Possible values will look similar
 	// to this but are different for each cloud (i.e. "blob.core.govcloudapi.net" for USGovernment).
 	// Check the Azure developer guide for the cloud environment where your bucket resides.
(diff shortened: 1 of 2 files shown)


marwan-at-work (approved):
Thank you for the quick turnaround!

eliben (approved):

PR #2962, opened 2021-02-16
blob/fileblob: try copy by reflink first

wmark:
Following #2954 enables a faster file duplication on filesystems that
support clone/copy-on-write.

The speed advantage is particularly beneficial if the operator merely
intends to rename a key, in which case going through Copy is currently needed.

### DIFF
--- a/blob/fileblob/fileblob.go
+++ b/blob/fileblob/fileblob.go
@@ -63,6 +63,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync/atomic"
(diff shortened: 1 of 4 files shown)


wmark:
It's the default in **coreutils** after v8.32. https://github.com/coreutils/coreutils/commit/25725f9d41735d176d73a757430739fb71c7d043

wmark:
This is obsoleted by 9529703990f0bf720684a98c872e1c983a3abf61, which will use `copy_file_range` if available.

PR #2966, opened 2021-02-21
blob/fileblob: utilize syscall openat2 on Linux as another guard

wmark:
Extending syscall 'openat', by using flag RESOLVE_BENEATH we can involve the kernel in pathname resolution as additional (and final) guard against descending below the Bucket's configured root directory.


wmark:
Thanks for the encouragement, as expressed by the `Merge branch 'master' into…'.

A better design is to have *fileblob* be backed by a `fs.FS` (not available before Go 1.16), which would use this call. The latter's implementation then ideally comes with Go's stdlib.

Issue #2877, opened 2020-11-10
all: Migrate to OpenTelemetry


tonyhhyip:
Please use a title starting with the name of the affected package, or \"all\",
followed by a colon, followed by a short summary of the feature request.
Example: `blob/gcsblob: add support for more blobbing`.

### Is your feature request related to a problem? Please describe.

OpenCensus and OpenTracing have merged to form OpenTelemetry, considering of migrate towards OpenTelemetry should be taken

### Describe the solution you'd like

Make plans on migrate to OpenTelemetry from OpenCensus.

### Describe alternatives you've considered

Keep using OpenCensus as most of the exporter is supported on OpenCensus

### Additional context

Both [Google Cloud Trace](https://cloud.google.com/trace/docs/setup/go-ot) and [AWS](https://aws.amazon.com/otel/) begins with support on OpenTelemetry.

saeidakbari:
Is there any migration plan? it would look better if migration is done section by section as it will not introduce too many changes at once.

vangent:
There are no active plans to do this.

hsblhsn:
This issue is over one year old. But there is no progress. Both OpenCensus and OpenTracing have merged into OpenTelemetry! So, it's better to move to OpenTelemetry.

tmc:
@vangent can you share if anything has changed here?

vangent:
No, nothing has changed.

https://github.com/google/go-cloud/pull/3085 exists, but hasn't seen any action in months.

I still don't understand whether this conversion would be a backwards-compatible change or not.

vangent:
I included an announcement in todays v0.27.0 release. Let's wait a couple of weeks to see if there's any objection; if not, I'll accept the PR. We'll want to commit to updating all of the packages in this library before the next release.

saeidakbari:
Fortunately, most packages using OpenCensus in this library have an alternative compatible version in OpenTelemetry. I don't think updating them would be challenging.

vangent:
Update: stuck on this. @saeidakbari hasn't made any progress in a while. I cloned the PR, but it's not trivial to get tests to pass. I'd also want to update the entire library in one release,  and I also haven't been able to figure out how to update the rest of the code base (e.g., `blob/blob.go`) to `OpenTelemetry`.

Fryuni:
[...]

PR #2983, opened 2021-03-31
server/requestlog: Add Request Context in requestlog.Entry

mickeyreiss:
I would like to have access to the request's Context in the request logger, so that I can include metadata and also reference logging implementation details that might be stashed away in the context.

This is particularly useful for request loggers that aim to perform tracing with frameworks other than opencensus. It's also helpful for integrating with certain logging frameworks. I've also noticed that it is a good way to pass information from the `http.Server`s (e.g. [`BaseContext`](https://github.com/golang/go/blob/87c6fa4f473f178f7d931ddadd10c76444f8dc7b/src/net/http/server.go#L2627-L2638)) down to the request logger.

To solve this, I've added a new field to `Entry` that is set based on the context from the request. I believe this change is relatively simple and backwards compatible.

### DIFF
--- a/server/requestlog/requestlog.go
+++ b/server/requestlog/requestlog.go
@@ -58,6 +58,7 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	start := time.Now()
 	sc := trace.FromContext(r.Context()).SpanContext()
 	ent := &Entry{
+		Request:           cloneRequestWithoutBody(r),
 		ReceivedTime:      start,
 		RequestMethod:     r.Method,
 		RequestURL:        r.URL.String(),
@@ -95,19 +96,21 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	h.log.Log(ent)
 }
 
+func cloneRequestWithoutBody(r *http.Request) *http.Request {
+	r = r.Clone(r.Context())
+	r.Body = nil
+	return r
+}
+
 // Entry records information about a completed HTTP request.
 type Entry struct {
-	ReceivedTime      time.Time
-	RequestMethod     string
-	RequestURL        string
-	RequestHeaderSize int64
-	RequestBodySize   int64
-	UserAgent         string
-	Referer           string
-	Proto             string
+	// Request is the http request that has been completed.
+	//
+	// This request's Body is always nil, regardless of the actual request body.
+	Request *http.Request
 
-	RemoteIP string
-	ServerIP string
+	ReceivedTime    time.Time
+	RequestBodySize int64
 
 	Status             int
 	ResponseHeaderSize int64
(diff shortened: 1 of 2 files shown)


mickeyreiss:
I am not familiar with the test frameworks in this library. If this is PR is looking good, I would be happy to add a line of code somewhere to maintain the test coverage.

vangent:
@zombiezen thoughts?

Add a context.Context to a struct is generally discouraged:
https://blog.golang.org/context-and-structs

mickeyreiss:
> Add a context.Context to a struct is generally discouraged:
> https://blog.golang.org/context-and-structs

Fair enough, I generally agree. I did it this way based on the `Exception to the rule: preserving backwards compatibility` recommendation in that blog post. Is there different design that will work with the `Logger` interface?

jba:
I think it's okay here, but how about a change that's both more general and less controversial: adding the `Request` to the `Entry`? After all, it is a _request_ logger.

mickeyreiss:
> I think it's okay here, but how about a change that's both more general and less controversial: adding the `Request` to the `Entry`? After all, it is a _request_ logger.

@jba That would solve everything I am looking for. Would you recommend deprecating the redundant fields (`Referer`, `UserAgent`, etc.)? Do you have something in mind for handling the `Body`, which is in an undefined state at this point (depending on whether or not it's been read)?

jba:
You can document the redundant fields as deprecated. Also document that the implementor of Logger shouldn't touch the body.

This is all subject to @vangent's OK, BTW.

zombiezen:
It's been a while, but IIRC the original reason we did not take in a `Context` in the `Logger` interface was to push for implementations to not do something long-running because `Log` does block request completion currently. It also permits the `requestlog.Handler` implementation to log asynchronously in the future.

I'll throw in another couple of implementation ideas here:

1) Rather than having a field that stores a `Context`, have the field type be `interface { Value(interface{}) interface{} }` so as to disentangle the `Done` channel. This does unfortunately prevent future attempts at logging asynchronously and is a little awkward to use with APIs that access a `Context` (but can be worked around with struct embedding).
2) Add a method to `requestlog.Handler` like `SetContextTransform(f func(context.Context) interface{})` that invokes a callback to snapshot the data from the `Context` and then adds it to a new `interface{}` field in `Entry`. This would fit within the existing design goals while not running afoul of the `Context`-within-struct guidelines.

As an aside, migrating to OpenTelemetry would probably be a good idea to avoid users needing to deal with this in the common case. I believe OpenTelemetry is gaining more traction that OpenCensus did. I see this is already tracked as #2877.

vangent:
I like @jba's suggestion; it's the most flexible. We can document the `Request` field with the caveats (don't touch `Body`, don't modify anything, and note that `Log` is synchronous so don't do anything long-running).

mickeyreiss:
Thanks for the feedback, folks. Please take another look, as I've switched from `Context` to `Request`, added unit tests and deprecated the redundant fields with comments.

vangent (approved):

PR #2984, opened 2021-04-03
internal: Correct Docs of ADC for GCP

### LINKED ISSUE #2973: blob/gcsblob: gcloud auth login does not set ADC
### Describe the bug
Docs [here](https://gocloud.dev/howto/blob/#gcs) say that `gocloud.dev/blob` should pick up `gcloud auth login` creds, but that is not the case.

On digging a little deeper into the code, I noticed that gocloud internally uses `golang.org/x/oauth2/google` for authentication to GCP, which searches for `Application Default Credentials (ADC)`. As expected `gcloud auth application-default login` worked. This was a breaking [change](https://cloud.google.com/sdk/docs/release-notes#breaking_changes_119) in gcloud version 128.

### To Reproduce
- Login using `gcloud auth login`
- Use OpenBucket method on GCP Storage which would use Default Credentials.
- See the error.

## Expected behavior
Documentation should be updated with the correct command `gcloud auth application-default login`. 

### Version
gocloud.dev v0.20.0

### Additional context
Would like to work on this issue.

prateekj117:
Fixes #2973 

Update Documentation regarding `gcloud auth login` as it no longer (after gcloud version 128) sets Application Default Credentials.

### DIFF
--- a/internal/admin/README.md
+++ b/internal/admin/README.md
@@ -3,7 +3,7 @@
 This is a [Terraform][] configuration for the Go CDK open source project. It
 manages GitHub ACLs, issue labels, and the module proxy buckets on GCS. To apply
 the configuration to the project's resources, [sign into the gcloud
-CLI][gcloud auth login], grab a [GitHub access token][], and then do the
+CLI][gcloud auth application-default login], grab a [GitHub access token][], and then do the
 following:
 
 ```bash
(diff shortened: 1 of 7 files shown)


prateekj117:
@googlebot I signed it!

vangent (approved):
Thanks for the cleanup!

Issue #2982, opened 2021-03-31
blob/azblob: Permissions error when using SAS Token


stack72:
Since upgrading to go-cloud v0.22.0, we have started receiving permissions errors when using SAS Tokens and azblobs. The error is as follows:

```
error: saving update info: blob (key ".pulumi/stacks/dev.json -> .pulumi/history/dev/dev-1617188138003597000.checkpoint.json") (code=Unknown): -> github.com/Azure/azure-storage-blob-go/azblob.newStorageError, /Users/brew/Library/Caches/Homebrew/go_mod_cache/pkg/mod/github.com/!azure/azure-storage-blob-go@v0.13.0/azblob/zc_storage_error.go:42
===== RESPONSE ERROR (ServiceCode=AuthorizationPermissionMismatch) =====
Description=This request is not authorized to perform this operation using this permission.
RequestId:5ead053e-501e-007f-031c-261d34000000
Time:2021-03-31T10:55:38.2878720Z, Details:
   Code: AuthorizationPermissionMismatch
   PUT https://stack72testing.blob.core.windows.net/statestore/.pulumi/history/dev/dev-1617188138003597000.checkpoint.json?se=2021-03-31t18%3A48%3A01z&sig=REDACTED&sp=rw&spr=https&sr=c&st=2021-03-31t10%3A48%3A01z&sv=2020-02-10&timeout=61
   User-Agent: [go-cloud/blob/0.1.0 Azure-Storage/0.13 (go1.16.2; darwin)]
   X-Ms-Client-Request-Id: [ca9eef4c-da54-4e82-4871-c5e37fbdc99c]
   X-Ms-Copy-Source: [https://stack72testing.blob.core.windows.net/statestore/.pulumi/stacks/dev.json?se=2021-03-31t18%3A48%3A01z&sig=REDACTED&sp=rw&spr=https&sr=c&st=2021-03-31t10%3A48%3A01z&sv=2020-02-10]
   X-Ms-Tags: []
   X-Ms-Version: [2019-12-12]
   --------------------------------------------------------------------------------
   RESPONSE Status: 403 This request is not authorized to perform this operation using this permission.
   Content-Length: [279]
   Content-Type: [application/xml]
   Date: [Wed, 31 Mar 2021 10:55:37 GMT]
   Server: [Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0]
[...]

PR #2988, opened 2021-04-16
`blob/azblob`: Pass nil instead of an empty map for tags on Copy operation

leezen:
Fixes https://github.com/google/go-cloud/issues/2982

Providing an empty map ends up causing the underlying SDK to want to have permissions to write tags against the destination blob. However, we don't actually need this permission since we're not really writing any tags. Instead, we should just pass in `nil`.

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -541,7 +541,6 @@ func (b *bucket) Copy(ctx context.Context, dstKey, srcKey string, opts *driver.C
 	mac := azblob.ModifiedAccessConditions{}
 	bac := azblob.BlobAccessConditions{}
 	at := azblob.AccessTierNone
-	btm := azblob.BlobTagsMap{}
 	if opts.BeforeCopy != nil {
 		asFunc := func(i interface{}) bool {
 			switch v := i.(type) {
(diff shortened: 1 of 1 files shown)


blob/azureblob/azureblob.go
```diff
@@ -561,7 +560,7 @@ func (b *bucket) Copy(ctx context.Context, dstKey, srcKey string, opts *driver.C
 			return err
 		}
 	}
-	resp, err := dstBlobURL.StartCopyFromURL(ctx, srcURL, md, mac, bac, at, btm)
+	resp, err := dstBlobURL.StartCopyFromURL(ctx, srcURL, md, mac, bac, at, nil /* blogsTagMap */)
```
vangent, line 563: nit: BlobTagsMap (typo)

vangent:
Hmm this looks like it's a change to the golden file, which is hard for you to do. If you don't mind I'll clone it and send it to you.

vangent:
Cloned into #2989.

blob/azureblob/azureblob.go
leezen: Haha, not sure how I mistyped that. Guess I must be thinking about blogs...

leezen:
@vangent Sounds good.

PR #2992, opened 2021-05-06
mempubsub/mem: Add support for AsFunc for message type in mempubsub

anjanarvarma:
This is a PR to add support for AsFunc for message types in mempubsub,  so that while running unit tests, we can cover all the test cases which involve gocloud message type conversions to their respective driver message types.

### DIFF
--- a/pubsub/mempubsub/mem.go
+++ b/pubsub/mempubsub/mem.go
@@ -36,7 +36,7 @@
 //
 // As
 //
-// mempubsub does not support any types for As.
+// mempubsub support As for message types.
 package mempubsub // import "gocloud.dev/pubsub/mempubsub"
 
 import (
@@ -51,6 +51,8 @@ import (
 	"gocloud.dev/gcerrors"
 	"gocloud.dev/pubsub"
 	"gocloud.dev/pubsub/driver"
+
+	pb "google.golang.org/genproto/googleapis/pubsub/v1"
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


anjanarvarma:
@googlebot I signed it!

vangent:
Sorry, I don't understand what you are trying to accomplish here. The type you are exposing via `As` is not one that is used anywhere else in the package, and the only fields that it includes are ones that are exposed in the public type, so you don't need `As` to access them.

`As` is specifically a mechanism to give access to driver-specific types that may contain fields or functionality that is not exposed via the Go CDK portable types (in this case, `pubsub`).

anjanarvarma:
The reason for this change is I am trying to unit test a pubsub logic using mempubsub package. The logic I am trying to unit test has an As() which performs conversion to driver message type, and the rest of the logic uses this driver message type for processing. When I use the mempubsub package to test it with the current implementation, As() does not perform the conversion(returns nil) due to which my logic is not getting tested entirely. 

For instance, if below is a sample code that I want to unit test using mempubsub :

  func Receive(msg *pubsub.Message){
	var driverMsg *pbapi.PubsubMessage
	if ok := msg.As(&driverMsg); ok {	
	//get messageId or publish time from driver msg
	// further processing		
	}
} 

How will I be able to get past the As() with a valid driver message returned?
If there is any other solution, please let me know. I would be happy to try them.

vangent:
> logic I am trying to unit test has an As() which performs conversion to driver message type, and the
> rest of the logic uses this driver message type for processing

Why? Can you just use the `pubsub.Message` from the gocloud.dev `pubsub` package?

If the answer is "no, I need driver-specific fields that are not exposed in `pubsubMessage`", then yes, you have to use `As` for that, but `mempubsub` doesn't have any concept of those driver-specific fields from another driver. Your unit test will have to handle both cases; this is the down side of using `As` to access non-portable fields.

anjanarvarma:
What I wanted to access are MessageId and PublishTime. Is there any other way other than using As()?

vangent:
`MessageId` and `PublishTime` are driver-specific concepts. The `mempubsub` driver doesn't have them.

PR #3004, opened 2021-05-20
pubsub/gcppubsub: Allow configuration of batch sizes

### LINKED ISSUE #3003: pubsub/gcppubsub: Allow configuration of batch size

The current gcppubsub driver does not have a way of configuring the amount of messages pulled at once, it's 1000 by default. With long running processing of messages, a single subscriber can sit with all the messages, while other subscribers may have waits between receiving messages.

Add the ability to configure the batch size for the gcppubsub driver. Seems that the `SubscriptionOptions` type is currently empty, and it could be added there and used to populate the `recvBatchOpts` parameter, which is always passed as `nil` in `OpenSubscription`.

Only alternative is to reimplement the driver locally or fork the repo to add support for this.

davidsbond:
Fixes #3003 

This PR allows users of the `pubsub/gcppubsub` package to configure the receive batch size via a query parameter. 

Signed-off-by: David Bond <davidsbond93@gmail.com>

### DIFF
--- a/pubsub/gcppubsub/gcppubsub.go
+++ b/pubsub/gcppubsub/gcppubsub.go
@@ -56,6 +56,7 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -210,8 +211,63 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
 
 // OpenSubscriptionURL opens a pubsub.Subscription based on u.
 func (o *URLOpener) OpenSubscriptionURL(ctx context.Context, u *url.URL) (*pubsub.Subscription, error) {
-	for param := range u.Query() {
-		return nil, fmt.Errorf("open subscription %v: invalid query parameter %q", u, param)
+	// Set subscription options to use defaults
+	o.SubscriptionOptions = SubscriptionOptions{
+		ReceiveMaxHandlers:  recvBatcherOpts.MaxHandlers,
+		ReceiveMinBatchSize: recvBatcherOpts.MinBatchSize,
+		ReceiveMaxBatchSize: recvBatcherOpts.MaxBatchSize,
+		AckMaxHandlers:      ackBatcherOpts.MaxHandlers,
+		AckMinBatchSize:     ackBatcherOpts.MinBatchSize,
+		AckMaxBatchSize:     ackBatcherOpts.MaxBatchSize,
+	}
+
+	for param, value := range u.Query() {
+		switch param {
+		case "receive-max-handlers":
+			maxHandlers, err := queryParameterInt(value)
+			if err != nil {
+				return nil, fmt.Errorf("open subscription %v: invalid query parameter %q: %v", u, param, err)
+			}
+
+			o.SubscriptionOptions.ReceiveMaxHandlers = int(maxHandlers)
+		case "receive-min-batch-size":
+			minBatchSize, err := queryParameterInt(value)
+			if err != nil {
+				return nil, fmt.Errorf("open subscription %v: invalid query parameter %q: %v", u, param, err)
+			}
+
+			o.SubscriptionOptions.ReceiveMinBatchSize = int(minBatchSize)
+		case "receive-max-batch-size":
+			maxBatchSize, err := queryParameterInt(value)
+			if err != nil {
+				return nil, fmt.Errorf("open subscription %v: invalid query parameter %q: %v", u, param, err)
+			}
+
+			o.SubscriptionOptions.ReceiveMaxBatchSize = int(maxBatchSize)
+		case "ack-max-handlers":
+			maxHandlers, err := queryParameterInt(value)
+			if err != nil {
+				return nil, fmt.Errorf("open subscription %v: invalid query parameter %q: %v", u, param, err)
+			}
+
+			o.SubscriptionOptions.AckMaxHandlers = int(maxHandlers)
+		case "ack-min-batch-size":
+			minBatchSize, err := queryParameterInt(value)
+			if err != nil {
+				return nil, fmt.Errorf("open subscription %v: invalid query parameter %q: %v", u, param, err)
+			}
+
+			o.SubscriptionOptions.AckMinBatchSize = int(minBatchSize)
+		case "ack-max-batch-size":
+			maxBatchSize, err := queryParameterInt(value)
+			if err != nil {
+				return nil, fmt.Errorf("open subscription %v: invalid query parameter %q: %v", u, param, err)
+			}
+
+			o.SubscriptionOptions.AckMaxBatchSize = int(maxBatchSize)
+		default:
+			return nil, fmt.Errorf("open subscription %v: invalid query parameter %q", u, param)
+		}
 	}
 	sc, err := SubscriberClient(ctx, o.Conn)
 	if err != nil {
@@ -392,14 +448,34 @@ type subscription struct {
 }
 
 // SubscriptionOptions will contain configuration for subscriptions.
-type SubscriptionOptions struct{}
+type SubscriptionOptions struct {
+	ReceiveMaxHandlers  int
+	ReceiveMinBatchSize int
+	ReceiveMaxBatchSize int
+	AckMaxHandlers      int
+	AckMinBatchSize     int
+	AckMaxBatchSize     int
+}
 
 // OpenSubscription returns a *pubsub.Subscription backed by an existing GCP
 // PubSub subscription subscriptionName in the given projectID. See the package
 // documentation for an example.
 func OpenSubscription(client *raw.SubscriberClient, projectID gcp.ProjectID, subscriptionName string, opts *SubscriptionOptions) *pubsub.Subscription {
 	path := fmt.Sprintf("projects/%s/subscriptions/%s", projectID, subscriptionName)
-	return pubsub.NewSubscription(openSubscription(client, path), nil, ackBatcherOpts)
+
+	recvOpts := &batcher.Options{
+		MaxHandlers:  opts.ReceiveMaxHandlers,
+		MinBatchSize: opts.ReceiveMinBatchSize,
+		MaxBatchSize: opts.ReceiveMaxBatchSize,
+	}
+
+	ackOpts := &batcher.Options{
+		MaxHandlers:  opts.AckMaxHandlers,
+		MinBatchSize: opts.AckMinBatchSize,
+		MaxBatchSize: opts.AckMaxBatchSize,
+	}
+
+	return pubsub.NewSubscription(openSubscription(client, path), recvOpts, ackOpts)
 }
 
 var subscriptionPathRE = regexp.MustCompile("^projects/.+/subscriptions/.+$")
@@ -412,7 +488,20 @@ func OpenSubscriptionByPath(client *raw.SubscriberClient, subscriptionPath strin
 	if !subscriptionPathRE.MatchString(subscriptionPath) {
 		return nil, fmt.Errorf("invalid subscriptionPath %q; must match %v", subscriptionPath, subscriptionPathRE)
 	}
-	return pubsub.NewSubscription(openSubscription(client, subscriptionPath), nil, ackBatcherOpts), nil
+
+	recvOpts := &batcher.Options{
+		MaxHandlers:  opts.ReceiveMaxHandlers,
+		MinBatchSize: opts.ReceiveMinBatchSize,
+		MaxBatchSize: opts.ReceiveMaxBatchSize,
+	}
+
+	ackOpts := &batcher.Options{
+		MaxHandlers:  opts.AckMaxHandlers,
+		MinBatchSize: opts.AckMinBatchSize,
+		MaxBatchSize: opts.AckMaxBatchSize,
+	}
+
+	return pubsub.NewSubscription(openSubscription(client, subscriptionPath), recvOpts, ackOpts), nil
 }
 
 // openSubscription returns a driver.Subscription.
@@ -525,3 +614,11 @@ func (*subscription) ErrorCode(err error) gcerrors.ErrorCode {
 
 // Close implements driver.Subscription.Close.
 func (*subscription) Close() error { return nil }
+
+func queryParameterInt(value []string) (int64, error) {
+	if len(value) > 1 {
+		return 0, fmt.Errorf("expected only one parameter value, got: %v", len(value))
+	}
+
+	return strconv.ParseInt(value[0], 10, 64)
+}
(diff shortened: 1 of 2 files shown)


pubsub/gcppubsub/gcppubsub.go
```diff
@@ -210,8 +211,63 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
...
+
+			o.SubscriptionOptions.ReceiveMaxBatchSize = int(maxBatchSize)
+		case "ack-max-handlers":
+			maxHandlers, err := queryParameterInt(value)
+			if err != nil {
+				return nil, fmt.Errorf("open subscription %v: invalid query parameter %q: %v", u, param, err)
+			}
+
+			o.SubscriptionOptions.AckMaxHandlers = int(maxHandlers)
+		case "ack-min-batch-size":
```
vangent, line 254: I don't think we need to make the ack batchers configurable. They should be driven by the receive batches.

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -210,8 +211,63 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
...
+	for param, value := range u.Query() {
+		switch param {
+		case "receive-max-handlers":
+			maxHandlers, err := queryParameterInt(value)
+			if err != nil {
+				return nil, fmt.Errorf("open subscription %v: invalid query parameter %q: %v", u, param, err)
+			}
+
+			o.SubscriptionOptions.ReceiveMaxHandlers = int(maxHandlers)
+		case "receive-min-batch-size":
```
vangent, line 233: I don't think we should make this configurable. What's the use case? The concrete class already starts at `1` and dynamically ramps up, it's not obvious to me how a default other than 1 would behave.

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -210,8 +211,63 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
...
+		ReceiveMinBatchSize: recvBatcherOpts.MinBatchSize,
+		ReceiveMaxBatchSize: recvBatcherOpts.MaxBatchSize,
+		AckMaxHandlers:      ackBatcherOpts.MaxHandlers,
+		AckMinBatchSize:     ackBatcherOpts.MinBatchSize,
+		AckMaxBatchSize:     ackBatcherOpts.MaxBatchSize,
+	}
+
+	for param, value := range u.Query() {
+		switch param {
+		case "receive-max-handlers":
```
vangent, line 226: Not sure this needs to be configurable either, is there a use case?

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -210,8 +211,63 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
...
+
+			o.SubscriptionOptions.ReceiveMaxHandlers = int(maxHandlers)
+		case "receive-min-batch-size":
+			minBatchSize, err := queryParameterInt(value)
+			if err != nil {
+				return nil, fmt.Errorf("open subscription %v: invalid query parameter %q: %v", u, param, err)
+			}
+
+			o.SubscriptionOptions.ReceiveMinBatchSize = int(minBatchSize)
+		case "receive-max-batch-size":
```
vangent, line 240: Need to ensure that it's not > 1000, since GCP doesn't support anything larger.

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -210,8 +211,63 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
 
 // OpenSubscriptionURL opens a pubsub.Subscription based on u.
 func (o *URLOpener) OpenSubscriptionURL(ctx context.Context, u *url.URL) (*pubsub.Subscription, error) {
-	for param := range u.Query() {
-		return nil, fmt.Errorf("open subscription %v: invalid query parameter %q", u, param)
+	// Set subscription options to use defaults
+	o.SubscriptionOptions = SubscriptionOptions{
+		ReceiveMaxHandlers:  recvBatcherOpts.MaxHandlers,
```
vangent, line 216: If we do this, let's rename `recvBatchOpts` -> `defaultRecvBatchOpts` to make clear it's not used explicitly.

In fact, if you do that you fill find that it *is* used explicitly, on line 519; that will need to be updated to use the saved `SubscriptionOptions`.

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -392,14 +448,34 @@ type subscription struct {
 }
 
 // SubscriptionOptions will contain configuration for subscriptions.
-type SubscriptionOptions struct{}
+type SubscriptionOptions struct {
+	ReceiveMaxHandlers  int
```
vangent, line 452: All of these that remain would need to be documented.

Some of them are not easy to explain because they are used in non-trivial ways by the concrete implementation.

I think we can keep only `MaxBatchSize` (no need for the `Receive` prefix), with the docstring something like:

"MaxBatchSize caps the maximum batch sized used when retrieving messages. It defaults to 1000."

vangent (changes requested):

pubsub/gcppubsub/gcppubsub.go
davidsbond: I mostly went for "If I'm configuring one, I can configure them all". But I'll omit them all besides the max batch size

davidsbond:
@vangent Thanks for taking the time to look at my PR. I've implemented your feedback and also added some extra `nil` checks after seeing errors in the tests suite.

davidsbond:
It also seems like the project has ran out of credits for TravisCI:

https://travis-ci.com/github/google/go-cloud/requests

So the build hasn't run

davidsbond:
Did a force push to get the Travis CI to run, seems there's credits again now

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -210,8 +211,27 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
 
 // OpenSubscriptionURL opens a pubsub.Subscription based on u.
 func (o *URLOpener) OpenSubscriptionURL(ctx context.Context, u *url.URL) (*pubsub.Subscription, error) {
-	for param := range u.Query() {
-		return nil, fmt.Errorf("open subscription %v: invalid query parameter %q", u, param)
+	// Set subscription options to use defaults
+	o.SubscriptionOptions = SubscriptionOptions{
```
vangent, line 215: This doesn't look right, because 
a) it will override any `SubscriptionOptions` that the user actually provided if they are using a custom `Opener`.
b) Down below, the struct is updated based on the URL parameter; since we pass a pointer to the struct to `OpenSubscription`, that essentially updates all of the subscribers to whatever the latest URL parameter was.

I think this should clone the struct and update the `MaxBatchSize` based on the URL parameter. There's no need to worry about the default here since it is handled later.

Something like:

```
opts := o.SubscriberOptions // clone the options
```

and then update `ops.MaxBatchSize` down below as well.

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -210,8 +211,27 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
...
+			maxBatchSize, err := queryParameterInt(value)
+			if err != nil {
+				return nil, fmt.Errorf("open subscription %v: invalid query parameter %q: %v", u, param, err)
+			}
+
+			if maxBatchSize <= 0 || maxBatchSize > 1000 {
+				return nil, fmt.Errorf("open subscription %v: invalid query parameter %q: must be between 1 and 1000", u, param)
+			}
+
+			o.SubscriptionOptions.MaxBatchSize = int(maxBatchSize)
```
vangent, line 231: Is the cast to `int` needed here?

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -210,8 +211,27 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
...
-	for param := range u.Query() {
-		return nil, fmt.Errorf("open subscription %v: invalid query parameter %q", u, param)
+	// Set subscription options to use defaults
+	o.SubscriptionOptions = SubscriptionOptions{
+		MaxBatchSize: defaultRecvBatcherOpts.MaxBatchSize,
+	}
+
+	for param, value := range u.Query() {
+		switch param {
+		case "max-batch-size":
```
vangent, line 221: Please update the comments near line 182 to document this URL parameter.

See https://github.com/google/go-cloud/blob/master/blob/gcsblob/gcsblob.go#L204 for an example. Make sure it's clear that it only affects Subscriber.

vangent, line 221: Please modify this to `max_recv_batch_size` to match the style used elsewhere in the Go CDK, and to make it clear that it's specific to subscriptions.

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -387,19 +407,34 @@ func (*topic) ErrorCode(err error) gcerrors.ErrorCode {
...
-	path   string
+	client  *raw.SubscriberClient
+	path    string
+	options *SubscriptionOptions
 }
 
 // SubscriptionOptions will contain configuration for subscriptions.
-type SubscriptionOptions struct{}
+type SubscriptionOptions struct {
+	// MaxBatchSize caps the maximum batch sized used when retrieving messages. It defaults to 1000.
```
vangent, line 417: sized -> size

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -387,19 +407,34 @@ func (*topic) ErrorCode(err error) gcerrors.ErrorCode {
...
+}
 
 // OpenSubscription returns a *pubsub.Subscription backed by an existing GCP
 // PubSub subscription subscriptionName in the given projectID. See the package
 // documentation for an example.
 func OpenSubscription(client *raw.SubscriberClient, projectID gcp.ProjectID, subscriptionName string, opts *SubscriptionOptions) *pubsub.Subscription {
 	path := fmt.Sprintf("projects/%s/subscriptions/%s", projectID, subscriptionName)
-	return pubsub.NewSubscription(openSubscription(client, path), nil, ackBatcherOpts)
+
+	if opts == nil {
```
vangent, line 427: To make sure we're always getting the default as documented, and to share code better between the two constructors, how about:

```
dsub := openSubscription(client, path, opts)
recvOpts := *defaultRecvBatcherOpts
recvOpts.MaxBatchSize = dsub.opts.MaxBatchSize
return pubsub.NewSubscription(dsub, recvOpts, ackBatcherOpts)
```

here and in `OpenSubscriptionByPath`.

Then in `openSubscription`:

```
if opts == nil {
  opts = &SubscriptionOptions{}
}
if opts.MaxBatchSize == 0 {
  opts.MaxBatchSize = defaultRecvBatchOpts.MaxBatchSize
}
return &subscription{client, subscriptionPath, opts}
```

I think for that to work you'll need to change the return value for `openSubscription` to `*subscription`, but that should be fine.
?

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -412,12 +447,27 @@ func OpenSubscriptionByPath(client *raw.SubscriberClient, subscriptionPath strin
 	if !subscriptionPathRE.MatchString(subscriptionPath) {
 		return nil, fmt.Errorf("invalid subscriptionPath %q; must match %v", subscriptionPath, subscriptionPathRE)
 	}
-	return pubsub.NewSubscription(openSubscription(client, subscriptionPath), nil, ackBatcherOpts), nil
+
+	if opts == nil {
```
vangent, line 451: See above.

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -525,3 +575,11 @@ func (*subscription) ErrorCode(err error) gcerrors.ErrorCode {
 
 // Close implements driver.Subscription.Close.
 func (*subscription) Close() error { return nil }
+
+func queryParameterInt(value []string) (int64, error) {
+	if len(value) > 1 {
+		return 0, fmt.Errorf("expected only one parameter value, got: %v", len(value))
+	}
+
+	return strconv.ParseInt(value[0], 10, 64)
```
vangent, line 584: Maybe use `strconv.Atoi` here and return an `int` instead of `int64`?

pubsub/gcppubsub/gcppubsub_test.go
```diff
@@ -408,6 +408,8 @@ func TestOpenSubscriptionFromURL(t *testing.T) {
 		{"gcppubsub://projects/myproject/subscriptions/mysub", false},
 		// Invalid parameter.
 		{"gcppubsub://myproject/mysub?param=value", true},
+		// Valid parameters
+		{"gcppubsub://projects/myproject/subscriptions/mysub?max-batch-size=1", false},
```
vangent, line 412: Can you add a test with an invalid value for the new parameter?

vangent (changes requested):

pubsub/gcppubsub/gcppubsub.go
davidsbond: Nope, I'll make the function use `strconv.Atoi` instead as suggested. Not sure why I forgot about that function to begin with.

pubsub/gcppubsub/gcppubsub.go
davidsbond: Yep, this makes sense to me

davidsbond:
@vangent thanks for the second round of review. I think I've implemented everything as you've suggested. Let me know if there's more :)

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -212,13 +216,12 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
 // OpenSubscriptionURL opens a pubsub.Subscription based on u.
 func (o *URLOpener) OpenSubscriptionURL(ctx context.Context, u *url.URL) (*pubsub.Subscription, error) {
 	// Set subscription options to use defaults
-	o.SubscriptionOptions = SubscriptionOptions{
-		MaxBatchSize: defaultRecvBatcherOpts.MaxBatchSize,
-	}
+	opts := o.SubscriptionOptions
+	opts.MaxBatchSize = defaultRecvBatcherOpts.MaxBatchSize
```
vangent, line 220: Shouldn't need this (and in fact don't want it, in case there is a custom `URLOpener` with it set we don't want to overwrite it). If the value is 0, it will get the default via openSubscriber later.

davidsbond:
@vangent I rebased onto master and made your change

vangent (approved):

davidsbond:
@vangent thanks for the reviews! Looking forward to this being released, been a pleasure to contribute.

vangent:
No problem, and thanks for the contribution. You can use it right now if you like, by referencing the commit:

PR #3005, opened 2021-05-20
pubsub/gcppubsub: Consider publish request body size limitation when batching

zapo:
This allows configuring a maximum batch size in bytes on the batcher in an attempt to respect the GCP 10MB limitation per publish request_size as documented in https://cloud.google.com/pubsub/quotas.
Happy to add tests if this limitation should be enforced at the batcher level like proposed in this PR.

### DIFF
--- a/pubsub/batcher/batcher.go
+++ b/pubsub/batcher/batcher.go
@@ -74,6 +74,10 @@ type Batcher struct {
 	shutdown  bool
 }
 
+type sizableItem interface {
+	ByteSize() int
+}
+
 type waiter struct {
 	item interface{}
 	errc chan error
@@ -87,6 +91,8 @@ type Options struct {
 	MinBatchSize int
 	// Maximum size of a batch. 0 means no limit.
 	MaxBatchSize int
+	// Maximum bytesize of a batch. 0 means no limit.
+	MaxBatchByteSize int
 }
 
 // newOptionsWithDefaults returns Options with defaults applied to opts.
@@ -174,14 +180,26 @@ func (b *Batcher) nextBatch() []waiter {
 	if len(b.pending) < b.opts.MinBatchSize {
 		return nil
 	}
-	if b.opts.MaxBatchSize == 0 || len(b.pending) <= b.opts.MaxBatchSize {
-		// Send it all!
-		batch := b.pending
-		b.pending = nil
-		return batch
+
+	batch := make([]waiter, 0, len(b.pending))
+	batchByteSize := 1
+	for _, msg := range b.pending {
+		itemByteSize := 0
+		if sizable, ok := msg.item.(sizableItem); ok {
+			itemByteSize = sizable.ByteSize()
+		}
+		reachedMaxSize := b.opts.MaxBatchSize > 0 && len(batch)+1 > b.opts.MaxBatchSize
+		reachedMaxByteSize := b.opts.MaxBatchByteSize > 0 && batchByteSize+itemByteSize > b.opts.MaxBatchByteSize
+
+		if reachedMaxSize || reachedMaxByteSize {
+			break
+		} else {
+			batch = append(batch, msg)
+			batchByteSize = batchByteSize + itemByteSize
+		}
 	}
-	batch := b.pending[:b.opts.MaxBatchSize]
-	b.pending = b.pending[b.opts.MaxBatchSize:]
+
+	b.pending = b.pending[len(batch):]
 	return batch
 }
 
--- a/pubsub/batcher/batcher.go
+++ b/pubsub/batcher/batcher.go
@@ -148,12 +148,21 @@ func (b *Batcher) Add(ctx context.Context, item interface{}) error {
 func (b *Batcher) AddNoWait(item interface{}) <-chan error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
+
 	// Create a channel to receive the error from the handler.
 	c := make(chan error, 1)
 	if b.shutdown {
 		c <- errors.New("batcher: shut down")
 		return c
 	}
+
+	if sizable, ok := item.(sizableItem); b.opts.MaxBatchByteSize > 0 && ok {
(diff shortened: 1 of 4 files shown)


pubsub/batcher/batcher.go
```diff
@@ -174,14 +192,26 @@ func (b *Batcher) nextBatch() []waiter {
 	if len(b.pending) < b.opts.MinBatchSize {
 		return nil
 	}
-	if b.opts.MaxBatchSize == 0 || len(b.pending) <= b.opts.MaxBatchSize {
```
vangent, line 177: Same here, let's keep this optimization:

```
if b.opts.MaxByteSize == 0 && (b.opts.MaxBatchSize == 0 || len(b.pending) <= b.opts.MaxBatchSize) {
  ...
```

pubsub/batcher/batcher.go
```diff
@@ -174,14 +192,26 @@ func (b *Batcher) nextBatch() []waiter {
...
-		// Send it all!
-		batch := b.pending
-		b.pending = nil
-		return batch
+
+	batch := make([]waiter, 0, len(b.pending))
+	batchByteSize := 1
```
vangent, line 197: 0?

pubsub/batcher/batcher.go
```diff
@@ -174,14 +192,26 @@ func (b *Batcher) nextBatch() []waiter {
...
+		itemByteSize := 0
+		if sizable, ok := msg.item.(sizableItem); ok {
+			itemByteSize = sizable.ByteSize()
+		}
+		reachedMaxSize := b.opts.MaxBatchSize > 0 && len(batch)+1 > b.opts.MaxBatchSize
+		reachedMaxByteSize := b.opts.MaxBatchByteSize > 0 && batchByteSize+itemByteSize > b.opts.MaxBatchByteSize
+
+		if reachedMaxSize || reachedMaxByteSize {
+			break
+		} else {
```
vangent, line 208: Nit: don't need the `else`, just close the `if`.

pubsub/batcher/batcher.go
```diff
@@ -174,14 +192,26 @@ func (b *Batcher) nextBatch() []waiter {
...
+			batch = append(batch, msg)
+			batchByteSize = batchByteSize + itemByteSize
+		}
 	}
-	batch := b.pending[:b.opts.MaxBatchSize]
-	b.pending = b.pending[b.opts.MaxBatchSize:]
+
+	b.pending = b.pending[len(batch):]
 	return batch
```
vangent, line 215: Yes please add tests for this.

pubsub/driver/driver.go
```diff
@@ -78,6 +78,11 @@ type Message struct {
 	AfterSend func(asFunc func(interface{}) bool) error
 }
 
+// Implement batcher item for byte sized based batching
```
vangent, line 81: Nit: "ByteSize estimates the size in bytes of the message for the purpose of restricting batch sizes."

pubsub/driver/driver.go
```diff
@@ -78,6 +78,11 @@ type Message struct {
...
+func (m *Message) ByteSize() int {
+	return len(m.Body)
```
vangent, line 83: Is it just the `Body` that's included? What about the attributes?

pubsub/gcppubsub/gcppubsub.go
```diff
@@ -80,8 +80,9 @@ import (
 var endPoint = "pubsub.googleapis.com:443"
 
 var sendBatcherOpts = &batcher.Options{
-	MaxBatchSize: 1000, // The PubSub service limits the number of messages in a single Publish RPC
-	MaxHandlers:  2,
+	MaxBatchSize:     1000, // The PubSub service limits the number of messages in a single Publish RPC
+	MaxHandlers:      2,
+	MaxBatchByteSize: 10 * 1000 * 1000, // The PubSub service limits the size of request body in a single Publish RPC
```
vangent, line 85: MB is presumably 1024 * 1024.

vangent (changes requested):

pubsub/gcppubsub/gcppubsub.go
fsaintjacques: Be ready for more confusion: https://en.wikipedia.org/wiki/Byte#Multiple-byte_units

pubsub/batcher/batcher.go
zapo: Oh wow, thanks, yes leftover from testing manually.

pubsub/gcppubsub/gcppubsub.go
zapo: That's what I did initially and actually hit and other request size validation error on the pubsub API when using 1024*1024.
I feel there might be different units used in the pubsub API itself (maybe at loadbalancer level vs grpc handlers ?)

vangent: I suspect they are enforcing the total bytes for the request, while this only measures the sum of the Body fields for each message (ignores Attributes plus all other metadata). It is not obvious from the GCP documentation though.

How about we use 9 * 1024 * 1024 as the limit and document that the actual limit is 10 * 1024 * 1024 but we're giving ourselves some headroom for metadata?

pubsub/driver/driver.go
zapo: Yes only the `Data` field is contributing, the `Attributes` one (from Metadata field) is subject to other limitations as described in https://cloud.google.com/pubsub/quotas#resource_limits which can probably be left as "server side" validation errors.

zapo: However I can add an estimated size based on the attributes as well so it contributes to the batching, even though GCP doesn't really care, since that limitation could technically be used for other providers.

zapo:
Thank you for the review @vangent, I believe I addressed all the feedback in my latest push. Please let me know if anything else needed here.

vangent (approved):

vangent:
Thanks for the contribution!

PR #3006, opened 2021-05-26
internal/testing: provided script to create release git tags 

kucjac:
This pull request provides a release script mentioned in the [Release Docs](https://github.com/google/go-cloud/blame/master/internal/docs/release.md#L55), that is creating git tags for all released modules listed in the `allmodules` file.

The script takes an input tag as an argument i.e. `v0.24.0` which would be used as a git tag.
It reads the `allmodules` file line by line, checking if the modules are meant to be released, and if so it creates a suitable git tag.


vangent (changes requested):
Thanks!

Can you also update the Release Docs you referenced to remove the TODO and reference this script instead?

kucjac:
> Thanks!
> 
> Can you also update the Release Docs you referenced to remove the TODO and reference this script instead?

Oh, yeah sure. It is a good idea to mention the changes there.

vangent (approved):

PR #3008, opened 2021-05-27
blob/s3: Adds NewSessionFromURLParams for AWS to set profile in the URL

### LINKED ISSUE #2865: blob/s3: allow to specify profile via query string
### Problem

It would be nice to be able to specify AWS profile via query string, like `?profile=my_profile`.

At the moment the only possible way to do it is via an env var `AWS_PROFILE` (or `AWS_DEFAULT_PROFILE`), which is very inconvenient and doesn't work well if you have multiple AWS accounts and multiple projects.

### Solution

It won't require any SDK updates, just a couple of code changes.
As per [the doc](https://docs.aws.amazon.com/sdk-for-go/api/aws/session/), it will need `Profile: xxx` [here](https://github.com/google/go-cloud/blob/master/aws/aws.go#L41).

The profile can be parsed in [here](https://github.com/google/go-cloud/blob/master/aws/aws.go#L81) with renaming the `ConfigFromURLParams` function to something more appropriate like `SessionFromURLParams` .

I'd be keen to make a PR if the idea sounds good.

ekini:
It is used for s3 blob to enable setting profile from S3 URLs like: s3://bucket_name?profile=main

Fixes #2865

The first cut. I had to replace `lazySessionOpener` private class because after this change it is possible to open several s3 buckets using different AWS profiles, so creating a session with `init.Do` doesn't make much sense anymore.

Basically, the idea is to get profile from the url and pass the rest of the parameters to `ConfigFromURLParams` afterwards.

Maybe it would be a good idea to refactor those functions that parse the URL somehow, so they are not repeated in several places in this repository (i.e. https://github.com/google/go-cloud/blob/master/blob/azureblob/azureblob.go#L318 does pretty much the same thing, but it's private, and also doesn't allow multiple values for the same parameter).

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -105,3 +105,33 @@ func ConfigFromURLParams(q url.Values) (*aws.Config, error) {
 	}
 	return &cfg, nil
 }
+
+// NewSessionFromURLParams returns an session.Session with session.Options initialized based on the URL
+// parameters in q. It is intended to be used by URLOpeners for AWS services.
+// https://docs.aws.amazon.com/sdk-for-go/api/aws/session/#Session
+//
+// It should be used before ConfigFromURLParams as it strips the query
+// parameters it knows about
+//
+// The following query options are supported:
+//  - profile: The AWS profile to use from the AWS configs (shared config file and
+//             shared credentials file)
+func NewSessionFromURLParams(q url.Values) (*session.Session, url.Values, error) {
+	// always enable shared config (~/.aws/config by default)
+	opts := session.Options{SharedConfigState: session.SharedConfigEnable}
+	rest := url.Values{}
+	for param, values := range q {
+		value := values[0]
+		switch param {
+		case "profile":
+			opts.Profile = value
+		default:
+			rest.Add(param, value)
+		}
+	}
+	sess, err := session.NewSessionWithOptions(opts)
+	if err != nil {
+		return nil, nil, fmt.Errorf("couldn't create session %w", err)
+	}
+	return sess, rest, nil
+}
--- a/blob/s3blob/s3blob.go
+++ b/blob/s3blob/s3blob.go
@@ -69,7 +69,6 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	"sync"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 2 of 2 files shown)


ekini:
@googlebot I signed it!

vangent (changes requested):
Can you add a test that uses "profile"?

Should be near here:
https://github.com/google/go-cloud/blob/master/blob/s3blob/s3blob_test.go#L293

It won't really verify the semantics, but at least verify that
1) using a "profile" query parameter works.
2) using a "profile" query parameter along with another query parameter works.
3) using a "profile" query parameter along with an invalid query parameter fails (this will at least partly test the "rest" code path).

ekini:
Thanks for the feedback! Changes added.

vangent (approved):

ianhmarkowitz:
I believe this change introduced a race

This is the (partial) error message:

`WARNING: DATA RACE
Write at 0x00c0001a4d00 by goroutine 61:
  gocloud.dev/blob/s3blob.(*urlSessionOpener).OpenBucketURL()
      [Removed]/vendor/gocloud.dev/blob/s3blob/s3blob.go:108 +0x1e4
  gocloud.dev/blob.applyPrefixParam()
      [Removed]/vendor/gocloud.dev/blob/blob.go:1270 +0x210
  gocloud.dev/blob.(*URLMux).OpenBucket()
      [Removed]/vendor/gocloud.dev/blob/blob.go:1247 +0xd2
  gocloud.dev/blob.OpenBucket()
      [Removed]/vendor/gocloud.dev/blob/blob.go:1301 +0x32d
      
...

Previous write at 0x00c0001a4d00 by goroutine 60:
  gocloud.dev/blob/s3blob.(*urlSessionOpener).OpenBucketURL()
      [Removed]/vendor/gocloud.dev/blob/s3blob/s3blob.go:108 +0x1e4
  gocloud.dev/blob.applyPrefixParam()
      [Removed]/vendor/gocloud.dev/blob/blob.go:1270 +0x210
  gocloud.dev/blob.(*URLMux).OpenBucket()
      [Removed]/vendor/gocloud.dev/blob/blob.go:1247 +0xd2
  gocloud.dev/blob.OpenBucket()
      [Removed]/vendor/gocloud.dev/blob/blob.go:1301 +0x32d
      
...`

vangent:
Thanks for reporting, sent https://github.com/google/go-cloud/pull/3101 to
fix.

PR #3010, opened 2021-06-07
blob/s3blob: Don't set ContentType in the S3 input for signing if it is an empty string

praneetloke:
Including an empty ContentType during signing causes the signed URL to also include `content-type` as one of the signed headers. This doesn't appear to be a problem with AWS S3 itself. However, when used with Minio, it seems that Minio is strict about including the content-type when a client tries to use a signed URL to upload objects. That is, it seems to expect the client upload the object to also including an empty `content-type` header. Otherwise, a signature mismatch error is encountered. I have worked around this issue by using the `BeforeSign` callback to override the `ContentType` of `s3.PutObjectInput` to `nil` and then things work as expected.

### DIFF
--- a/blob/s3blob/s3blob.go
+++ b/blob/s3blob/s3blob.go
@@ -748,9 +748,11 @@ func (b *bucket) SignedURL(_ context.Context, key string, opts *driver.SignedURL
 		req, _ = b.client.GetObjectRequest(in)
 	case http.MethodPut:
 		in := &s3.PutObjectInput{
-			Bucket:      aws.String(b.name),
-			Key:         aws.String(key),
-			ContentType: aws.String(opts.ContentType),
+			Bucket: aws.String(b.name),
+			Key:    aws.String(key),
+		}
+		if opts.ContentType != "" {
+			in.ContentType = aws.String(opts.ContentType)
 		}
 		if opts.BeforeSign != nil {
 			asFunc := func(i interface{}) bool {


vangent:
Looks like tests are not passing for this. I thought maybe it was because we'd need to re-record the replay files, so I cloned locally and tried that, but they still fail.

The semantics for an empty ContentType in `SignedURLOptions` are relatively tricky. They are documented here:
https://github.com/google/go-cloud/blob/master/blob/blob.go#L1099

I will need to take some time to determine if this proposed change violates that contract, or if there is a bug in the driver tests.

praneetloke:
> so I cloned locally and tried that, but they still fail.

Curious as to what the failure was? Was it the same as the ones in Travis? I expected no changes in S3 behavior based on my tests with an S3 and a Minio backend.

The most surprising failure is this one: https://travis-ci.com/github/google/go-cloud/jobs/511865465#L360
>  GET with PUT URL got status code 200, want 2xx? false

I don't understand how GET'ing a PUT signed URL can pass? Maybe because the content-type header is not included, S3 allows downloads? That sort of defeats the purpose of having signed URLs for a specific HTTP method. I assume these tests are running against a real AWS S3 account.

> I will need to take some time to determine if this proposed change violates that contract, or if there is a bug in the driver tests.

Sounds good. Happy to assist if I can, so please let me know.

vangent:
I forked this and made a minor change, please see #3011. The change was also setting the `ContentType` header if `SignedURLOptions.EnforceAbsentContentType=true`.

It turns out there was a bug in the test, so I fixed that and re-recorded the .replay files.

PR #3011, opened 2021-06-08
blob/s3blob: Don't set ContentType in the S3 input for signing if it is an empty string and EnforcedAbsentContentType is false

vangent:
See #3010 , forked from there and fixing tests.

@praneetloke
 
Including an empty ContentType during signing causes the signed URL to also include content-type as one of the signed headers. This doesn't appear to be a problem with AWS S3 itself. However, when used with Minio, it seems that Minio is strict about including the content-type when a client tries to use a signed URL to upload objects. That is, it seems to expect the client upload the object to also including an empty content-type header. Otherwise, a signature mismatch error is encountered. I have worked around this issue by using the BeforeSign callback to override the ContentType of s3.PutObjectInput to nil and then things work as expected.


blob/drivertest/drivertest.go
```diff
@@ -2215,42 +2224,49 @@ func testSignedURL(t *testing.T, newHarness HarnessMaker) {
...
+		tests = append(tests, signedURLTest{"putURLWithContentType", putURLWithContentType, allowedContentType, true})
+		tests = append(tests, signedURLTest{"putURLWithContentType", putURLWithContentType, differentContentType, false})
+		tests = append(tests, signedURLTest{"putURLWithContentType", putURLWithContentType, "", false})
 	}
+	/*
+	*/
 	if putURLEnforcedAbsentContentType != "" {
-		tests = append(tests, signedURLTest{http.MethodPut, "", putURLWithoutContentType, true})
-		tests = append(tests, signedURLTest{http.MethodPut, differentContentType, putURLWithoutContentType, false})
+		tests = append(tests, signedURLTest{"putURLEnforcedAbsentContentType", putURLEnforcedAbsentContentType, "", true})
```
vangent, line 2244: This was an actual bug in the test -- it was using putURLWithoutContentType instead of putURLEnforcedContentType.

The rest of the changes here are cosmetic to make it easier to see which test failed.

praneetloke: Thank you for doing this! I spent quite a bit of time last night trying to correlate the tests with the failures in Travis.

vangent: Yea, me too :-).

vangent:
Every few months we'll tag a release. You can use it with this commit before then though.

Issue #3016, opened 2021-06-24
secrets: Provide API to sign digests


maguro:
It would be handy to be able to sign and verify digests.  Not sure if this would go into `secrets`.  I'm thinking that there's enough here to warrant a new `signatures` package.

maguro:
fyi, I've started to flesh out what would go into this on my fork.

vangent:
This sounds like a bigger feature change than we're going to accept here, but we're happy to link to your fork if you add it there.
https://github.com/google/go-cloud#project-status

PR #3017, opened 2021-06-26
Add signatures capability

maguro:
Adds signing digests and verification of digital signatures with keys from the cloud; https://github.com/google/go-cloud/issues/3016.


vangent:
Per separate discussion, please host this in a separate repo. We're happy to add links.

PR #3018, opened 2021-07-05
pubsub/kafkapubsub: Configure initial offset in URL.

davidsbond:
This commit adds the ability for users of the `kafkapubsub` package to specify the initial consumer offset
when using a URL.

When creating a kafka consumer, you sometimes want to specify at which point the consumer should begin reading from
a topic. This is specified as either `oldest` or `newest`, where `oldest` is the oldest committed offset by the
consumer group and `newest` is the most recent offset.

Currently, there is no way to configure this via the URL, you'd have to use the `kafkapubsub` package directly to
build your consumer. So when you use a URL, you get the default sarama behaviour and you start consuming from
the newest point in the stream. This doesn't allow you to read streams from the start, which is a very common
usecase for consumers.

Signed-off-by: David Bond <davidsbond93@gmail.com>

### DIFF
--- a/pubsub/kafkapubsub/kafka.go
+++ b/pubsub/kafkapubsub/kafka.go
@@ -139,7 +139,10 @@ const Scheme = "kafka"
 //
 // For subscriptions, the URL's host+path is used as the group name,
 // and the "topic" query parameter(s) are used as the set of topics to
-// subscribe to.
+// subscribe to. The "offset" parameter is available to subscribers to set
+// the Kafka consumer's initial offset. Where "oldest" starts consuming from
+// the oldest offset of the consumer group and "newest" starts consuming from
+// the most recent offset on the topic.
 type URLOpener struct {
 	// Brokers is the slice of brokers in the Kafka cluster.
 	Brokers []string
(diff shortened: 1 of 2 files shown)


pubsub/kafkapubsub/kafka.go
```diff
@@ -164,12 +164,30 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
 
 // OpenSubscriptionURL opens a pubsub.Subscription based on u.
 func (o *URLOpener) OpenSubscriptionURL(ctx context.Context, u *url.URL) (*pubsub.Subscription, error) {
-	q := u.Query()
-	topics := q["topic"]
-	q.Del("topic")
-	for param := range q {
-		return nil, fmt.Errorf("open subscription %v: invalid query parameter %q", u, param)
+	var topics []string
```
vangent, line 167: Please document the new URL options (near line 142).

pubsub/kafkapubsub/kafka_test.go
```diff
@@ -499,6 +499,11 @@ func TestOpenSubscriptionFromURL(t *testing.T) {
 	}{
 		// OK, but still error because broker doesn't exist.
 		{"kafka://mygroup?topic=mytopic", true},
+		// OK, specifying initial offset, but still error because broker doesn't exist.
```
vangent, line 502: These tests are terrible (not your fault), since they all expect an error. What do you think about checking more explicitly for the "invalid query parameter" error so that we can distinguish the "good" URL parameters from the bad ones?

davidsbond: Yeah I can do that. How about modifying the tests so you can specify the type of error you get back? Effectively ignoring the "has your client ran out of brokers to talk to?" one, since that will happen every time?

vangent: Not sure exactly what you mean by "type of error", but that sounds right.

(I was thinking of changing `WantErr` to be a `string` instead of a `bool`, and the test would check that the returned `err.Error()` had that string in it. Checking error strings is generally a bad idea, but I think it would improve the coverage here.

davidsbond: That's what I meant, using `type` was the wrong word since we have actual types. I'm just talking about string comparisons. I'll double check what sarama returns isn't actually a sentinel error but it might just come down to comparing strings.

davidsbond:
@vangent I've made your suggested changes :)

vangent (approved):

vangent:
Thanks for the contribution!

davidsbond:
Thanks @vangent, is there any chance of a tagged release soon? It'd be great to have this and #3004 on a fixed version. I've accidentally rolled back on myself while doing dependency upgrades once already haha.

vangent:
Probably not for a while, but you can reference it right now using pseudo version based on the commit:

PR #3027, opened 2021-08-05
pubsub/gcppubsub: Tell gRPC to allow messages up to 11MB

### LINKED ISSUE #2947: GCP pubsub dials without specifying correct max message size
### Describe the bug

By default GRPC message size is 4MB, GCP PubSub official docs say it's 10MB.  When you dial here https://github.com/google/go-cloud/blob/master/pubsub/gcppubsub/gcppubsub.go#L237-L247, you need specify the max message size or allow to specify the variables yourself.
.
### To Reproduce

Subscribe, send a 5-9MB message, see how the subscriber errors. 

## Expected behavior

To receive messages >4MB.

Dial can look like:
```
	conn, err := grpc.DialContext(ctx, endPoint,
		grpc.WithTransportCredentials(credentials.NewClientTLSFromCert(nil, "")),
		grpc.WithPerRPCCredentials(oauth.TokenSource{TokenSource: ts}),

		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(1024*1024*10)),
		useragent.GRPCDialOption("pubsub"),
	)
```

Multiply:
Fixes #2947.

### DIFF
--- a/pubsub/gcppubsub/gcppubsub.go
+++ b/pubsub/gcppubsub/gcppubsub.go
@@ -267,10 +267,11 @@ func Dial(ctx context.Context, ts gcp.TokenSource) (*grpc.ClientConn, func(), er
 		grpc.WithTransportCredentials(credentials.NewClientTLSFromCert(nil, "")),
 		grpc.WithPerRPCCredentials(oauth.TokenSource{TokenSource: ts}),
 		// The default message size limit for gRPC is 4MB, while GCP
-		// PubSub supports messages up to 10MB. Tell gRPC to support up
-		// to 10MB.
+		// PubSub supports messages up to 10MB. Aside from the message itself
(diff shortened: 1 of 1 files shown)


rusenask:
looks good, actually got the same issue as https://github.com/google/go-cloud/issues/2947 few days ago and just had to modify a forked version as a hotfix. Is there any info on how much bigger can the message be? I haven't collected the data in my case but if it's more than 11MB it might make sense to bump it even higher (even to 15 or 20MB) as we are on the receiving side here anyway.

Multiply:
We're not using metadata ourselves, so I'm unsure how big it can be.
I have not encountered a single message above 11MB yet, but I don't see the harm in bumping it further

vangent (approved):

vangent:
Fine to change to 12 as well if you want to be more cautious.

PR #3028, opened 2021-08-05
pubsub/natspubsub: update dependency per dependabot suggestion

vangent:
See https://github.com/google/go-cloud/pull/3024, which failed because dependabot didn't clean the go.mod/go.sum files correctly.


eliben (approved):

mmorel-35:
Maybe a workflow on push to dependabot branches with
- gomodcleanup.sh 
- internal/testing/listdeps.sh > internal/testing/alldeps
- commit changes to the dependabot branch

would avoid that kind of problem?

PR #3032, opened 2021-08-06
blob/azureblob: update azure-storage-blob-go dep

vangent:
Also change dependabot to monthly.


eliben (approved):
Just a question: any reason to change the replay files here?

vangent:
> Just a question: any reason to change the replay files here?

Yes -- it looks like the new version uses a different `X-Ms-Version` header; the old replay files were matching on that so the tests fail.

I updated the TestPipeline for Azure to clear the value of this header going forward, so if you look at the diffs for a `.replay` file you'll see the old values `2019-xx-yy` being replaced by `CLEARED_VALUE` or something like that.

Issue #3026, opened 2021-07-30
all/s3: Upgrade to AWS SDK Version 2


richardartoul:
Has there been any further discussion about migrating this library to use AWS SDK Version 2? I know there was some hesitation in the past because the library was still in developer preview, but it looks like they moved to GA awhile back: https://aws.amazon.com/blogs/developer/aws-sdk-for-go-version-2-general-availability/

vangent:
I don't think we've investigated. It doesn't look trivial because it's not a backwards-compatible change, and so we'd need to support both versions; it's not obvious what the best way to do that is.

We could either do that by creating new providers for each AWS service (e.g., `blob/s3blobv2` instead of `blob/s3blob`), which means a lot of code duplication. Or, by trying to support both in a single provider (e.g., for `blob/s3blob`, we could add a new `OpenBucketV2` constructor and (possibly) a new URL opener, which would mean more code complexity.

I would lean toward the former, but it's a significant amount of work and we haven't heard much demand. Let's leave this bug open to see if anyone really wants this.

vangent:
I've sent a PR to add a `secrets/awskmsv2` provider, which is basically the same as the existing `secrets/awskms` provider except that it uses the V2 SDK.

It also adds a bunch of scaffolding in terms of the shared code used to create an `aws.Config`, etc.

Anyone who is familiar/interested in `V2`, please take a look and leave comments.

I expect once this is submitted, it should be relatively straightforward to do the same for the other AWS providers.

vangent:
https://github.com/google/go-cloud/pull/3034 is an alternative approach, adding support to the existing provider instead of adding a new one. I think I like this way better, but opinions are welcome.

PR #3034, opened 2021-08-07
secrets/awskms: add support for using AWS SDK V2

vangent:
Updates #3026.

This is an alternative approach, where we support both SDKs in the same driver.

I think I'm leaning toward this one, but opinions welcome.


aws/aws.go
```diff
@@ -135,3 +142,55 @@ func NewSessionFromURLParams(q url.Values) (*session.Session, url.Values, error)
...
+func UseV2(q url.Values) bool {
+	for param, values := range q {
+		if param == "awssdk" {
+			if values[0] == "v2" || values[0] == "V2" {
+				return true
+			}
+			return false
+		}
+	}
+	return false
```
servusdei2018, line 161: Rather than looping over all the values, this may be simplified to a single parameter lookup, like so:
```suggestion
if values, ok := q["awssdk"]; ok {
	if values[0] == "v2" || values[0] == "V2" {
		return true
	}
}
return false
```

servusdei2018 (changes requested):

eliben (approved):

eliben:
This approach does seem better in terms of overload for the user, who in most cases shouldn't care whether Go CDK uses v1 and v2, right?

aws/aws.go
vangent: Thanks, done!

PR #3051, opened 2021-08-31
secrets/hashivault: add support for setting the secrets engine

### LINKED ISSUE #3049: secrets/hashivault: allow usage of transit engines with a custom name
The `Encrypt` and `Decrypt` methods in the hashivault secrets provider use a hardcoded secrets engine called "transit". https://github.com/google/go-cloud/blob/master/secrets/hashivault/vault.go#L197

A "transit" secrets engine is just a type of secrets engine, there's no reason to assume it will be called "transit"; and in fact there's potentially reason to have multiple transit secrets engines, meaning you can't assume the "transit secrets engine" used to encrypt/decrypt data will always be named "transit".

I'd like to be able to provide a custom name to use for the transit secrets engine in the Encrypt/Decrypt methods. Since Go doesn't support method overloads, to avoid breaking changes to the API there could be new methods `EncryptWithEngine` and `DecryptWithEngine` that take the transit engine name as a parameter.

This is the necessary first step for a similar feature request in https://github.com/pulumi/pulumi.

I don't use go-cloud directly; therefore I don't have even a baseline understanding of the structure of the project or the design goals. Perhaps there are technical reasons the secrets/hashivault package can't have new methods, I don't know. In that case maybe there could be another environment variable for `VAULT_TRANSIT_ENGINE_NAME` or something; although that solution wouldn't allow for using multiple transit engines. I'm open to suggestions. The idea is to allow a differently-named transit engine to be used, rather than hardcoding the name "transit".

vangent:
Fixes #3049.

@leviathanbadger

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,6 +17,7 @@ on: [push, pull_request]
 jobs:
   build:
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         go-version: [1.17.x]
--- a/secrets/hashivault/vault.go
+++ b/secrets/hashivault/vault.go
@@ -144,7 +144,8 @@ const Scheme = "hashivault"
 //
 // The URL Host + Path are used as the keyID.
 //
-// No query parameters are supported.
+// The following query parameters are supported:
+// - engine: The secrets engine to use; defaults to "transit".
 type URLOpener struct {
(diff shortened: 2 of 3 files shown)


vangent:
@leviathanbadger can you confirm this works for what you need?

leviathanbadger:
This will work for my use case. Thank you for the quick turnaround!

eliben (approved):

PR #3085, opened 2021-12-19
server/*: migrate server tracing to OpenTelemetry

saeidakbari:
This PR is a step toward migrating Tracing and Metrics from OpenCensus to OpenTelemetry. discussed at #2877 

Packages changed are:
server
server/xrayserver
server/sdserver
internal/trace


saeidakbari:
Please rescan for CLA to take effect

vangent:
Sorry, I'm not super familiar with OpenCensus or OpenTelemetry.

Is this a backwards-compatible change?

tmc:
@saeidakbari can you rebase this and sign the CLA?

saeidakbari:
Sorry for the very late check on this PR.
These changes are not backward compatiable. Is it OK to proceed?

vangent:
I don't see how it's possible to make this change if it's not backwards compatible. Wouldn't it potentially break all existing users?

vangent:
(Doesn't OpenTelemetry have a compatibility story with OpenCensus?)

saeidakbari:
Practically yes, OpenTelemetry is a successor of OpenCensus but the minimum Go version required for it is `1.16`.

vangent:
The minimum Go version is not a problem.

I think https://opentelemetry.io/docs/migration/opentracing/ describes how to make it backward compatible, by having the OpenTelemetry code also emit OpenCensus events.

saeidakbari:
OK, then I think there isn't any issue to continue this PR.
I will make some changes in the next few days.

saeidakbari:
@vangent Conflict resolved, but this PR only change the server package to OpenTelemetry.
I did it with the purpose of not making a big PR with lots of code changes.

internal/trace/trace.go
```diff
@@ -18,27 +18,29 @@ package trace
 import (
 	"context"
 
-	"go.opencensus.io/trace"
+	"go.opentelemetry.io/otel/codes"
```
vangent, line 21: This package doesn't actually appear to be used anywhere; I deleted it so you can drop these files and merge.

server/requestlog/requestlog.go
```diff
@@ -25,7 +25,7 @@ import (
 	"net/http"
 	"time"
 
-	"go.opencensus.io/trace"
+	"go.opentelemetry.io/otel/trace"
```
vangent, line 28: This doesn't look like it's using the bridge that IIUC is required to make this backwards compatible.

vangent: https://pkg.go.dev/go.opentelemetry.io/otel/bridge/opentracing#section-readme

vangent (changes requested):

internal/trace/trace.go
saeidakbari: Rebased

server/requestlog/requestlog.go
saeidakbari: @vangent Link you provide is talking about OpenTracing but SDK uses OpenCensus, take a look at this:
https://pkg.go.dev/go.opentelemetry.io/otel/bridge/opencensus
Anyway, this change will not be backwards compatible, because applications using the server package have to change the server `New` function parameter from OpenCensus to OpenTelemetry.

vangent: I don't want to move ahead with this if it's not backwards compatible for users who still have OpenCensus monitoring integrations.

I'm a bit less concerned about the `New` call if it's straightforward for a user to update their code to compile and then it will continue to work with their OpenCensus integrations.

I'm more concerned about changing the application to no longer emit OpenCensus events, and only OpenTelemetry ones; that would require a user to update their monitoring stack as part of the upgrade. It's especially not OK to update some packages in this library in that way, but not others.

saeidakbari: The `New` call is straightforward enough for users IMO and I don't think users need to change their integrated monitoring facilities because OpenTelemetary is fully compatible in the case of exploring metrics and traces.

If bridge added, also all spans created with openTracing or OpenCensus will be exported by otel

saeidakbari: Also, known limitation to migrating is explained here:
https://opentelemetry.io/docs/migration/opentracing/#limits-on-compatibility

vangent: Maybe I'm not understanding, but it seems like:

* Right now, with no changes, users can migrate to OpenTelemetry by creating a bridge as described in https://pkg.go.dev/go.opentelemetry.io/otel/bridge/opencensus ("Since OpenCensus telemetry uses globals, installing a bridge will result in telemetry collection from all libraries that use OpenCensus, including some you may not expect. For example ([#1928](https://github.com/open-telemetry/opentelemetry-go/issues/1928)), if a client library generates traces with OpenCensus, installing the bridge will cause those traces to be exported by OpenTelemetry."). They can use the bridge to create an OpenCensus to pass to `New` ("Override OpenCensus' DefaultTracer with the bridge").

So, with no changes, we currently support OpenCensus and OpenTelemetry (through a bridge).

* If we submit this CL now, then users who are using OpenTelemetry will have an easy switch by updating their call to `New`. However, users that are using OpenCensus will be forced to migrate; I don't think that OpenCensus dashboards will work with OpenTelemetry-only events.


Another way of saying this is looking at the migration journey:
https://opentelemetry.io/docs/migration/opentracing/#migration-overview

"Progressively re-instrument your application using the OpenTelemetry API. There is no hard requirement to remove existing OpenTracing API calls from your application, they will continue to work." is step 5 of 5; it happens after all of the other migration work is complete. I think this implies that shared libraries like this one should wait until a high percentage of its users have done steps 1-4.

vangent:
Related question: why are you motivated to make this change? Can't you use the bridge to get OpenTelemetry events from this library?

saeidakbari:
@vangent Accessing OpenCensus data is possible through migrating to OpenTelemetry and implementing the bridge but not vice versa.
Basically, OpenTelemetry is the future of tracing, metrics and monitoring and migrating to it can bring new functionalities in different aspects.

server/requestlog/requestlog.go
saeidakbari: If we go by the OpenTelemetry and the bridge, users with OpenCensus tracing spans will still have access to their events.

vangent:
Yes, I think that agrees with what I said. What functionality do you expect to get by doing the migration in this library that you wouldn't get by using the bridge?

My understanding of the compatibility requirements as stated above imply that shared libraries like this one should continue emitting OpenCensus events, supporting both their OpenCensus and OpenTelemetry users, until such a point where there are basically no more OpenCensus users.

I'm not sure how to determine when that point is unfortunately, any ideas? Maybe I can make an announcement that we're planning on making the non-backward-compatible change as part of the next release announcement and see if anyone objects.

saeidakbari:
Mostly third-party libraries are more available and maintained for OpenTelemetry rather than deprecated OpenCensus. Also, this migration is inevitable as the process of migration has to happen sometime.

As this project appears to be used in many different places and has a good reputation I think an announcement before the non-backwards change would be good to see if we have any complaints.

naveensrinivasan:
> Related question: why are you motivated to make this change? Can't you use the bridge to get OpenTelemetry events from this library?

We depend on `go-cloud`,and this is causing the fuzz testing to fail in [oss-fuzz](http://github.com/google/oss-fuzz/)
https://github.com/ossf/scorecard-webapp/issues/220

vangent:
I did a release today, and included an announcement that we'll be moving to OpenTelemetry in the next release. Unless there are objections over the next couple of weeks, I'll re-review this and plan to get it merged in after that. Thanks for your patience!

Note that I believe you can use the links in comments above to get OpenCensus libraries to emit OpenTelemetry events.

server/server.go
```diff
@@ -115,10 +124,7 @@ func (srv *Server) init() {
 		if srv.reqlog != nil {
 			h = requestlog.NewHandler(srv.reqlog, h)
 		}
-		h = &ochttp.Handler{
-			Handler:          h,
-			IsPublicEndpoint: true,
-		}
+		h = otelhttp.NewHandler(h, srv.handlerName)
```
vangent, line 127: Does this work with empty string? Should we provide a default?

vangent (changes requested):
Let's go ahead with this. Can you sync to HEAD?

naveensrinivasan:
> Let's go ahead with this. Can you sync to HEAD?

A friendly ping.

vangent:
Cloned to #3189 , but tests don't pass.

PR #3122, opened 2022-04-16
postgres: Support IAM auth for GCP.

### LINKED ISSUE #3121: postgres: Support IAM auth for GCP

The cloudsql-proxy supports using [IAM database auth](https://cloud.google.com/sql/docs/postgres/iam-logins) but this package does not. This also means that packages that rely on this, such as the postgresql terraform module, must copy passwords around, as in the [example](https://registry.terraform.io/providers/cyrilgdn/postgresql/latest/docs#gcp).

I should be able to pass in an enable_iam_login parameter, and have it passed through to cloudsql-proxy. I really would love to be able to use that with the terraform postgresql package which relies on this.

Otherwise, we have to keep throwing SQL password strings around, which is annoying and less secure. Alternately, switch to AWS, where this is [supported](https://github.com/cyrilgdn/terraform-provider-postgresql/pull/134).

### Additional context

I believe this can be implemented by doing the following (based on a cursory review of the code):

1. Add an iam auth option to `go-cloud/gcp/cloudsql/cloudsql.go`, and use that to set `EnableIAMLogin: true,` in `certs.RemoteOpts{}`
2. Add a url parameter for iam auth, and read it in from `lazyCredsOpener.OpenPostgresURL` (in `gcppostgres/gcppostgres.go`)
3. Filter out the url parameter from the previous step in `URLOpener.OpenPostgresURL` (in `gcppostgres/gcppostgres.go`) when doing parameter validation.

aebrahim:
This fixes #3121.

This always enables the IAM login option for GCP.

We could also have this configurable through a URL parameter, but that seems like the wrong approach. If you truly for some reason didn't want IAM authentication for cloud sql, you would have already declined to enable the flag it in the GCP setup for that database instance. If you did want to use IAM authentication, it would be annoyingly hard to debug why it wasn't working as it may surprise you to learn that the proxy has it off by default. So, it makes the most sense to just always enable it in the proxy.

### DIFF
--- a/gcp/cloudsql/cloudsql.go
+++ b/gcp/cloudsql/cloudsql.go
@@ -32,5 +32,5 @@ var CertSourceSet = wire.NewSet(
 // NewCertSource creates a local certificate source that uses the given
 // HTTP client. The client is assumed to make authenticated requests.
 func NewCertSource(c *gcp.HTTPClient) *certs.RemoteCertSource {
-	return certs.NewCertSourceOpts(&c.Client, certs.RemoteOpts{})
+	return certs.NewCertSourceOpts(&c.Client, certs.RemoteOpts{EnableIAMLogin: true})
 }


jba:
Is this a backwards-compatible change? Does adding this option cause other forms of login to fail?

Also, reword your commit message:

- Start with the directory and a short summary: "gcp/cloudsql: enable IAM login"
- Then the description, after a blank line.
- Another blank line, then "Fixes #XXXX."

Lastly, a small typo: remove "it" in "enable the flag it in the..."

aebrahim:
Yes, the change is backwards compatible.

Reworded the commit message, PTAL @jba

jba (approved):

vangent:
Thanks for the contribution!

PR #3137, opened 2022-06-14
blob/s3blob.ExampleOpenBucketV2: removed duplicate error handler in documentation

fabiant7t:
Removed the duplicate, identical error handler in the example of `gocloud.dev/blob/s3blob.ExampleOpenBucketV2`:
```
cfg, err := awsv2cfg.LoadDefaultConfig(ctx)
if err != nil {
	return err
}
if err != nil {
	return err
}
```


vangent:
Thanks for the PR! It turns out that just editing the JSON file is not sufficient, that files is generated from source. I'm fixing this in https://github.com/google/go-cloud/pull/3138.

PR #3159, opened 2022-08-11
blob/azureblob: add storage_account query parameter

ekini:
This PR adds "storage_account" as possible query string parameter to the URL.

It'll allow opening multiple storage accounts using the same credentials.

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -28,7 +28,9 @@
 // credentials and a service URL; see
 // https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/storage/azblob
 // for a more complete descriptions of each approach.
-//   - AZURE_STORAGE_ACCOUNT: The service account name. Required.
+//   - AZURE_STORAGE_ACCOUNT: The service account name. Required if used along with AZURE_STORAGE KEY, because it defines
+//     authentication mechanism to be azblob.NewSharedKeyCredential, which creates immutable shared key credentials.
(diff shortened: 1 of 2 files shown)


blob/azureblob/azureblob.go
```diff
@@ -216,6 +216,8 @@ func (o *ServiceURLOptions) withOverrides(urlValues url.Values) (*ServiceURLOpti
 				return nil, err
 			}
 			retval.IsLocalEmulator = isLocalEmulator
+		case "storage_account":
```
vangent, line 219: So, this will override the AccountName for the ServiceURLOptions, and thus in the ServiceURL, but not in the credentials; see newCredInfoFromEnv below.

Does overriding this only make sense if you're using an authentication mechanism that doesn't include AccountName (e.g., not shared key)?

blob/azureblob/azureblob_test.go
```diff
@@ -624,6 +624,8 @@ func TestOpenBucketFromURL(t *testing.T) {
 		{"azblob://mybucket?protocol=http", false},
 		// With invalid protocol.
 		{"azblob://mybucket?protocol=ftp", true},
+		// With Account.
+		{"azblob://mybucket?storage_account=test", false},
```
vangent, line 628: Please also add a test to TestNewServiceURL above, similar to "Setting domain from the URL".

ekini: added!

vangent (approved):

PR #3160, opened 2022-08-11
blob/azureblob: Adds CLI auth

### LINKED ISSUE #3012: blob/azureblob: add a way to use CLI auth
As requested in https://github.com/pulumi/pulumi/issues/7251, it would be helpful to support `AZURE_STORAGE_AUTH_VIA_CLI` so that blob storage could be authenticated via CLI auth, similar to the existing `AZURE_KEYVAULT_AUTH_VIA_CLI` that was added in https://github.com/google/go-cloud/issues/2462.

ekini:
Contrary to the closed #3012 the cli auth didn't work for me so I've taken a look a made a working version.

However! Here https://github.com/google/go-cloud/blob/master/blob/azureblob/azureblob.go#L42 it says it should fall back to `NewDefaultAzureCredentials` but I don't see it happening in the code below.

But then again after looking at https://github.com/Azure/azure-sdk-for-go/blob/sdk/azidentity/v1.1.0/sdk/azidentity/default_azure_credential.go#L45-L93 it seems it does something similar as code here https://github.com/google/go-cloud/blob/master/blob/azureblob/azureblob.go#L320-L331 so maybe we should just use `NewDefaultAzureCredentials` straight away?

Hence, I'm creating this PR as a draft to discuss the approach first.

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -293,6 +293,7 @@ const (
 	credTypeSASViaNone
 	credTypeConnectionString
 	credTypeIdentityFromEnv
+	credTypeIdentityFromCli
 )
 
 type credInfoT struct {
@@ -317,6 +318,9 @@ func newCredInfoFromEnv() *credInfoT {
 	credInfo := &credInfoT{
 		AccountName: accountName,
 	}
+	// set the default first
+	credInfo.CredType = credTypeIdentityFromEnv
+
 	if accountName != "" && accountKey != "" {
 		credInfo.CredType = credTypeSharedKey
 		credInfo.AccountKey = accountKey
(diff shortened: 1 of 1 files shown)


vangent:
Thanks for trying it out and your suggestions!

You are right, the comment says it defaults to `NewDefaultAzureCredential` while the code uses `NewEnvironmentCredential`. The former seems better because it tries the latter and then also tries other things.

So, can you update the PR to rename `credTypeIdentityFromEnv` to `credTypeIdentityDefault` and have it call `NewDefaultAzureCredential`? Then see if that works for CLI credentials; based on https://github.com/Azure/azure-sdk-for-go/blob/sdk/azidentity/v1.1.0/sdk/azidentity/default_azure_credential.go#L74 it seems like it should.

I'm less confident that `NewDefaultAzureCredential` handles the shared key, connection string, and SAS token approaches in the same way as `azureblob` does, and so I think it makes sense to detect and handle them explicitly the way it's done now, for backwards compatibility if nothing else.

vangent:
I'm doing this in #3161, PTAL and let me know if it doesn't work.

ekini:
Thank you, that worked!

PR #3174, opened 2022-10-21
feat: Update keyvault impl to new sdk module

AaronFriel:
This removes a ~500MiB dependency on github.com/Azure/azure-sdk-for-go, moving to their new finely divided Go module per SDK libraries. This will have nice benefits for Go module caching for downstream packages, as well as update to the latest SDK standards and [security guidance](https://devblogs.microsoft.com/azure-sdk/guidance-for-applications-using-the-key-vault-libraries/).

I was a bit unsure of how to integrate and test this, being unfamiliar with the testing harness in this repo. Happy to take any pointers on this.


secrets/azurekeyvault/akv.go
```diff
@@ -219,20 +215,20 @@ var (
 // The "/{key-version}" suffix is optional; it defaults to the latest version.
 // See https://docs.microsoft.com/en-us/azure/key-vault/about-keys-secrets-and-certificates
 // for more details.
-func OpenKeeper(client *keyvault.BaseClient, keyID string, opts *KeeperOptions) (*secrets.Keeper, error) {
+func OpenKeeper(client azcore.TokenCredential, keyID string, opts *KeeperOptions) (*secrets.Keeper, error) {
```
vangent, line 218: Can we keep this as the Client? The URLOpener can create a default client.

AaronFriel: What's the purpose of passing in a default client here? We'd be discarding it to create one with the vault URL.

vangent: The reason for this is similar to the rationale behind dependency injection. There are a million and three options for how to create a `Client`, so we want callers to be able to provide one explicitly instead of us creating one internally and thus only supporting options that we explicitly enable.

You are right though, that the constructor `NewClient` in the new package requires a vault URL, which we don't know until `OpenKeeper`.

One approach is to modify the constructors to take all of the parameters needed by `NewClient` (the credentials, plus `ClientOptions`). Another approach is to modify the constructors to take a function, e.g., `makeClient func(vaultURL string) *azkeys.Client`. It could default to `Dial`.

At this point I think it probably makes more sense for me to patch your Pull Request and modify it; I suspect that will be needed anyway to update the test goldens etc. I'll cc you on the PR to get your feedback.

vangent: @AaronFriel  see PR #3175

AaronFriel: Thank you!

PR #3175, opened 2022-10-21
secrets/azurekeyvault: Update to the latest Azure SDK BREAKING_CHANGE_OK

vangent:
Used #3174 as a starting point.


vangent:
FYI @AaronFriel

PR #3178, opened 2022-11-14
blob/memblob: fix copy with options

emcfarlane:
Fix copy for memblob with options. Current behaviour always returns the option error value, nil with no copy done.


vangent:
Thanks for the fix!

PR #3200, opened 2023-01-04
pubsub/awssqs Fix BeforeSend to Enable changes to the sqs input message

### LINKED ISSUE #3199: pubsub/awssqs BeforeSend pass a reference to the message
### Describe the bug

the pubsub BeforeSend function implementation by  `awssqs` driver using aws sdk v2 is not working. The BeforeSend function is getting a copy of the original message not a reference to it, so any updates or changes to the message occur on the copy and not the original message. 

### To Reproduce

Steps to reproduce the behavior.

1. Create/Use aws sqs queue of type fifo which requires a message group id attribute to be sent alongside the original message
2. Create a publisher (open a pubsub topic with awssqs driver)
3. call pubsub.Send and pass a message 
4. Provide a BeforeSend func to inject the message group id as it's not a default parameter to pubsub
5. aws api will reject the message with error 
Send Error:  pubsub (code=Unknown): operation error SQS: SendMessageBatch, https response error StatusCode: 400, RequestID: 6dqabc73-bcx4-5bv0-a55c-66663018j7cb, api error InvalidParameterValue: The request must contain the parameter MessageGroupId.

```
 msgAttrs := sqstypesv2.SendMessageBatchRequestEntry{
		Id:                      nil,
		MessageBody:             nil,
		DelaySeconds:            30,
		MessageAttributes:       nil,
		MessageDeduplicationId:  nil,
		MessageGroupId:          aws.String(uuid.NewString()),
		MessageSystemAttributes: nil,
	}

	fmt.Println("Message Group ID", msgAttrs.MessageGroupId)

	err = publisher.Send(ctx, &pubsub.Message{
		Body:     []byte("test message"),
		Metadata: nil,
		BeforeSend: func(asFunc func(interface{}) bool) error {
			ok := asFunc(&msgAttrs)
			msgAttrs.MessageGroupId = aws.String(uuid.NewString())
			fmt.Println("Body .... ", *msgAttrs.MessageBody)
			fmt.Println("Message Group ID", msgAttrs.MessageGroupId)
			fmt.Println("Ok??? ", ok)
			return nil
		},
		AfterSend: nil,
	})
```
## Expected behavior

The BeforeSend function is expected to pass a reference to the original message so changes/updates can be added to the message.

### Version

v0.27.0

szaher:
Fixes #3199

### DIFF
--- a/pubsub/awssnssqs/awssnssqs.go
+++ b/pubsub/awssnssqs/awssnssqs.go
@@ -678,15 +678,14 @@ func (t *sqsTopic) SendBatch(ctx context.Context, dms []*driver.Message) error {
 			if len(attrs) == 0 {
 				attrs = nil
 			}
-			entry := sqstypesv2.SendMessageBatchRequestEntry{
+			entry := &sqstypesv2.SendMessageBatchRequestEntry{
 				Id:                aws.String(strconv.Itoa(len(req.Entries))),
 				MessageAttributes: attrs,
 				MessageBody:       aws.String(body),
 			}
-			req.Entries = append(req.Entries, entry)
 			if dm.BeforeSend != nil {
 				asFunc := func(i interface{}) bool {
-					if p, ok := i.(*sqstypesv2.SendMessageBatchRequestEntry); ok {
+					if p, ok := i.(**sqstypesv2.SendMessageBatchRequestEntry); ok {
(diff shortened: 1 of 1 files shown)


szaher:
recreating to fix cla

Issue #3218, opened 2023-01-23
aws sdk v2 setting custom endpoint missing SigningRegion


szaher:
### Describe the bug

AWS SDK v2 while setting custom endpoint is missing SigningRegion which must be set to correctly query custom AWS endpoints or services like https://localstack.cloud

```go
			customResolver := awsv2.EndpointResolverWithOptionsFunc(
				func(service, region string, options ...interface{}) (awsv2.Endpoint, error) {
					return awsv2.Endpoint{
						PartitionID:   "aws",
						URL:           value,
					}, nil
				})
			opts = append(opts, awsv2cfg.WithEndpointResolverWithOptions(customResolver))
```

### To Reproduce

append custom endpoint query param to queue url ?endpoint=http://custom-endpoint

## Expected behavior

initialized AWS clients being able to connect to the custom endpoints with no error like resource not found since client might not be connecting to the right region.

### Version

Which version(s) of the Go Cloud modules are you seeing the bug with?

szaher:
I am sorry I have missed that in previous PR #3215

PR #3219, opened 2023-01-23
Set SigningRegion when using custom endpoint url

szaher:
make sure to set `SigningRegion` while using custom endpoints for aws sdk v2 

fixes: #3218

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -177,6 +177,7 @@ func NewDefaultV2Config(ctx context.Context) (awsv2.Config, error) {
 // The following query options are supported:
 //   - region: The AWS region for requests; sets WithRegion.
 //   - profile: The shared config profile to use; sets SharedConfigProfile.
+//   - endpoint: The AWS service endpoint to send HTTP request.
 func V2ConfigFromURLParams(ctx context.Context, q url.Values) (awsv2.Config, error) {
 	var opts []func(*awsv2cfg.LoadOptions) error
 	for param, values := range q {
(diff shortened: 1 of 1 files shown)


aws/aws.go
```diff
@@ -188,8 +189,9 @@ func V2ConfigFromURLParams(ctx context.Context, q url.Values) (awsv2.Config, err
 			customResolver := awsv2.EndpointResolverWithOptionsFunc(
 				func(service, region string, options ...interface{}) (awsv2.Endpoint, error) {
```
vangent, line 190: Do we not need to use "service"?

szaher: I don't think we need to pass the service parameter to any other function or add it to an object. 
the service parameter is the name of the service and can be used for custom handling https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/endpoints/

szaher: It's being used only to check if the custom endpoint should be use for specific service
```
customResolver := aws.EndpointResolverWithOptionsFunc(func(service, region string, options ...interface{}) (aws.Endpoint, error) {
    if service == dynamodb.ServiceID && region == "us-west-2" {
        return aws.Endpoint{
            PartitionID:   "aws",
            URL:           "https://test.us-west-2.amazonaws.com",
            SigningRegion: "us-west-2",
            HostnameImmutable: true,
        }, nil
    }
    return aws.Endpoint{}, fmt.Errorf("unknown endpoint requested")
})

cfg, err := config.LoadDefaultConfig(context.TODO(), config.WithEndpointResolverWithOptions(customResolver))

```

PR #3228, opened 2023-02-20
blob/gcsblob: update proto import

Multiply:
Similar to the fix in #3227

I chose to split the remaining fixes up into 4 smaller PRs, the others being: #3229 #3230 #3231


Multiply:
I force pushed to retest, as I couldn't seem to trigger it any other way. No changes.

vangent (approved):

PR #3232, opened 2023-02-24
blob/s3blob - fix: S3 upload with io.Multireader fail on exceeded total allowed configured MaxUploadParts

javiersigler:
== Description

S3 set a maximum number of 10K upload parts, setting a maximum file size of `BufferSize` * 10K, which, by default, is 10K *5MiB ~= 50GiB. 

Reported error
```
MultipartUpload: upload multipart failed
	upload id: XXXX
caused by: TotalPartsExceeded: exceeded total allowed configured MaxUploadParts (10000). Adjust PartSize to fit in this limit
```

**Proposed Change**
* New writer option: `MaxUploadParts` specifying the maximum number of uploaded parts.

== References
https://github.com/aws/aws-sdk-go/issues/2557

### DIFF
--- a/blob/driver/driver.go
+++ b/blob/driver/driver.go
@@ -61,6 +61,9 @@ type WriterOptions struct {
 	// write in a single request, if supported. Larger objects will be split into
 	// multiple requests.
 	BufferSize int
+	// MaxUploadParts specifies the maximum number of uploaded parts.
+	// No all drivers set this value. Default S3 value is 10_000.
+	MaxUploadParts int
 	// MaxConcurrency changes the default concurrency for uploading parts.
 	MaxConcurrency int
 	// CacheControl specifies caching attributes that services may use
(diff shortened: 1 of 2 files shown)


blob/driver/driver.go
```diff
@@ -61,6 +61,9 @@ type WriterOptions struct {
 	// write in a single request, if supported. Larger objects will be split into
 	// multiple requests.
 	BufferSize int
+	// MaxUploadParts specifies the maximum number of uploaded parts.
```
vangent, line 64: Let's model this comment after the previous one:


// MaxUploadParts changes the maximum number of uploaded parts, if supported.

No need to mention the S3 default.

vangent, line 64: You will also need to add this option to WriterOptions as defined in blob.go, and copy the value to driver.WriterOptions. Search for MaxConcurrency in blob.go.

vangent (changes requested):

Issue #3214, opened 2023-01-17
blob/s3: add support for per-request options


dkolbly:
### Is your feature request related to a problem? Please describe.

I have an application where I need to use AWS STS (Security Token Service) to obtain
credentials to access an S3 bucket.  The way these credentials are supplied to s3v2
is via an options list on the call to, e.g., ListObjectsV2 or GetObject.  In order to use
the blob API I need a way to supply the appropriate option on the call.

### Describe the solution you'd like

The solution I have in mind (and have prototyped on a fork) is that As() would
accept a `*[]func(*s3v2.Options)` argument which would allow the injection
of arguments to the call.

### Describe alternatives you've considered

Considered a solution where the options could be expressed at the constructor
(`s3blob.OpenBucketV2`).  This approach requires an extra level of conceptual
indirection, and in my situation passing information via the context
because the necessary information is not known until the call site.

### Additional context

As mentioned above, I have a prototype of the `As()` approach that I've used
in my application to good effect.  Patch is here:
https://gist.github.com/dkolbly/d5deae7f6efd942bed899e8b8c795495

vangent:
Can these credentials be used at bucket instantiation time? If so, could you instantiate buckets lazily? I don't think there's much cost to doing so.

dkolbly:
Yeah, I think the bucket instantiation is fairly cheap compared to network calls.  I briefly considered an approach similar to that, but it would kind of turn my application on its head and doesn't seem to be how go-cloud is intended to be used.  I'm using google wire, and the bucket configuration is bound at wiring time.

As a workaround, I could probably create my own bucket abstraction and open the bucket at request time.  It would be easier if _all_ my AWS (or even S3) calls were using the same STS auth, then I could just build an aws.Config with the right CredentialsProvider, but that doesn't wind up working out either.

How do people normally deal with setting s3.Options on requests in go-cloud?  🤔  When I was using `minio` I have to set s3.Options.UsePathStyle, and I think that would run into the same problem trying to use s3/blob.

vangent:
Extending `As` to support this seems OK. Can you send a PR, including a test? You should be able to extend the existing tests (e.g., https://github.com/google/go-cloud/blob/master/blob/s3blob/s3blob_test.go#L236) with the new type; just verify the type conversion works.

PR #3238, opened 2023-03-06
blob/s3blob: adds support for per-request s3v2 Options, implementing #3214

dkolbly:
This PR adds support in the **blob/s3blob** driver for per-request s3 options when using the underlying v2 AWS SDK.

Specifically, adds **As** support for:

 * `ListOptions.BeforeList` (V2) for the type `*[]func(*s3v2.Options)`
 * `ReaderOptions.BeforeRead` (V2) for the type `*[]func(*s3v2.Options)`

which gives the application the ability to supply per-request options when using the V2 SDK.

There does not appear to be any utility in doing the same for `BeforeWrite` because we already expose access to the `s3v2manager.Uploader` which includes the `ClientOptions` list.

### DIFF
--- a/blob/s3blob/s3blob.go
+++ b/blob/s3blob/s3blob.go
@@ -968,13 +968,16 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
 		}
 		if opts.BeforeWrite != nil {
 			asFunc := func(i interface{}) bool {
-				// Note that since GCS does not expose AWS's
-				// Uploader abstraction, there does not appear
-				// to be any utility in exposing the options
-				// list to the v2 Uploader's Upload() method.
-				// Instead, applications can manipulate the
-				// exposed *Uploader directly, including by
-				// setting ClientOptions if needed.
+				// Note that since the Go CDK Blob
+				// abstraction does not expose AWS's
+				// Uploader concept, there does not
+				// appear to be any utility in
+				// exposing the options list to the v2
+				// Uploader's Upload() method.
+				// Instead, applications can
+				// manipulate the exposed *Uploader
+				// directly, including by setting
+				// ClientOptions if needed.
 				if p, ok := i.(**s3managerv2.Uploader); ok {
 					*p = uploaderV2
 					return true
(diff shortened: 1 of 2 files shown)


blob/s3blob/s3blob.go
```diff
@@ -960,14 +968,19 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
 		}
 		if opts.BeforeWrite != nil {
 			asFunc := func(i interface{}) bool {
-				pu, ok := i.(**s3managerv2.Uploader)
-				if ok {
-					*pu = uploaderV2
+				// Note that since GCS does not expose AWS's
```
vangent, line 971: What is "GCS"?

blob/s3blob/s3blob_test.go
```diff
@@ -233,10 +239,16 @@ func (v verifyContentLanguage) BeforeList(as func(interface{}) bool) error {
...
 			}
 		} else {
-			var req *s3v2.ListObjectsV2Input
-			if !as(&req) {
-				return errors.New("List.As failed")
+			var (
+				list *s3v2.ListObjectsV2Input
+				opts *[]func(*s3v2.Options)
+			)
+			// make sure we can extract ALL the underlying GCS types
```
vangent, line 246: I think you can drop this comment and make the code look more like above, e.g.:

if !as(&list) || !as(&opts) {
  return errors.New(...)
}
return nil

vangent (changes requested):
Mostly looks great! Couple of small suggestions. Sorry for the delay.

blob/s3blob/s3blob.go
rusq: Google Cloud Storage, I'd think. (I'm just passing by)

vangent: That's what GCS means to me too, but this is the AWS S3 driver, not the GCS one.

dkolbly: Interesting... yeah, I picked up "GCS" from [some code in the test file](https://github.com/google/go-cloud/blob/master/blob/s3blob/s3blob_test.go#L132), but I bet that code actually _did_ come from GCS.

Will fix!

dkolbly:
I believe I have addressed the feedback, thanks :+1:

vangent (approved):

PR #3247, opened 2023-05-08
blob: add Bucket.{Upload,Download} APIs

### LINKED ISSUE #2807: blob/s3: High memory usage during multi-part uploads

### LINKED ISSUE #3245: blob: allow driver.Bucket to upload data by pulling from an `io.Reader` instead of pushing to an `io.Writer`
followed by a colon, followed by a short summary of the feature request.
Example: `blob/gcsblob: add support for more blobbing`.

Unless I am mistaken, the only way to upload data to a `driver.Bucket` is using `driver.Bucket.NewTypedWriter`. While this is fine for most scenarios--users can just `io.Copy` an `io.Reader` into the `driver.Writer`--it can cause performance problems in cases where the API of the underlying storage works in terms of `io.Reader`. In those cases, the driver must use something like `io.Pipe` to bridge between the `driver.Writer` interface and the `io.Reader` expected by the underlying storage API. For an example of the problems this can cause, see #2807, which details allocation volume issues for `s3blob`.

We really need a method whose semantics are "upload the data from this reader" s.t. we can pass the reader directly to the underlying storage. Maybe we could add `Upload` methods to `blob.Bucket` and `driver.Bucket`:

    func (b *blob.Bucket) Upload(ctx context.Context, key string, body io.Reader, opts *blob.WriterOptions)
    func Upload(ctx context.Context, key string, body io.Reader, opts *driver.WriterOptions)

For parity, we could also add `Download` methods:

    func (b *blob.Bucket) Download(ctx context.Context, w io.Writer, key string, opts *blob.ReaderOptions)
    func Download(ctx context.Context, w io.Writer, key string, opts *driver.ReaderOptions)

`blob.Bucket.Writer` already exposes `ReadFrom`. My first thought was that we could change the implementation of that method to check whether or not the underlying writer is an `io.ReaderFrom` and if so call its `ReadFrom` method. Unfortunately, I don't think that really helps, as I think the semantics of `blob.ReadFrom` are such that it can be called repeatedly.

### Additional context

In order to work around this issue, we've specialized some of out internal code to use `s3manager.Uploader` directly instead of `bucket.NewWriter` because profiles indicated that we were spending a disproportionate amount of time zeroing and collecting temporary buffers due to #2807. The workaround isn't too annoying, but we were pretty shocked when we discovered that we were hitting #2807 and that it was costing us so much CPU time.

pgavlin:
Add Upload and Download methods to blob.Bucket and optional Uploader and Downloader interfaces to driver. If a driver.Bucket implements those interfaces, Upload and Download will call through those interfaces. This allows backends that can implement pull-style uploads and push-style downloads more efficiently than push-style uploads and pull-style downloads to do so.

These changes include an implementation of driver.Uploader for s3blob.Bucket. That implementation allows s3blob.Bucket to avoid allocating temporary buffers if the input is an io.ReaderAt and an io.Seeker. This can dramatically reduce allocation volume for services that upload large amounts of data to S3.

Fixes #3245
Fixes #2807

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -623,6 +623,12 @@ func newBucket(b driver.Bucket) *Bucket {
 	}
 }
 
+func (b *Bucket) isClosed() bool {
+	b.mu.RLock()
+	defer b.mu.RUnlock()
+	return b.closed
+}
+
 // As converts i to driver-specific types.
 // See https://gocloud.dev/concepts/as/ for background information, the "As"
 // examples in this package for examples, and the driver package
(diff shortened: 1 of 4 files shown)


vangent:
See https://github.com/google/go-cloud/pull/3248.

PR #3258, opened 2023-06-14
chore: remove refs to deprecated  xerrors

testwill:


vangent:
Thanks for the cleanups!

PR #3270, opened 2023-07-07
all: update go.mod Go version

### LINKED ISSUE #3269: all: go.sum monotonically increasing
I was [updating my version of gocloud.dev from 0.24.0 to 0.30.0](https://github.com/spotlightpa/sheets-uploader/pull/1/files), and I see that it has increased the length of my go.sum file by 2,600+ lines! There seems to be some issue with how you're doing upgrades where old versions never drop out of the go.sum file. For example, there are 40 different versions of cloud.google.com/go listed in my go.sum file, which all seem to be copied over from yours.

carlmjohnson:
Fixes #3269


carlmjohnson:
I conservatively set it to Go 1.18, which is 2 versions (one year) behind at this point. The Go team actually only supports Go 1.19+, and they're going to switch to Go 1.20+ in a month or so when Go 1.21 is released.

vangent:
Thanks for figuring this out!  I'm going to reject this and do it in a separate PR, as we have more than just this one go.mod file, and go 1.19 is OK.

PR #3282, opened 2023-07-21
pubsub/natspubsub: NATS v2.2.0+ native message headers and message encoding

kucjac:
This PR implements the native NATS v2.2.0+ message headers as metadata. 

This PR **doesn't break** the API of the `github.com/google/go-cloud/pubsub/natspubsub` package. It provides additional functionality that could be chosen by the user of the library.

For servers that support it (NATS Server 2.2.0 or later), messages could
be encoded using native NATS message headers, and native message content.
This provides full support for non-Go clients. Versions prior to 2.2.0 uses `gob.Encoder` to encode the message metadata and content, which limits the subscribers only to Go clients.
To use this feature, set the query parameter "natsv2" in the URL. 
The value needs to either be empty value or boolean parsable. For example:
  - `nats://mysubject?natsv2`
  - `nats://mysubject?natsv2=true`

This feature could also be enabled by setting the  `URLOpener.UseV2` field manually.

This implementation automatically determines if the server is capable of using v2.2.0+ headers.
If the server does not support it, any attempt to use it results in an error.

Using **native** NATS message headers and content is more efficient than using
`gob.Encoder`, and allows **non-Go clients** to subscribe to the topic and
receive messages.
The benchmarks provided in the test files shows that using v2.2.0+ API is about three times faster on both `Send` and  `Receive` operations than the original implementation:
```
goos: linux
goarch: amd64
pkg: gocloud.dev/pubsub/natspubsub
cpu: AMD Ryzen 7 PRO 4750U with Radeon Graphics
BenchmarkNatsQueuePubSub
BenchmarkNatsQueuePubSub/V1
BenchmarkNatsQueuePubSub/V1/BenchmarkReceive
BenchmarkNatsQueuePubSub/V1/BenchmarkReceive-16	1	1368954367 ns/op 7304.85 MB/s
BenchmarkNatsQueuePubSub/V1/BenchmarkSend
BenchmarkNatsQueuePubSub/V1/BenchmarkSend-16	12	102359676 ns/op	97694.72 MB/s
BenchmarkNatsQueuePubSub/V2
BenchmarkNatsQueuePubSub/V2/BenchmarkReceive
BenchmarkNatsQueuePubSub/V2/BenchmarkReceive-16	3	439734654 ns/op	22740.99 MB/s
BenchmarkNatsQueuePubSub/V2/BenchmarkSend
BenchmarkNatsQueuePubSub/V2/BenchmarkSend-16	28	44177761 ns/op	226358.23 MB/s
PASS
```
where the `MB/s` is the actual number of messages send/received per second.


pubsub/natspubsub/nats.go
```diff
@@ -23,6 +23,27 @@
 // for the scheme "nats".
 // The default URL opener will connect to a default server based on the
 // environment variable "NATS_SERVER_URL".
+//
+// For servers that support it (NATS Server 2.2.0 or later), messages could
```
vangent, line 27: nit: could -> can

pubsub/natspubsub/nats.go
```diff
@@ -23,6 +23,27 @@
...
+// This provides full support for non-Go clients. Versions prior to 2.2.0
+// uses gob.Encoder to encode the message headers and content, which limits
+// the subscribers only to Go clients.
+// To use this feature, set the query parameter "natsv2" in the URL.
+// If no value is provided, it assumes the value is true. Otherwise, the value
+// needs to be parsable as a boolean. For example:
+//   - nats://mysubject?natsv2
+//   - nats://mysubject?natsv2=true
+//
+// This feature could also be enabled by setting the UseV2 field in the
```
vangent, line 38: nit: could -> can

pubsub/natspubsub/nats.go
```diff
@@ -23,6 +23,27 @@
...
+// URLOpener.
+// If the server does not support this feature, any attempt to use it will
+// result in an error.
+// Using native NATS message headers and content is more efficient than using
+// gob.Encoder, and allows non-Go clients to subscribe to the topic and
+// receive messages. It is recommended to use this feature if only the server
```
vangent, line 44: nit: drop "only"

pubsub/natspubsub/nats.go
```diff
@@ -93,12 +117,15 @@ func init() {
 // defaultDialer dials a default NATS server based on the environment
 // variable "NATS_SERVER_URL".
 type defaultDialer struct {
-	init   sync.Once
-	opener *URLOpener
-	err    error
+	init sync.Once
+	conn *nats.Conn
```
vangent, line 121: Do we need this? I see you setting it on line 140, but I don't think it's ever used...?

vangent (changes requested):

pubsub/natspubsub/nats.go
kucjac: Done.

pubsub/natspubsub/nats.go
kucjac: Done.

pubsub/natspubsub/nats.go
kucjac: Done.

pubsub/natspubsub/nats.go
kucjac: That's true, the `conn` is used only during the `sync.Once.Do` and it is passed to both url openeres, so there is no need to keep it in the `defaultDialer`.

kucjac:
Thank you, @vangent for your review. I've already applied requested changes.

pubsub/natspubsub/nats.go
vangent: I don't see this change in the commit?

pubsub/natspubsub/nats.go
vangent: same

pubsub/natspubsub/nats.go
kucjac: Oh, I somehow didn't include it in the previous commit. Done.

pubsub/natspubsub/nats.go
kucjac: Same as above. Done.

vangent (approved):

vangent:
Thanks for the contribution!

PR #2811, opened 2020-06-05
secrets/awskms: support ARNs in URLs by allowing empty Host

vangent:
For example, "awskms:///my:arn" now works.

Fixes #2808

Issue #2808, opened 2020-06-03
secrets/awskms: AWS KMS keys cannot be specified by ARN


csssuf:
### Describe the bug

While AWS allows referencing KMS keys by the ARN of the key or alias, doing so via go-cloud is not possible. This seems to be because go-cloud attempts to [parse the ARN as a URL](https://github.com/google/go-cloud/blob/8ac135d7ab6bf0f1787c7d21bc3b4a2613478b16/internal/openurl/openurl.go#L57), which isn't possible, since ARNs contain multiple colons and are not URLs.

### To Reproduce

```go
package main

import (
	"context"
	"fmt"

	"gocloud.dev/secrets"
	_ "gocloud.dev/secrets/awskms"
)

func tryUrl(s string) {
	ctx := context.Background()
	plaintext := []byte("test plaintext")

	keeper, err := secrets.OpenKeeper(ctx, s)
	if err != nil {
		fmt.Println("ERR opening:", s, err)
		return
	}

	cipherText, err := keeper.Encrypt(ctx, plaintext)
	if err != nil {
		fmt.Println("ERR encrypting:", err)
		return
	}

	fmt.Println("OK", s, cipherText)
}

func main() {
	tryUrl("awskms://alias/gocloud-test?region=us-east-1")
	tryUrl("awskms://arn:aws:kms:us-east-1:932528106278:alias/gocloud-test?region=us-east-1")
	tryUrl("awskms://8be0dcc5-da0a-4164-a99f-649015e344b5?region=us-east-1")
	tryUrl("awskms://arn:aws:kms:us-east-1:932528106278:key/8be0dcc5-da0a-4164-a99f-649015e344b5?region=us-east-1")
}
```

This results in:
```
[...]

PR #3320, opened 2023-10-11
secrets/awskms: update the example for the AWS ARN use case

h27771420:
As the title.

_AWS ARN_ can't be parsed by `url.Parse()` directly, because there are couple `:` in it. (e.g., `"awskms://arn:aws:kms:us-east-1:932528106278:key/8be0dcc5-da0a-4164-a99f-649015e344b5?region=us-east-1"`)
Then I fell into that trap💀, so here is a PR to avoid further victims.

The issue has been solved, but the use case isn't updated. 
1. issue: https://github.com/google/go-cloud/issues/2808
2. fix: https://github.com/google/go-cloud/pull/2811


vangent:
Please run 

internal/website/gatherexamples/run.sh > internal/website/data/examples.json

and add the commit.

vangent:
Thanks for the fix!

PR #3340, opened 2023-11-21
blob/s3blob: Support S3 server side encryption headers for Write and Copy

### LINKED ISSUE #3337: blob/s3: Support Server Side encryption settings in bucket url

I am attempting to use go-cloud  with an S3 bucket that has a bucket policy that enforces Server Side Encryption and a specific KMS key for uploads. 

I am unable to use go-cloud to do this in the context of third party tool, since it does not currently implement support for S3 specific settings using `As` and `BeforeWrite`. See also https://github.com/pulumi/pulumi/pull/14476

Allow configuring of server side encryption settings that are added to the S3 headers via new query parameters in the bucket URL. This would be similar to #3008

In https://github.com/pulumi/pulumi/pull/14476, I implemented support using `As` and `Before`,  but the maintainer preference if this is supported directly in go-cloud via url params

### Additional context
See also https://github.com/google/go-cloud/issues/2802. I am mainly checking to see if there is appetite to support this sort of feature now (I am happy to submit the contribution) since it seems like a couple of third party tools have not added support for this, and it would be handy if it was handled automatically by go-cloud

tristan-newmann:
Adds support to s3blob for setting AWS S3 server side encryption headers when making requests that require such headers. The additional settings can be specified with the `ssetype` and `kmskeyid` URL params, similar to the other configuration settings

Fixes #3337

#### SDKv2
Bucket policy enforces KMS headers - none provided (effectively the issue described in #3337)
Expect: `AccessDenied` failure
image

Bucket policy enforces KMS headers - values provided
Expect: No errors. Object is uploaded
image

Bucket policy doesn't enforce KMS - no values provided 
Expect: No errors. Object is uploaded
image

#### SDKv1
Bucket policy enforces KMS headers - none provided (effectively the issue described in #3337)
Expect: `AccessDenied` failure
image

Bucket policy enforces KMS headers - values provided
Expect: No errors. Object is uploaded
image

Bucket policy doesn't enforce KMS - no values provided 
Expect: No errors. Object is uploaded
image

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -103,6 +103,8 @@ func ConfigFromURLParams(q url.Values) (*aws.Config, error) {
 			}
 			cfg.S3ForcePathStyle = aws.Bool(b)
 		case "awssdk":
+		case "ssetype":
+		case "kmskeyid":
 			// ignore, should be handled before this
 		default:
 			return nil, fmt.Errorf("unknown query parameter %q", param)
@@ -198,6 +200,8 @@ func V2ConfigFromURLParams(ctx context.Context, q url.Values) (awsv2.Config, err
 		case "profile":
 			opts = append(opts, awsv2cfg.WithSharedConfigProfile(value))
 		case "awssdk":
+		case "ssetype":
+		case "kmskeyid":
 			// ignore, should be handled before this
 		default:
 			return awsv2.Config{}, fmt.Errorf("unknown query parameter %q", param)
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -103,8 +103,6 @@ func ConfigFromURLParams(q url.Values) (*aws.Config, error) {
 			}
 			cfg.S3ForcePathStyle = aws.Bool(b)
 		case "awssdk":
-		case "ssetype":
-		case "kmskeyid":
 			// ignore, should be handled before this
 		default:
 			return nil, fmt.Errorf("unknown query parameter %q", param)
@@ -200,8 +198,6 @@ func V2ConfigFromURLParams(ctx context.Context, q url.Values) (awsv2.Config, err
 		case "profile":
 			opts = append(opts, awsv2cfg.WithSharedConfigProfile(value))
 		case "awssdk":
-		case "ssetype":
-		case "kmskeyid":
 			// ignore, should be handled before this
 		default:
 			return awsv2.Config{}, fmt.Errorf("unknown query parameter %q", param)
--- a/blob/s3blob/s3blob.go
+++ b/blob/s3blob/s3blob.go
@@ -154,7 +154,7 @@ func (o *URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket
 		clientV2 := s3v2.NewFromConfig(cfg)
 
 		o.Options.EncryptionType = u.Query().Get("ssetype")
-		o.Options.KMSEncryptionId = u.Query().Get("kmskeyid")
+		o.Options.KMSEncryptionID = u.Query().Get("kmskeyid")
 
 		return OpenBucketV2(ctx, clientV2, u.Host, &o.Options)
 	}
@@ -168,7 +168,7 @@ func (o *URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket
 	configProvider.Configs = append(configProvider.Configs, overrideCfg)
 
 	o.Options.EncryptionType = u.Query().Get("ssetype")
-	o.Options.KMSEncryptionId = u.Query().Get("kmskeyid")
+	o.Options.KMSEncryptionID = u.Query().Get("kmskeyid")
 
 	return OpenBucket(ctx, configProvider, u.Host, &o.Options)
 }
@@ -180,9 +180,15 @@ type Options struct {
 	// ListObjectsV2.
 	UseLegacyList bool
 
+	// EncryptionType sets the encryption type headers when making write or
+	// copy calls. This is required if the bucket has a restrictive bucket
+	// policy that enforces a specific encryption type
 	EncryptionType string
 
-	KMSEncryptionId string
+	// KMSEncryptionID sets the kms key id header for write or copy calls.
+	// This is required when a bucket policy enforces the use of a specific
+	// KMS key for uploads
+	KMSEncryptionID string
 }
 
 // openBucket returns an S3 Bucket.
@@ -210,7 +216,7 @@ func openBucket(ctx context.Context, useV2 bool, sess client.ConfigProvider, cli
 		client:         client,
 		clientV2:       clientV2,
 		useLegacyList:  opts.UseLegacyList,
-		kmsKeyId:       opts.KMSEncryptionId,
+		kmsKeyId:       opts.KMSEncryptionID,
 		encryptionType: opts.EncryptionType,
 	}, nil
 }
--- a/blob/s3blob/s3blob.go
+++ b/blob/s3blob/s3blob.go
@@ -64,13 +64,6 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"io"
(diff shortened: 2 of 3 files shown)


aws/aws.go
```diff
@@ -103,6 +103,8 @@ func ConfigFromURLParams(q url.Values) (*aws.Config, error) {
 			}
 			cfg.S3ForcePathStyle = aws.Bool(b)
 		case "awssdk":
+		case "ssetype":
```
vangent, line 106: These shouldn't be needed here. The caller should handle and remove any known query parameters before calling this function.

Can you revert edits from this file, and then edit s3blob.go to do that? https://github.com/google/go-cloud/blob/6d5d2890816abc88cd11ef13d6cd6506eb5b5537/runtimevar/awsparamstore/awsparamstore.go#L142 is an example.

blob/s3blob/s3blob.go
```diff
@@ -171,6 +179,16 @@ type Options struct {
 	// Some S3-compatible services (like CEPH) do not currently support
 	// ListObjectsV2.
 	UseLegacyList bool
+
+	// EncryptionType sets the encryption type headers when making write or
+	// copy calls. This is required if the bucket has a restrictive bucket
+	// policy that enforces a specific encryption type
+	EncryptionType string
```
vangent, line 186: Let's make this the actual type, and do the conversion from a string in the URL handler. Ideally it would return an error if it's an invalid string value at that time.

blob/s3blob/s3blob.go
```diff
@@ -973,6 +996,12 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key string, contentType str
 		if len(opts.ContentMD5) > 0 {
 			reqV2.ContentMD5 = aws.String(base64.StdEncoding.EncodeToString(opts.ContentMD5))
 		}
+		if b.encryptionType != "" {
+			reqV2.ServerSideEncryption = typesv2.ServerSideEncryption(b.encryptionType)
```
vangent, line 1000: As noted above, let's get rid of the conversion here.

blob/s3blob/s3blob_test.go
```diff
@@ -466,6 +466,8 @@ func TestOpenBucketFromURL(t *testing.T) {
 		{"s3://mybucket?profile=main&region=us-west-1", false},
 		// OK, use V2.
 		{"s3://mybucket?awssdk=v2", false},
+		// OK, use KMS Server Side Encryption
```
vangent, line 469: If possible, add a test for an invalid encryptionType value.

vangent (changes requested):

tristan-newmann:
@vangent Thanks for the feedback! I implemented the requested changes. I left a couple unresolved since I wasn't sure if I did it exactly how you had in mind. This is ready for re-review

Let me know if I can force push to fix the CLA check, one of the commits has a slightly incorrect version of my email due to a local misconfiguration

vangent:
Please go ahead and force-push and I'll re-review.

tristan-newmann:
Ready for re-review

blob/s3blob/s3blob.go
```diff
@@ -64,13 +64,6 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"io"
```
vangent, line 67: Can you revert this header change? I don't think it is necessary or related.

blob/s3blob/s3blob.go
```diff
@@ -144,24 +143,58 @@ type URLOpener struct {
 	Options Options
 }
 
+const (
+	sseTypeParamKey  = "ssetype"
+	kmsKeyIdParamKey = "kmskeyid"
+)
+
+func IsValidServerSideEncryptionType(value string) bool {
```
vangent, line 151: This doesn't need to be exported -> isValidServerSideEncryptionType

blob/s3blob/s3blob.go
```diff
@@ -144,24 +143,58 @@ type URLOpener struct {
...
+	// Surely there is a better way to get the values
```
vangent, line 152: Nope, please remove :-).

blob/s3blob/s3blob.go
vangent: Actually let's change it to "toServerSideEncryptionType" and have it return (typesv2.ServerSideEncryption, error).

blob/s3blob/s3blob.go
```diff
@@ -144,24 +143,58 @@ type URLOpener struct {
...
+	for _, v := range typesv2.ServerSideEncryptionAes256.Values() {
+		if string(v) == value {
+			return true
```
vangent, line 155: Here you can return v.

blob/s3blob/s3blob.go
```diff
@@ -144,24 +143,58 @@ type URLOpener struct {
...
+		}
+	}
+	return false
```
vangent, line 158: Here you can return "", fmt.Errorf(...)

blob/s3blob/s3blob.go
```diff
@@ -144,24 +143,58 @@ type URLOpener struct {
...
+
 // OpenBucketURL opens a blob.Bucket based on u.
 func (o *URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket, error) {
+	q := u.Query()
+
+	if sseType := q.Get(sseTypeParamKey); sseType != "" {
+		q.Del(sseTypeParamKey)
+
+		if !IsValidServerSideEncryptionType(sseType) {
+			return nil, fmt.Errorf("ssetype invalid %v", sseType)
```
vangent, line 169: Nit: use the constant here.

blob/s3blob/s3blob.go
```diff
@@ -144,24 +143,58 @@ type URLOpener struct {
...
 
+const (
+	sseTypeParamKey  = "ssetype"
+	kmsKeyIdParamKey = "kmskeyid"
+)
+
+func IsValidServerSideEncryptionType(value string) bool {
+	// Surely there is a better way to get the values
+	for _, v := range typesv2.ServerSideEncryptionAes256.Values() {
+		if string(v) == value {
```
vangent, line 154: Optional: Make the comparison case insensitive using strings.ToLower on both sides.

tristan-newmann:
@vangent Sorry for the delay on this one, priorities forced my to put this aside for a moment, but I am back now. This is ready for re-review, and I re-did the smoke testing as well.

# Unit tests pass
image

# Redoing the smoke testing after the adjustments:
## SDKv2
### Bucket policy enforces KMS headers - none provided (effectively the issue described in https://github.com/google/go-cloud/issues/3337)
Expect: AccessDenied failure
image

### Bucket policy enforces KMS headers - values provided
Expect: No errors. Object is uploaded
image
image

### Bucket policy doesn't enforce KMS - no values provided
Expect: No errors. Object is uploaded
image

## SDKv1
### Bucket policy enforces KMS headers - none provided (effectively the issue described in https://github.com/google/go-cloud/issues/3337)
Expect: AccessDenied failure
image

### Bucket policy enforces KMS headers - values provided
Expect: No errors. Object is uploaded
image

### Bucket policy doesn't enforce KMS - no values provided
Expect: No errors. Object is uploaded
image

blob/s3blob/s3blob.go
```diff
@@ -144,24 +144,58 @@ type URLOpener struct {
...
+
+		sseType, err := toServerSideEncryptionType(sseTypeParam)
+		if err != nil {
+			return nil, err
+		}
+
+		o.Options.EncryptionType = sseType
+	}
+
+	if kmsKeyID := u.Query().Get(kmsKeyIdParamKey); kmsKeyID != "" {
```
vangent, line 176: use q here.

vangent (changes requested):

vangent (approved):

Issue #3339, opened 2023-11-17
docstore/gcpfirestore: ought to support non-default databases


coryschwartz:
### Is your feature request related to a problem? Please describe.

I want to use go-cloud doctore for production and staging environments, but this is difficult to do because the docstore only supports the default database.

### Describe the solution you'd like

Ideally, we should be able to use go-cloud with any database in firestore.

```
gcloud firestore database create --region [region] --database [whatever]
```

And then create the collection in go-cloud

```
import "gocloud.dev/docstore/gcpfirestore"

client, cleanup, err := gcpfirestore.Dial(cix, token)
// This ought to work.
gcpfirestore.OpenCollection(client, "project/databases/whatever/documents/collection", "index", nil)

```

Currently, this does not work when the default database is not used.
See the regex which is checked here.
https://github.com/google/go-cloud/blob/v0.34.0/docstore/gcpfirestore/fs.go#L196

### Describe alternatives you've considered

Currently, I'm using different GCP projects to differentiate between environments, so I'm able to avoid the restriction this way.

vangent:
Is relaxing that regex to replace "default" with ".*" all that's needed to fix this?

coryschwartz:
I suspect so -- I'll give this a try. I wasn't sure if there was some reason I hadn't thought of this regex is there.

coryschwartz:
I created a PR with a changed regex.

It seems like it works, tested with this little snippet.

```
package main

import (
	"context"

	"gocloud.dev/docstore"
	_ "gocloud.dev/docstore/gcpfirestore"
)

const (
	// url = "firestore://projects/XXXX/databases/(default)/documents/my-collection?name_field=userID"
	url = "firestore://projects/XXXX/databases/gocloudtest/documents/my-collection?name_field=userID"
)

type doc struct {
	UserID string
}

func main() {
	ctx := context.Background()
	coll, err := docstore.OpenCollection(ctx, url)
	if err != nil {
		panic(err)
	}
	coll.Create(ctx, &doc{UserID: "test"})
}
```

and then just peering into the Firestore web console to make sure those are created.

I lightly updated the test to make it not fail.

I wonder if you can just remove this regex entirely? You chance sending a bogus API call and would get an API error, but that seems reasonable in my view.

vangent:
> I wonder if you can just remove this regex entirely?

I don't think we can, we use the regex to pull out the database path: https://github.com/google/go-cloud/blob/master/docstore/gcpfirestore/fs.go#L207

PR #3345, opened 2023-11-22
docstore/gcpfilestore: add support for non-default databases.

coryschwartz:
Fixes: https://github.com/google/go-cloud/issues/3339

This edits the regex to allow non-default database names when creating Firestore collections.

In line with prior art, this also creates a helper function to assist with creating a valid scheme.

### DIFF
--- a/docstore/gcpfirestore/fs.go
+++ b/docstore/gcpfirestore/fs.go
@@ -149,6 +149,10 @@ func CollectionResourceID(projectID, collPath string) string {
 	return fmt.Sprintf("projects/%s/databases/(default)/documents/%s", projectID, collPath)
 }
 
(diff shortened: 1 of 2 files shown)


docstore/gcpfirestore/fs.go
```diff
@@ -149,6 +149,10 @@ func CollectionResourceID(projectID, collPath string) string {
 	return fmt.Sprintf("projects/%s/databases/(default)/documents/%s", projectID, collPath)
 }
 
+func CollectionResourceIDWithDatabase(projectID, databaseID, collPath string) string {
```
vangent, line 152: Please add a docstring for this exported function.

vangent (changes requested):

PR #3348, opened 2023-11-25
mysql: pass TLS config directly to MySQL's config

giautm:


mysql/gcpmysql/gcpmysql.go
```diff
@@ -98,11 +100,10 @@ func (uo *URLOpener) OpenMySQLURL(ctx context.Context, u *url.URL) (*sql.DB, err
 		return nil, fmt.Errorf("gcpmysql: URLOpener CertSource is nil")
 	}
 	// TODO(light): Avoid global registry once https://github.com/go-sql-driver/mysql/issues/771 is fixed.
-	dialerCounter.mu.Lock()
-	dialerNum := dialerCounter.n
-	dialerCounter.mu.Unlock()
```
giautm, line 103: No increment,

giautm:
Hello @vangent, can you please review? thanks!

mysql/gcpmysql/gcpmysql.go
```diff
@@ -97,12 +98,8 @@ func (uo *URLOpener) OpenMySQLURL(ctx context.Context, u *url.URL) (*sql.DB, err
 	if uo.CertSource == nil {
 		return nil, fmt.Errorf("gcpmysql: URLOpener CertSource is nil")
 	}
-	// TODO(light): Avoid global registry once https://github.com/go-sql-driver/mysql/issues/771 is fixed.
```
vangent, line 100: Can you explain why this TODO can be addressed while the issue referenced is still open?

giautm: For this part, it really doesnt have another solution, and the package itself will not support to change this to a custom dialer function. So, I think the current implement is correct and no need to change in future. The package sql-cloud-proxy also used this way without comment

vangent: It would help if you would explain what problem you are trying to solve. Does the current code not work? Or are you trying to simplify it?

Our testing for this part of the codebase is not good, and I am hesitant to make changes that I don't fully understand without any rationale, especially when there's a TODO in the code that hasn't been addressed.

vangent: It would help if you would explain what problem you are trying to solve. Does the current code not work? Or are you trying to simplify it?

giautm: Simplify, and fix one bug.

giautm:
Hello, please let me know if you need anything to review/merge this PR.

Issue #3352, opened 2023-12-10
docstore:  should support (in) comparison operator


coryschwartz:
The query operators supported by go-cloud is a more restrictive set than any of its drivers.

In my application, I want to query for documents with a key that matches any of several keys, and I am surprised that I cannot do this in a single call. I want to do something similar to this:

SELECT * WHERE KEY IN ('val1', 'val2'...)

I can think of a couple of ways this could be done:

I can add additional `.Where` clauses, but this has the effect of AND'ing the clauses together, so I cannot use multiple Where's to select for multiple keys with the existing operators. The docstring tells us drivers should AND these together. https://github.com/google/go-cloud/blob/master/docstore/driver/driver.go#L192

One solution, although I suspect this isn't the most efficient way to do this, there could be an `Or`. If there was an or operator, then we could do this.

```
iter := coll.Query().Or("Key", "=", "val1").Or("Key", "=", "val2")
```

Maybe a better way to do this is just to support additional operators. Personally, this would be my preferred solution. 
currently, only the following are supported: https://github.com/google/go-cloud/blob/master/docstore/query.go#L65

```
var validOp = map[string]bool{
	"=":  true,
	">":  true,
	"<":  true,
	">=": true,
	"<=": true,
}
```
We could simply throw an 'in' operator into this list, and then adjust the individual drivers.

When I look at the query operators supported by each of the supported drivers, they all have a richer set of operators and they all include the 'in' operator I'm hoping to use.

https://www.mongodb.com/docs/manual/reference/operator/query/#std-label-query-projection-operators-top
https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/query/keywords#in
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.OperatorsAndFunctions.html#Expressions.OperatorsAndFunctions.Syntax
https://firebase.google.com/docs/firestore/query-data/queries#in_not-in_and_array-contains-any

I can think of at least one reason why you *shouldn't* do this request. If you're going to support adapters any less-featured docstores in the future, you might want to keep the supported features to a minimum subset.

If you'd accept it, I'd be happy to attempt a PR to add an 'in' operator to the validOp and adjust the drivers, although I don't have easy access to all of the supported databases to properly test every driver.

jba:
You should be looking at [Firestore](https://cloud.google.com/firestore/docs/query-data/queries#in_and_array-contains-any) rather than Firebase. But Firestore has `in` too.

[...]

PR #3358, opened 2023-12-15
feat/docstore in query operation

coryschwartz:
fixes: https://github.com/google/go-cloud/issues/3352

This PR updates the docstore to allow in/not-in queries.

This isn't yet merge-able, I need some assistance to get this into shape.

Test replay data does not exist for this type of query, so tests that rely on that don't work currently. Am I able to generate those myself somehow?

I haven't tested the dynamodb changes at all.

I updated the local mongodb test, but haven't tried this against cosmo.

I tested firstore functionality with this snipit, and it seems to function, but the real tests continue to fail due to the missing replays. https://gist.github.com/coryschwartz/80c37f15f58ddf0d8c979073f30d6ccf

### DIFF
--- a/docstore/awsdynamodb/query.go
+++ b/docstore/awsdynamodb/query.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"reflect"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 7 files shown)


vangent:
Thanks for this! I don't think you can generate the replay files. I'll either have to figure out how to get access to this PR so I can run it locally and generate them, then update the PR, or clone the PR entirely. It might take a bit as this is a larger effort than I have time for right now.

vangent:
The code changes look fine to me, so I cloned your PR and updated the various .replay files (https://github.com/google/go-cloud/pull/3372).

vangent:
(The drivertest you added passes for all the providers, so LGTM!)

vangent:
Thanks for the contribution!

PR #3371, opened 2023-12-21
blob/all: Allow disabling of ContentType auto-detection during writes

### LINKED ISSUE #3298: blob/gcsblob: Allow blank Content-Type to be set
### Describe the bug

Google Cloud Storage allows a client to request a blob with a different `Content-Type` via the `response-content-type` query parameter *only if* the blob's `Content-Type` metadata has NOT already been set.

This differs from Amazon S3, which allows `response-content-type` to be requested regardless of the blob metadata. 

With Go Cloud, It's currently not possible to keep the `Content-Type` value blank for GCS. https://github.com/google/go-cloud/blob/v0.30.0/internal/docs/design.md#portable-types-and-drivers says:

> Consider the Bucket.NewWriter method, which infers the content type of the blob based on the first bytes written to it. 

That seems to be happening here: https://github.com/google/go-cloud/blob/d2d5bedb50683e2a6b893b75aafc193eca2715db/blob/blob.go#L1109-L1121

### To Reproduce

1. Create an object via `Bucket.NewWriter` with `Content-Type` blank in `WriterOptions` (https://pkg.go.dev/github.com/google/go-cloud/blob#WriterOptions).
2. Observe in the bucket that the file has a `Content-Type` already set.

## Expected behavior

The metadata should remain blank so that clients can override the type on request.

### Version

v0.30.0, but this has been true for a while.

vangent:
Fixes #3298.


stanhu:
@vangent Thanks so much for this!

PR #3381, opened 2024-02-23
blob/gcsblob: Ensure driver sets Content-Type auto-detection properly

stanhu:
https://github.com/google/go-cloud/pull/3371 made it possible to disable Go Cloud's Content-Type auto-detection via the
`DisableContentTypeDetection` option. However, the Google Cloud Storage (GCS) driver still performed auto-detection even if this option were enabled, so it was previously not possible to keep `Content-Type` blank. This commit adds the
`DisableContentTypeDetection` option to the the driver and passes along the value to make it possible to keep Content-Type blank in the GCS object metadata. This enables the use of the `Response-Content-Type` override in a signed URL.

This commit needs https://github.com/googleapis/google-cloud-go/pull/9431 to work. `go.mod` has been updated to use pseudo-version until cloud.google.com/go/storage v1.39 is available (https://github.com/googleapis/google-cloud-go/pull/9457).

PR #3382, opened 2024-02-23
all: update deps for gcsblob

stanhu:
This is needed in preparation for #3381.

https://github.com/googleapis/google-cloud-go/pull/9211 introduced the `includeFoldersAsPrefixes=false` query string in the URLs.


vangent:
I would rather not use an unreleased version, can we wait until there's a real one? Since I may have to update the test goldens anyway (sometimes that happens, sometimes not), you can just ping me when it happens on the other PR.

stanhu:
@vangent This is actually a released version. I was just updating to the latest version so that the upgrade to 1.39.0 would be easier.

PR #3385, opened 2024-02-26
docstore/all: add Offset method implementation

bartventer:
This pull request adds an `Offset` method to the docstore package for all providers. This method allows users to skip a specified number of documents in the results of a query, which is useful for implementing pagination in applications that use docstore. This complements the existing `Limit` method.

I've tested this implementation across all providers and generated new replay files for `awsdynamodb` and `gcpfirestore`.


vangent (approved):
Thanks for the contribution!

PR #3386, opened 2024-02-27
pubsub: Ensure batcher flushes on shutdown, even if min batch size isn't met

tonyhb:
This PR ensures that the batcher flushes on shutdown, even if the pending length is less than the min batch size specified.  Sending events is preferred to dropping, even if limits are not obeyed.

Related to #3383, but doesn't necessarily replace.

### DIFF
--- a/pubsub/batcher/batcher.go
+++ b/pubsub/batcher/batcher.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"reflect"
 	"sync"
+	"sync/atomic"
 )
 
 // Split determines how to split n (representing n items) into batches based on
@@ -70,7 +71,7 @@ type Batcher struct {
 
 	mu        sync.Mutex
 	pending   []waiter // items waiting to be handled
-	nHandlers int      // number of currently running handler goroutines
+	nHandlers int32    // number of currently running handler goroutines
 	shutdown  bool
 }
 
@@ -197,7 +198,7 @@ func (b *Batcher) AddNoWait(item interface{}) <-chan error {
 
 	// Add the item to the pending list.
 	b.pending = append(b.pending, waiter{item, c})
-	if b.nHandlers < b.opts.MaxHandlers {
+	if atomic.LoadInt32(&b.nHandlers) < int32(b.opts.MaxHandlers) {
 		// If we can start a handler, do so with the item just added and any others that are pending.
 		batch := b.nextBatch()
 		b.handleBatch(batch)
@@ -217,7 +218,7 @@ func (b *Batcher) handleBatch(batch []waiter) {
 		b.callHandler(batch)
 		b.wg.Done()
 	}()
-	b.nHandlers++
+	atomic.AddInt32(&b.nHandlers, 1)
 }
 
 // nextBatch returns the batch to process, and updates b.pending.
@@ -284,7 +285,7 @@ func (b *Batcher) callHandler(batch []waiter) {
 		// always get to run.
 		batch = b.nextBatch()
 		if batch == nil {
-			b.nHandlers--
+			atomic.AddInt32(&b.nHandlers, -1)
 		}
 		b.mu.Unlock()
 	}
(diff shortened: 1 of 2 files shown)


vangent:
The tests are failing with a data race:

WARNING: DATA RACE
Write at 0x00c00044e3f0 by goroutine 978:
  gocloud.dev/pubsub/batcher.(*Batcher).callHandler()
      /home/runner/work/go-cloud/go-cloud/pubsub/batcher/batcher.go:287 +0x1ca
  gocloud.dev/pubsub/batcher.(*Batcher).handleBatch.func1()
      /home/runner/work/go-cloud/go-cloud/pubsub/batcher/batcher.go:217 +0x54

Previous read at 0x00c00044e3f0 by goroutine 979:
  gocloud.dev/pubsub/batcher.(*Batcher).Shutdown()
      /home/runner/work/go-cloud/go-cloud/pubsub/batcher/batcher.go:303 +0x75
  gocloud.dev/pubsub.(*Subscription).Shutdown.func2()
      /home/runner/work/go-cloud/go-cloud/pubsub/pubsub.go:697 +0xb2

Goroutine 978 (running) created at:
  gocloud.dev/pubsub/batcher.(*Batcher).handleBatch()
      /home/runner/work/go-cloud/go-cloud/pubsub/batcher/batcher.go:216 +0x11d
  gocloud.dev/pubsub/batcher.(*Batcher).AddNoWait()
      /home/runner/work/go-cloud/go-cloud/pubsub/batcher/batcher.go:203 +0x3b1
  gocloud.dev/pubsub.(*Subscription).Receive.func3()
      /home/runner/work/go-cloud/go-cloud/pubsub/pubsub.go:608 +0xcc
  gocloud.dev/pubsub.(*Message).Ack()
      /home/runner/work/go-cloud/go-cloud/pubsub/pubsub.go:161 +0xe1
  gocloud.dev/pubsub/natspubsub.TestInteropWithDirectNATS.deferwrap4()
      /home/runner/work/go-cloud/go-cloud/pubsub/natspubsub/nats_test.go:259 +0x33
  runtime.deferreturn()
      /opt/hostedtoolcache/go/1.22.0/x64/src/runtime/panic.go:602 +0x5d
  testing.tRunner()
      /opt/hostedtoolcache/go/1.22.0/x64/src/testing/testing.go:1689 +0x21e
  testing.(*T).Run.gowrap1()
      /opt/hostedtoolcache/go/1.22.0/x64/src/testing/testing.go:1742 +0x44

tonyhb:
Ah, thanks.  I ran without -race 🤦.  Fixing!

tonyhb:
@vangent fixed, thanks for the notification.

pubsub/batcher/batcher.go
```diff
@@ -283,5 +298,13 @@ func (b *Batcher) Shutdown() {
 	b.mu.Lock()
 	b.shutdown = true
 	b.mu.Unlock()
+
+	// On shutdown, ensure that we attempt to flush any pending items
+	// if there's a minimum batch size.
+	if atomic.LoadInt32(&b.nHandlers) < int32(b.opts.MaxHandlers) {
```
vangent, line 304: It shouldn't be necessary to use `atomic` to check/modify nHandlers. There's a mutex on the struct to protect it. IIUC, the problem is that you added a read on this line outside of the lock. I think if you move the b.mu.Unlock a few lines up to below this new stanza, it will be fine.

vangent, line 304: What happens if nHandlers == MaxHandlers? Won't we drop some messages then?

vangent (changes requested):

pubsub/batcher/batcher.go
tonyhb: At this point, `nextBatch` in the handlers call will return the remaining items as `shutdown` is set to true, so everything will be processed as expected.

vangent: If nHandlers == MaxHandlers, `nextBatch` won't even be called...?

vangent: It's not obvious to me that we should be checking `nHandlers` here at all. Seems like during shutdown we need to choose between possibly creating more than MaxHandlers handlers, or dropping messages.

vangent:
Ping?

vangent:
I patched this with changes in https://github.com/google/go-cloud/pull/3543.

Issue #3393, opened 2024-03-02
docstore/mongodocstore: Error update with only unset or increment modifier (recurring issue)


bartventer:
### Describe the bug

I'm encountering the same issue as [#2979](https://github.com/google/go-cloud/issues/2979), which was previously closed. When the update function is called with only `unset` or `inc` modifiers (i.e., no `set` operations included), I get an error saying `Modifiers operate on fields but we found type null instead. For example: {$mod: {<field>: ...}} not {$set: null} (code=Unknown)`.

### To Reproduce

```go
import (
    "context"
    "fmt"

    "gocloud.dev/docstore"
    _ "gocloud.dev/docstore/mongodocstore"
)

func main() {
    type Player struct {
        Name  string `docstore:"Name"`
        Score int    `docstore:"Score"`
    }

    // Open a collection.
    ctx, cancel := context.WithCancel(context.Background())
    defer cancel()
    coll, err := docstore.OpenCollection(ctx, "mongo://testdb/my-collection?id_field=Name")
    if err != nil {
        fmt.Println(err)
    }

    // Create a player.
    pat := &Player{Name: "Pat", Score: 0}
    if err := coll.Create(ctx, pat); err != nil {
        panic(err)
    }

    // Increment the score or unset the score (i.e. no set modifier).
    pat2 := &Player{Name: "Pat"}
    err = coll.Actions().Update(pat, docstore.Mods{"Score": docstore.Increment(5)}).Get(pat2).Do(ctx)
    if err != nil {
        panic(err)
    }

}
```

## Expected behavior

I expect the Update function to successfully increment or unset the Score field without requiring a set operation. The unset and inc modifiers should work independently of the set modifier.

### Version

This issue occurs with gocloud.dev/docstore/mongodocstore v0.36.0.

### Additional context

A current workaround for this issue is to include a set modifier in the update function.

The problem appears to be in the newUpdateDoc function where the length of sets is not checked. If there are no set operations, the $set field in the updateDoc map is set to an empty bson.D, which leads to the error. A potential fix could be to only add the $set field to updateDoc if len(sets) > 0.

PR #3394, opened 2024-03-02
fix(docstore/mongodocstore): check length of sets in newUpdateDoc function

bartventer:
This PR resolves issue [#3393](https://github.com/google/go-cloud/issues/3393). 

This update modifies the newUpdateDoc function in the docstore/mongodocstore package to check the length of the 'sets' variable before adding the '$set' field to the 'updateDoc' map. This prevents an error when calling the update function with only 'unset' or 'inc' modifiers and no 'set' modifiers.


vangent (approved):
Thanks for the repro and fix!

PR #3412, opened 2024-03-31
pubsub/awssnssqs: aws sqs expose receiver max batch

pxp928:
This PR exposes the receiver max batch size as a URL parameter for AWS SQS via `receivermaxbatch`.

For example: `awssqs://sqs.us-east-2.amazonaws.com/99999/my-queue?receivermaxbatch=5`

Based on the `recvBatcherOpts`: https://github.com/google/go-cloud/blob/be1b4aee38955e1b8cd1c46f8f47fb6f9d820a9b/pubsub/awssnssqs/awssnssqs.go#L118-L123
and the limitations of SQS, any value above 10 would default back to 10.

### DIFF
--- a/pubsub/awssnssqs/awssnssqs.go
+++ b/pubsub/awssnssqs/awssnssqs.go
@@ -305,13 +305,16 @@ func (o *URLOpener) OpenSubscriptionURL(ctx context.Context, u *url.URL) (*pubsu
 		}
 		q.Del("waittime")
 	}
+	// SQS supports receiving at most 10 messages at a time:
+	// https://godoc.org/github.com/aws/aws-sdk-go/service/sqs#SQS.ReceiveMessage
+	// if value is higher than 10, it will default to 10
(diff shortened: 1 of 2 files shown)


pxp928:
oh, interesting I did not see that in https://gocloud.dev/howto/pubsub/subscribe/. Is there any documentation around this behavior? Is there a way to know what the current batch size is set to or changes to via logs? Thank You for pointing this out.

vangent:
I don't think it's well-documented, as it is not really part of the public interface, it's internal implementation detail. 

You can see constants for the algorithm here:
https://github.com/google/go-cloud/blob/master/pubsub/pubsub.go#L397

and the main code is here:
https://github.com/google/go-cloud/blob/master/pubsub/pubsub.go#L462

No, the batch size isn't currently logged, but you can patch a local copy and add some logging.

pxp928:
Thanks @vangent for the information!

PR #3429, opened 2024-04-30
docstore/mongodocstore: Update Mongo dialer when MONGO_SERVER_URL rotates

concaf:
Prior to this commit, the dialer for MongoDB was generated once from MONGO_SERVER_URL environment variable but was never updated even when the environment variable was updated in subsequent calls. While this works fine when MONGO_SERVER_URL is not expected to update, but as MONGO_SERVER_URL also contains the credentials to connect to MongoDB, it's a fairly common use case to rotate these credentials (and hence the environment variable) at regular intervals.

This commit fixes that and updates the dialer when MONGO_SERVER_URL is updated.

This PR blocks https://github.com/tektoncd/chains/issues/1089

### DIFF
--- a/docstore/mongodocstore/urls.go
+++ b/docstore/mongodocstore/urls.go
@@ -34,27 +34,31 @@ func init() {
 // defaultDialer dials a default Mongo server based on the environment variable
 // MONGO_SERVER_URL.
 type defaultDialer struct {
-	init   sync.Once
-	opener *URLOpener
-	err    error
+	mongoServerURL string
+	mu             sync.Mutex
+	opener         *URLOpener
+	err            error
 }
 
 func (o *defaultDialer) OpenCollectionURL(ctx context.Context, u *url.URL) (*docstore.Collection, error) {
-	o.init.Do(func() {
-		serverURL := os.Getenv("MONGO_SERVER_URL")
-		if serverURL == "" {
-			o.err = errors.New("MONGO_SERVER_URL environment variable is not set")
-			return
-		}
-		client, err := Dial(ctx, serverURL)
-		if err != nil {
-			o.err = fmt.Errorf("failed to dial default Mongo server at %q: %v", serverURL, err)
-			return
-		}
-		o.opener = &URLOpener{Client: client}
-	})
-	if o.err != nil {
+	o.mu.Lock()
+	defer o.mu.Unlock()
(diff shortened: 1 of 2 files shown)


docstore/mongodocstore/urls.go
```diff
@@ -34,27 +33,25 @@ func init() {
...
-	err    error
+	mongoServerURL string
+	opener         *URLOpener
+	err            error
 }
 
 func (o *defaultDialer) OpenCollectionURL(ctx context.Context, u *url.URL) (*docstore.Collection, error) {
-	o.init.Do(func() {
-		serverURL := os.Getenv("MONGO_SERVER_URL")
+	serverURL := os.Getenv("MONGO_SERVER_URL")
```
vangent, line 42: This is not thread-safe (OpenCollectionURL can be called concurrently on the same `o`).

You can protect it with a sync.Mutex.

docstore/mongodocstore/urls.go
```diff
@@ -34,27 +33,25 @@ func init() {
...
+	if serverURL != o.mongoServerURL {
```
vangent, line 43: This doesn't work. If MONGO_SERVER_URL isn't set, then both of these will be the empty string and we'll fall through without returning an error.

How about something like `if o.opener == nil || serverURL != o.openerServerURL` ... ?

vangent (changes requested):

docstore/mongodocstore/urls.go
concaf: makes sense, updated 👍🏼

docstore/mongodocstore/urls.go
concaf: good catch 👍🏼

i've updated this to check for an empty `MONGO_SERVER_URL` outside the if condition, so it should fail each time the env var is empty

```go
	if currentEnv == "" {
		o.err = errors.New("MONGO_SERVER_URL environment variable is not set")
		return nil, fmt.Errorf("open collection %s: %v", u, o.err)
	}

	// If MONGO_SERVER_URL has been updated, then update o.opener as well
	if currentEnv != o.mongoServerURL {
		client, err := Dial(ctx, currentEnv)
...
...
```

wdyt?

concaf:
@vangent thanks for the quick review 🚀 🙂 - i've updated the PR and also added some tests.

vangent (approved):

PR #3431, opened 2024-05-27
 pubsub/rabbitpubsub: add query string set the qos prefetch count

### LINKED ISSUE #3430: pubsub/rabbitpubsub: add support to qos prefetch count via url query string

Today we can't set the [QoS prefetch count](https://www.rabbitmq.com/docs/consumer-prefetch) when we open a subscription to a rabbitmq queue.

I check how kafka handle query strings and I add support to `prefetch_count=<prefetch count value>` that should be transparent when using a `rabbit://queue` url format like `rabbit://queue?prefetch_coint=1000` to set the prefetch count equal to 1000.

I find no alternatives, since to set the QoS we need to call a specific method on the amqp channel object. See https://pkg.go.dev/github.com/rabbitmq/amqp091-go#Channel.Qos

This development is not needed if we use the following configuration in the rabbitmq server:
https://www.rabbitmq.com/docs/consumer-prefetch#default-limit

### Additional context

Usuall we only set the consumer prefetch count, this is why I am ignoring the other parameters on the Qos method and setting default value.

peczenyj:
Fixes #3430

Since we usually want to set the prefetch count, I simplify the rabbit url to support a single query string

we can use `rabbit://myqueue` (continues backward compatible) or the new one `rabbit://myqueue?prefetch_count=<prefetch-count-value>`

I tested locally using a rabbitmq docker image and it seems to work properly.

### DIFF
--- a/pubsub/rabbitpubsub/rabbit.go
+++ b/pubsub/rabbitpubsub/rabbit.go
@@ -97,7 +97,9 @@ const Scheme = "rabbit"
 //
 // For subscriptions, the URL's host+path is used as the queue name.
 //
-// No query parameters are supported.
+// An optional query string can be used to set the Qos consumer prefetch on subscriptions
+// like "rabbit://myqueue?qos=1000" to set the consumer prefetch count to 1000
+// see also https://www.rabbitmq.com/docs/consumer-prefetch
 type URLOpener struct {
 	// Connection to use for communication with the server.
 	Connection *amqp.Connection
--- a/pubsub/rabbitpubsub/rabbit.go
+++ b/pubsub/rabbitpubsub/rabbit.go
@@ -98,7 +98,7 @@ const Scheme = "rabbit"
 // For subscriptions, the URL's host+path is used as the queue name.
 //
 // An optional query string can be used to set the Qos consumer prefetch on subscriptions
-// like "rabbit://myqueue?qos=1000" to set the consumer prefetch count to 1000
+// like "rabbit://myqueue?prefetch_count=1000" to set the consumer prefetch count to 1000
 // see also https://www.rabbitmq.com/docs/consumer-prefetch
 type URLOpener struct {
 	// Connection to use for communication with the server.
@@ -121,28 +121,27 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
 
 // OpenSubscriptionURL opens a pubsub.Subscription based on u.
 func (o *URLOpener) OpenSubscriptionURL(ctx context.Context, u *url.URL) (*pubsub.Subscription, error) {
+	opts := o.SubscriptionOptions
 	for param, value := range u.Query() {
 		switch param {
-		case "qos":
+		case "prefetch_count":
 			if len(value) == 0 {
 				return nil, fmt.Errorf("open subscription %v: invalid query parameter %q", u, param)
 			}
 			count := value[0]
-			if o.SubscriptionOptions.Qos == nil {
-				o.SubscriptionOptions.Qos = new(Qos)
-			}
+
 			prefetchCount, err := strconv.Atoi(count)
 			if err != nil {
 				return nil, fmt.Errorf("open subscription %v: invalid query parameter %q", u, count)
 			}
 
-			o.SubscriptionOptions.Qos.PrefetchCount = prefetchCount
+			opts.PrefetchCount = prefetchCount
 		default:
 			return nil, fmt.Errorf("open subscription %v: invalid query parameter %q", u, param)
 		}
 	}
 	queueName := path.Join(u.Host, u.Path)
-	return OpenSubscription(o.Connection, queueName, &o.SubscriptionOptions), nil
+	return OpenSubscription(o.Connection, queueName, &opts), nil
 }
 
 type topic struct {
@@ -163,12 +162,7 @@ type TopicOptions struct{}
 // SubscriptionOptions sets options for constructing a *pubsub.Subscription
 // backed by RabbitMQ.
 type SubscriptionOptions struct {
-	// Qos properties
-	Qos *Qos
-}
-
-// Qos options to be used when we create a subscription.
-type Qos struct {
+	// Qos property prefetch count.
 	PrefetchCount int
 }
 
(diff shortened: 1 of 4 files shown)


pubsub/rabbitpubsub/rabbit.go
```diff
@@ -118,8 +121,25 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
 
 // OpenSubscriptionURL opens a pubsub.Subscription based on u.
 func (o *URLOpener) OpenSubscriptionURL(ctx context.Context, u *url.URL) (*pubsub.Subscription, error) {
-	for param := range u.Query() {
-		return nil, fmt.Errorf("open subscription %v: invalid query parameter %q", u, param)
+	for param, value := range u.Query() {
+		switch param {
+		case "qos":
```
vangent, line 126: Shouldn't this be called "prefetch_count"?

pubsub/rabbitpubsub/rabbit.go
```diff
@@ -142,7 +162,15 @@ type TopicOptions struct{}
 
 // SubscriptionOptions sets options for constructing a *pubsub.Subscription
 // backed by RabbitMQ.
-type SubscriptionOptions struct{}
+type SubscriptionOptions struct {
+	// Qos properties
+	Qos *Qos
+}
+
+// Qos options to be used when we create a subscription.
```
vangent, line 170: Let's drop this extra struct, just put PrefetchCount in SubscriptionOptions.

pubsub/rabbitpubsub/rabbit.go
```diff
@@ -118,8 +121,25 @@ func (o *URLOpener) OpenTopicURL(ctx context.Context, u *url.URL) (*pubsub.Topic
 
 // OpenSubscriptionURL opens a pubsub.Subscription based on u.
 func (o *URLOpener) OpenSubscriptionURL(ctx context.Context, u *url.URL) (*pubsub.Subscription, error) {
-	for param := range u.Query() {
-		return nil, fmt.Errorf("open subscription %v: invalid query parameter %q", u, param)
```
vangent, line 122: You should clone SubscriptionOptions here, and modify and use the clone below, otherwise setting a URL parameter on one OpenSubscriptionURL will affect the next OpenSubscriptionURL as well.

pubsub/rabbitpubsub/rabbit.go
```diff
@@ -564,15 +595,37 @@ func (s *subscription) establishChannel(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
+		// applay subscription options to channel.
```
vangent, line 598: applay -> Apply

pubsub/rabbitpubsub/rabbit.go
```diff
@@ -564,15 +595,37 @@ func (s *subscription) establishChannel(ctx context.Context) error {
...
+		err = applyOptionsToChannel(s.opts, ch)
```
vangent, line 599: Nit: 

if err := applyOptionsToChannel(s.opts, ch); err != nil {
  return err
}

vangent (changes requested):

pubsub/rabbitpubsub/rabbit.go
peczenyj: That can be a good idea.

peczenyj: done via 1c10a049fd44a97a27978c4bbf5ea9466886b8bf

pubsub/rabbitpubsub/rabbit.go
peczenyj: Simplified

pubsub/rabbitpubsub/rabbit.go
peczenyj: Ok I follow what natspubsub is doing

pubsub/rabbitpubsub/rabbit.go
peczenyj: thanks

pubsub/rabbitpubsub/rabbit.go
```diff
@@ -533,11 +557,12 @@ type subscription struct {
 
 var nextConsumer int64 // atomic
 
-func newSubscription(conn amqpConnection, name string) *subscription {
+func newSubscription(conn amqpConnection, name string, opts *SubscriptionOptions) *subscription {
 	return &subscription{
```
vangent, line 561: Let's make sure `opts` is non-nil here:

```
if opts == nil {
		opts = &SubscriptionOptions{}
	}
```

pubsub/rabbitpubsub/rabbit.go
```diff
@@ -564,15 +589,34 @@ func (s *subscription) establishChannel(ctx context.Context) error {
...
 	}
+
 	s.ch = ch
 	s.closec = ch.NotifyClose(make(chan *amqp.Error, 1)) // closec will get at most one element
+
+	return nil
+}
+
+func applyOptionsToChannel(opts *SubscriptionOptions, ch amqpChannel) error {
+	if opts == nil {
```
vangent, line 612: After following my suggestion above, you can remove these 3 lines, `opts` will never be `nil`.

pubsub/rabbitpubsub/rabbit.go
```diff
@@ -564,15 +589,34 @@ func (s *subscription) establishChannel(ctx context.Context) error {
...
+		return nil
+	}
+
+	if err := ch.Qos(opts.PrefetchCount, 0, false); err != nil {
```
vangent, line 616: Is it safe to set 0 as the default? What's the Rabbit default if we don't set it?

I think 0 is a valid value, which makes it a bad default :-|.

One approach is to add a separate boolean, eg.

int PrefetchCount
bool SetPrefetchCount

and only set it if SetPrefetchCount is true.

Another options is to use `int *PrefetchCount` and only set it if non-nil.

vangent (changes requested):

pubsub/rabbitpubsub/rabbit.go
peczenyj: done in c47bd5488d052b276f080f8c4130edcbc2aa361d

pubsub/rabbitpubsub/rabbit.go
peczenyj: done on c47bd5488d052b276f080f8c4130edcbc2aa361d

pubsub/rabbitpubsub/rabbit.go
peczenyj: done on 3cd7cf194059d19f9350be849630351dfdcda7d8

vangent (approved):

PR #3437, opened 2024-06-02
pubsub/rabbitpubsub: wrap pubsub rabbitmq errors

peczenyj:
I note that we mere add a %v reference to the original rabbitmq errors, and since we may need to unwrap such errors, it is better use %w

I also create few global errors to return instead allocate a new error in some trivial cases

Enjoy


vangent (changes requested):
I like the %v -> %w change, but not really the constant errors; they just seem like they add a level of indirection that makes the code less readable. Can you revert that change?

peczenyj:
> I like the %v -> %w change, but not really the constant errors; they just seem like they add a level of indirection that makes the code less readable. Can you revert that change?

Done

vangent (approved):

PR #3438, opened 2024-06-02
all: fix deprecated code and small lint issues detected by staticcheck

peczenyj:
by running `staticcheck ./...` I found several small issues including

- usage of deprecated ioutil package
- grpc.Insecure() option
- explicit temp dir creation inside tests -- instead use t.TempDir()
- deprecated amqp Publish instead PublishWithContext
- and few other small changes

should not change any existing functionality.


vangent:
Thanks for the cleanup!

PR #3446, opened 2024-06-17
docstore/all - add support for boolean filter

ybourgery:
This PR updates Docstore to allow boolean filters in the queries, as it's supported by all the different implementations.

This isn't yet merge-able as the replay data needs to be re-generated, and I don't have the permission to do it.

I have tested it with Firestore, the local version of MongoDB but haven't tested it with AWS or Azure as I don't have an account.

### DIFF
--- a/docstore/awsdynamodb/create_tables.sh
+++ b/docstore/awsdynamodb/create_tables.sh
@@ -42,7 +42,6 @@ aws dynamodb create-table \
         AttributeName=Player,AttributeType=S \
         AttributeName=Score,AttributeType=N \
         AttributeName=Time,AttributeType=S \
-        AttributeName=WithGlitch,AttributeType=BOOL \
   --key-schema AttributeName=Game,KeyType=HASH AttributeName=Player,KeyType=RANGE \
   --provisioned-throughput ReadCapacityUnits=5,WriteCapacityUnits=5 \
   --local-secondary-indexes \
--- a/docstore/awsdynamodb/create_tables.sh
+++ b/docstore/awsdynamodb/create_tables.sh
@@ -42,6 +42,7 @@ aws dynamodb create-table \
         AttributeName=Player,AttributeType=S \
         AttributeName=Score,AttributeType=N \
         AttributeName=Time,AttributeType=S \
+        AttributeName=WithGlitch,AttributeType=BOOL \
   --key-schema AttributeName=Game,KeyType=HASH AttributeName=Player,KeyType=RANGE \
   --provisioned-throughput ReadCapacityUnits=5,WriteCapacityUnits=5 \
   --local-secondary-indexes \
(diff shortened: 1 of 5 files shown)


ybourgery:
Hey @vangent, would it be possible to have a review for my CR, please?

docstore/awsdynamodb/create_tables.sh
```diff
@@ -42,6 +42,7 @@ aws dynamodb create-table \
         AttributeName=Player,AttributeType=S \
         AttributeName=Score,AttributeType=N \
         AttributeName=Time,AttributeType=S \
+        AttributeName=WithGlitch,AttributeType=BOOL \
```
vangent, line 45: This doesn't work, the only possible values are S, N, B. I am not sure what to use for a BOOL value....

Although, the test seems to pass without this so maybe it's not necessary?

docstore/drivertest/drivertest.go
```diff
@@ -1485,6 +1486,11 @@ func testGetQuery(t *testing.T, _ Harness, coll *docstore.Collection) {
 			q:    coll.Query().Where("Player", "not-in", []string{"pat", "billie"}),
 			want: func(h *HighScore) bool { return h.Player != "pat" && h.Player != "billie" },
 		},
+		{
+			name: "WithGlitch",
```
vangent, line 1490: Can you also add tests for "in" and "not-in"?

docstore/query.go
```diff
@@ -38,7 +38,7 @@ func (c *Collection) Query() *Query {
 
 // Where expresses a condition on the query.
 // Valid ops are: "=", ">", "<", ">=", "<=, "in", "not-in".
-// Valid values are strings, integers, floating-point numbers, and time.Time values.
+// Valid values are strings, integers, floating-point numbers, time.Time and boolean (only for "=", "in" and "not-in") values.
```
vangent, line 41: It currently doesn't work for "in" or "not-in", please fix.

vangent (changes requested):
Sorry, I didn't have access to AWS for a while so I wasn't able to run the tests.

I will have to copy this PR into a new one in order to update the goldens for you, but will leave comments here so you can iterate.

docstore/awsdynamodb/create_tables.sh
ybourgery: Yes you are correct, I have removed it

docstore/drivertest/drivertest.go
ybourgery: Yes it's done

docstore/query.go
ybourgery: I have added a test, and it's working for Firestore, MongoDB, but I don't have access to AWS, so I wasn't able to test it

vangent: Ah, it's working. My attempt used `true` not `[]bool{true}` for the `not-in`. Thanks!

vangent:
I have cloned this PR to https://github.com/google/go-cloud/pull/3464 and added the AWS and GCP golden test file updates.

vangent:
Thanks for the contribution!

Issue #3456, opened 2024-08-04
pubsub/all: Using long polling when consuming from a FIFO SQS queue ends up blocking requests


mitsos1os:
### Describe the bug

The `pubsub` package has trouble consuming from a **FIFO enabled SQS** queue when using **long polling** on messages with the same `MessageGroupID` and blocks requests by the amount of wait time configured for **long polling**.

The reason this is happening is due to the way the `pubsub` library is structured and also due to the basic operation mode of **FIFO** SQS queues.

#### Details about SQS FIFO queue operation
When an SQS queue has **FIFO** enabled, then as the name suggests, it guarantees that the messages belonging to the same **group** (using the same `MessageGroupID`), will be delivered in-order and with an exactly-once delivery guarantee. This means, that unless a message is acknowledged, the next in-line will not be made available when reading from the queue. Sending a **poll** request when the last message has not been acknowledged, will return 0 messages from the queue. More info [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-fifo-queues.html)

**Long Polling** is a mode during which you configure a **Wait Time**, which shows how much time a poll request will wait before returning until it gets a message or the configured wait time expires. This is used to reduce the number of repeated requests that might return 0 messages. For example, if you configure a wait time of 10 seconds and then send a **poll** request, it will wait up to 10 seconds, either returning as soon as a message is available or when the 10 seconds have passed by (returning 0 messages again). If **short polling** is used and a poll request is sent when 0 messages, are available the request will return immediately, sending an other request right after, repeating until a message is available. It is generally advised to use long polling. More info can be found [here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-short-and-long-polling.html)

#### `pubsub` operation
The package supports an internal auto-scaling mechanism that tries to increase the message consumption throughput based on the current speed of message processing (time from the message received, until it is acknowledged).

When the message rate is increased, the so-called **batch size** of messages requested is increased and requests are split into **batches** depending on the allowed limits of each driver/cloud-provider.

[...]

PR #3457, opened 2024-08-04
pubsub: Make batch request results independent

mitsos1os:
Allow batch requests to return results independently without one blocking the other.

This will in turn allow `Subscription.Receive` to return a message from a request (since the `Receive` function signature returns only one message), without waiting for all the requests from the batch to complete.

Except for this, package behavior should remain as is...

Fixes: #3456

### DIFF
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -562,37 +562,31 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
 			// log.Printf("BATCH SIZE %d", batchSize)
 
 			go func() {
-				defer func() {
-					close(s.waitc)
-					s.waitc = nil
-				}()
 				if s.preReceiveBatchHook != nil {
 					s.preReceiveBatchHook(batchSize)
 				}
 				resultChannel := s.getNextBatch(batchSize)
-				for {
-					select {
-					case msgs, ok := <-resultChannel.msgs:
-						if !ok {
-							// batch reception finished
-							return
-						} else if len(msgs) > 0 {
-							// messages received from channel
-							s.mu.Lock()
-							s.q = append(s.q, msgs...)
-							s.mu.Unlock()
-						}
-					case err := <-resultChannel.err:
+				for msgsOrError := range resultChannel {
+					if msgsOrError.msgs != nil && len(msgsOrError.msgs) > 0 {
+						// messages received from channel
+						s.mu.Lock()
+						s.q = append(s.q, msgsOrError.msgs...)
+						s.mu.Unlock()
+						// notify that queue should now have messages
+						s.waitc <- struct{}{}
+					} else if msgsOrError.err != nil {
 						// err can receive message only after batch group completes
-						if err != nil {
-							// Non-retryable error from ReceiveBatch -> permanent error.
-							s.mu.Lock()
-							s.err = err
-							s.mu.Unlock()
-						}
-						return
+						// Non-retryable error from ReceiveBatch -> permanent error
+						s.mu.Lock()
+						s.err = msgsOrError.err
+						s.mu.Unlock()
 					}
 				}
+				// batch reception finished
+				s.mu.Lock()
+				close(s.waitc)
+				s.waitc = nil
+				s.mu.Unlock()
 			}()
 		}
 		if len(s.q) > 0 {
@@ -640,33 +634,33 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
 			})
 			return m2, nil
 		}
+		// A call to ReceiveBatch must be in flight. Wait for it.
+		waitc := s.waitc
 		s.mu.Unlock() // unlock to allow message or error processing from background goroutine
 		select {
+		case <-waitc:
+			// Continue to top of loop.
+			s.mu.Lock()
 		case <-ctx.Done():
 			s.mu.Lock()
 			return nil, ctx.Err()
-		default:
-			// Continue to top of loop.
-			s.mu.Lock()
 		}
 	}
 }
 
-type batchChannelResult struct {
-	msgs chan []*driver.Message
-	err  chan error
+type msgsOrError struct {
+	msgs []*driver.Message
+	err  error
 }
 
-// getNextBatch gets the next batch of messages from the server and returns it.
-func (s *Subscription) getNextBatch(nMessages int) *batchChannelResult {
+// getNextBatch gets the next batch of messages from the server. It will return a channel that will itself return the
+// messages as they come from each independent batch, or an operation error
+func (s *Subscription) getNextBatch(nMessages int) chan msgsOrError {
 	// Split nMessages into batches based on recvBatchOpts; we'll make a
 	// separate ReceiveBatch call for each batch, and aggregate the results in
 	// msgs.
 	batches := batcher.Split(nMessages, s.recvBatchOpts)
-	result := batchChannelResult{
-		msgs: make(chan []*driver.Message, len(batches)),
-		err:  make(chan error),
-	}
+	result := make(chan msgsOrError, len(batches))
 	g, ctx := errgroup.WithContext(s.backgroundCtx)
 	for _, maxMessagesInBatch := range batches {
 		// Make a copy of the loop variable since it will be used by a goroutine.
@@ -683,19 +677,18 @@ func (s *Subscription) getNextBatch(nMessages int) *batchChannelResult {
 			if err != nil {
 				return wrapError(s.driver, err)
 			}
-			result.msgs <- msgs
+			result <- msgsOrError{msgs: msgs}
 			return nil
 		})
 	}
 	go func() {
 		// wait on group completion on the background and proper channel closing
 		if err := g.Wait(); err != nil {
-			result.err <- err
+			result <- msgsOrError{err: err}
 		}
-		close(result.err)
-		close(result.msgs)
+		close(result)
 	}()
-	return &result
+	return result
 }
 
 var errSubscriptionShutdown = gcerr.Newf(gcerr.FailedPrecondition, nil, "pubsub: Subscription has been Shutdown")
--- a/pubsub/pubsub.go
+++ b/pubsub/pubsub.go
@@ -562,20 +562,37 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
 			// log.Printf("BATCH SIZE %d", batchSize)
 
 			go func() {
+				defer func() {
+					close(s.waitc)
+					s.waitc = nil
+				}()
 				if s.preReceiveBatchHook != nil {
 					s.preReceiveBatchHook(batchSize)
 				}
-				msgs, err := s.getNextBatch(batchSize)
-				s.mu.Lock()
-				defer s.mu.Unlock()
-				if err != nil {
-					// Non-retryable error from ReceiveBatch -> permanent error.
-					s.err = err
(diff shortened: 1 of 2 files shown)


pubsub/pubsub.go
```diff
@@ -562,20 +562,37 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
...
-				if err != nil {
-					// Non-retryable error from ReceiveBatch -> permanent error.
-					s.err = err
-				} else if len(msgs) > 0 {
-					s.q = append(s.q, msgs...)
+				resultChannel := s.getNextBatch(batchSize)
+				for {
+					select {
+					case msgs, ok := <-resultChannel.msgs:
+						if !ok {
```
vangent, line 576: Can't this lose errors?

I.e., if `getNextBatch` gets an error, and writes it to resultChannel.err, and then immediately closes resultChannel.msgs, I think this `select` could end up here, and we'd never receive the error.

Instead of returning a struct that contains two channels, I think it might be better to return a single channel, which delivers structs (where the struct can either have messages or an error).

pubsub/pubsub.go
```diff
@@ -562,20 +562,37 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
 			// log.Printf("BATCH SIZE %d", batchSize)
 
 			go func() {
+				defer func() {
+					close(s.waitc)
+					s.waitc = nil
```
vangent, line 567: I'm a bit concerned that previously this happened with the lock, and now it doesn't.

If I'm following the code correctly, I think the only way out of this infinite `for` loop is when the channel is closed, so can't we put this cleanup code there and make sure we have the lock when we do it?

pubsub/pubsub.go
```diff
@@ -562,20 +562,37 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
...
-					s.err = err
-				} else if len(msgs) > 0 {
-					s.q = append(s.q, msgs...)
+				resultChannel := s.getNextBatch(batchSize)
+				for {
+					select {
+					case msgs, ok := <-resultChannel.msgs:
+						if !ok {
+							// batch reception finished
+							return
```
vangent, line 578: Per comment above, move the cleanup code out of the defer and put it here.

pubsub/pubsub.go
```diff
@@ -562,20 +562,37 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
...
+						}
+					case err := <-resultChannel.err:
+						// err can receive message only after batch group completes
+						if err != nil {
+							// Non-retryable error from ReceiveBatch -> permanent error.
+							s.mu.Lock()
+							s.err = err
+							s.mu.Unlock()
+						}
+						return
```
vangent, line 593: I think you can drop this `return` to make the flow simpler; the next time through the `select` should get `ok=false` and exit there.

pubsub/pubsub.go
```diff
@@ -623,30 +640,33 @@ func (s *Subscription) Receive(ctx context.Context) (_ *Message, err error) {
...
+		s.mu.Unlock() // unlock to allow message or error processing from background goroutine
 		select {
-		case <-waitc:
-			s.mu.Lock()
-			// Continue to top of loop.
 		case <-ctx.Done():
 			s.mu.Lock()
 			return nil, ctx.Err()
+		default:
+			// Continue to top of loop.
```
vangent, line 649: Doesn't this spin-loop while it's waiting? I think you've removed any use of `s.waitc`.

Maybe you can put back the `waitc` case, and up above every time we get some messages, write something to `waitc`? Only close it / set to null when all batches are complete.

vangent (changes requested):

pubsub/pubsub.go
vangent: In fact, tests failed with a DATA RACE due to this:

WARNING: DATA RACE
Read at 0x00c000482088 by goroutine 20:
  gocloud.dev/pubsub.(*Subscription).Receive()
      /home/runner/work/go-cloud/go-cloud/pubsub/pubsub.go:554 +0x312
  gocloud.dev/samples/order.(*processor).handleRequest()
      /home/runner/work/go-cloud/go-cloud/samples/order/processor.go:63 +0xc7
  gocloud.dev/samples/order.TestHandleRequest()
      /home/runner/work/go-cloud/go-cloud/samples/order/processor_test.go:47 +0x52a
  testing.tRunner()
      /opt/hostedtoolcache/go/1.22.5/x64/src/testing/testing.go:1689 +0x21e
  testing.(*T).Run.gowrap1()
      /opt/hostedtoolcache/go/1.22.5/x64/src/testing/testing.go:1742 +0x44

Previous write at 0x00c000482088 by goroutine 22:
  gocloud.dev/pubsub.(*Subscription).Receive.func2.1()
      /home/runner/work/go-cloud/go-cloud/pubsub/pubsub.go:567 +0x64
  runtime.deferreturn()
      /opt/hostedtoolcache/go/1.22.5/x64/src/runtime/panic.go:602 +0x5d

Goroutine 20 (running) created at:
  testing.(*T).Run()
      /opt/hostedtoolcache/go/1.22.5/x64/src/testing/testing.go:1742 +0x825
  testing.runTests.func1()
      /opt/hostedtoolcache/go/1.22.5/x64/src/testing/testing.go:2161 +0x85
  testing.tRunner()
      /opt/hostedtoolcache/go/1.22.5/x64/src/testing/testing.go:1689 +0x21e
  testing.runTests()
      /opt/hostedtoolcache/go/1.22.5/x64/src/testing/testing.go:2159 +0x8be
  testing.(*M).Run()
      /opt/hostedtoolcache/go/1.22.5/x64/src/testing/testing.go:2027 +0xf17
  main.main()
      _testmain.go:89 +0x2e4

Goroutine 22 (finished) created at:
  gocloud.dev/pubsub.(*Subscription).Receive()
      /home/runner/work/go-cloud/go-cloud/pubsub/pubsub.go:564 +0x493
  gocloud.dev/samples/order.(*processor).handleRequest()
      /home/runner/work/go-cloud/go-cloud/samples/order/processor.go:63 +0xc7
  gocloud.dev/samples/order.TestHandleRequest()
      /home/runner/work/go-cloud/go-cloud/samples/order/processor_test.go:47 +0x52a
  testing.tRunner()
      /opt/hostedtoolcache/go/1.22.5/x64/src/testing/testing.go:1689 +0x21e
  testing.(*T).Run.gowrap1()
      /opt/hostedtoolcache/go/1.22.5/x64/src/testing/testing.go:1742 +0x44

mitsos1os: You are right about this. I concentrated on the msg return and missed that `waitc` was handled while in lock. WIll update that

pubsub/pubsub.go
mitsos1os: I think there might be a bit more to it...

So, given that `g.Wait` in `getNextBatch` will return only after each goroutine has exited, it means that any messages sent, will certainly happen before `Wait` returns.

So when `Wait` returns, if there was an error it will be written to the error channel which is **not buffered** . So, this will block the execution until it is consumed. So it won't lose any errors since it won't move to close the channel unless the error gets consumed.

But you may be right about the flow being simpler returning a single channel. I will try this out

mitsos1os:
@vangent Thank you for your suggestions! I have refactored the execution flow to address them.

This is what you get when coding tired after midnight 😄

vangent (changes requested):
I haven't taken another look, but the tests are still failing with a data race.

You should be able to reproduce this locally with `-race`....

WARNING: DATA RACE
Read at 0x00c0000bc6b8 by goroutine 2559:
  gocloud.dev/pubsub.(*secondReceiveBlockedDriverSub).ReceiveBatch()
      /Users/runner/work/go-cloud/go-cloud/pubsub/pubsub_test.go:291 +0x38
  gocloud.dev/pubsub.(*Subscription).getNextBatch.func1.1()
      /Users/runner/work/go-cloud/go-cloud/pubsub/pubsub.go:674 +0x128
  gocloud.dev/internal/retry.call()
      /Users/runner/work/go-cloud/go-cloud/internal/retry/retry.go:52 +0x10c
  gocloud.dev/internal/retry.Call()
      /Users/runner/work/go-cloud/go-cloud/internal/retry/retry.go:40 +0x140
  gocloud.dev/pubsub.(*Subscription).getNextBatch.func1()
      /Users/runner/work/go-cloud/go-cloud/pubsub/pubsub.go:670 +0x58
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      /Users/runner/go/pkg/mod/golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78 +0x7c

Previous write at 0x00c0000bc6b8 by goroutine 2560:
  gocloud.dev/pubsub.(*secondReceiveBlockedDriverSub).ReceiveBatch()
      /Users/runner/work/go-cloud/go-cloud/pubsub/pubsub_test.go:291 +0x4c
  gocloud.dev/pubsub.(*Subscription).getNextBatch.func1.1()
      /Users/runner/work/go-cloud/go-cloud/pubsub/pubsub.go:674 +0x128
  gocloud.dev/internal/retry.call()
      /Users/runner/work/go-cloud/go-cloud/internal/retry/retry.go:52 +0x10c
  gocloud.dev/internal/retry.Call()
      /Users/runner/work/go-cloud/go-cloud/internal/retry/retry.go:40 +0x140
  gocloud.dev/pubsub.(*Subscription).getNextBatch.func1()
      /Users/runner/work/go-cloud/go-cloud/pubsub/pubsub.go:670 +0x58
  golang.org/x/sync/errgroup.(*Group).Go.func1()
      /Users/runner/go/pkg/mod/golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78 +0x7c

mitsos1os:
@vangent the race condition was on test code now... I had forgotten to save the `-race` flag when running tests locally
It is now fixed as well

pubsub/pubsub_test.go
```diff
@@ -281,6 +282,51 @@ func TestCancelTwoReceives(t *testing.T) {
...
+	// not block the Subscription.Receive result
+	receiveWaitDuration := 200 * time.Millisecond
+	s := NewSubscription(
+		&secondReceiveBlockedDriverSub{waitDuration: receiveWaitDuration},
+		&batcher.Options{MaxBatchSize: 1, MaxHandlers: 2}, // force 2 batches, by allowing 2 handlers and 1 msg per batch
+		nil,
+	)
+	// set the false calculated subscription batch size to force 2 batches to be called
+	s.runningBatchSize = 2
+	ctx, cancel := context.WithTimeout(context.Background(), 300*time.Millisecond)
```
vangent, line 315: The hardcoded timings might be racy.

Could `secondReceiveBlockedDriverSub` just wait forever (or maybe until the passed-in `ctx` is `Done`) on the second call, and `s.Receive` wait forever (no `context.WithTimeout`)? Without your change, the test would hang and eventually timeout; after it, I think the receive should work ~right away (getting the first message) and then exit successfully.

vangent (changes requested):
pubsub.go looks good now, just a nit on the test. Thanks!

pubsub/pubsub_test.go
mitsos1os: Yeah, I opted in for context timeout because in the case of failure, I didn't want the test to hang for a long amount of time and get a faster failure.

Anyway, I changed it to how you proposed with the wait-forever approach and made sure that on master the test hangs.

vangent (approved):

Issue #3472, opened 2024-08-18
blob/s3blob: s3ForcePathStyle stopped working in v0.39.0


caarlos0:
### Describe the bug

This was flagged by [goreleaser's tests][build] when dependabot opened the pr updating from v0.38 to v0.39.

[build]: https://github.com/goreleaser/goreleaser/pull/5080

I'm not sure if this is expected and just not documented, or if it is in fact a bug.

### To Reproduce

open a bucket (can be a local minio for example) using `s3ForcePathStyle`.

## Expected behavior

To work.

### Version

v0.39.0

### Additional context

Error:

```
open bucket s3://basic?endpoint=http%3A%2F%2Flocalhost%3A32812&region=us-east&s3ForcePathStyle=true: unknown query parameter "s3ForcePathStyle"
```

Also, forcing `awssdk=v1` makes it work again.

vangent:
Please see the release notes (https://github.com/google/go-cloud/releases/tag/v0.39.0). This release changed the default for AWS URLs from their v1 SDK to their v2 SDK. Some parameters are different, `s3ForcePathStyle` is one that doesn't seem to exist for v2.

You should either update the URL to work with v2 (possibly just dropping that parameter will fix it?), or add `awssdk=v1` (which will work for now, but support will be dropped at some point, the release notes proposed 6 months).

caarlos0:
yes, I saw the release notes, and also that forcing v1 works, my point was that the specific breaking change is not documented anywhere.

also, just dropping that part of the URL doesn't work (at least not with minio).

vangent:
OK. Yes, I didn't try to document every specific URL parameter that wouldn't work anymore, just "Most URLs should continue to work, but in some cases you may need to add awssdk=v1 to force V1 explicitly."

vangent:
> just dropping that part of the URL doesn't work (at least not with minio

Please try to figure out what's needed to get V2 to work with Minio (I'm not sure). It's possible we need to add support for more URL parameters for V2.

dmotylev:
It seems `s3ForcePathStyle` has been moved to https://github.com/aws/aws-sdk-go-v2/blob/0cbb5aa17f9078cb45dc0e82d3e1d0abee3744a9/service/s3/options.go#L147-L150

vangent:
Added, see https://github.com/google/go-cloud/pull/3491.

stanhu:
[...]

PR #3473, opened 2024-08-19
blob/s3blob: custom endpoints with s3 and aws sdk v2

caarlos0:
It seems that this is needed in aws-sdk-v2 for minio to work.

see: https://github.com/minio/docs/issues/406#issuecomment-1246316964
refs https://github.com/google/go-cloud/issues/3472

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -202,19 +202,18 @@ func V2ConfigFromURLParams(ctx context.Context, q url.Values) (awsv2.Config, err
 		default:
 			return awsv2.Config{}, fmt.Errorf("unknown query parameter %q", param)
 		}
-
-		if endpoint != "" {
-			customResolver := awsv2.EndpointResolverWithOptionsFunc(
-				func(service, region string, options ...interface{}) (awsv2.Endpoint, error) {
-					return awsv2.Endpoint{
-						PartitionID:       "aws",
-						URL:               endpoint,
-						SigningRegion:     region,
-						HostnameImmutable: hostnameImmutable,
-					}, nil
-				})
-			opts = append(opts, awsv2cfg.WithEndpointResolverWithOptions(customResolver))
-		}
+	}
+	if endpoint != "" {
+		customResolver := awsv2.EndpointResolverWithOptionsFunc(
+			func(service, region string, options ...interface{}) (awsv2.Endpoint, error) {
+				return awsv2.Endpoint{
+					PartitionID:       "aws",
+					URL:               endpoint,
+					SigningRegion:     region,
+					HostnameImmutable: hostnameImmutable,
+				}, nil
+			})
+		opts = append(opts, awsv2cfg.WithEndpointResolverWithOptions(customResolver))
 	}
 	return awsv2cfg.LoadDefaultConfig(ctx, opts...)
 }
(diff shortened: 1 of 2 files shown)


aws/aws.go
```diff
@@ -188,9 +188,10 @@ func V2ConfigFromURLParams(ctx context.Context, q url.Values) (awsv2.Config, err
 			customResolver := awsv2.EndpointResolverWithOptionsFunc(
 				func(service, region string, options ...interface{}) (awsv2.Endpoint, error) {
 					return awsv2.Endpoint{
-						PartitionID:   "aws",
-						URL:           value,
-						SigningRegion: region,
+						PartitionID:       "aws",
+						URL:               value,
+						SigningRegion:     region,
+						HostnameImmutable: true,
```
vangent, line 194: The documentation says "Setting an endpoint as immutable may prevent some service client features from functioning correctly, and could result in undefined behavior. Caution should be taken when defining an endpoint as immutable."

https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/endpoints/#immutable-endpoints

Should we make this an option instead of always setting it to true? Or is it reasonable to expect it to always be true when someone is explicitly setting an endpoint?

SebouChu: I think it should be an option, preventing unexpected behavior for others users

caarlos0: yeah, I was thinking it too, gonna work on it later today

caarlos0: done

aws/aws.go
```diff
@@ -178,29 +178,42 @@ func NewDefaultV2Config(ctx context.Context) (awsv2.Config, error) {
 //   - profile: The shared config profile to use; sets SharedConfigProfile.
 //   - endpoint: The AWS service endpoint to send HTTP request.
```
vangent, line 179: Please add hostname_immutable here.

aws/aws.go
```diff
@@ -178,29 +178,42 @@ func NewDefaultV2Config(ctx context.Context) (awsv2.Config, error) {
...
 func V2ConfigFromURLParams(ctx context.Context, q url.Values) (awsv2.Config, error) {
+	var endpoint string
+	var hostnameImmutable bool
 	var opts []func(*awsv2cfg.LoadOptions) error
 	for param, values := range q {
 		value := values[0]
 		switch param {
+		case "awsHostnameImmutable":
```
vangent, line 187: Let's use "hostname_immutable", there's no need for "aws" prefix and URL params aren't case sensitive.

aws/aws.go
```diff
@@ -178,29 +178,42 @@ func NewDefaultV2Config(ctx context.Context) (awsv2.Config, error) {
...
+			var err error
+			hostnameImmutable, err = strconv.ParseBool(value)
+			if err != nil {
+				return awsv2.Config{}, err
```
vangent, line 191: Please add something to the error, e.g., `fmt.Errorf("invalid value for hostname_immutable: %w", err)`

vangent (changes requested):

vangent (approved):

SebouChu:
Thanks @caarlos0!

PR #3480, opened 2024-08-29
blob/s3blob: fix data race

arjunnair1997:
w.pr cannot be used to indicate that the pipe has been closed. Instead, we use an atomic.Bool. The problem with using w.pr is that a write to it in the goroutine launched by writer.open can race with a read on w.pr in Close.

Note that it's still possible that w.pr.CloseWithError runs concurrently with w.pr.Close. But this is fine.

### DIFF
--- a/blob/s3blob/s3blob.go
+++ b/blob/s3blob/s3blob.go
@@ -380,7 +380,6 @@ func (w *writer) open(r io.Reader, closePipeOnError bool) {
 		if err != nil {
 			if closePipeOnError {
 				w.pr.CloseWithError(err)
-				w.pr = nil
 			}
 			w.err = err
 		}


richardartoul (approved):
LGTM, although wondering if we could just allow `w.pr` to be closed twice because it looks like thats concurrency safe:

```golang
func (p *pipe) closeRead(err error) error {
	if err == nil {
		err = ErrClosedPipe
	}
	p.rerr.Store(err)
	p.once.Do(func() { close(p.done) })
	return nil
}
```

vangent (changes requested):
I see the race you are describing, but I don't think this PR completely fixes it. I think it changes the race from "might result in a nil access" to "might result in a double-Close".

It looks like double-Close is actually fine though (https://play.golang.com/p/WvceCRttXVU).

So I think it should be sufficient to drop the "prClosed" entirely, and just remove the "w.pr = nil" line (line 383). WDYT?

vangent:
Ha, my review raced with @richardartoul  :-). Glad we came to the same conclusion though.

arjunnair1997:
Yea, I think it could double close the read end of the pipe before my change to, but it's just simpler to not use an atomic.Bool. Updated the pr, please take a look.

vangent (approved):

vangent:
Thanks!

Issue #3484, opened 2024-09-09
s3blob/blob: support additional endpoint parameters: UseDualStack, UseFips, and UseAccelerate options


stanhu:
### Is your feature request related to a problem? Please describe.

https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/endpoints/ mentions additional parameters that can be configured:

* `UseDualStack`
* `UseFips`
* `UseAccelerate`

These were added upstream in https://github.com/aws/aws-sdk-go-v2/pull/836.

### Describe the solution you'd like

The AWS SDK v2 URL probably should accept the following query parameters:

1. `dualstack`
2. `fips`
3. `accelerate`

As described in https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/endpoints/#migration, we probably should migrate to v2 of the endpoint resolution.  Note that `EndpointResolverWithOptionsFunc` is deprecated and should likely be replaced with the v2 mechanism:  https://github.com/google/go-cloud/blob/d2adbc5c0014d2b7de556e837a508ba999b08f51/aws/aws.go#L207 

### Describe alternatives you've considered

While the `endpoint` can probably be used to support this functionality, users would have to know their region-specific endpoints. For example, if my AWS S3 bucket is `my-bucket` in `us-east-1`, and I want to enable transfer acceleration, dual-stack support, and FIPS, I would need to configure `endpoint` with one of the following:

1. `my-bucket.s3-accelerate.amazonaws.com`
5. `my-bucket.s3-accelerate.dualstack.amazonaws.com`
6. `my-bucket.s3-fips.us-gov-east-1.amazonaws.com` 
7. `my-bucket.s3-fips.dualstack.us-east-1.amazonaws.com`

This forces the application to build the hostname, when this seems more appropriately handled by the SDK.

References:

1. https://docs.aws.amazon.com/AmazonS3/latest/userguide/transfer-acceleration.html
1. https://aws.amazon.com/compliance/fips/

### Additional context

I believe `UseAccelerate` can be added via something like:

```diff
diff --git a/blob/s3blob/s3blob.go b/blob/s3blob/s3blob.go
index d3e80cf0..7353c6db 100644
--- a/blob/s3blob/s3blob.go
+++ b/blob/s3blob/s3blob.go
@@ -145,8 +145,9 @@ type URLOpener struct {
 }
 
 const (
-	sseTypeParamKey  = "ssetype"
-	kmsKeyIdParamKey = "kmskeyid"
[...]

PR #3486, opened 2024-09-11
s3blob/blob: support additional endpoint query parameters

stanhu:
This commit adds the following query parameters for AWS:

1. `dualstack`
2. `fips`
3. `accelerate` (S3-only)

This avoids the need for users to specify `endpoint`. For example, if my AWS S3 bucket is `my-bucket` in `us-east-1`, and you want to enable transfer acceleration, dual-stack support, and/or FIPS, you would need to configure `endpoint` with one of the following:

1. `my-bucket.s3-accelerate.amazonaws.com`
2. `my-bucket.s3-accelerate.dualstack.amazonaws.com`
3. `my-bucket.s3-fips.us-gov-east-1.amazonaws.com`
4. `my-bucket.s3-fips.dualstack.us-east-1.amazonaws.com`

For example, for the last option, users can use `s3://my-bucket?fips=true&dualstack=true`.

Closes https://github.com/google/go-cloud/issues/3484

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -26,6 +26,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/google/wire"
 )
@@ -81,6 +82,8 @@ func (co ConfigOverrider) ClientConfig(serviceName string, cfgs ...*aws.Config) 
 //   - endpoint: The endpoint URL (hostname only or fully qualified URI); sets aws.Config.Endpoint.
 //   - disableSSL: A value of "true" disables SSL when sending requests; sets aws.Config.DisableSSL.
 //   - s3ForcePathStyle: A value of "true" forces the request to use path-style addressing; sets aws.Config.S3ForcePathStyle.
+//   - dualstack: A value of "true" enables dual stack (IPv4 and IPv6) endpoints
+//   - fips: A value of "true" enables the use of FIPS endpoints
 func ConfigFromURLParams(q url.Values) (*aws.Config, error) {
 	var cfg aws.Config
 	for param, values := range q {
@@ -102,6 +105,20 @@ func ConfigFromURLParams(q url.Values) (*aws.Config, error) {
 				return nil, fmt.Errorf("invalid value for query parameter %q: %v", param, err)
 			}
 			cfg.S3ForcePathStyle = aws.Bool(b)
+		case "dualstack":
+			b, err := strconv.ParseBool(value)
+			if err != nil {
+				return nil, fmt.Errorf("invalid value for query parameter %q: %v", param, err)
+			}
+			cfg.UseDualStack = aws.Bool(b)
+		case "fips":
+			b, err := strconv.ParseBool(value)
+			if err != nil {
+				return nil, fmt.Errorf("invalid value for query parameter %q: %v", param, err)
+			}
+			if b {
+				cfg.UseFIPSEndpoint = endpoints.FIPSEndpointStateEnabled
+			}
 		case "awssdk":
 			// ignore, should be handled before this
 		default:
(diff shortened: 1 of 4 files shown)


aws/aws.go
```diff
@@ -102,6 +103,22 @@ func ConfigFromURLParams(q url.Values) (*aws.Config, error) {
 				return nil, fmt.Errorf("invalid value for query parameter %q: %v", param, err)
 			}
 			cfg.S3ForcePathStyle = aws.Bool(b)
+		case "dualstack":
```
vangent, line 106: Please document the new query options up above in the docstring.

aws/aws.go
```diff
@@ -197,6 +214,26 @@ func V2ConfigFromURLParams(ctx context.Context, q url.Values) (awsv2.Config, err
 			endpoint = value
 		case "profile":
 			opts = append(opts, awsv2cfg.WithSharedConfigProfile(value))
+		case "dualstack":
```
vangent, line 217: Same.

aws/aws.go
```diff
@@ -197,6 +214,26 @@ func V2ConfigFromURLParams(ctx context.Context, q url.Values) (awsv2.Config, err
...
+			opts = append(opts, awsv2cfg.WithUseDualStackEndpoint(dualStack))
+		case "fips":
+			fipsParam, err := strconv.ParseBool(value)
+			if err != nil {
+				return awsv2.Config{}, fmt.Errorf("invalid value for fips: %w", err)
+			}
+			fips := awsv2.FIPSEndpointStateDisabled
+			if fipsParam {
+				fips = awsv2.FIPSEndpointStateEnabled
+			}
```
vangent, line 235: For hostnameImmutable, we kept it so that it could be added to the customResolver below, should we do that for these 2 as well?

aws/aws.go
```diff
@@ -102,6 +103,22 @@ func ConfigFromURLParams(q url.Values) (*aws.Config, error) {
...
+				return nil, fmt.Errorf("invalid value for query parameter %q: %v", param, err)
+			}
+			cfg.UseDualStack = aws.Bool(b)
+		case "fips":
+			b, err := strconv.ParseBool(value)
+			if err != nil {
+				return nil, fmt.Errorf("invalid value for query parameter %q: %v", param, err)
+			}
+			fips := endpoints.FIPSEndpointStateDisabled
+			if b {
```
vangent, line 118: Please add some tests for these 2 new parameters in aws_test.go (for both functions you've added them to).

blob/s3blob/s3blob_test.go
```diff
@@ -476,6 +476,18 @@ func TestOpenBucketFromURL(t *testing.T) {
 		{"s3://mybucket?awssdk=v2", false},
 		// OK, use KMS Server Side Encryption
 		{"s3://mybucket?ssetype=aws:kms&kmskeyid=arn:aws:us-east-1:12345:key/1-a-2-b", false},
+		// OK, use S3 Transfer acceleration
+		{"s3://mybucket?accelerate=true", false},
+		// OK, use dual stack endpoints
+		{"s3://mybucket?dualstack=true", false},
```
vangent, line 482: Nit: You don't necessarily need to test all the combinations of things that are supported in aws.go. Just one is OK, plus "accelerate" since it's specific to s3blob.

vangent (changes requested):

blob/s3blob/s3blob_test.go
stanhu: Ok. AWS v2 is unhappy if `accelerate=true&fips=true`, I believe:

```
operation error S3: ListObjectsV2, resolve auth scheme: resolve endpoint: endpoint rule error, Accelerate cannot be used with FIPS
```

aws/aws.go
stanhu: Sure, how does this look now? This will probably change with the move to service-specifc endpoint resolvers.

aws/aws.go
```diff
@@ -215,5 +246,12 @@ func V2ConfigFromURLParams(ctx context.Context, q url.Values) (awsv2.Config, err
 			})
 		opts = append(opts, awsv2cfg.WithEndpointResolverWithOptions(customResolver))
 	}
+	if dualStack {
```
vangent, line 249: Hmm if you're just going to add them down here, I think it was better before, doing it right after the parameter was processed. I thought maybe they needed to be set in the awsv2.Endpoint like HostnameImmutable, but if that's not the case then my apologies, please change it back.

vangent (changes requested):

aws/aws.go
stanhu: Yes, interestingly enough, they aren't passed in with the resolver.

vangent (approved):

Issue #3490, opened 2024-10-04
No way to support blob/blob in compatible storage like minio with v2 aws sdkgo


khrm:
Please use a title starting with the name of the affected package, or \"all\",
followed by a colon, followed by a short summary of the issue. Example:
`blob/gcsblob: not blobby enough`.

You can use `go bug` to have a cool, automatically filled out bug template, or
fill out the template below.

### Describe the bug
It's more of a regression than a bug. If we remove aws sdk v1, we won't have a way to support the minio or other compatible storage using blob/blob. Obviously, blob/s3blob will work but we lost out on blob abstraction.

### To Reproduce
You can use `play.min.io:9443` and try to run with aws sdk v2 and blob. There's no way. We definitely need a query param to solve this.

## Expected behavior

A query param like we have in aws sdk v1 `s3ForcePathStyl`.

vangent:
See https://github.com/google/go-cloud/pull/3473...?

PR #3491, opened 2024-10-04
Add disableHTTPS and usePathStyle s3v2.Options as query param

khrm:
This ensures that we can use blob/blob with s3 compatible storage like minio.
Pass `usePathStyle=true` to force PathStyle for s3. Pass `disableHTTPS=true` to disable tls for endpoint.

Fixes: https://github.com/google/go-cloud/issues/3490

### DIFF
--- a/blob/s3blob/s3blob.go
+++ b/blob/s3blob/s3blob.go
@@ -150,9 +150,11 @@ type URLOpener struct {
 }
 
 const (
-	sseTypeParamKey    = "ssetype"
-	kmsKeyIdParamKey   = "kmskeyid"
-	accelerateParamKey = "accelerate"
+	sseTypeParamKey      = "ssetype"
+	kmsKeyIdParamKey     = "kmskeyid"
+	accelerateParamKey   = "accelerate"
+	usePathStyleParamkey = "use_path_style"
+	disableHTTPSParamKey = "disable_https"
 )
 
 func toServerSideEncryptionType(value string) (typesv2.ServerSideEncryption, error) {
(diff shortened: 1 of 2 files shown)


aws/aws.go
```diff
@@ -253,3 +254,49 @@ func V2ConfigFromURLParams(ctx context.Context, q url.Values) (awsv2.Config, err
...
+// UseV2 returns true.
+//
+// https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3#Options
+//
+// It returns an error if q contains any unknown query parameters; callers
+// should remove any query parameters they know about from q before calling
+// V2s3OptionsFromURLParams.
+//
+// The following query options are supported:
+//   - usePathStyle: A value of "true" forces the request to use path-style addressing; sets s3v2.Options.UsePathStyle.
```
vangent, line 271: Just so I understand -- do these options make sense for all AWS services, or just for S3?

If they make sense for all AWS services, let's update the existing V2ConfigFromURLParams to return an awsv2.Config and a []func(*s3v2.Options), instead of creating a new one.

aws/aws.go
```diff
@@ -253,3 +254,49 @@ func V2ConfigFromURLParams(ctx context.Context, q url.Values) (awsv2.Config, err
...
+//   - disableHTTPS: A value of "true" disables SSL when sending requests; sets s3v2.Options.EndpointOptions.DisableHTTPS.
```
vangent, line 272: Please follow the existing style for URL parameters, e.g., "use_path_style" and "disable_https".

Let's add support for "no_accelerate" too? (leave the default as true, but allow people to turn it off).

blob/s3blob/s3blob.go
```diff
@@ -199,10 +199,16 @@ func (o *URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket
 		if err != nil {
```
vangent, line 199: Do the tests pass? I would expect V2ConfigFromURLParams to fail if you passed it a URL with these new parameters in it.

vangent (changes requested):

aws/aws.go
khrm: `s3v2.Options` options are used by s3 only.

khrm: So I think we need to have only a separate function.

aws/aws.go
khrm: Aren't we following camelcase everywhere else?

Reference:
https://github.com/google/go-cloud/pull/3491/files#diff-6fe59643f73b478394f267669c70f00520f5902ea6d880ea1806169a2e687296L102

https://github.com/google/go-cloud/pull/3491/files#diff-6fe59643f73b478394f267669c70f00520f5902ea6d880ea1806169a2e687296L96

blob/s3blob/s3blob.go
khrm: I have added the following tests to catch this issue:
https://github.com/google/go-cloud/pull/3491/files#diff-0520a4d136ddd4348403e11bf2af92e43dd4dae87d9d970370c795cb966614f6R311

Similarly, in NewConfig:
https://github.com/google/go-cloud/pull/3491/files#diff-0520a4d136ddd4348403e11bf2af92e43dd4dae87d9d970370c795cb966614f6R222

Now if someone add a new query param in any of the place, this test will be able to catch this issue.

khrm:
@vangent Can you review this again?

aws/aws.go
vangent: If it's only used by S3, then let's move this code into blob/s3blob. I.e., see how the "accelerate" option is parsed and used.

aws/aws.go
vangent: Hmm, you're right. But most existing URL parameters are not camelcase, so let's follow that.

I can send a separate PR to fix those.

vangent: Ignore my comment about "no_accelerate" -- that's already supported.

aws/aws.go
khrm: Done that. Please review again.

vangent (approved):

Issue #3496, opened 2024-10-09
aws: disable rate limits by default with AWS


stanhu:
### Describe the bug

After upgrading to go-cloud v0.39.0, one of our users reported:

```
operation error S3: PutObject, failed to get rate limit token, retry quota exceeded, 0 available, 5 requested
```

Terraform also has more details in https://github.com/hashicorp/terraform-provider-aws/issues/36024#issue-2161155725. It seems the AWS team discussed this specific issue in https://github.com/aws/aws-sdk-go-v2/issues/1665#issuecomment-2004339077. 

The bottom line is that SDK v2 adds new client rate limits, and users may hit this after switching from SDK v1. I recommend disabling these rate limits by default (as done in https://github.com/hashicorp/aws-sdk-go-base/pull/977/files) and provide a way to configure them if desired.

### To Reproduce

I haven't tried this yet, but I imagine if you make lots of PutObject requests in quick succession you will hit these client rate limits.

## Expected behavior

No rate limit errors.

### Version

Which version(s) of the Go Cloud modules are you seeing the bug with?

v0.39.0

stanhu:
@vangent Thank you for the quick review and merge!

Do you know when you might want to cut a new tag? Otherwise I'll just use the pseudo-version.

vangent:
done

stanhu:
@vangent Thank you!

PR #3497, opened 2024-10-09
aws: add rate_limiter_capacity to configure client-side rate limits

stanhu:
Issuing many S3 uploads in quick succession causes the following error:                                                                                                                         
                                                                                                                  
```             
operation error S3: PutObject, failed to get rate limit token, retry quota exceeded, 0 available, 5 requested                  
```                                                                                                                          

By default, this rate limit is disabled to restore the same behavior as SDK v1. If `rate_limiter_capacity` is set, then client-side rate limits will take effect.                                                                                                       
                                                                                                                               
Closes https://github.com/google/go-cloud/issues/3496

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -22,6 +22,9 @@ import (
 	"strconv"
 
 	awsv2 "github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/ratelimit"
+	"github.com/aws/aws-sdk-go-v2/aws/retry"
+	"github.com/aws/aws-sdk-go-v2/config"
 	awsv2cfg "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/client"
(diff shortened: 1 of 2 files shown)


aws/aws.go
```diff
@@ -251,5 +254,14 @@ func V2ConfigFromURLParams(ctx context.Context, q url.Values) (awsv2.Config, err
 		opts = append(opts, awsv2cfg.WithEndpointResolverWithOptions(customResolver))
 	}
 
+	// SDK v2 adds client-side rate limiting: https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/retries-timeouts/#client-side-rate-limiting
```
vangent, line 257: Presumably AWS has some good reason for adding this, although they don't explain it in the doc you pointed to. So I'm not sure it's a great idea to disable it by default.

Can you make this a query parameter? (e.g., "disable_ratelimiting") and handle it similar to the others above (add it to the docstring too).

stanhu: > Presumably AWS has some good reason for adding this, although they don't explain it in the doc you pointed to. So I'm not sure it's a great idea to disable it by default.

I think these client side rate limits are causing more harm than good. The Terraform AWS provider disabled this with consultation with AWS: https://github.com/hashicorp/terraform-provider-aws/issues/36024#issuecomment-2008001009

Perhaps a better strategy is to leave this off by default and add the ability to configure the token bucket rate.

vangent: How about we follow what Terraform did, and add a `rate_limiter_capacity` parameter. If the default is 0, then use `None` (I'm fine with changing that to be the default). If there's a value, then create a rate limiter with that capacity. Please link to https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/retries-timeouts/#client-side-rate-limiting in the documentation for the parameter.

stanhu: 👍 Thanks, that's exactly what I think we should do.

PR #3500, opened 2024-10-11
Support atomic writes in the docstore

sandeepvinayak:
docstore/all

Details in the issue : https://github.com/google/go-cloud/issues/3501

See

### DIFF
--- a/docstore/awsdynamodb/dynamo.go
+++ b/docstore/awsdynamodb/dynamo.go
@@ -616,7 +616,7 @@ func revisionPrecondition(doc driver.Document, revField string) (*expression.Con
 	return &cb, nil
 }
 
-// transactWrite execute the write actions in an atomic manner, either they all succeed or they all fail together.
+// transactWrite executes the write actions atomically: either they all succeed or they all fail together.
 func (c *collection) transactWrite(ctx context.Context, actions []*driver.Action, errs []error, opts *driver.RunActionsOptions) {
 	if len(actions) == 0 {
 		return
--- a/docstore/awsdynamodb/dynamo.go
+++ b/docstore/awsdynamodb/dynamo.go
@@ -162,10 +162,12 @@ func (c *collection) RevisionField() string { return c.opts.RevisionField }
 
 func (c *collection) RunActions(ctx context.Context, actions []*driver.Action, opts *driver.RunActionsOptions) driver.ActionListError {
 	errs := make([]error, len(actions))
-	beforeGets, gets, writes, afterGets := driver.GroupActions(actions)
+	beforeGets, gets, writes, writesTx, afterGets := driver.GroupActions(actions)
 	c.runGets(ctx, beforeGets, errs, opts)
 	ch := make(chan struct{})
+	ch2 := make(chan struct{})
 	go func() { defer close(ch); c.runWrites(ctx, writes, errs, opts) }()
+	go func() { defer close(ch2); c.transactWrite(ctx, writesTx, errs, opts) }()
 	c.runGets(ctx, gets, errs, opts)
 	<-ch
 	c.runGets(ctx, afterGets, errs, opts)
@@ -613,25 +615,26 @@ func revisionPrecondition(doc driver.Document, revField string) (*expression.Con
 	return &cb, nil
 }
 
-// TODO(jba): use this if/when we support atomic writes.
-func (c *collection) transactWrite(ctx context.Context, actions []*driver.Action, errs []error, opts *driver.RunActionsOptions, start, end int) {
+func (c *collection) transactWrite(ctx context.Context, actions []*driver.Action, errs []error, opts *driver.RunActionsOptions) {
+	if len(actions) == 0 {
+		return
+	}
 	setErr := func(err error) {
-		for i := start; i <= end; i++ {
-			errs[actions[i].Index] = err
+		for _, a := range actions {
+			errs[a.Index] = err
 		}
 	}
 
+	tws := make([]*dyn.TransactWriteItem, 0, len(actions))
 	var ops []*writeOp
-	tws := make([]*dyn.TransactWriteItem, 0, end-start+1)
-	for i := start; i <= end; i++ {
-		a := actions[i]
-		op, err := c.newWriteOp(a, opts)
+	for _, w := range actions {
+		op, err := c.newWriteOp(w, opts)
 		if err != nil {
-			setErr(err)
-			return
+			errs[w.Index] = err
+		} else {
+			ops = append(ops, op)
+			tws = append(tws, op.writeItem)
 		}
-		ops = append(ops, op)
-		tws = append(tws, op.writeItem)
 	}
 
 	in := &dyn.TransactWriteItemsInput{
--- a/docstore/awsdynamodb/dynamo.go
+++ b/docstore/awsdynamodb/dynamo.go
@@ -170,6 +170,7 @@ func (c *collection) RunActions(ctx context.Context, actions []*driver.Action, o
 	go func() { defer close(ch2); c.transactWrite(ctx, writesTx, errs, opts) }()
 	c.runGets(ctx, gets, errs, opts)
 	<-ch
+	<-ch2
 	c.runGets(ctx, afterGets, errs, opts)
 	return driver.NewActionListError(errs)
 }
@@ -615,10 +616,12 @@ func revisionPrecondition(doc driver.Document, revField string) (*expression.Con
 	return &cb, nil
 }
 
+// transactWrite execute the write actions in an atomic manner, either they all succeed or they all fail together.
 func (c *collection) transactWrite(ctx context.Context, actions []*driver.Action, errs []error, opts *driver.RunActionsOptions) {
 	if len(actions) == 0 {
 		return
 	}
+	// all actions will fail atomically even if a single action fails
 	setErr := func(err error) {
 		for _, a := range actions {
 			errs[a.Index] = err
@@ -630,7 +633,8 @@ func (c *collection) transactWrite(ctx context.Context, actions []*driver.Action
 	for _, w := range actions {
 		op, err := c.newWriteOp(w, opts)
 		if err != nil {
-			errs[w.Index] = err
+			setErr(err)
+			return
 		} else {
 			ops = append(ops, op)
 			tws = append(tws, op.writeItem)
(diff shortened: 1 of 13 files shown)


sandeepvinayak:
@eliben @vangent We are trying to add the transactions support/atomic writes in the docstore and this is the high level idea of how can we support that in go-cloud. https://docs.google.com/document/d/1UVj1kmwDfrs5qm8r7X1p4fAFmsdPEeBHcvWJ8zWF1dY/edit?tab=t.0

The PR is just at a high level to support the idea. 
Let me know what you think about it and we can help to drive it further.

docstore/awsdynamodb/dynamo.go
```diff
@@ -613,25 +615,26 @@ func revisionPrecondition(doc driver.Document, revField string) (*expression.Con
 	return &cb, nil
 }
 
-// TODO(jba): use this if/when we support atomic writes.
-func (c *collection) transactWrite(ctx context.Context, actions []*driver.Action, errs []error, opts *driver.RunActionsOptions, start, end int) {
+func (c *collection) transactWrite(ctx context.Context, actions []*driver.Action, errs []error, opts *driver.RunActionsOptions) {
```
jba, line 618: This function has never been tested or documented properly, so let's go carefully.

First, it needs a doc comment.

docstore/awsdynamodb/dynamo.go
```diff
@@ -613,25 +615,26 @@ func revisionPrecondition(doc driver.Document, revField string) (*expression.Con
...
+	if len(actions) == 0 {
+		return
+	}
 	setErr := func(err error) {
```
jba, line 622: The idea here is that if anything in this function fails, everything fails (since it's atomic). Document that here.

docstore/awsdynamodb/dynamo.go
```diff
@@ -613,25 +615,26 @@ func revisionPrecondition(doc driver.Document, revField string) (*expression.Con
...
-	tws := make([]*dyn.TransactWriteItem, 0, end-start+1)
-	for i := start; i <= end; i++ {
-		a := actions[i]
-		op, err := c.newWriteOp(a, opts)
+	for _, w := range actions {
+		op, err := c.newWriteOp(w, opts)
 		if err != nil {
-			setErr(err)
-			return
+			errs[w.Index] = err
```
jba, line 633: You've changed the semantics here: this is no longer atomic, since if one newWriteOp fails the others still happen.

docstore/docstore.go
```diff
@@ -534,6 +537,12 @@ func (l *ActionList) String() string {
 	return "[" + strings.Join(as, ", ") + "]"
 }
 
+// AtomicWrites causes all following writes in the list to execute atomically.
```
jba, line 540: "to execute as a single atomic operation"

(to avoid confusion with the fact that each write already happens atomically by itself)

docstore/driver/util.go
```diff
@@ -69,7 +70,7 @@ func GroupActions(actions []*Action) (beforeGets, getList, writeList, afterGets
 		} else if a.Kind == Get {
 			// If there was a prior write with this key, make sure this get
 			// happens after the writes.
-			if _, ok := writes[a.Key]; ok {
+			if valueExistsInMaps(a.Key, writes, writesTx) {
```
jba, line 73: We don't need a whole function.
```
if _, ok := writes[a.Key]; ok {
    agets[a.Key] = a
} else if _, ok :=  writesTx[a.Key]; ok {
   agets[a.Key] = a
} else ...
```

docstore/driver/util.go
```diff
@@ -95,7 +100,16 @@ func GroupActions(actions []*Action) (beforeGets, getList, writeList, afterGets
 		return as
 	}
 
-	return vals(bgets), vals(cgets), append(vals(writes), nilkeys...), vals(agets)
+	return vals(bgets), vals(cgets), append(vals(writes), nilkeys...), vals(writesTx), vals(agets)
+}
+
+func valueExistsInMaps(key interface{}, maps ...map[interface{}]*Action) bool {
```
jba, line 106: delete

docstore/driver/util_test.go
```diff
@@ -79,7 +79,7 @@ func TestGroupActions(t *testing.T) {
 	}{
 		{
```
jba, line 80: Please add various cases involving atomic writes, mixed with other writes and gets before, after, and concurrently.

docstore/drivertest/drivertest.go
```diff
@@ -29,6 +29,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
```
jba, line 32: Do not use this package or any other assertion framework.

docstore/drivertest/drivertest.go
```diff
@@ -1900,6 +1901,81 @@ func testMultipleActions(t *testing.T, coll *docstore.Collection, revField strin
...
+	doc := docmap{KeyField: docs[6][KeyField]}
+	_ = coll.Get(ctx, doc)
+	assert.Equal(t, "66", doc["s"])
+	doc = docmap{KeyField: docs[7][KeyField]}
+	_ = coll.Get(ctx, doc)
+	assert.Equal(t, "77", doc["s"])
+	doc = docmap{KeyField: docs[8][KeyField]}
+	_ = coll.Get(ctx, doc)
+	assert.Equal(t, "88", doc["s"])
+}
```
jba, line 1977: Can you add a test where one atomic write fails, and verify that they all fail?

docstore/memdocstore/mem.go
```diff
@@ -191,7 +191,7 @@ func (c *collection) RunActions(ctx context.Context, actions []*driver.Action, o
 		}
 	}
 
-	beforeGets, gets, writes, afterGets := driver.GroupActions(actions)
+	beforeGets, gets, writes, _, afterGets := driver.GroupActions(actions)
```
jba, line 194: You need to implement atomic writes for this, mongo and firestore.

jba (changes requested):

docstore/memdocstore/mem.go
sandeepvinayak: yes, as we discussed planned, we will be delivering it immediately after aws and driver is in. We will do the azure, gco and mongo
https://github.com/google/go-cloud/issues/3501#issuecomment-2435614003

jba: OK, thanks, I forgot.

docstore/awsdynamodb/dynamo.go
sandeepvinayak: Sure, let me revert it, this was to handle the cases when the write operation is non-write kind. But I believe it makes to fail here to be semantically correct.

sandeepvinayak: Sure, let me revert it. It was added to handle the case if the action is non-write kind which will be somebug in the code. But I got it that it's making it semantically incorrect.

docstore/awsdynamodb/dynamo.go
```diff
@@ -162,10 +162,12 @@ func (c *collection) RevisionField() string { return c.opts.RevisionField }
...
 	errs := make([]error, len(actions))
-	beforeGets, gets, writes, afterGets := driver.GroupActions(actions)
+	beforeGets, gets, writes, writesTx, afterGets := driver.GroupActions(actions)
 	c.runGets(ctx, beforeGets, errs, opts)
 	ch := make(chan struct{})
+	ch2 := make(chan struct{})
 	go func() { defer close(ch); c.runWrites(ctx, writes, errs, opts) }()
+	go func() { defer close(ch2); c.transactWrite(ctx, writesTx, errs, opts) }()
 	c.runGets(ctx, gets, errs, opts)
 	<-ch
```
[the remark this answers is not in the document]
sandeepvinayak: done.

sandeepvinayak:
@jba thanks for your feedback, I addressed all of the comments, please take another look

docstore/awsdynamodb/dynamo.go
```diff
@@ -615,10 +616,12 @@ func revisionPrecondition(doc driver.Document, revField string) (*expression.Con
 	return &cb, nil
 }
 
+// transactWrite execute the write actions in an atomic manner, either they all succeed or they all fail together.
```
jba, line 619: s/execute/executes/
s/in an atomic manner/atomically/
s/,/:/

docstore/drivertest/drivertest.go
```diff
@@ -1967,13 +1966,90 @@ func testAtomicWrites(t *testing.T, coll *docstore.Collection, revField string)
 
 	doc := docmap{KeyField: docs[6][KeyField]}
 	_ = coll.Get(ctx, doc)
-	assert.Equal(t, "66", doc["s"])
+	cmpDiff("66", doc["s"])
```
jba, line 1969: You have to check the return value of cmpDiff.

docstore/drivertest/drivertest.go
```diff
@@ -1967,13 +1966,90 @@ func testAtomicWrites(t *testing.T, coll *docstore.Collection, revField string)
...
+	}
+	compare(gdocs, docs[3:6])
+
+	// Get the docs updated as part of atomic writes and verify that it didn't get written.
+	actions = coll.Actions()
+
+	// validate that the mods didn't get applied to all docs in transaction.
+	doc := docmap{KeyField: docs[6][KeyField]}
+	_ = coll.Get(ctx, doc)
+	cmpDiff("6", doc["s"])
```
jba, line 2049: First, change the "6" to "66" and make sure your test fails.
It won't.
Now check the results of cmpDiff.

I recommend you use this technique all the time: first make sure the test fails, to prove that you're testing what you think. Then make it pass.

jba (changes requested):

sandeepvinayak:
@jba, I addressed your latest 2 comments

docstore/drivertest/drivertest.go
sandeepvinayak: sounds good, validated that the test didn't fail and fixed it the diff check.

jba (approved):

sandeepvinayak:
@jba the codecov report is under-reporting because the conformance tests are disabled for now but I verified the tests we added are good. I will enable the 2 conformance tests added along with with the other providers in the follow up PR.

jba:
Thanks for the PR! Looking forward to the follow-ups.

PR #3508, opened 2024-11-19
Docstore/memdocstore: nested query

### LINKED ISSUE #3506: docstore/memdocstore query for nested documents with slices does not work
### Describe the bug

Documents like 
```go
docmap{
	"goals": []any{
		docmap{"id": "1145755"},
		docmap{"id": "8336474"},
		docmap{"id": "1474479"},
	},
}
```
are not found by the following query:
```go
collection.Query().Where("goals.id", "=", "1145755")
```

This is working fine when using `docstore/mongodocstore`.

### To Reproduce

I forked this repository and added a unit test to show the issue eqinox76/go-cloud@7f3ef044a00cc76d258147fd9e1b529e0b4f8ad9.

## Expected behavior

Find the document.

### Version

v0.39.0

### Additional context

eqinox76/go-cloud@7f3ef044a00cc76d258147fd9e1b529e0b4f8ad9 contains a potential fix. If it looks good to you i could make a pr.

eqinox76:
Fixes #3506 docstore/memdocstore query for nested documents with slices does not work

Added a test to the drivertest. It is working fine with the fixed memdocstore implementation but i can't test it against the other cloud providers.


vangent:
can you update your PR to remove the drivertest changes and add that option? You should be able to add it to the existing Options struct

docstore/memdocstore/mem.go
```diff
@@ -68,6 +68,12 @@ type Options struct {
 	// When the collection is closed, its contents are saved to the file.
 	Filename string
 
+	// AllowNestedSlicesQuery allows querying with nested slices.
```
vangent, line 71: Nit: How about "AllowNestedSliceQueries"?

docstore/memdocstore/mem.go
```diff
@@ -399,16 +405,34 @@ func (c *collection) checkRevision(arg driver.Document, current storedDoc) error
 
 // getAtFieldPath gets the value of m at fp. It returns an error if fp is invalid
 // (see getParentMap).
-func getAtFieldPath(m map[string]interface{}, fp []string) (interface{}, error) {
-	m2, err := getParentMap(m, fp, false)
-	if err != nil {
-		return nil, err
+func getAtFieldPath(m map[string]interface{}, fp []string, nested bool) (result interface{}, err error) {
+
+	var get func(m interface{}, name string) interface{}
```
vangent, line 410: Nit: You can drop this "var" line and just do "get := ..." on the next line.

docstore/memdocstore/mem.go
```diff
@@ -399,16 +405,34 @@ func (c *collection) checkRevision(arg driver.Document, current storedDoc) error
...
+			return concrete[name]
+		case []interface{}:
+			if !nested {
+				return nil
+			}
+			result := []interface{}{}
+			for _, e := range concrete {
+				result = append(result, get(e, name))
+			}
+			return result
```
vangent, line 423: So if there are no matches, this will return an empty slice, which is not caught by the "== nil" check below so it will be returned.

Is that what you want?

docstore/memdocstore/mem_test.go
```diff
@@ -129,6 +130,49 @@ func TestUpdateAtomic(t *testing.T) {
...
+	doc := docmap{drivertest.KeyField: "TestQueryNested",
+		"list":             []any{docmap{"a": "A"}},
+		"map":              docmap{"b": "B"},
+		dc.RevisionField(): nil,
+	}
+	if err := coll.Put(ctx, doc); err != nil {
+		t.Fatal(err)
+	}
+
+	got := count(coll.Query().Where("list.a", "=", "A").Get(ctx))
```
vangent, line 166: How about a couple of nested tests (i.e., a list of maps, and a map of lists)?

And a test for what happens when you query a nested list but there aren't any matchers (see my other question).

docstore/memdocstore/query.go
```diff
@@ -138,6 +138,19 @@ func compare(x1, x2 interface{}) (int, bool) {
 		}
 		return -1, true
 	}
```
vangent, line 140: Can you add a comment here similar to the one above? I don't follow why you added this. What breaks if you don't?

vangent (changes requested):

jba (changes requested):

docstore/memdocstore/mem.go
```diff
@@ -422,14 +446,6 @@ func setAtFieldPath(m map[string]interface{}, fp []string, val interface{}) erro
 	return nil
 }
 
-// Delete the value from m at the given field path, if it exists.
```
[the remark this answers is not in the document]
eqinox76: According to go-staticcheck it is not used. As it is a private function i trusted that check. Shall we keep it?

docstore/memdocstore/mem.go
```diff
@@ -399,16 +399,31 @@ func (c *collection) checkRevision(arg driver.Document, current storedDoc) error
 
 // getAtFieldPath gets the value of m at fp. It returns an error if fp is invalid
 // (see getParentMap).
-func getAtFieldPath(m map[string]interface{}, fp []string) (interface{}, error) {
-	m2, err := getParentMap(m, fp, false)
-	if err != nil {
-		return nil, err
+func getAtFieldPath(m map[string]interface{}, fp []string) (result interface{}, err error) {
```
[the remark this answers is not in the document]
eqinox76: Done

docstore/memdocstore/mem.go
eqinox76: Sounds good.

docstore/memdocstore/mem.go
eqinox76: Done

docstore/memdocstore/mem_test.go
eqinox76: I added more tests. And yes your other point should have been covered by a test.

eqinox76:
Thank you @vangent and @jba for your reviews! I amended the pr and i think only the question about the removed function is open. Please let me know if i missed anything.

docstore/memdocstore/mem.go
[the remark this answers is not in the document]
jba: No, that's fine.

vangent:
@jba I'd like to wait for your review/LGTM.

jba:
I'll try to get to it this weekend/early next week.

docstore/memdocstore/mem.go
```diff
@@ -68,6 +68,12 @@ type Options struct {
 	// When the collection is closed, its contents are saved to the file.
 	Filename string
 
+	// AllowNestedSliceQueries allows querying with nested slices.
+	// This makes the memdocstore more compatible with MongoDB,
+	// but other providers may not support this feature.
+	// See https://github.com/google/go-cloud/pull/3511 for more details.
```
jba, line 74: Don't refer to a PR for documentation. Make the doc self-contained by explaining how this changes the behavior, that is, precisely what "querying with nested slices" means.

You can refer to the PR in an implementation comment so developers can understand the context.

docstore/memdocstore/mem.go
```diff
@@ -399,16 +405,33 @@ func (c *collection) checkRevision(arg driver.Document, current storedDoc) error
 
 // getAtFieldPath gets the value of m at fp. It returns an error if fp is invalid
 // (see getParentMap).
-func getAtFieldPath(m map[string]interface{}, fp []string) (interface{}, error) {
-	m2, err := getParentMap(m, fp, false)
-	if err != nil {
-		return nil, err
+func getAtFieldPath(m map[string]any, fp []string, nested bool) (result any, err error) {
```
jba, line 408: Add doc explaining `nested`.

docstore/memdocstore/mem_test.go
```diff
@@ -129,6 +130,72 @@ func TestUpdateAtomic(t *testing.T) {
...
+		"listOfMaps":       []any{docmap{"id": "1"}, docmap{"id": "2"}, docmap{"id": "3"}},
+		"mapOfLists":       docmap{"ids": []any{"1", "2", "3"}},
+		"deep":             []any{docmap{"nesting": []any{docmap{"of": docmap{"elements": "yes"}}}}},
+		dc.RevisionField(): nil,
+	}
+	if err := coll.Put(ctx, doc); err != nil {
+		t.Fatal(err)
+	}
+
+	got := count(coll.Query().Where("list.a", "=", "A").Get(ctx))
```
jba, line 169: I think the tests would be improved by actually comparing the results. Also, use table-driven style. See [testGetQuery](https://github.com/google/go-cloud/blob/master/docstore/drivertest/drivertest.go#L1419) for an example.

docstore/memdocstore/query.go
```diff
@@ -138,6 +138,20 @@ func compare(x1, x2 interface{}) (int, bool) {
...
+	// for AllowNestedSliceQueries
+	// when querying for x2 in the document and x1 is a list of values we only need one value to match
+	if v1.Kind() == reflect.Slice {
+		for i := 0; i < v1.Len(); i++ {
+			if c, ok := compare(x2, v1.Index(i).Interface()); ok {
+				if !ok {
```
jba, line 146: We already know ok is true from the previous line.

jba (changes requested):
Thanks for doing this work. Sorry it took me a while to give it careful look. The only major change is improving the test, otherwise looks good!

jba (approved):

vangent:
@jba when you have a chance can you re-review, sounds like there have been non-trivial changes since your approval.

eqinox76:
Thanks for the review. Shadowing the variable name makes it actually more readable. I commit the changes and fixed the comment style. Please let me know when there is something else that i missed.

PR #3511, opened 2024-11-22
docstore: Clarify docs for nested field paths

vangent:


docstore/doc.go
```diff
@@ -212,6 +212,12 @@
 //
 //	iter := coll.Query().Where("Author.Last", "=", "Bulgakov").Limit(3).Get(ctx)
 //
+// Dot-separated paths work as you'd expect for nested structs and maps, addressing
+// the nested struct field or map key. For nested slices, some providers (notably
+// MongoDB) treat s.f, where s is a sequence, as the slice of e.f for each e in s.
+// However, others do not support field selectors inside slices. If you need fully
+// portable code, you cannot rely on this feature.
```
jba, line 219: We ended up not merging the PR, so I would say: "However, others do not support field selectors inside slices, so docstore does not support them."

jba (approved):

docstore/doc.go
jba: Sorry, I hadn't seen your comment on the bug. I think this should say we don't support it, but then the memdocstore option should explain that it's not portable and will only work on mongodb.

PR #3513, opened 2024-12-01
aws, blob: Add support for anonymous query param for gcsblob and all AWSv2 providers

### LINKED ISSUE #3512: s3blob: Need a way to set custom credentials provider for S3 when using v2
### Describe the bug

The aws-sdk-go (v1) module is being removed soon, and so I'm trying to migrate to aws-sdk-go-v2 (see also #3489).

I have a need to sometimes use the anonymous CredentialsProvider, and with V1 I could achieve this by setting the `ConfigProvider` field with an AWS session properly configured with the anonymous provider.

I'm using the `blob.BucketURLOpener` interface to support non-S3 buckets as well via URLs, and so using `s3blob.OpenBucketV2` with a custom client isn't an option.

### To Reproduce

Attempt to use `&s3blob.URLOpener{UseV2: true}` to open S3 buckets with anonymous credential...

## Expected behavior

There are a few ways to achieve this - one way would be to allow adding additional `config.LoadOptions` items to be provided to `s3blob.V2ConfigFromURLParams` via a new field in the `s3blob.URLOpener` struct.

Alternately, a new URL query parameter could be supported - something like `anonymous=true`. FWIW, that would be useful in the `gcsblob` package as well.

### Version

v0.40.0 (latest)

### Additional context

n/a

vangent:
Fixes #3512.

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -195,6 +195,7 @@ func NewDefaultV2Config(ctx context.Context) (awsv2.Config, error) {
 //
 // The following query options are supported:
 //   - region: The AWS region for requests; sets WithRegion.
+//   - anonymous: A value of "true" forces use of anonymous credentials.
 //   - profile: The shared config profile to use; sets SharedConfigProfile.
 //   - endpoint: The AWS service endpoint to send HTTP request.
 //   - hostname_immutable: Make the hostname immutable, only works if endpoint is also set.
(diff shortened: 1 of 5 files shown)


hairyhenderson (approved):
Thanks for adding this so quickly @vangent! 🙏

Issue #3489, opened 2024-10-04
Removing AWS SDK v1 (targeted early 2025)


khrm:
Amazon has announced the End of Support for AWS SDK for Go (v1). Can we now drop this library from the dependencies? I see that the default behaviour is now sdk v2.

Is any help required to do this before July 31st 2025?

https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/

vangent:
I announced on the latest release that we plan to remove V1 support in early 2025:

https://groups.google.com/g/go-cloud/c/I5XUddBHLCU

We can use this bug to track it.

S-YOU:
@vangent, @khrm 

Any update on this? some vulnerability scans raising some CVEs on SDKv1 that depends on this library.

![Image](https://github.com/user-attachments/assets/96d85814-fe72-46e6-a94a-07e4037bc8de)

vangent:
No progress as of yet. I am hoping https://github.com/google/go-cloud/pull/3519 will land soon so v1 can be removed entirely.

Note that if you don't use AWSv1, you shouldn't be exposed to any vulnerabilities in v1 libraries.

S-YOU:
> Note that if you don't use AWSv1, you shouldn't be exposed to any vulnerabilities in v1 libraries.

Thats correct but by term of vulnerabilities checkers like the one AWS provided, they check signatures on binary, having a Struct reference to AWS V1 will include all related dependencies from AWS V1 and that will still triggered those alerts.

But hopfully there seems to be some progress. Looking forward to it.

jba:
I have bandwidth this week to get this done.

vangent:
Support is dropped for everything except `docstore` at this point.

vangent:
`docstore` just got `v2` support, so I'll make a release with that, and then drop AWSv1 for the next release.

PR #3519, opened 2025-01-31
Docstore/awsdynamodb: Supporting the AWS V2 SDK

### LINKED ISSUE #3458: docstore/awsdynamodb: Support AWS SDK V2

The V1 AWS SDK has officially entered maintenance mode (31st July 2024).
In addition, the vendor specific constructors for AWS are inconsistent between using the V1 and V2 SDK.

I propose migrating all modules to use the V2 SDK both for consistency when using the vendor specific constructors, but also to address future maintenance considerations as the V1 SDK continues to be deprecated and diverges from V2.

The V1 support could be retained for some time, as the V1 SDK is flagged as receiving maintenance for another year.
But this seems the kind of edit likely to accumulate in the amount of effort needed to make the migration over time.

### Additional context

I've created a [fork](https://github.com/Arlula/go-cloud) and will be investigating the necessary changes there. My intent is to try and keep any necessary changes as minimal as feasible, and created this issue for discussion of anything I find along the way needing input, and to provide advance warning for a future PR.

Maldris:
Resolved #3458 

A somewhat quick and dirty conversion to using the AWS V2 SDK rather than the deprecated V1.

Due to the delays in my finding time to work on this I prioritised getting the functionality migrated to mitigate any risk from API drift form the deprecated V1 SDK.

The main sections that warrant further discussion:
 - codec behaviour could be handled better with type switches now, but may be a follow up task
 - url opener doesn't use a configuration override any more, most of this is handled by `V2ConfigFromURLParams`, but is worth review
 - error handling using an As pattern and I've declared all the as targets in a block which is likely less memory efficient, but simple to read, should another approach be used?


docstore/awsdynamodb/create_tables.sh
```diff
@@ -36,6 +37,7 @@ aws dynamodb create-table \
 # The docstore-test-2 table has both a partition and a sort key, and two indexes.
 
 aws dynamodb create-table \
+  --region us-east-2 \
```
jba, line 40: Why is this line different?

docstore/awsdynamodb/urls.go
```diff
@@ -75,8 +69,6 @@ const Scheme = "dynamodb"
 // See https://godoc.org/gocloud.dev/aws#ConfigFromURLParams for supported query
 // parameters for overriding the aws.Session from the URL.
 type URLOpener struct {
-	// ConfigProvider must be set to a non-nil value.
```
jba, line 78: This is a breaking change.

docstore/awsdynamodb/urls.go
```diff
@@ -115,25 +107,21 @@ func (o *URLOpener) processURL(u *url.URL) (db *dyn.DynamoDB, tableName, partiti
...
 	if err != nil {
 		return nil, "", "", "", nil, fmt.Errorf("open collection %s: %v", u, err)
 	}
 	return db, tableName, partitionKey, sortKey, opts, nil
 }
 
 // Dial gets an AWS DynamoDB service client.
-func Dial(p client.ConfigProvider) (*dyn.DynamoDB, error) {
-	if p == nil {
+func Dial(p aws.Config) (*dyn.Client, error) {
```
jba, line 122: This is a breaking change.

jba (commented):
We appreciate the effort you put into this. However, there is a lot of changed code here, including at least two breaking changes. How does the docstore user benefit from the switch to the new API?

Note that the breaking changes are unlikely to be approved, because they would require a v2 of the main module.

Maldris:
Hi Johnathan,

Last year the AWS SDK V1 was deprecated ([see announcement about maintenance and end of support in project readme](https://github.com/aws/aws-sdk-go)) with its maintenance grace period set to end on 7/31/2025, this PR was created to address that and move the package to using the new SDK as all the other modules already have.

This was discussed in the flagged issue #3458
Forms part of the larger body of work supporting #3489 
And follows all the other modules being moved to use V2 as the default (#3465 and #3466)

`docstore/awsdynamodb` was the only module without a V2 implementation, and as seen in #3458 was discussed as being better served as a breaking replacement ahead of removing V1 from other modules, rather than the more complicated task of building a complete abstraction of the behaviour to support both, or having duplicated code for both.

I'll be adding comments on the three sections you've specifically flagged as well.

docstore/awsdynamodb/urls.go
Maldris: This has changed to the equivalent type used for AWS SDK V2, again, please see the comment below on why that change has already been discussed and is necessary.

docstore/awsdynamodb/urls.go
Maldris: This was flagged in the original message soliciting feedback on the best way to address, as omitting it is in keeping with several other V2 implementations, but regardless the type will change to support V2

docstore/awsdynamodb/create_tables.sh
Maldris: The unit tests hard code the region as us-east-2, but the creation script used to set up the test db's don' t specify a region, falling back to the user default, this was to improve consistency in testing

Maldris:
Apologies for the mentions but as its been a week and I've been seeing notification issues in other repo's this week;
so just to be sure @jba 
and because the prior conversation about this was with you, @vangent

jba:
@Maldris, I read #3458 as concluding that a [v2 fork](https://github.com/google/go-cloud/issues/3458#issuecomment-2276383789) was the right approach. I'm in favor of that.

Maldris:
ok, did a little history re-writing to move things into `docstore/awsdynamodbv2` while preserving change history a bit better (though it forced changing all the commit dates to today to keep it simpler)

I still have the other variant on a backup branch locally, but does this pass requirements better @jba ?

jba:
Can we make the path `docstore/awsdynamodb/v2`? That's typically what we do. For example, `math/rand` and `math/rand/v2`.

vangent:
Can you merge this with HEAD and push it forward? It's the only thing blocking removal of V1 entirely.

Maldris:
merged and updated with the changes to the docstore test harness added in V1 that needed to be replicated in V2.

jba:
@Maldris looks like we still have some failing checks.

youngbupark:
@Maldris @jba Do we have any update for this PR? [aws-sdk-go will be end of support](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/) in < 2months.

jba (approved):

vangent:
I cloned this and made some fixes, see https://github.com/google/go-cloud/pull/3602.

vangent:
No worries, and thanks for the contribution! Hope things are solid enough you can get back to work soon.

PR #3524, opened 2025-02-27
all: upgrade dependencies to resolve `golang.org/x/oauth2/jws` vulnerability

thomasschafer:
As shown [here](https://security.snyk.io/vuln/SNYK-GOLANG-GOLANGORGXOAUTH2JWS-8749594), `golang.org/x/oauth2/jws` versions below 0.27.0 have a sev. 8.7 vulnerability. This PR bumps the versions of those dependencies.

I also had to bump the Go version to enable the upgrade:

```sh
go: cloud.google.com/go/auth@v0.15.0 requires go@1.23.0, but 1.21.0 is requested
```


thomasschafer:
Getting a 400 error when I try and sign the CLA, I agree to the terms of the CLA if it's possible to override the blocked pipeline, otherwise I'm happy to sign when the form is fixed!

Screenshot 2025-02-27 at 13 33 05

vangent:
We generally prefer to update dependencies in bulk periodically; you can of course update the dependency in your application to avoid the vulnerability.

Issue #3518, opened 2025-01-29
blob/s3blob: add support for conditional writes


SoMuchForSubtlety:
### Is your feature request related to a problem? Please describe.

S3 recently added support for conditional writes by setting the `If-None-Match` or `If-Match`  headers. This is great for managing concurrent operations.

### Describe the solution you'd like

Exposing this option through the standard API would be great

### Describe alternatives you've considered

GCS also supports request preconditions[^2].  `If-None-Match: $key` and `ifSourceGenerationMatch=0` could be mapped to a common API, but this will only cover a subset of the possible use cases.

[^1]: https://aws.amazon.com/about-aws/whats-new/2024/11/amazon-s3-functionality-conditional-writes/
[^2]: https://cloud.google.com/storage/docs/request-preconditions

vangent:
Does Azure support it?

SoMuchForSubtlety:
[Azure](https://learn.microsoft.com/en-us/rest/api/storageservices/specifying-conditional-headers-for-blob-service-operations#:~:text=specified%20in%20quotes.-,If%2DNone%2DMatch,-An%20ETag%20value) and [Minio](https://blog.min.io/leading-the-way-minios-conditional-write-feature-for-modern-data-workloads/) both support `If-None-Match` and `If-Match`.

ocampeau:
I also vote for this issue. Is this something that would be accepted? I can submit a PR if no one else is working on this.

vangent:
Fixed

PR #3533, opened 2025-03-31
blob: introduce conditional write option for storage

ocampeau:
This commit introduces support for conditional writes for major cloud providers: aws, gcp, azure.

It does so by adding a new write options: `IfNotExist`. 

When set to true, conditions are set for each cloud storage driver to enable the desired behavior.

Addresses #3518

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -1098,6 +1098,7 @@ func (b *Bucket) NewWriter(ctx context.Context, key string, opts *WriterOptions)
 		MaxConcurrency:              opts.MaxConcurrency,
 		BeforeWrite:                 opts.BeforeWrite,
 		DisableContentTypeDetection: opts.DisableContentTypeDetection,
+		IfNotExist:                  opts.IfNotExist,
 	}
 	if len(opts.Metadata) > 0 {
 		// Services are inconsistent, but at least some treat keys
@@ -1422,6 +1423,13 @@ type WriterOptions struct {
 	// asFunc converts its argument to driver-specific types.
 	// See https://gocloud.dev/concepts/as/ for background information.
 	BeforeWrite func(asFunc func(any) bool) error
+
+	// IfNotExist is used for conditional writes. When set to 'true',
+	// if a blob exists for the same key in the bucket, the write
+	// operation won't succeed and the current blob for the key will
+	// be left untouched. An error for which gcerrors.Code will return
+	// gcerrors.PreconditionFailed will be returned by Write or Close.
+	IfNotExist bool
 }
 
 // CopyOptions sets options for Copy.
--- a/blob/driver/driver.go
+++ b/blob/driver/driver.go
@@ -110,6 +110,11 @@ type WriterOptions struct {
 	// asFunc allows drivers to expose driver-specific types;
 	// see Bucket.As for more details.
 	BeforeWrite func(asFunc func(any) bool) error
+
+	// IfNotExist is used for conditional writes.
+	// When set to true, if a blob exists for the same key in the bucket, the write operation
+	// won't take place.
+	IfNotExist bool
 }
 
 // CopyOptions controls options for Copy.
--- a/blob/drivertest/drivertest.go
+++ b/blob/drivertest/drivertest.go
@@ -260,6 +260,9 @@ func RunConformanceTests(t *testing.T, newHarness HarnessMaker, asTests []AsTest
 	t.Run("TestSignedURL", func(t *testing.T) {
 		testSignedURL(t, newHarness)
 	})
+	//t.Run("TestIfNotExist", func(t *testing.T) {
+	//	testIfNotExist(t, newHarness)
+	//})
 	asTests = append(asTests, verifyAsFailsOnNil{})
 	t.Run("TestAs", func(t *testing.T) {
 		for _, st := range asTests {
@@ -2731,6 +2734,73 @@ func testAs(t *testing.T, newHarness HarnessMaker, st AsTest) {
 		if err != nil && gcerrors.Code(err) != gcerrors.Unimplemented {
 			t.Errorf("got err %v when signing url with method %q", err, method)
 		}
+	}
+}
+
+func testIfNotExist(t *testing.T, newHarness HarnessMaker) {
+	t.Helper()
+
+	const key = "blob-for-if-not-exist"
+	const contents = "up and down"
+
+	ctx := context.Background()
+	h, err := newHarness(ctx, t)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer h.Close()
+	drv, err := h.MakeDriver(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	b := blob.NewBucket(drv)
+	defer func() { _ = b.Close() }()
+
+	opts := blob.WriterOptions{
+		ContentType: "text",
+		IfNotExist:  true,
+	}
+
+	// Create one file for the key
+	w1, err := b.NewWriter(ctx, key, &opts)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer func() {
+		_ = b.Delete(ctx, key)
+	}()
+
+	// Write to the file
+	if _, err := w1.Write([]byte(contents)); err != nil {
+		t.Fatal(err)
+	}
+
+	// Closing the file (ie: thus writing the content to the file)
+	// should not return an error
+	if err := w1.Close(); err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a new writer for the same key
+	w2, err := b.NewWriter(ctx, key, &opts)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Write to the file
+	// We expect an error either from `Write` or `Close`
+	if _, err = w2.Write([]byte(contents)); err == nil {
+		err = w2.Close()
+	} else {
+		_ = w2.Close()
+	}
+
+	if err == nil {
+		t.Error("expected error rewriting key with IfNotExist, got nil")
+	}
+	if code := gcerrors.Code(err); code != gcerrors.FailedPrecondition {
+		t.Errorf("expected FailedPrecondition error, got %v", code)
 	}
 }
 
--- a/blob/fileblob/fileblob.go
+++ b/blob/fileblob/fileblob.go
@@ -78,6 +78,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"gocloud.dev/blob"
@@ -739,9 +740,11 @@ func (b *bucket) NewTypedWriter(ctx context.Context, key, contentType string, op
 
 	if b.opts.Metadata == MetadataDontWrite {
 		w := &writer{
-			ctx:  ctx,
-			File: f,
-			path: path,
+			ctx:        ctx,
+			File:       f,
+			path:       path,
+			ifNotExist: opts.IfNotExist,
+			mu:         &sync.Mutex{},
 		}
 		return w, nil
 	}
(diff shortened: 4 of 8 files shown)


blob/blob.go
```diff
@@ -1422,6 +1423,13 @@ type WriterOptions struct {
 	// asFunc converts its argument to driver-specific types.
 	// See https://gocloud.dev/concepts/as/ for background information.
 	BeforeWrite func(asFunc func(any) bool) error
+
+	// IfNotExist is used for conditional writes. When set to true, if
+	// a blob exists for the same key in the bucket, the write operation
+	// won't take place.
+	// It only works for AWS S3, Google Storage and Azure. Setting this to
```
vangent, line 1430: This generally isn't OK; this is supposed to be a portable library. Can you implement this option for fileblob and memblob too? It shouldn't be too hard.

blob/driver/driver.go
```diff
@@ -110,6 +110,13 @@ type WriterOptions struct {
 	// asFunc allows drivers to expose driver-specific types;
 	// see Bucket.As for more details.
 	BeforeWrite func(asFunc func(any) bool) error
+
+	// IfNotExist is used for conditional writes. When set to true, if
+	// a blob exists for the same key in the bucket, the write operation
+	// won't take place.
+	// It only works for AWS S3, Google Storage and Azure. Setting this to
```
vangent, line 117: Same comment, please delete these two lines.

vangent (changes requested):
Can you add a drivertest that exercises this new feature?

It'll be tricky for you to verify on GCP/Azure/AWS, but you should be able to get it working with memblob and fileblob. Once you've done that, you can comment it out, we can get this merged, and then I can generate the goldens for the GCP/Azure/AWS packages.

blob/blob.go
```diff
@@ -1422,6 +1423,13 @@ type WriterOptions struct {
 	// asFunc converts its argument to driver-specific types.
 	// See https://gocloud.dev/concepts/as/ for background information.
 	BeforeWrite func(asFunc func(any) bool) error
+
+	// IfNotExist is used for conditional writes. When set to true, if
+	// a blob exists for the same key in the bucket, the write operation
```
vangent, line 1428: Please specify exactly what happens (and verify in the drivertest).

* Does the write operation succeed and just silently not write anything?
* Or does the write operation return an error? If so, what gcerrors error code does it return?

blob/blob.go
ocampeau: Sure! Will do!

blob/driver/driver.go
ocampeau: Sure

blob/blob.go
ocampeau: Done!

ocampeau:
Thanks @vangent  for the quick feedback. I have provided the implementation for the `fileblob` and `memblob` driver. I also added a test in the `drivertest` package.

Following your suggestion, I ran the tests for the `memblob` and `fileblob` package, and they all passed. I haven't been able to test for the cloud providers though. I have commented out the test so you can run them in your CI as mentioned.

Let me know if there is anything else that needs to be fixed.

blob/drivertest/drivertest.go
```diff
@@ -2734,6 +2737,70 @@ func testAs(t *testing.T, newHarness HarnessMaker, st AsTest) {
...
+//
+//	// Create a new writer for the same key
+//	w2, err := b.NewWriter(ctx, key, &opts)
+//	if err != nil {
+//		t.Fatal(err)
+//	}
+//
+//	// Write to the file
+//	if _, err := w2.Write([]byte(contents)); err != nil {
+//		t.Fatal(err)
```
vangent, line 2793: The test needs to handle the case where Write returns the error, and also verify the PreconditionFailed.

Something like:

var err error
if _, err = w2.Write([]byte(contents)); err == nil {
  err = w2.Close()
}
if err == nil {
  t.Error("expected error rewriting key with IfNotExist, got nil")
}
if code := gcerrors.Code(err); code != gcerrors.PreconditionFailed {
t.Errorf("expected PreconditionFailed error, got %v", code)
}

(I believe this test fails right now for memblob, because you didn't return PreconditionFailed).

vangent (changes requested):

blob/fileblob/fileblob.go
```diff
@@ -794,33 +797,55 @@ func (w *writerWithSidecar) Write(p []byte) (n int, err error) {
...
+		// Set offset at the beginning of the file
+		_, err = w.f.Seek(0, 0)
+		if err != nil {
+			return err
+		}
+
+		// Copy content from the temp file to the new file
+		_, err = w.f.WriteTo(fd)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
ocampeau: Unless I am mistaken, your suggestion could lead to a race condition if multiple process concurrently tries to write to the same key. I implemented it in a way that prevents this, using the `O_EXCL` flag when opening the file.

I can also add a mutex to the writer if that is better?

blob/fileblob/fileblob.go
```diff
@@ -794,33 +797,55 @@ func (w *writerWithSidecar) Write(p []byte) (n int, err error) {
 	return n, nil
 }
 
-func (w *writerWithSidecar) Close() error {
-	err := w.f.Close()
-	if err != nil {
-		return err
-	}
+func (w *writerWithSidecar) Close() (err error) {
```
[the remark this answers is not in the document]
ocampeau: This is related to this comment: https://github.com/google/go-cloud/pull/3533/files#r2025692083

With the implementation I went for, it is required to not close this file before writing to the new file.

blob/fileblob/fileblob.go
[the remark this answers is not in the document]
vangent: Hmm, interesting, I see why you did that.

I don't think adding a mutex helps, as you say it could be another process.

The problem is that opening the file directly for write (rather than writing a temp file and moving it) has other down sides. E.g., you could create the file and then crash, leaving a partially-written file. That's why we're doing the tempfile/rename thing. And I think it would be weird if that behavior changed based on whether you enable IfNotExists.

I'm not sure I have a better answer though; maybe we can just accept and document the race.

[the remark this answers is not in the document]
ocampeau: I just submitted a new implementation using a mutex. The `memblob` package was already using a mutex, so I based myself off of that. let me know what you think.

blob/fileblob/fileblob.go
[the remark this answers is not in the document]
ocampeau: I have put it back, since I changed the implementation. See comment here: https://github.com/google/go-cloud/pull/3533#discussion_r2031429879

blob/memblob/memblob.go
```diff
@@ -355,6 +357,9 @@ func (w *writer) Close() error {
 	w.b.mu.Lock()
 	defer w.b.mu.Unlock()
 	if prev := w.b.blobs[w.key]; prev != nil {
+		if w.ifNotExist {
+			return errors.New("'IfNotExists' is set and blob already exists for key")
```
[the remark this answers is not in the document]
ocampeau: True! Updated!

blob/blob.go
```diff
@@ -1422,6 +1423,20 @@ type WriterOptions struct {
...
+	// bucket, the write operation won't succeed and the current blob
+	// for the key will be left untouched.
+	// For 'fileblob' and 'memblob', an error of type 'PreconditionFailed'
+	// will be returned by the 'Close' method. This is due to how those driver
+	// are implemented, which buffers the content during writes and perform
+	// the actual write at close time.
+	// For cloud providers (aws, azure, google), an error of type 'PreconditionFailed'
+	// will be returned by the 'Write' method, since a call to 'Write' results
+	// in an API call to the cloud provider.
+	// Default is set to 'false'.
```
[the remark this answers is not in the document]
ocampeau: Done!

blob/drivertest/drivertest.go
```diff
@@ -2734,6 +2737,70 @@ func testAs(t *testing.T, newHarness HarnessMaker, st AsTest) {
 	}
 }
 
+//func testIfNotExist(t *testing.T, newHarness HarnessMaker) {
```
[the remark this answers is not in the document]
ocampeau: I wasn't sure, since you previously said:

> Once you've done that, you can comment it out, we can get this merged, and then I can generate the goldens for the GCP/Azure/AWS packages.

Anyhow, it is uncommented now.

blob/drivertest/drivertest.go
ocampeau: Done! Good catch!

blob/blob.go
```diff
@@ -1422,6 +1423,14 @@ type WriterOptions struct {
 	// asFunc converts its argument to driver-specific types.
 	// See https://gocloud.dev/concepts/as/ for background information.
 	BeforeWrite func(asFunc func(any) bool) error
+
+	// IfNotExist is used for conditional writes. When set to 'true',
+	// if a blob exists for the same key in the bucket, the write
+	// operation won't succeed and the current blob for the key will
+	// be left untouched. n error for which gcerrors.Code will return
+	// gcerrors.PreconditionFailed will be returned by Write or Close.
+	// Default is set to 'false'.
```
vangent, line 1432: You can leave this line out ("Default is set to 'false').

blob/blob.go
```diff
@@ -1422,6 +1423,14 @@ type WriterOptions struct {
 	// asFunc converts its argument to driver-specific types.
 	// See https://gocloud.dev/concepts/as/ for background information.
 	BeforeWrite func(asFunc func(any) bool) error
+
+	// IfNotExist is used for conditional writes. When set to 'true',
+	// if a blob exists for the same key in the bucket, the write
+	// operation won't succeed and the current blob for the key will
+	// be left untouched. n error for which gcerrors.Code will return
```
vangent, line 1430: n -> An

blob/drivertest/drivertest.go
```diff
@@ -260,6 +260,9 @@ func RunConformanceTests(t *testing.T, newHarness HarnessMaker, asTests []AsTest
 	t.Run("TestSignedURL", func(t *testing.T) {
 		testSignedURL(t, newHarness)
 	})
+	t.Run("TestIfNotExist", func(t *testing.T) {
+		testIfNotExist(t, newHarness)
```
vangent, line 264: You will probably have to comment out this single line to get this to pass the AWS/GCP tests. That's what I meant by commenting out -- just this one line, not the whole test.

blob/fileblob/fileblob.go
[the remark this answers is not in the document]
vangent: This seems OK; it doesn't solve the problem really, but it's better than nothing. I'll add some documentation for this later.

blob/drivertest/drivertest.go
```diff
@@ -2734,6 +2737,72 @@ func testAs(t *testing.T, newHarness HarnessMaker, st AsTest) {
...
+	}
+
+	// Create a new writer for the same key
+	w2, err := b.NewWriter(ctx, key, &opts)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Write to the file
+	if _, err := w2.Write([]byte(contents)); err != nil {
```
vangent, line 2792: This still isn't right because w2.Write could fail, which is expected. You want something like:

// We expect an error from this write, either during Write or during Close.
var err error
if _, err = w2.Write([]byte(contents)); err == nil {
  err = w2.Close()
} else {
  _ = w2.Close()
}
if err == nil {
t.Error("expected error rewriting key with IfNotExist, got nil")
}
if code := gcerrors.Code(err); code != gcerrors.PreconditionFailed {
t.Errorf("expected PreconditionFailed error, got %v", code)
}

vangent (changes requested):

blob/blob.go
ocampeau: Done

blob/drivertest/drivertest.go
ocampeau: Oh got it! Done!

blob/drivertest/drivertest.go
ocampeau: True! I have applied your suggestion. Thanks!

vangent:
For future reference, please don't rebase/force-push. It makes it so I have to review the whole PR again every time rather than just the commits/changes since my last review.

vangent:
Thanks for the contribution!

PR #3539, opened 2025-04-05
all : Opentelemetry migration away from opencensus

pitabwire:
Pr is an effort to resolve issue #2877 all use of opencensus has been replaced with opentelemetry.


vangent:
Thanks for taking this on! This PR is too big for me to be able to usefully review it. Is there any reason we can't do this migration one package at a time? Maybe start with an easy one like `secrets`.

pitabwire:
@vangent other than time constraints, getting the tests to work was fairly non trivial when you maintain a compatibility shim. The same application needs to be monitored by two different services together hence the decision to just make everything switched at once.

vangent:
Sorry, I'm not sure I follow your comment.

* What are the implications for this change on Go CDK users? Will they have to change their monitoring setup, or is the change backwards compatible?
* Why can't we make the change a package at a time? I can hold off on doing a release until they are all done.

pitabwire:
> Sorry, I'm not sure I follow your comment.

@vangent that is fine
> 
> * What are the implications for this change on **GoCDK** users? Will they have to change their monitoring setup, or is the change backwards compatible?

- This change is not backwards compatible for **GoCDK** users as they are different projects. 
- The monitoring solutions are not changed however how data is exported generally will require some change 
- Opencensus is being sunset see [here]( https://opentelemetry.io/blog/2023/sunsetting-opencensus/ ) the go version of OC repository is also already archived.
- Opencensus has alot of direct integrations to APM providers from within the code because of fragmented monitoring service APIs hence it was important to maintain alot of different custom sdks to support the different providers. 
- Opentelemetry on the other hand is more standardized around **OTLP** only requiring minimal changes to the monitoring backend endpoint the trace and metrics exporter points to.
- AWS users may have to do some work as xray doesn't have a compliant **OTLP** endpoint like other providers. They recommend using the **ADOT** collector](https://aws-otel.github.io/docs/introduction) alongside the app which is not very intuitive like say Google cloud trace which has not much friction.
- Back to **GoCDK**, it is a well written project with tonnes of tests to verify functionality including those of the tracing functionality, these were the guard rails and had to pass before I even considered openning this PR. 
 


> * Why can't we make the change a package at a time? I can hold off on doing a release until they are all done.

- This is just a matter of lack of enough resources and time to continue working on this feature. As it is I have spent too much time on it already since last year. Starting all over with a compatibility shim and ensuring the tests are ok will require more effort in my opinion than spending more time reviewing. Some help from the community to get it over the hump is really welcome.

vangent:
We can do this a package at a time. You say that will require starting all over with a compatibility shim -- I'm not suggesting that., just splitting this PR into smaller ones. I am not worried about the intermediate state where some packages are using OpenCensus and others are using OpenTelemetry -- I can hold off on making a release until the transition is complete.

Let's start with the `internal` supporting files -- please just send one PR that adds `otel/*` and `oteltest/*` and I can review that. Once that PR is done, we can do the remaining packages (`blob`, `secrets`, `server`, etc.) each separately, in parallel if you want.

pitabwire:
@vangent as requested, I have embarked on splitting the PR. In hind sight you are right on pushing to do it in small bits. I prefer we do it piece wise just to ensure all corners are covered.

pitabwire:
@vangent when reviewing please leave server & samples to be the last one to review.

PR #3542, opened 2025-04-05
natspubsub: Introduce Jetstream to nats

### LINKED ISSUE #3544: pubsub/natspubsub: Introduce support for jetstream
Nats has several ways of operation as a pubsub provider, depending on the configuration provided it allows non persisted messaging or persisted messaging via jetstream. The in tree implementation provides a way of working with nats without persisting the messages. 

Some usecases require that persistance is supported, hence this issue seeks to enable this driver provide the persisted messaging via jetstream. 

More discussions that inspired the issue can be found here : #3001  #1559 and #3353

pitabwire:
PR introduces support for jetstream when using natspubsub. It will resolve #3544


vangent:
Again, this PR is way too big for me to usefully review.

Can you open an Issue describing the problem you are trying to solve, and propose a solution at a high level so we can agree on that before you send a concrete PR?

Per https://github.com/google/go-cloud?tab=readme-ov-file#project-status, we are unlikely to accept new drivers.

vangent:
See comments on the Issue.

PR #3547, opened 2025-04-11
secrets: migrate secrets to utilize opentelemetry

pitabwire:
Micro PR to migrate monolith : https://github.com/google/go-cloud/pull/3539

PR affects :

secrets


secrets/secrets.go
```diff
@@ -71,30 +74,91 @@ var NewKeeper = newKeeper
...
+	var err error
+	latencyHistogram, err = meter.Float64Histogram(
+		pkgName+"/latency",
+		metric.WithDescription("Latency distribution of method calls"),
+	)
+	if err != nil {
+		otel.Handle(err)
+	}
+
+	completedCallsCounter, err = meter.Int64Counter(
```
vangent, line 105: Why do we need this? Doesn't the latency histogram implicitly have a call count?

secrets/secrets.go
```diff
@@ -71,30 +74,91 @@ var NewKeeper = newKeeper
...
 )
 
+func init() {
+	var err error
+	latencyHistogram, err = meter.Float64Histogram(
+		pkgName+"/latency",
+		metric.WithDescription("Latency distribution of method calls"),
+	)
+	if err != nil {
+		otel.Handle(err)
```
vangent, line 102: What does this do, and in what state does it leave the latencyHistogram?

secrets/secrets.go
```diff
@@ -71,30 +74,91 @@ var NewKeeper = newKeeper
...
+	ctx, span := k.tracer.Start(ctx, "Encrypt")
+	// Set span attributes for testing
+	span.SetAttributes(
+		attribute.String("gocdk.package", pkgName),
+		attribute.String("gocdk.method", "Encrypt"),
+	)
+	defer func() {
+		// Set status on span before ending
+		if err != nil {
+			span.SetAttributes(attribute.String("gocdk.status", "13")) // Internal error
```
vangent, line 140: Where did you get 13? Isn't there a constant for it somewhere?

secrets/secrets.go
```diff
@@ -111,8 +175,38 @@ func (k *Keeper) Encrypt(ctx context.Context, plaintext []byte) (ciphertext []by
 
 // Decrypt decrypts the ciphertext and returns the plaintext.
 func (k *Keeper) Decrypt(ctx context.Context, ciphertext []byte) (plaintext []byte, err error) {
-	ctx = k.tracer.Start(ctx, "Decrypt")
-	defer func() { k.tracer.End(ctx, err) }()
+	start := time.Now()
+	ctx, span := k.tracer.Start(ctx, "Decrypt")
+	// Set span attributes for testing
```
vangent, line 180: So, it's not great that we went from 2 lines of code to record metrics to about 30. I suspect this will happen in every package. It's repetitive even inside this package (lines 130-160, lines 178-209 are basically the same).

a) Why does OpenTelemetry need so much more? For example, you're already passing the method name to the OpenSpan, why do we need to provide it again as an attribute?

b) Can you add a wrapper Start/End in the otel/ package that could be called here so that secrets/blob/pubsub/etc. only have to call Start and defer End?

vangent (changes requested):
Let's iterate on this one until it looks good, then you can update the others as needed and I'll look at those.

secrets/secrets.go
pitabwire: @vangent I tried to maintain existing functionality here, I have no strong opinion on it and can get rid of it if its preferred not to exist
.

secrets/secrets.go
pitabwire: The otle.Handle(err) is a global handler for errors so for sure there will be an error, however now I see we could endup with a partially configured  latencyHistogram.  Just looking further into this.

vangent: What does "there will be an error" mean though? Does the application crash? Or does it just log something and return? If the latter, what happens later one when we try to record metrics?

secrets/secrets.go
vangent: In the existing functionality, I only see one metric, a "latencyMeasure"....?

pitabwire: So the other "completed_calls" was being added through a view, I have tried to spend sometime to replicate the functionality.  To be able to use the latency measure as its being used now.

secrets/secrets.go
pitabwire: This is cleaned up

secrets/secrets.go
pitabwire: This is fixed now

PR #3581, opened 2025-07-03
runtimevar: switch to utilising opentelemetry

pitabwire:
PR to slowly migrate go-cloud into utilising open telemetry see:

https://github.com/google/go-cloud/pull/3539
https://github.com/google/go-cloud/issues/2877
PR affects :

runtimevar


vangent (changes requested):
also please run ./internal/testing/gomodcleanup.sh, looks like it wants to change go.mod for etcd.

internal/testing/oteltest/diff.go
```diff
@@ -86,7 +86,7 @@ func mapStatusCode(code gcerrors.ErrorCode) codes.Code {
 	return codes.Error
 }
 
-func diffSpans(got []sdktrace.ReadOnlySpan, prefix string, want []Call) string {
+func DiffSpans(got []sdktrace.ReadOnlySpan, prefix string, want []Call) string {
```
vangent, line 89: Instead of exposing these separately (and having 3 overlapping functions in the package), can you keep diffSpans and diffMetrics private, and use Diff from runtimevar tests?

pitabwire: @vangent So the challange is Diff is usefull only when you have spans that endup generating metrics, however in this case the metric is generated independent of any spans. We can at minimum keep diffSpans private but DiffMetrics will need to remain public.

pitabwire: Alternatively in the diffSpans we can add a check on supplied calls such that if the Call.Method is empty we assume it was not intended to be used for checking Diff on the span. Personally I feel that makes the function unnecessarily convoluted just to avoid exposing required functionality in a test package.

vangent: OK, let's make `DiffSpans` -> `diffSpans` then, as it sounds like a caller would always use `Diff` instead.

Agreed on not making Diff more complex.

vangent (approved):

PR #3608, opened 2025-07-26
blob: add ListIterator.All, a Go 1.23 iterator

### LINKED ISSUE #3488: blob: `Bucket.List()` with Go 1.23 iterators
## Is your feature request related to a problem? Please describe.

`Bucket.List()` is somewhat cumbersome to use since it was designed long before Go 1.23. Now that Go 1.23 standardized what iterators should look like, there's an opportunity to design simpler iteration APIs.

For details, see https://go.dev/blog/range-functions

### Option 1: `ListIterator.All()`
```go
func (i *ListIterator) All(ctx context.Context, err *error) iter.Seq2[*ListObject, func(io.Writer)]
// Note: The iter.Seq2 type cannot be used until Go CDK phases out support for Go versions older than 1.23.
// For the time being, this function will need to return func(func(*ListObject, func(io.Writer)) bool)
```
Consumes the remaining objects in the `ListIterator`. The second value is a convenient way to call `Bucket.Download()` on the current object without the typical overhead of error handling, as this is already handled by `All()`
```go
	var err error
	iter := bucket.List(options)
	for obj, download := range iter.All(ctx, &err) {
		fmt.Println(obj.Key)
		if obj.Size < 1024 {
			var buf bytes.Buffer
			download(&buf)
			fmt.Println(buf.String())
		}
	}
	// loop terminates early on error
	if err != nil {
		// also happens if download() fails or ctx is cancelled
		log.Fatal(err)
	}
```

### Option 2: `Bucket.All()`
Avoids the need for the `ListIterator` object entirely.
```go
	var err error
	for obj, download := range bucket.All(ctx, options, &err) {
		fmt.Println(obj.Key)
		if obj.Size < 1024 {
			var buf bytes.Buffer
			download(&buf)
			fmt.Println(buf.String())
		}
	}
	if err != nil {
		log.Fatal(err)
	}
```

One potential downside to this is that if a standard `io.FS.All()` method is ever introduced, it will almost certainly have an incompatible signature with whatever is decided here. It seems safer to avoid adding iterator methods directly to `Bucket` at this time.

vangent:
Fixes #3488.

### DIFF
--- a/blob/blob.go
+++ b/blob/blob.go
@@ -70,6 +70,7 @@ import (
 	"fmt"
 	"hash"
 	"io"
+	"iter"
 	"log"
 	"mime"
 	"net/http"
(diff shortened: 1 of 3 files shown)


blob/blob.go
```diff
@@ -611,6 +612,28 @@ func (i *ListIterator) Next(ctx context.Context) (*ListObject, error) {
 	return i.Next(ctx)
 }
 
+// All iterates over the iterator, returning a *ListObject and a download function for each entry.
+func (i *ListIterator) All(ctx context.Context, err *error) iter.Seq2[*ListObject, func(io.Writer, *ReaderOptions) error] {
```
jba, line 616: You need to say more about the err arg. 
But there is a better way: instead of the arg, return a `func() error`.
- The compiler won't let you ignore it, as it would with the arg.
- You can catch early calls (before the iterator is done).

PR #3624, opened 2025-10-24
mysql: revert TLSConfig change

giautm:
See: https://github.com/google/go-cloud/pull/3348

In the old PR, I was make mistake when changed the config object to use TLS directly.  This TLS setting should use directly with `mysql.NewConnector()`, if we encoding the config again with `FormatDSN()`, the TLS will have no effect.

In summary, the old PR https://github.com/google/go-cloud/pull/3348 disabled TLS config for both AWS/Azure in 2 years.

The issue only found when adding IAM authenticate, which require TLS config.

I'm apologize for my mistake.

### DIFF
--- a/mysql/awsmysql/awsmysql.go
+++ b/mysql/awsmysql/awsmysql.go
@@ -33,6 +33,7 @@ import (
 	"database/sql/driver"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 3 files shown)


giautm:
cc @vangent , please review.

PR #3631, opened 2025-10-30
mysql: migrate to use `otelsql.OpenDB()` for wrap the connector

giautm:


giautm:
cc @vangent for review all openning PR there. 🙏

PR #3636, opened 2025-11-05
blob/azureblob: improve error handling

SoMuchForSubtlety:
Follow-up for #3635

This cleans up the error handling and removes code duplication.

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -655,11 +655,7 @@ func (b *bucket) ErrorCode(err error) gcerrors.ErrorCode {
 		switch bloberror.Code(rErr.ErrorCode) {
 		case bloberror.AuthenticationFailed:
 			return gcerrors.PermissionDenied
-		case bloberror.BlobAlreadyExists,
-			bloberror.ConditionNotMet,
-			bloberror.TargetConditionNotMet,
-			bloberror.SourceConditionNotMet:
-			// the documented error code is a variation of "ConditionNotMet", but "BlobAlreadyExists" has also been observed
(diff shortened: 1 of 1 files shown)


vangent (changes requested):
I can't tell if this is an improvement or not, and I no longer have the ability to run tests, so I'd rather not accept it.

If you want to add a specific check to map an error better, I'm likely to be OK with that.

SoMuchForSubtlety:
> For example, why are you removing the "bloberror.HasCode" calls? It's not obvious to me whether those are exactly the same as the lower checks (otherwise why would they have been made separate to begin with?).

The duplications were introduced during a migration to a newer client version[^1]. They seem to be accidental. If you look at the implementation of `HasCode`[^2], you will see that it is equivalent to the check we now use. 

> If you want to do a separate refactor that doesn't change any behavior, do that in a separate PR and explain in detail why the behavior doesn't change. Again, I no longer have the ability to run tests against Azure, so my bar for ensuring we're not breaking it is super high.

Sure, I'm fine with creating a separate PR.

I'll also run the conformance tests against azure to make sure there is no breakage.

[^1]: https://github.com/google/go-cloud/commit/e1a8e6f0c1d3ba55d8839194b16e0674d519f731#diff-9ee8a1ad2dadc385fec784185cfeacacb8796a02b4da3b576dfe5e3a9346d753R630-R653
[^2]: https://github.com/Azure/azure-sdk-for-go/blob/c0ac8eac0bcf293604a62be6c222981953f01c49/sdk/storage/azblob/bloberror/error_codes.go#L18-L31

Issue #3702, opened 2026-04-09
gcerrors: Support unwrapping error types


Megakuul:
### Is your feature request related to a problem? Please describe.

Certain APIs like in my situation the `docstore` Action() return non-unwrappable error types like `ActionListError`.
This results in a lot of boilerplate code to achieve something for which go offers an idiomatic approach (errors.As/Is): 
```go
if err = coll.Actions().Create(a).Create(b).Create(c).Do(ctx); err != nil {
	if aErrs, ok := errors.AsType[docstore.ActionListError](err); ok {
		for _, aErr := range aErrs {
			if gcerrors.Code(aErr.Err) == gcerrors.AlreadyExists {
				return "", nil
			}
		}
	}
	return "", err
}
```

### Describe the solution you'd like

Is there a specific reason for the usage of message codes here? Would it be viable to use error structs instead so that the go `errors` utilities can be used for comparison / unwrapping instead of the gcerrors package?

### Describe alternatives you've considered

Writing helper utilities for the boilerplate code shown above.

### Additional context

-

vangent:
Pretty sure we did this before Go's `errors` existed (https://go.dev/blog/go1.13-errors). It's probably not super easy to change now.

Feel free to write helpers for the boilerplate.

vangent:
Sorry, I meant helpers in your codebase... If you want to make a contribution, I'd suggest doing a high-level overview of what you're going to do and running it by me first.

Megakuul:
Sorry for the late response; but let me quickly push back on the errors situation. Would it possibly be a viable solution to not replace gcerrors with custom error structs but rather support both.

Again just speaking for the docstore package I used so far, but returning a custom error struct that wraps the `gcerr.Error` should keep the current pattern mostly intact (if someone uses gcerrors.Code(...) nothing changes, it would only break if someone directly type asserted `gcerrs.Error`).

While writing helpers is a practical workaround for this, I believe _slowly_ moving to error structs is beneficial for this module.

I would be willing to create a pr with the suggested changes if you consider it a viable option for the module / the other affected packages.

vangent:
I'd prefer to keep the library consistent, so if it's going to support error structs, it should do so everywhere.

Can you rewrite the code snippet in the original comment with what you think it should look like after the proposed changes? 
[...]

PR #3705, opened 2026-04-14
docstore: Support unwrapping action list errors

Megakuul:
Example changes to use struct error types (see #3702 ).

While creating this PR I realized that there is already a concrete infrastructure for unwrapping driver specific errors (e.g. `docstore.ErrorAs()`). Besides, I also realized that rewriting this part is currently not very simple because it would inflict breaking changes in the driver<->interface API.

Therefore, I as initial step I added a new Wrap() function which simply wraps a gcerrs.Error into a code-specific struct (currently only implemented AlreadyExistsError).

The idea behind such an approach would be to slowly migrate to struct errors like this:

1. Slowly deprecate `gcerror.ErrorAs()` as it would now be possible to extract driver specific errors by simply using `errors.As()`.
2. Instead of wrapping error codes, change the driver API to convert driver-errors to struct error wrappers (this step would break the API between driver<->sdk interface).
3. Remove legacy error helpers and deprecate error codes.

As already explained, I do now see that this is a long, complex and especially breaking process, therefore it is up to you to decide if the effort is worth going this way.

### DIFF
--- a/docstore/docstore.go
+++ b/docstore/docstore.go
@@ -299,13 +299,12 @@ func (e ActionListError) Error() string {
 // Unwrap returns the error in e, if there is exactly one. If there is more than one
 // error, Unwrap returns nil, since there is no way to determine which should be
 // returned.
-func (e ActionListError) Unwrap() error {
-	if len(e) == 1 {
-		return e[0].Err
+func (e ActionListError) Unwrap() []error {
+	errs := []error{}
+	for _, err := range e {
+		errs = append(errs, err.Err)
 	}
-	// Return nil when e is nil, or has more than one error.
-	// When there are multiple errors, it doesn't make sense to return any of them.
-	return nil
+	return errs
 }
 
 // BeforeDo takes a callback function that will be called before the ActionList is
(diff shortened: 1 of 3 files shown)


vangent:
PTAL https://github.com/google/go-cloud/pull/3708 and see what you think. I think this solves the problem of being able to use `errors.Is`, taking advantage of the fact that Go CDK errors are already wrapped.

gcerrors/errors.go
```diff
@@ -68,6 +68,33 @@ const (
 	DeadlineExceeded ErrorCode = gcerr.DeadlineExceeded
 )
 
+// TODO: just as an example, hypothetically this would be implemented for all errors
+type AlreadyExistsError struct {
```
vangent: Please revert this change.

docstore/docstore.go
```diff
@@ -698,9 +679,9 @@ func wrapError(c driver.Collection, err error) error {
 		return err
 	}
 	if _, ok := err.(*gcerr.Error); ok {
-		return err
+		return gcerrors.Wrap(err)
```
vangent: Revert this line.

docstore/docstore.go
```diff
@@ -698,9 +679,9 @@ func wrapError(c driver.Collection, err error) error {
...
 	}
-	return gcerr.New(c.ErrorCode(err), err, 2, "docstore")
+	return gcerrors.Wrap(gcerr.New(c.ErrorCode(err), err, 2, "docstore"))
```
vangent: And this one.

docstore/docstore.go
```diff
@@ -299,13 +299,12 @@ func (e ActionListError) Error() string {
 // Unwrap returns the error in e, if there is exactly one. If there is more than one
```
vangent: Can you update the comment?

vangent (changes requested):

vangent:
Please also update the PR title to reflect this is just about docstore's ActionList.

vangent (approved):

PR #3709, opened 2026-04-19
all: support errors.Is directly with ErrorCode

Megakuul:


Megakuul:
closed in favor to #3710.

-> Proposed solution removes a lot of code but also makes error sentinels non-unique which should not be done (errors.Is() compares by value)

PR #3712, opened 2026-05-16
secrets/localsecrets: add encryption context support via context_ URL parameters

giautm:
Add EncryptionContext support to localsecrets, mirroring the awskms interface. Encryption context key/value pairs can be provided via URL parameters prefixed with "context_" (e.g., context_tenant=abc) or programmatically through KeeperOptions.EncryptionContext.

When an encryption context is set, a subkey is derived from the base secret key using HKDF (RFC 5869) with the context as the info parameter. This ensures that ciphertext can only be decrypted with the same encryption context. A configurable salt (via KeeperOptions.Salt or the "salt" URL parameter) provides domain separation, defaulting to "gocloud.dev/secrets/localsecrets".

New public API:
- OpenKeeper(sk [32]byte, opts *KeeperOptions) *secrets.Keeper
- KeeperOptions{EncryptionContext, Salt}

NewKeeper remains unchanged for backwards compatibility.

### DIFF
--- a/secrets/localsecrets/localsecrets.go
+++ b/secrets/localsecrets/localsecrets.go
@@ -31,14 +31,18 @@ package localsecrets // import "gocloud.dev/secrets/localsecrets"
 import (
 	"context"
 	"crypto/rand"
+	"crypto/sha256"
 	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
 	"net/url"
+	"sort"
+	"strings"
 
 	"gocloud.dev/gcerrors"
 	"gocloud.dev/secrets"
+	"golang.org/x/crypto/hkdf"
 	"golang.org/x/crypto/nacl/secretbox"
 )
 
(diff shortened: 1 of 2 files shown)


vangent:
Can you explain what problem you're trying to solve? It's not obvious to me how this adds any value, given that the key is derived from the encryption context params statically and locally; it's always going to be the same. Couldn't you do the equivalent of `deriveKey` before calling `NewKeeper`?

If you're trying to make it so that `localsecrets` has the same constructor API as `awskms`, that's not a goal of this project; it's not possible because there are different drivers. For example, `gcpkms` does this differently with `additional_authenticated_data`.

giautm:
> It's not obvious to me how this adds any value, given that the key is derived from the encryption context params statically and locally;

Valid point, sorry for don't think about it before opening the PR. 😄, the master key should be delivered from local file or some external systems. For  static keys like localsecrets, just generate a new key from the master key. 👍

PR #3721, opened 2026-06-04
blob/fileblob: parse dir_file_mode as octal

LeSingh1:
The fileblob URL opener accepts a dir_file_mode query parameter that sets the os.FileMode used when creating directories. It was parsed with strconv.ParseUint using base 10:

    fm, err := strconv.ParseUint(fms, 10, 32)

File modes are conventionally written in octal, and the doc comment for this parameter gives the default as 0777. A user who follows that convention and passes dir_file_mode=0777 ends up with decimal 777, which is os.FileMode 01411 (-r----x--x), not the intended -rwxrwxrwx. The leading zero is also silently dropped, so the value never round-trips the way it reads.

This changes the parse to base 0 so the value is interpreted by its prefix: a leading 0 or 0o means octal (the usual file-mode notation), while a plain decimal value keeps working exactly as before. The existing valid case in the tests uses dir_file_mode=666 with no prefix, so it is unaffected.

The doc comment is updated to describe the actual parsing (it previously referred to a nonexistent "os.Parseuint").

A new test, TestDirFileModeFromURL, drives the opener's parameter parsing and asserts the resulting DirFileMode for octal, 0o-prefixed, and decimal inputs. Before the change it fails:

    dir_file_mode=0777: got DirFileMode -r----x--x (01411), want -rwxrwxrwx (0777)

After the change the full blob/fileblob package passes, with gofmt and go vet clean.

### DIFF
--- a/blob/fileblob/fileblob.go
+++ b/blob/fileblob/fileblob.go
@@ -112,7 +112,9 @@ const Scheme = "file"
 //     if it does not already exist.
 //   - dir_file_mode: any directories that are created (the base directory when create_dir
 //     is true, or subdirectories for keys) are created using this os.FileMode, parsed
-//     using os.Parseuint. Defaults to 0777.
+//     using strconv.ParseUint with base 0. A leading "0" or "0o" is interpreted as
+//     octal (e.g. "0777"), matching the conventional notation for file modes.
+//     Defaults to 0777.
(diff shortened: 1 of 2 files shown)


vangent (approved):
Thanks for the clear problem description and PR!

PR #3724, opened 2026-06-07
mysql/awsmysql: fixed panic when `HTTPClient` is typed-nil

giautm:


giautm:
cc @vangent, please help.

PR #3727, opened 2026-06-12
blob/memblob: fix Copy aliasing attributes between source and destination

LeSingh1:
memblob.Copy was assigning the source *blobEntry pointer directly to the destination key. After Copy(dst, src), both keys store the same *blobEntry, which holds a *driver.Attributes pointer and a Metadata map. Mutating the attributes of one key (the driver's Attributes method returns the stored pointer) silently changes the other.

The fix makes Copy deep-copy the Attributes struct and Metadata map into a fresh blobEntry for the destination.

The aliasing is covered by a new drivertest subtest, TestCopy/DoesNotAliasAttributes. It writes a blob with metadata, copies it, mutates the source's attributes through the driver, and verifies the copy is unchanged. It fails for memblob without this fix and passes with it, and it passes for fileblob in both conformance configurations. The subtest is commented out in drivertest.go for now since enabling it requires regenerating the record/replay files for providers like S3 and GCS.

An earlier version of this PR also made Copy preserve the destination's CreateTime when copying onto an existing key. That was dropped after review since the behavior is not consistent across providers.

### DIFF
--- a/blob/memblob/memblob.go
+++ b/blob/memblob/memblob.go
@@ -411,22 +411,14 @@ func (b *bucket) Copy(ctx context.Context, dstKey, srcKey string, opts *driver.C
 		return errNotFound
 	}
 
-	// Deep-copy the attributes so that src and dst entries are independent.
-	// In particular, the Metadata map must not be shared: a later write to
-	// either key would create a new entry, but until then both entries point
-	// at the same map and a mutation of one would corrupt the other.
+	// Deep-copy the attributes so that the source and destination entries
+	// are independent. In particular, the Metadata map must not be shared:
+	// mutating the attributes of one key must not affect the other.
 	md := make(map[string]string, len(v.Attributes.Metadata))
 	maps.Copy(md, v.Attributes.Metadata)
 
-	dstAttrs := *v.Attributes // shallow copy of the value
+	dstAttrs := *v.Attributes
 	dstAttrs.Metadata = md
-
-	// Preserve the original CreateTime of the destination blob if it already
-	// exists. This matches the behavior of fileblob (which goes through
-	// NewTypedWriter) and real object stores like GCS.
-	if prev := b.blobs[dstKey]; prev != nil {
-		dstAttrs.CreateTime = prev.Attributes.CreateTime
-	}
 
 	b.blobs[dstKey] = &blobEntry{
 		Content:    v.Content,
--- a/blob/memblob/memblob.go
+++ b/blob/memblob/memblob.go
@@ -410,7 +410,28 @@ func (b *bucket) Copy(ctx context.Context, dstKey, srcKey string, opts *driver.C
 	if v == nil {
 		return errNotFound
 	}
-	b.blobs[dstKey] = v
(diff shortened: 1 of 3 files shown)


blob/memblob/memblob.go
```diff
@@ -410,7 +410,28 @@ func (b *bucket) Copy(ctx context.Context, dstKey, srcKey string, opts *driver.C
...
+	// at the same map and a mutation of one would corrupt the other.
+	md := make(map[string]string, len(v.Attributes.Metadata))
+	maps.Copy(md, v.Attributes.Metadata)
+
+	dstAttrs := *v.Attributes // shallow copy of the value
+	dstAttrs.Metadata = md
+
+	// Preserve the original CreateTime of the destination blob if it already
+	// exists. This matches the behavior of fileblob (which goes through
+	// NewTypedWriter) and real object stores like GCS.
```
vangent: Are you sure about this? I tried adding a drivertest for this; the existing TestCopy already mostly does the setup here (e.g., copying onto an existing key). See https://github.com/google/go-cloud/pull/3729.

As you suggest in this PR, it fails for memblob and passes for fileblob, but interestingly it also passes for s3blob but fails for gcsblob. This seems to imply that this behavior is not consistent across blob providers and shouldn't be relied on.

blob/memblob/memblob_test.go
```diff
@@ -76,6 +77,88 @@ func BenchmarkMemblob(b *testing.B) {
...
+		t.Fatal(err)
+	}
+
+	if err := bkt.Copy(ctx, "dst", "src", &driver.CopyOptions{}); err != nil {
+		t.Fatal(err)
+	}
+
+	// Directly check that src and dst do not share the same *blobEntry or
+	// *driver.Attributes pointer.
+	bkt.mu.Lock()
```
vangent: Let's make this a drivertest instead of a memblob-specific one.
It will have to test it through the API rather than through checking the pointers (e.g., make a blob, copy it, mutate the attributes of one, verify it didn't change the other).

You should be able to add a test to ../drivertest/drivertest.go, and ensure it fails for memblob without your fix, and that it passes with your fix. You will have to disable it (you can just comment it out in drivertest.go) in order to submit, because I'll have to update the record/replay files for S3 and GCS.

vangent (changes requested):

LeSingh1:
Thanks for testing this across providers, and for putting together #3729. I dropped the CreateTime preservation from Copy since the behavior clearly isn't consistent enough to rely on, and removed both memblob-specific tests.

The aliasing check is now a drivertest subtest under testCopy. It writes a blob with metadata, copies it, mutates the source's attributes through the driver, and verifies the copy is unchanged. I confirmed locally that it fails for memblob without the fix and passes with it, and that it passes for fileblob in both conformance configurations (the metadata mutation is guarded because the MetadataDontWrite configuration returns no metadata map). Per your note about the S3 and GCS record/replay files, the subtest is commented out in drivertest.go. I also updated the PR description to match.

PR #3731, opened 2026-06-13
blob/fileblob: fix handling `file:///`

lpiepiora:
This PR fixes a missing test for the `b.dir` being `/`, which caused problems when using `file:///` as a bucket root.

### DIFF
--- a/blob/fileblob/fileblob_test.go
+++ b/blob/fileblob/fileblob_test.go
@@ -551,7 +551,7 @@ func TestEscapeBucketRoot(t *testing.T) {
 	}
 }
 
-func TestListAtRoot(t *testing.T) {
+func TestEscapeBucketRootAtRoot(t *testing.T) {
 	if runtime.GOOS == "windows" {
 		t.Skip("/ as root is a unix concept")
 	}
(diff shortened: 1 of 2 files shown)


blob/fileblob/fileblob_test.go
```diff
@@ -551,6 +551,39 @@ func TestEscapeBucketRoot(t *testing.T) {
 	}
 }
 
+func TestEscapeBucketRootAtRoot(t *testing.T) {
```
vangent, line 554: What does this test do that the one below doesn't? Does it fail without the proposed changes?

lpiepiora: It does fail, but actually you're right - part of the test is the same as the one below (listing). I just wanted to have a full test - create the file and list the created file. Should I remove the listing part? (I could possibly change the old test name, and add the creation via `WriteAll` which should also trigger the bug.)

The difference between the tests is that the new test creates a file through `WriteAll` and the old test creates it bypassing the lib's code. The `WriteAll` path didn't check for `b.dir != string(os.PathSeparator)`.

vangent: Yea, please remove the new test, and just modify the old one to use `WriteAll`. Feel free to update the name and add comments explaining what's being tested too.

lpiepiora: Done.

PR #3732, opened 2026-06-16
aws: ensure role assumption is done after other options are parsed

troyready:
This fixes an issue introduced with the new feature in #3716 . Without this change, the options map is iterated over in a random order, which will cause issues if another parameter (e.g. region) is provided and needed.

Current tests don't catch this, but this part of the PR:

```
			if baseConfig.Region == "" {
				return aws.Config{}, fmt.Errorf("STS client created with empty region: all other parameters must be collected before assuming role")
			}
```

does exercise the tests -- if that same check was added to the current code, the current tests will fail sporadically.

### DIFF
--- a/aws/aws.go
+++ b/aws/aws.go
@@ -94,6 +94,7 @@ func V2ConfigFromURLParams(ctx context.Context, q url.Values) (aws.Config, error
 	var endpoint string
 	var hostnameImmutable bool
 	var rateLimitCapacity int64
+	var roleARN string
 	var opts []func(*config.LoadOptions) error
 	for param, values := range q {
 		value := values[0]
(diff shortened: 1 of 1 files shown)


aws/aws.go
```diff
@@ -176,6 +166,23 @@ func V2ConfigFromURLParams(ctx context.Context, q url.Values) (aws.Config, error
...
+		}
+		if baseConfig.Region == "" {
+			return aws.Config{}, fmt.Errorf("STS client created with empty region: all other parameters must be collected before assuming role")
+		}
+		stsClient := sts.NewFromConfig(baseConfig)
+		provider := stscreds.NewAssumeRoleProvider(stsClient, roleARN)
+		opts = []func(*config.LoadOptions) error{
+			config.WithCredentialsProvider(provider),
+		}
+		if baseConfig.Region != "" {
```
vangent: Should be able to delete this "if", since already checked it above, or (probably better), remove the check above down to here.

troyready:
Good thought, I agree -- @vangent ; updated.

PR #3737, opened 2026-07-02
blob/azureblob: prevent SAS token endpoint overrides

huynhtrungcip:
This change prevents the default `azblob://` opener from changing the service origin through bucket URL parameters when `AZURE_STORAGE_SAS_TOKEN` is set.

The SAS token is appended to the service URL. The revised check compares the configured and overridden origins instead of maintaining a domain allowlist.

Existing behavior is preserved when no SAS token is configured, when URL parameters leave the origin unchanged, and for endpoints configured through environment variables or a custom `URLOpener`.

Tests:
- `go test ./blob/azureblob -count=1`
- `go test ./blob/... -count=1`
- `go test ./... -count=1`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.0 run`


vangent:
Can you explain why you think this change is needed?

huynhtrungcip:
Thanks for taking a look.

The reason for this change is that the registered `azblob://` opener combines endpoint selection from the bucket URL with credentials loaded from the environment/default Azure configuration.

Today, if an application accepts or composes an `azblob://` URL from lower-trust configuration, the URL can set `domain` and `protocol` while the opener still applies credentials from `AZURE_STORAGE_SAS_TOKEN`, shared key, or default Azure credentials. That means a bucket URL can choose the service endpoint that receives credential-bearing Azure Blob requests.

This PR keeps custom endpoints available through trusted `ServiceURLOptions` or environment configuration, but stops the bucket URL query string from selecting arbitrary service domains or downgrading the endpoint to HTTP. I used that boundary because `blob.OpenBucket` URLs are commonly treated as configuration strings, while `ServiceURLOptions`/environment setup is controlled by the application deployer.

I rechecked the patch locally:

- `go test ./blob/azureblob -count=1`
- `go test ./blob/... -count=1`
- `go test ./... -count=1`
- `(cd internal/website/gatherexamples && go test -count=1)`
- `internal/website/gatherexamples/run.sh | diff -u internal/website/data/examples.json -`

If you prefer a narrower compatibility-preserving approach, I can adjust the patch so the restriction only applies when the default registered opener is using environment/default credentials.

blob/azureblob/azureblob.go
```diff
@@ -257,9 +261,27 @@ func (o *ServiceURLOptions) withOverrides(urlValues url.Values) (*ServiceURLOpti
...
+	if domainFromURL && !allowedURLStorageDomain(retval.StorageDomain) {
+		return nil, fmt.Errorf("azureblob: domain %q is not allowed in bucket URLs", retval.StorageDomain)
+	}
+	if protocolFromURL && retval.Protocol == "http" {
+		return nil, fmt.Errorf("azureblob: protocol http is not allowed in bucket URLs")
+	}
 	return &retval, nil
 }
 
+func allowedURLStorageDomain(domain string) bool {
```
vangent: In particular, I don't like hardcoding the list of valid domains; this seems fragile and doesn't belong here.

vangent (commented):
I'm not sure I'm convinced this belongs here.

If applications are accepting arbitrary config URLs and passing them to this library, they can/should do some verification of them before that. I would expect most applications to have control over both the configuration and the environment.

huynhtrungcip:
PTAL. I removed the domain allowlist and narrowed the change to SAS tokens in the default opener. URL overrides are rejected only when they change the configured origin; existing custom-domain and Azurite behavior remains unchanged. All repository tests and golangci-lint v2.12 pass.

blob/azureblob/azureblob.go
huynhtrungcip: Agreed. I removed the hardcoded domain list. The revised change only applies to the default opener when AZURE_STORAGE_SAS_TOKEN is set, and rejects URL parameters only when they change the configured scheme or host. Custom domains, Azurite, and existing URL overrides remain unchanged otherwise.

PR #3738, opened 2026-07-03
mysql: preserve cleartext password auth setting

huynhtrungcip:
### Changes
- Preserve the existing Go CDK default `AllowCleartextPasswords` behavior for generic `mysql://` URLs when the option is omitted.
- Respect an explicit `allowCleartextPasswords=false` URL setting instead of overwriting it after DSN parsing.
- Continue allowing cleartext password authentication in `awsmysql`, where the connector configures TLS before connecting.
- Add coverage for default, explicit false, and explicit true `allowCleartextPasswords` URL settings.

### Tests
- `go test ./mysql -run TestConfigFromURL -count=1`
- `go test ./mysql ./mysql/awsmysql ./mysql/azuremysql -count=1`
- `go test ./mysql/... -count=1`
- `go test ./... -count=1`

### DIFF
--- a/mysql/awsmysql/awsmysql.go
+++ b/mysql/awsmysql/awsmysql.go
@@ -130,6 +130,7 @@ func (uo *URLOpener) OpenMySQLURL(ctx context.Context, u *url.URL) (*sql.DB, err
 	if err != nil {
 		return nil, err
 	}
+	cfg.AllowCleartextPasswords = true
 	c := &connector{
 		cfg:      cfg,
 		iam:      iam,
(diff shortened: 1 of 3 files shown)


vangent:
Can you explain why you think this change is necessary?
Isn't it a breaking change for any non-AWS mysql user that's got a cleartext password but doesn't have it explicitly in the URL?

huynhtrungcip:
Good point. The previous version would have changed the default behavior for generic `mysql://` URLs, so I adjusted the PR to avoid that compatibility break.

The updated patch keeps the existing Go CDK default when `allowCleartextPasswords` is omitted, but preserves an explicit URL setting such as `allowCleartextPasswords=false` instead of overwriting it after DSN parsing. `awsmysql` still sets `AllowCleartextPasswords` explicitly because that connector configures TLS before connecting.

Rechecked locally:

- `go test ./mysql -run TestConfigFromURL -count=1`
- `go test ./mysql ./mysql/awsmysql ./mysql/azuremysql -count=1`
- `go test ./mysql/... -count=1`
- `go test ./... -count=1`

PR #3752, opened 2026-07-23
postgres/gcppostgres: redact password from connection errors

herdiyana256:
## Summary
OSS VRP Issue 541240334

Both error paths in gcppostgres.go format the raw *url.URL with %v:

- lazyCredsOpener.OpenPostgresURL, when Application Default Credentials fail to load
- URLOpener.OpenPostgresURL, when the instance path fails to parse

Since url.URL's default String() method includes the userinfo password in plain text, any password supplied in the gcppostgres:// URL ends up in the returned error in both cases.

This switches both call sites to url.URL.Redacted(), which masks the password with "xxxxx" while keeping the rest of the URL for debugging. Added a test covering the malformed-path case.

go test ./postgres/... passes.


vangent:
Thanks for fixing these issues!

PR #3755, opened 2026-07-24
blob/all: add a BeforeDelete hook for conditional deletes

mustafabayar:
Add a `DeleteOptions.BeforeDelete` callback, mirroring the existing `BeforeRead`/`BeforeCopy` escape hatches, so callers can apply a provider-specific precondition to a delete (e.g. GCS `IfGenerationMatch`, S3 `If-Match`/`versionId`, Azure `If-Match`). This enables race-free "delete only if unchanged" without dropping down to the provider SDK.

Follows the escape-hatch precedent of #1851 (`BeforeRead`) and addresses the conditional-mutation need discussed in #3518.

## API changes

**Portable API (backward compatible):**
- New `blob.DeleteOptions` with a `BeforeDelete func(asFunc func(any) bool) error` field.
- New `Bucket.DeleteWithOptions(ctx, key, opts)`. The existing `Bucket.Delete` is unchanged and now delegates to it (nil options).

**Driver API (breaking for out-of-tree drivers):**
- `driver.Bucket.Delete` now takes a `*driver.DeleteOptions` parameter. External implementations of `driver.Bucket` will need to add the parameter; drivers with nothing useful to expose can ignore it (see fileblob/memblob, which invoke the hook once with a no-op `asFunc`).
- `drivertest.AsTest` has a new `BeforeDelete` method, so external users of the conformance tests will need to implement it.

## Types exposed via As, per driver

| Driver | `BeforeDelete` asFunc type | Example use |
|---|---|---|
| gcsblob | `**storage.ObjectHandle` | swap in the handle returned by `ObjectHandle.If(storage.Conditions{GenerationMatch: ...})` |
| s3blob | `*s3.DeleteObjectInput` | set `IfMatch` / `VersionId` |
| azureblob | `*azblobblob.DeleteOptions` | set `AccessConditions` (e.g. If-Match) |
| fileblob, memblob | none | hook is invoked once; `asFunc` always returns false |

Package doc "As" sections are updated accordingly.

Note on s3blob: `Delete` keeps its existing `Attributes` pre-check (S3 `DeleteObject` succeeds on nonexistent keys, so the check provides the portable NotFound semantics). It runs before `BeforeDelete`, so it always checks the current object even if the callback targets a specific `VersionId`; there's a code comment documenting this.

## Testing

- New drivertest conformance subtest `TestDelete/BeforeDelete`: the hook is called exactly once; an error returned from the hook aborts the delete, propagates through the portable wrapper (`errors.Is`), and the blob survives.
- `AsTest.BeforeDelete` is exercised in `TestAs` (the copied test blob is deleted via `DeleteWithOptions`), so each driver's documented As type is verified, including the nil-safety check in `verifyAsFailsOnNil`.
- `DeleteWithOptions` added to `TestErrorsAreWrapped` and `TestBucketIsClosed` in the portable-type tests.
- Verified locally: `go build`/`go vet ./blob/...` clean; `blob`, `memblob`, `fileblob`, and `azureblob` tests pass. gcsblob/s3blob `TestAs` passes against the **existing** replay recordings (the hook doesn't alter the outgoing request).
- Per CONTRIBUTING.md, this PR is sent without new replay files: the `TestDelete/BeforeDelete` golden files for gcsblob/s3blob need to be regenerated with `-record` against real accounts. The azureblob conformance tests are currently disabled upstream.

Let me also know if you prefer to discuss this through an Issue first.

Disclaimer: The code is generated with the help of Claude Fable & Opus 4.8

### DIFF
--- a/blob/azureblob/azureblob.go
+++ b/blob/azureblob/azureblob.go
@@ -84,6 +84,7 @@
 //   - Reader.BeforeRead: *azblob.DownloadStreamOptions
 //   - Attributes: azblobblob.GetPropertiesResponse
 //   - CopyOptions.BeforeCopy: *azblobblob.StartCopyFromURLOptions
+//   - DeleteOptions.BeforeDelete: *azblobblob.DeleteOptions
 //   - WriterOptions.BeforeWrite: *azblob.UploadStreamOptions
 //   - SignedURLOptions.BeforeSign: *sas.BlobPermissions
 package azureblob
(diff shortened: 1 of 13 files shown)


vangent:
Thanks!

PR #3757, opened 2026-07-24
blob/gcsblob: return Unimplemented for EnforceAbsentContentType in SignedURL

herdiyana256:
driver.SignedURLOptions documents that a backend which cannot enforce an absent Content-Type on a signed PUT URL must return Unimplemented. s3blob and azureblob both do this. gcsblob had no check at all and silently returned a working signed URL that does not enforce it, since the GCS SDK's SignedURLOptions has no field for forbidding a header's presence.

Added the same check gcsblob's siblings already have, right after the existing signer-validation check in SignedURL.


vangent:
Are you sure about this? IIRC, there is a test that verifies this works; search for "putURLEnforcedAbsentContentType" in blob/drivertest/drivertest.go.

herdiyana256:
You're right, thanks for pointing me at that test. I ran TestConformance/TestSignedURL against the unpatched code and the enforced-absent-content-type case already passes, the PUT with a mismatched Content-Type header does get rejected. I was going off the SignedURLOptions struct not exposing a way to express it and didn't check the existing conformance test first. Closing this, sorry for the noise.

PR #3765, opened 2026-08-07
blob/s3blob: fix third-party S3 compatible buckets when checksum calculation is when_required

tgummerer:
In #3736, the ChecksumSHA256 was unconditionally set to `UNSIGNED-PAYLOAD`, when the checksum calculation is set to `when_required`.  This works for S3, but doesn't work for all third party S3 compatible bucket implementations, e.g. Cloudflare R2 (see https://github.com/pulumi/pulumi/issues/24219).

Instead, make sure we thread the RequestChecksumCalculation from the buckets settings through, which to the best of my understanding should fix this issue.


vangent:
TBH I don't fully understand what's going on here so I'm a bit concerned to regress again.

I copied your change over to https://github.com/google/go-cloud/pull/3766 and added a test, can you review the ".replay" files and make sure they look right?

vangent:
Submitted with test in https://github.com/google/go-cloud/pull/3766

PR #3766, opened 2026-08-07
blob/s3blob: fix third-party S3 compatible buckets when checksum calculation is when_required

vangent:


tgummerer (approved):
My understanding is that this is correct, with no `X-Amz-Sdk-Checksum-*` header, which was the bit that broke Claudeflare R2.

PR #3767, opened 2026-08-13
internal/useragent: don't wrap a nil Transport

TuSKan:
`useragent.HTTPClient` wraps `client.Transport` without defaulting it:

```go
c := *client
c.Transport = &userAgentTransport{base: c.Transport, api: api}
```

A `*http.Client` with a nil `Transport` uses `http.DefaultTransport` implicitly, so the wrapper ends up with a nil `base` and `userAgentTransport.RoundTrip` panics with a nil pointer dereference on the first request. `http.DefaultClient` is such a client, and so is any zero-valued `&http.Client{}`.

I hit this from a driver that passes the caller's client straight through, but it is reachable from `gcsblob` too: `openBucket` does `useragent.HTTPClient(&client.Client, "blob")`, so a `*gcp.HTTPClient` whose embedded `http.Client` has no explicit `Transport` panics rather than using the default one.

The fix defaults `base` to `http.DefaultTransport`, which is what the unwrapped client would have used anyway.

Added `internal/useragent/useragent_test.go`, which had no test file before. It covers a nil `Transport`, `http.DefaultClient` and an explicit `Transport`, and checks that the caller's client is not modified — `HTTPClient` copies by value, and that copy should stay a copy.

Verified with `go test ./internal/useragent/` and `go build ./...`.

### DIFF
--- a/internal/useragent/useragent_test.go
+++ b/internal/useragent/useragent_test.go
@@ -1,0 +1,73 @@
+// Copyright 2026 The Go Cloud Development Kit Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package useragent
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+)
+
+// gotUserAgentHeader carries the User-Agent the test server saw back to the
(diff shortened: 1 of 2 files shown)


internal/useragent/useragent_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			gotUserAgent = ""
+			client := HTTPClient(test.client, "blob")
+			resp, err := client.Get(srv.URL)
+			if err != nil {
+				t.Fatal(err)
+			}
+			resp.Body.Close()
```
vangent: Please fix.

TuSKan: Done — the error is checked now:

```go
if err := resp.Body.Close(); err != nil {
	t.Errorf("failed to close the response body: %v", err)
}
```

While fixing it I noticed the test was also reading `gotUserAgent` from the test goroutine while the handler goroutine wrote it, with nothing synchronizing the two. The server now reports the User-Agent it saw in a response header instead, so the value arrives with the response and the shared variable is gone.

Also squashed to a single commit.

PR #3772, opened 2026-08-24
blob/gcsblob: support the gRPC API and Rapid Storage (zonal) buckets

stanhu:
This pull request adds support for the Cloud Storage gRPC API to `gcsblob`, and on top of it, support for Rapid Storage (zonal) buckets.

> **Updated after review.** `Options.UseGRPC` and `Options.UseZonalAPIs` are gone, replaced by a `Dial` plus a constructor following `secrets/gcpkms`. **`TestConformanceGRPC` is now a record/replay test.** It passes all 88 checks, but the golden files need recording against your bucket, so CI is red until then; see Testing. Details in "What changed since the review" at the bottom.

## API

```go
func DialGRPC(ctx context.Context, ts gcp.TokenSource, opts ...option.ClientOption) (*storage.Client, func(), error)
func OpenBucketGRPC(client *storage.Client, bucketName string, opts *Options) (*blob.Bucket, error)
```

```go
c, cleanup, err := gcsblob.DialGRPC(ctx, ts,
	storage.WithAppendableUploads(), storage.WithGRPCBidiReads())
if err != nil { ... }
defer cleanup()
b, err := gcsblob.OpenBucketGRPC(c, "my-rapid-bucket", nil)
```

`OpenBucketGRPC` takes no context, matching `gcpkms.OpenKeeper`: once a client exists there is nothing left to cancel. `Options.Client`, which has accepted a `*storage.Client` since v0.44.0, keeps working unchanged.

For callers whose entire configuration is a URL string, and who therefore have nowhere to put a client, there are two query parameters:

| Parameter | Effect |
|---|---|
| `grpc=true` | Uses the gRPC transport. |
| `zonal=true` | Additionally enables the zonal bucket APIs. Implies `grpc=true`. |

```go
b, err := blob.OpenBucket(ctx, "gs://my-rapid-bucket?zonal=true")
```

Combining `zonal=true` with an explicit `grpc=false` is rejected rather than silently overridden. `URLOpener` gains a `TokenSource`, populated by `lazyCredsOpener` from the credentials it already resolves, because a gRPC client cannot reuse an HTTP one.

## Why Rapid Storage needs more than a transport switch

Zonal buckets accept only appendable object uploads. Every ordinary write is rejected, on both transports:

```
gs://bucket?grpc=true   InvalidArgument: This bucket type only supports appendable objects
gs://bucket             googleapi: Error 400: This bucket requires appendable objects
```

An appendable object is uploaded over a bidirectional stream. It becomes visible as soon as the first bytes are flushed and stays open for further writes until something finalizes it. Ordinary uploads, one-shot or resumable, produce an object only once the whole payload has been sent. Zonal buckets support the appendable form and nothing else.

So two things are needed:

1. `storage.WithAppendableUploads()` and `storage.WithGRPCBidiReads()` on the client. The first makes `ObjectHandle.NewWriter` default `Writer.Append` to true, which selects the appendable upload path; the second switches reads to the bidirectional API. These were a single `experimental.WithZonalBucketAPIs()` until storage v1.67.0 split them and graduated both out of `experimental`.
2. `Writer.FinalizeOnClose` on the writer. An appendable object stays open by default, so `Close` leaves an unfinalized object exposing only the bytes that happened to be flushed. For a payload small enough to fit in one buffer that is a **zero-length object, even though `Close` returned no error**.

Reads already worked on both transports without any change.

## Notes for review

**`FinalizeOnClose` is set unconditionally in `NewTypedWriter`.** This looked risky to me too, so I traced it. The field never appears in `http_client.go`, so the JSON/HTTP writer cannot see it. In `grpc_writer.go` the only read is `gRPCAppendBidiWriteBufferSender.send`, and `pickBufferSender` returns that sender only when `Writer.Append` is set. It is dead code on every path `gcsblob` uses today.

**Buckets now close the client they created.** `Close` was a no-op. A gRPC client owns a connection pool, so every `OpenBucketURL` with `grpc`/`zonal` leaked one for the process lifetime. Clients supplied by the caller, through `OpenBucketGRPC` or `Options.Client`, are left alone.

**Emulator handling on the gRPC path is left to the storage library.** Reusing `STORAGE_EMULATOR_HOST` would not work: it is the HTTP endpoint, and a local emulator needs a separate port for gRPC. Passing it to `option.WithEndpoint` alongside `option.WithoutAuthentication` would also still dial over TLS, because skipping credentials does not make the transport plaintext. `defaultGRPCOptions` already reads `STORAGE_EMULATOR_HOST_GRPC`, strips the scheme, dials insecurely and disables client metrics, and those defaults are merged ahead of caller-supplied options. `lazyCredsOpener` checks that variable too.

**gRPC on its own is not a speedup.** From an `n2-standard-4` in `us-central1-c`, 1 KiB objects each read exactly once, n=1000:

| Bucket | JSON/HTTP p50 | JSON/HTTP p99 | gRPC p50 | gRPC p99 |
|---|---|---|---|---|
| Rapid Storage (zonal), `zonal=true` | * | * | **12.0 ms** | **34.5 ms** |
| NAM4 dual-region | 35.4 ms | 69.2 ms | 35.5 ms | 73.2 ms |
| US multi-region | 61.2 ms | 150.9 ms | 59.2 ms | 124.6 ms |

\* Not comparable: a zonal bucket rejects every write over JSON/HTTP.

On the standard buckets plain gRPC is a wash, and on a smaller `e2-standard-4` it was consistently slower, so the docs say to measure before enabling it. The win is the zonal bucket: 2.9x faster than NAM4 and 5.1x faster than multi-region.

## Testing

`go test ./blob/gcsblob/` passes and the existing replay tests are unaffected. New unit tests cover the `grpc` and `zonal` parameters, their invalid values, the `zonal=true` plus `grpc=false` conflict, and `OpenBucketGRPC` including that the caller's client survives `bucket.Close`.

`TestConformanceGRPC` runs the full drivertest suite over gRPC.

It records to and replays from `testdata/TestConformanceGRPC`, the same way `TestConformance` does. Recorded against a bucket I control it passes all 88 checks, over four consecutive offline replays.

**The golden files are not in this PR, so `TestConformanceGRPC` currently fails and CI will be red.** Recording embeds the bucket name in every request, so golden files only replay against the bucket they were made with, and I cannot write to `go-cloud-blob-test-bucket`. Committing mine would not have helped: with `bucketName` pointing at yours they fail to match, and each mismatch waits out a stream-selection timeout, so the suite takes four minutes to fail instead of half a second.

The procedure is in the test's doc comment:

```
1. Point bucketName at a bucket to which you can write.
2. rm -rf blob/gcsblob/testdata/TestConformanceGRPC
3. go test ./blob/gcsblob/ -run 'TestConformanceGRPC$' -record
```

Restricting `-run` keeps it from also re-recording `TestConformance`, which needs a private key for SignedURL. Note `bucketName` is shared with that test, so either re-record both or restore the constant afterwards.

Getting the suite recordable at all needed three fixes to `grpcreplay`, all merged and released in v1.5.0, which master already depends on: [#70](https://github.com/google/go-replayers/pull/70) for the zero-copy read codec, [#71](https://github.com/google/go-replayers/pull/71) for subtests that make no RPCs, and [#72](https://github.com/google/go-replayers/pull/72) for non-deterministic matching of concurrent bidirectional streams.

I also verified against a real Rapid Storage bucket that writes, reads, attributes, range reads and 1 MiB writes all succeed where they fail outright on master, and that objects written through `zonal=true` come back finalized, `size=23` with a non-zero `Finalized`, against `size=0` without the `FinalizeOnClose` line.

### TestConformanceGRPCZonal is skipped

It is present but skipped unconditionally, because it cannot pass. Deleting one `t.Skip` enables it once specific conformance tests can be disabled. Against a Rapid Storage bucket, 55 checks pass and 33 fail, none of them driver bugs:

- **`Rapid storage class objects do not support rewrite`** accounts for three of the five failing groups. Only `TestCopy` is about copying; `TestKeys` and `TestAs` copy incidentally, and `TestKeys` alone contributes 19 failures because it copies once per weird key.
- **Listing with a delimiter other than `/`** fails with `Invalid argument`. That is a hierarchical namespace restriction, which Rapid Storage inherits by requiring HNS, rather than anything to do with zonal buckets or gRPC.
- **`TestWrite` hits the per-object mutation rate limit.** Not a Rapid Storage limit at all, just the general GCS cap, and it only appeared when running from a VM in the bucket's zone, fast enough to trip it. The failure count varied with distance: 27 from a laptop, 33 from in-zone.

`drivertest` cannot express any of this. Its only opt-out is the `Unimplemented` error code, and all six places that honor it guard `SignedURL`; `testCopy` treats any error from `Copy` as a failure. The test's doc comment names the exact subtests to disable and why, so it can be switched on by deleting one `t.Skip` once selective disabling exists.

## What this PR does not do

It does not deliver the sub-millisecond reads Rapid Storage is advertised for.

Sub-millisecond is possible, but under certain conditions. You only get it on later reads of one particular object, from a process that has kept a bidirectional read stream open to that object. The first read of any given object costs 11 to 12 ms no matter which API you use.

Same 1 KiB object, same client, n=1000:

| Read | p50 | min |
|---|---|---|
| Never read before, via `NewReader`. This is what the driver does. | 12.0 ms | 6.7 ms |
| Never read before, by opening a `MultiRangeDownloader` for it | 11.1 ms | 6.3 ms |
| Same object again, on the stream already open | **0.95 ms** | 0.59 ms |
| Same object again, with a `ReadHandle` cached from the earlier read | 4.3 ms | 2.6 ms |

Getting there means keeping state alive between reads. Two ways:

- Cache open [`MultiRangeDownloader`](https://github.com/googleapis/google-cloud-go/blob/storage/v1.64.0/storage/reader.go#L269)s per object on the driver's `bucket`. Reaches the third row. Needs idle expiry, cleanup from `bucket.Close`, and an adapter from [`Add`](https://github.com/googleapis/google-cloud-go/blob/storage/v1.64.0/storage/reader.go#L575)'s callback to `io.Reader`.
- Cache [`ReadHandle`](https://github.com/googleapis/google-cloud-go/blob/storage/v1.64.0/storage/reader.go#L531)s per object. Much smaller, but only reaches the fourth row.

Neither has anywhere to live today. [`driver.Bucket.NewRangeReader`](https://github.com/google/go-cloud/blob/29ba3487e2e12a9a270aea9c74ca99bde8117885/blob/driver/driver.go#L305) hands back a fresh [`driver.Reader`](https://github.com/google/go-cloud/blob/29ba3487e2e12a9a270aea9c74ca99bde8117885/blob/driver/driver.go#L41-L51) per call, and the portable layer [makes a new one for every read](https://github.com/google/go-cloud/blob/29ba3487e2e12a9a270aea9c74ca99bde8117885/blob/blob.go#L1058) and [discards it on a Seek](https://github.com/google/go-cloud/blob/29ba3487e2e12a9a270aea9c74ca99bde8117885/blob/blob.go#L143-L147).

## What changed since the review

- **`Options.UseGRPC` and `Options.UseZonalAPIs` removed.** Replaced by `DialGRPC` and `OpenBucketGRPC`, per your `gcpkms` suggestion.
- **The half-used `gcp.HTTPClient` is gone from the gRPC path.** You were right that it was odd. It was not entirely ignored, its OAuth2 token source was reused, but everything else about it was dropped, and when the transport was not exactly an `*oauth2.Transport` the code silently fell back to `option.WithoutAuthentication`. A caller with a wrapped transport got an anonymous client and 403s at request time. Now a nil token source means unauthenticated explicitly, and a gRPC URL with neither credentials nor `anonymous=true` is an error.
- **Client ownership fixed.** See the note above; `Close` used to leak.
- **Rebased onto master**, which picks up `12ac9a59` and fixes the golangci-lint failure. That failure was never related to this change; it was Go 1.27 against golangci-lint v2.12.
- **`TestConformanceGRPC` and `TestConformanceGRPCZonal` added**, with the caveats above. The zonal one is skipped unconditionally with a comment naming the subtests that need disabling and why, per your suggestion.
- **`TestConformanceGRPC` is now a record/replay test**, passing 88 of 88 when recorded. This needed three `grpcreplay` fixes, all merged and released in v1.5.0. The golden files still need recording against your bucket.
- **Migrated off `experimental.WithZonalBucketAPIs`**, which storage v1.67.0 removed. It is now `storage.WithAppendableUploads()` plus `storage.WithGRPCBidiReads()`, both of which have graduated out of `experimental`, so this PR no longer depends on an experimental package.

Two things I would rather you decide:

1. **Should `Options.Client` be marked deprecated** now that `OpenBucketGRPC` exists? It is released API, added in `ab74300b`, so it cannot be withdrawn, but it is now a second way to do the same thing.
2. **Should the `grpc` and `zonal` URL parameters stay?** You said the URL option is fine, and I have kept them, but they are the part of this PR that adds surface area. The argument for them is that a caller configured purely by a connection string has no other way to reach a Rapid Storage bucket; anyone writing Go already has `OpenBucketGRPC`.

### DIFF
--- a/blob/gcsblob/gcsblob.go
+++ b/blob/gcsblob/gcsblob.go
@@ -75,6 +75,7 @@ import (
 
 	"cloud.google.com/go/compute/metadata"
 	"cloud.google.com/go/storage"
+	"cloud.google.com/go/storage/experimental"
 	"github.com/google/wire"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/googleapi"
@@ -155,7 +156,9 @@ func (o *lazyCredsOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.
 	o.init.Do(func() {
 		var opts Options
 		var creds *google.Credentials
-		if os.Getenv("STORAGE_EMULATOR_HOST") != "" {
+		// STORAGE_EMULATOR_HOST_GRPC is the gRPC equivalent; a local emulator
+		// needs separate ports for HTTP and gRPC, so either may be set.
+		if os.Getenv("STORAGE_EMULATOR_HOST") != "" || os.Getenv("STORAGE_EMULATOR_HOST_GRPC") != "" {
 			creds, _ = google.CredentialsFromJSON(ctx, []byte(`{"type": "service_account", "project_id": "my-project-id"}`))
 		} else {
 			var err error
@@ -205,7 +208,9 @@ func (o *lazyCredsOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.
 			o.err = err
 			return
 		}
-		o.opener = &URLOpener{Client: client, Options: opts}
+		// TokenSource is needed for grpc=true and zonal=true URLs, which build a
+		// separate gRPC client that cannot reuse the HTTP one.
+		o.opener = &URLOpener{Client: client, TokenSource: gcp.CredentialsTokenSource(creds), Options: opts}
 	})
 	if o.err != nil {
 		return nil, fmt.Errorf("open bucket %v: %w", u, o.err)
@@ -228,10 +233,24 @@ const Scheme = "gs"
 //     a value of "-" forces the use of an unauthenticated client.
 //   - private_key_path: Path to read for Options.PrivateKey; only used in SignedURL.
 //   - universe_domain: Sets the universe domain for the client.
+//   - grpc: A value of "true" uses the Cloud Storage gRPC API instead of the
+//     JSON/HTTP API. The bucket is opened with a client from DialGRPC.
+//   - zonal: A value of "true" additionally enables the zonal bucket APIs, which
+//     are required for Rapid Storage buckets. Implies grpc=true.
+//
+// gRPC is not automatically faster than JSON/HTTP. For small-object latency
+// against regional, dual-region and multi-region buckets it measured no better,
+// so prefer the default unless you have measured a win for your workload.
 type URLOpener struct {
 	// Client must be set to a non-nil HTTP client authenticated with
 	// Cloud Storage scope or equivalent (unless anonymous=true).
 	Client *gcp.HTTPClient
+
+	// TokenSource is used to authenticate the gRPC client built for URLs with
+	// grpc=true or zonal=true. The gRPC API cannot reuse an HTTP client, so
+	// Client is not enough on its own. It is not needed for URLs that use the
+	// default JSON/HTTP API, nor for anonymous=true.
+	TokenSource gcp.TokenSource
 
 	// Options specifies the default options to pass to OpenBucket.
 	Options Options
@@ -239,36 +258,105 @@ type URLOpener struct {
 
 // OpenBucketURL opens the GCS bucket with the same name as the URL's host.
 func (o *URLOpener) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket, error) {
-	opts, client, err := o.forParams(ctx, u.Query())
+	opts, client, params, err := o.forParams(ctx, u.Query())
 	if err != nil {
 		return nil, fmt.Errorf("open bucket %v: %w", u, err)
 	}
-	return OpenBucket(ctx, client, u.Host, opts)
+	if !params.useGRPC {
+		return OpenBucket(ctx, client, u.Host, opts)
+	}
+
+	// The gRPC API cannot reuse an HTTP client, so build a dedicated client from
+	// the same credentials. The bucket owns it and closes it, since nothing else
+	// holds a reference.
+	if opts.Client != nil {
+		return nil, fmt.Errorf("open bucket %v: Options.Client cannot be combined with grpc=true or zonal=true", u)
+	}
+	var clientOpts []option.ClientOption
+	if params.useZonal {
+		clientOpts = append(clientOpts, experimental.WithZonalBucketAPIs())
+	}
+	clientOpts = append(clientOpts, opts.ClientOptions...)
+	ts := o.TokenSource
+	if params.anonymous {
+		ts = nil
+	} else if ts == nil {
+		return nil, fmt.Errorf("open bucket %v: URLOpener.TokenSource is required for grpc=true or zonal=true", u)
+	}
+	gc, cleanup, err := DialGRPC(ctx, ts, clientOpts...)
+	if err != nil {
+		return nil, fmt.Errorf("open bucket %v: %w", u, err)
+	}
+	drv, err := openBucketGRPC(gc, u.Host, opts, true)
+	if err != nil {
+		cleanup()
+		return nil, fmt.Errorf("open bucket %v: %w", u, err)
+	}
+	return blob.NewBucket(drv), nil
 }
 
-func (o *URLOpener) forParams(ctx context.Context, q url.Values) (*Options, *gcp.HTTPClient, error) {
+// urlParams holds the transport choices parsed out of a bucket URL.
+type urlParams struct {
+	// useGRPC is set by grpc=true, and implied by zonal=true.
+	useGRPC bool
+	// useZonal is set by zonal=true.
+	useZonal bool
+	// anonymous is set when the URL asked for an unauthenticated client, via
+	// anonymous=true or access_id=-.
+	anonymous bool
+}
+
+func (o *URLOpener) forParams(ctx context.Context, q url.Values) (*Options, *gcp.HTTPClient, urlParams, error) {
+	var params urlParams
 	for k := range q {
-		if k != "access_id" && k != "private_key_path" && k != "anonymous" && k != "universe_domain" {
-			return nil, nil, fmt.Errorf("invalid query parameter %q", k)
+		if k != "access_id" && k != "private_key_path" && k != "anonymous" && k != "universe_domain" && k != "grpc" && k != "zonal" {
+			return nil, nil, params, fmt.Errorf("invalid query parameter %q", k)
 		}
 	}
 	opts := new(Options)
 	*opts = o.Options
 	client := o.Client
+	grpcInURL := false
+	if g := q.Get("grpc"); g != "" {
+		useGRPC, err := strconv.ParseBool(g)
+		if err != nil {
+			return nil, nil, params, fmt.Errorf("invalid value %q for query parameter \"grpc\": %w", g, err)
+		}
+		params.useGRPC = useGRPC
+		grpcInURL = true
+	}
+	if z := q.Get("zonal"); z != "" {
+		useZonal, err := strconv.ParseBool(z)
+		if err != nil {
+			return nil, nil, params, fmt.Errorf("invalid value %q for query parameter \"zonal\": %w", z, err)
+		}
+		params.useZonal = useZonal
+	}
+	// The zonal APIs exist only on the gRPC client, so zonal=true implies
+	// grpc=true. Asking for both zonal=true and grpc=false is a contradiction,
+	// so report it instead of picking a winner.
+	if params.useZonal {
+		if grpcInURL && !params.useGRPC {
+			return nil, nil, params, errors.New("query parameter \"zonal=true\" cannot be combined with \"grpc=false\"")
+		}
+		params.useGRPC = true
+	}
 	if anon := q.Get("anonymous"); anon != "" {
 		isAnon, err := strconv.ParseBool(anon)
 		if err != nil {
-			return nil, nil, fmt.Errorf("invalid value %q for query parameter \"anonymous\": %w", anon, err)
+			return nil, nil, params, fmt.Errorf("invalid value %q for query parameter \"anonymous\": %w", anon, err)
 		}
 		if isAnon {
 			opts.clear()
 			client = gcp.NewAnonymousHTTPClient(gcp.DefaultTransport())
+			params.anonymous = true
 		}
 	}
 	if accessID := q.Get("access_id"); accessID != "" && accessID != opts.GoogleAccessID {
 		opts.clear()
 		if accessID == "-" {
 			client = gcp.NewAnonymousHTTPClient(gcp.DefaultTransport())
+			params.anonymous = true
 		} else {
 			opts.GoogleAccessID = accessID
 		}
@@ -276,7 +364,7 @@ func (o *URLOpener) forParams(ctx context.Context, q url.Values) (*Options, *gcp
 	if keyPath := q.Get("private_key_path"); keyPath != "" {
 		pk, err := os.ReadFile(keyPath)
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, params, err
 		}
 		opts.PrivateKey = pk
 	} else if _, exists := q["private_key_path"]; exists {
@@ -285,7 +373,7 @@ func (o *URLOpener) forParams(ctx context.Context, q url.Values) (*Options, *gcp
 		// is intentional such as for tests or involving a key stored in a HSM/TPM.
 		opts.PrivateKey = nil
 	}
-	return opts, client, nil
+	return opts, client, params, nil
 }
 
 // Options sets options for constructing a *blob.Bucket backed by GCS.
@@ -315,7 +403,9 @@ type Options struct {
 	// Client provides a *storage.Client to use, instead of constructing one based on
 	// the HTTPClient. When set, you must pass nil as the gcp.HTTPClient to OpenBucket.
 	//
-	// For example, this can be used to create a Bucket backed by a gRPC client.
+	// Use this to open a bucket over the Cloud Storage gRPC API, including a
+	// Rapid Storage (zonal) bucket; see DialGRPC. The caller owns the client and
+	// is responsible for closing it.
 	Client *storage.Client
 
 	// ClientOptions are passed when constructing the storage.Client.
@@ -352,16 +442,7 @@ func openBucket(ctx context.Context, client *gcp.HTTPClient, bucketName string, 
 		return nil, errors.New("gcsblob.OpenBucket: client is required")
 	}
 
-	// We wrap the provided http.Client to add a Go CDK User-Agent.
-	clientOpts := []option.ClientOption{option.WithHTTPClient(useragent.HTTPClient(&client.Client, "blob"))}
-	if host := os.Getenv("STORAGE_EMULATOR_HOST"); host != "" {
-		clientOpts = []option.ClientOption{
-			option.WithoutAuthentication(),
-			option.WithEndpoint("http://" + host + "/storage/v1/"),
-			option.WithHTTPClient(http.DefaultClient),
-		}
-	}
-	clientOpts = append(clientOpts, opts.ClientOptions...)
+	clientOpts := append(httpClientOptions(client, os.Getenv("STORAGE_EMULATOR_HOST")), opts.ClientOptions...)
 	c, err := storage.NewClient(ctx, clientOpts...)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


vangent:
Can you merge with HEAD? I think that might fix the golangci-lint problem.

blob/gcsblob/gcsblob.go
```diff
@@ -352,23 +406,72 @@ func openBucket(ctx context.Context, client *gcp.HTTPClient, bucketName string,
...
 
-	// We wrap the provided http.Client to add a Go CDK User-Agent.
-	clientOpts := []option.ClientOption{option.WithHTTPClient(useragent.HTTPClient(&client.Client, "blob"))}
-	if host := os.Getenv("STORAGE_EMULATOR_HOST"); host != "" {
-		clientOpts = []option.ClientOption{
-			option.WithoutAuthentication(),
-			option.WithEndpoint("http://" + host + "/storage/v1/"),
-			option.WithHTTPClient(http.DefaultClient),
-		}
+	// storage.NewClient and storage.NewGRPCClient share the same signature; the
```
vangent: So, the "client *gcp.HTTPClient" passed in to the constructor here is getting ignored? That seems odd.

Maybe enforce that `client` is nil to make that more clear?

vangent: OpenKeeper for KMS has a constructor that takes a client (so that the caller can do whatever with it); maybe that's a better pattern here?

I.e., the grpc=true URL option is fine, and controls what the URL opener does, but there's no "UseGRPC" Option; instead, there are two separate OpenBucket constructors, one for HTTP and one for gRPC, where the latter takes a storage.Client, and we provide a Dial to create it pre-wrapped similar to KMS ("cloudkms.NewKeyManagementClient(ctx, option.WithTokenSource(ts), useragent.ClientOption("secrets"))").

blob/gcsblob/gcsblob_test.go
vangent, line 117: I don't want to merge this without running it through the conformance test.

You should be able to make a new function here, TestConformanceGRPC (and maybe another one, TestConformanceGRPCZonal), that uses a different newHarness-equivalent function (or refactor newHarness) that creates a gRPC client etc.

To generate the golden files locally you'll need to update the constants at the top of the file and run with --record. I'll ask you to revert the constant changes before merging, and I'll re-generate the golden files with our bucket after that.

vangent (changes requested):

blob/gcsblob/gcsblob_test.go
stanhu: `TestConformanceGRPCZonal` currently fails for a number of reasons:

```
TestCopy/Works
  got unexpected error copying blob: (code=InvalidArgument):
  Rapid storage class objects do not support rewrite.

TestListDelimiters/backslash
  (code=InvalidArgument): Invalid argument.        # non-"/" delimiter on an HNS bucket

TestWrite/write_with_explicit_ContentType_overrides_discovery
  NewWriter or Close got err (code=ResourceExhausted):
  The object <rapid bucket>/blob-for-reading exceeded the rate limit for object
  mutation operations (create, update, and delete).
```

https://docs.cloud.google.com/storage/docs/rapid/rapid-bucket mentions that object rewrites are not supported (https://docs.cloud.google.com/storage/docs/json_api/v1/objects/rewrite).

Rapid Storage also requires `/` as the delimeter (https://cloud.google.com/blog/products/storage-data-transfer/understanding-new-cloud-storage-hierarchical-namespace), so those tests fail as well.

For now I'll omit `TestConformanceGRPCZonal` until there's a better way to selectively disable conformance tests.

blob/gcsblob/gcsblob.go
stanhu: @vangent I've updated this pull request to have:

```go
func OpenBucket(ctx context.Context, client *gcp.HTTPClient, bucketName string, opts *Options) (*blob.Bucket, error)
func OpenBucketGRPC(client *storage.Client, bucketName string, opts *Options) (*blob.Bucket, error) {
```

I've kept `grpc=true` and `zonal=true` for query parameter support because this is the standard way to access a bucket with `OpenBucketURL`.

blob/gcsblob/gcsblob_test.go
vangent: Can you go ahead and add the test, but comment it out with some comments explaining the above? When I have time I'll try to make it easier to disable specific conformance tests (with explanation) so that they can be enabled.

vangent:
Let me know when you're ready for another round of review, you'll need to upload the golden files as part of the PR for it to pass I think.

blob/gcsblob/gcsblob_test.go
vangent: Or "skip" rather than comment out.

stanhu: @vangent Done!

> To generate the golden files locally you'll need to update the constants at the top of the file and run with --record. I'll ask you to revert the constant changes before merging, and I'll re-generate the golden files with our bucket after that.

In order to generate the golden files for gRPC, we need this fix for `go-replayers`: https://github.com/google/go-replayers/pull/70

vangent:
I updated deps so that we've got the latest go-replayers including your contribution, so you should be able to add the gRPC replay golden files now.

stanhu:
> I updated deps so that we've got the latest go-replayers including your contribution, so you should be able to add the gRPC replay golden files now.

Thanks! It looks like `go-replayers` still needs a few other fixes:

* https://github.com/google/go-replayers/pull/71
* https://github.com/google/go-replayers/pull/72

blob/gcsblob/gcsblob_test.go
vangent: Done.

vangent:
> > I updated deps so that we've got the latest go-replayers including your contribution, so you should be able to add the gRPC replay golden files now.
> 
> Thanks! It looks like `go-replayers` still needs a few other fixes:
> 
> * [grpcreplay: write the header when a recording captures no RPCs go-replayers#71](https://github.com/google/go-replayers/pull/71)
> * [grpcreplay: match bidi streams on their first send, not just the method go-replayers#72](https://github.com/google/go-replayers/pull/72)

Done

stanhu:
Thanks! I've updated the conformance tests. To re-record the golden files:

1. Point `grpcBucketName` at a bucket to which you can write
2. `rm -rf blob/gcsblob/testdata/TestConformanceGRPC`
3. `go test ./blob/gcsblob/ -run 'TestConformanceGRPC$' -record`

Recording needs Application Default Credentials with write access to that bucket. Restricting `-run` keeps it from also re-recording `TestConformance`, which needs a private key for SignedURL. The bucket name is part of every recorded request, so `grpcBucketName` and the golden files always have to be regenerated together.

blob/gcsblob/gcsblob_test.go
```diff
@@ -55,6 +57,13 @@ const (
 	//    name from the Terraform output instead (saving a copy of it for replay).
 	bucketName       = "go-cloud-blob-test-bucket"
 	serviceAccountID = "storage-updater@go-cloud-test-216917.iam.gserviceaccount.com"
+
+	// grpcBucketName is the bucket TestConformanceGRPC was recorded against.
+	// It is deliberately separate from bucketName: the bucket name is part of
+	// every recorded gRPC request, so a replay only matches when this is the
+	// bucket the golden files were made with, and regenerating one set of
+	// golden files must not invalidate the other.
+	grpcBucketName = "stanhu-test"
```
vangent: Sorry, I don't follow why this has to be a separate bucket name. The bucket is supposed to be re-entrant; you should be able to run the conformance test in -record mode multiple times (and some drivers already do that, e.g., the AWS one). Can you clarify?

I will have to re-record these with my bucket anyway, but I am pretty sure we shouldn't need this particular constant as long as you can create a bucket with the `bucketName` name in your account.

stanhu: Ok, I'll just use the same bucket name. I think if you wanted the zonal test you'd need to use a different bucket name, but we can do that later.

stanhu: I've had to delete the fixtures I generated with my test bucket because `TestConformanceGRPC` fails now, but it fails slowly: 250 seconds because every unmatched stream now waits out the `streamSelectTimeout `I added in https://github.com/google/go-replayers/pull/72 before falling back. Deleting the fixtures instead fails in 0.45s.

I'll let you regenerate them.

blob/gcsblob/gcsblob.go
```diff
@@ -228,55 +232,140 @@ const Scheme = "gs"
...
+	if params.anonymous {
+		ts = nil
+	} else if ts == nil {
+		return nil, fmt.Errorf("open bucket %v: URLOpener.TokenSource is required for grpc=true or zonal=true", u)
+	}
+	gc, cleanup, err := DialGRPC(ctx, ts, clientOpts...)
+	if err != nil {
+		return nil, fmt.Errorf("open bucket %v: %w", u, err)
+	}
+	opts.Client = gc
```
vangent: Can't this use OpenBucketGPC?

blob/gcsblob/gcsblob_test.go
```diff
@@ -116,6 +118,166 @@ func TestConformance(t *testing.T) {
...
+//  2. rm -rf blob/gcsblob/testdata/TestConformanceGRPC
+//  3. go test ./blob/gcsblob/ -run 'TestConformanceGRPC$' -record
+//
+// Recording needs Application Default Credentials with write access to that
+// bucket. The bucket name is part of every recorded request, so bucketName and
+// the golden files always have to be regenerated together. Note that it is
+// shared with TestConformance, whose golden files were recorded against the
+// same bucket, so either re-record both or restore bucketName afterwards.
+func TestConformanceGRPC(t *testing.T) {
+	drivertest.RunConformanceTests(t, newGRPCHarness(false), []drivertest.AsTest{verifyContentLanguage{}})
```
vangent, line 244: Can you "Skip" here as well? Since you haven't added the golden files this test fails.

I've confirmed that I can reproduce the golden files locally, so I'm OK with merging without them and me adding them later.

blob/gcsblob/gcsblob.go
stanhu: Yes.

blob/gcsblob/gcsblob_test.go
stanhu: Done.

blob/gcsblob/gcsblob.go
```diff
@@ -228,55 +232,138 @@ const Scheme = "gs"
...
+		ts = nil
+	} else if ts == nil {
+		return nil, fmt.Errorf("open bucket %v: URLOpener.TokenSource is required for grpc=true or zonal=true", u)
+	}
+	gc, cleanup, err := DialGRPC(ctx, ts, clientOpts...)
 	if err != nil {
 		return nil, fmt.Errorf("open bucket %v: %w", u, err)
 	}
-	return OpenBucket(ctx, client, u.Host, opts)
+	drv, err := openBucketGRPC(gc, u.Host, opts, true)
```
vangent, line 289: How important is it for the gRPC client to get closed?

E.g., nothing closes the HTTP client that's created in the old path.

If it's not too important, I think we can simplify a fair bit -- no need for "ownedClient", no neeed for protected override of openBucketGRPC, etc.

vangent: E.g., "secrets/gcpkms" has a similar pattern, and the URL opener just drops the `cleanup` return value from `Dial`.

vangent: If you think this is fine, I can make the changes if you prefer, otherwise this is looking pretty good. Thanks for the contribution!

stanhu: I think `storage.Client.Close()` on the gRPC path does three things. [Client.Close](https://github.com/googleapis/google-cloud-go/blob/storage/v1.67.0/storage/storage.go#L346-L356) delegates to the transport client, which for gRPC is [grpcStorageClient.Close](https://github.com/googleapis/google-cloud-go/blob/storage/v1.67.0/storage/grpc_client.go#L325-L333):

1. Shuts down the OpenTelemetry metrics context.
2. Runs its metrics cleanup.
3. Closes the gapic connection pool.

Client-side metrics are [on by default on the gRPC path](https://github.com/googleapis/google-cloud-go/blob/storage/v1.67.0/storage/grpc_client.go#L170-L178), and [enableClientMetrics](https://github.com/googleapis/google-cloud-go/blob/storage/v1.67.0/storage/grpc_client.go#L137-L154) builds a `MeterProvider` that exports to Cloud Monitoring. So every unclosed client keeps exporting once a minute, forever, on top of holding a gRPC connection and its goroutines. It is skipped only when no project can be resolved from the credentials.

If something called `OpenBucketURL` dynamically, we would accumulate those. `secrets/gcpkms` does not have this problem because it dials inside `lazyCredsOpener`’s `sync.Once` and so drops the clean-up exactly once per process; `OpenBucketURL` dials once per URL opened.

The HTTP path also builds a `storage.Client` per call and never closes it, but [httpStorageClient.Close](https://github.com/googleapis/google-cloud-go/blob/storage/v1.67.0/storage/http_client.go#L214-L220) only calls `CloseIdleConnections()` on a transport the caller owns and shares. There is no default metrics exporter on that path: the OTel one is [opt-in](https://github.com/googleapis/google-cloud-go/blob/storage/v1.67.0/storage/metrics.go#L90-L100), off unless `GCP_STORAGE_GO_ENABLE_OTEL_METRICS` or `WithEnableOtelMetrics` is set. 

In short, abandoning an HTTP client is close to free; abandoning a gRPC one is not.

vangent: OK, thanks! I'll probably fix secrets/gcpkms separately then :-).

vangent:
Thanks again for the contribution and the quick responses!

vangent:
Looking at secrets/kms more carefully, the default URLOpener uses Dial to create a single connection, and then all URLs opened use that one; it's never closed. I supposed we could follow the same pattern here, although we'd probably want to create the default URLOpener HTTPClient vs gRPC client lazily rather than always create both. Right now it always create the HTTPClient even if it's never going to be used.

WDYT?

stanhu:
@vangent Buckets may need different client options. For example, `zonal=true` needs `WithAppendableUploads()` and `WithGRPCBidiReads()` options enabled. A similar issue goes `anonymous=true`. You would need to cache clients based on options. Is this worth the effort?

vangent:
Yea, true. I don't love that I think we've done this 3 different ways in t3 different drivers:

1. Create a gRPC client per URL.
2. Create a single gRPC client and re-use it.
3. Create a single gRPC Connection and re-use it, with a separate gRPC client per URL.

But, it may be hard to fix at this point without breaking backwards compatibility, and it's probably not worth it.

PR #3791, opened 2026-09-12
blob/fileblob: don't end a list page before a key that sorts earlier

rootkiller6788:
Found this while poking at fileblob's listing with a delimiter set.

Write "dir/a", "dir/b", "dir-file" and then page through with pageSize=1: you get "dir/" and then an empty page. "dir-file" never shows up at all. A plain List on the same bucket returns "dir-file", "dir/", which is what I'd expect.

The cause is in ListPaged's early bail-out. It walks the tree with WalkDir, which visits the directory "dir" before the sibling file "dir-file", so the collapsed "dir/" key lands in the page first. When "dir-file" comes along the page is already full and it isn't a directory, so the old code stopped the walk there and set the page token to "dir/". The next page skips anything <= that token, and "dir-file" < "dir/" because '-' sorts before '/', so it gets skipped. Anything that sorts before a directory key it was walked after is affected - "foo-baz" next to "foo/", "n+a" next to "n/", and so on.

The fix just narrows when we're allowed to stop: only if the key we're looking at really does sort after the last key in the page. If it doesn't, we keep walking and let the existing trim at the end of the walk sort out the page. That's the same thing the code already does when the entry is a directory.

Test is in fileblob_test.go: it walks a few layouts with the page size from 1 up and checks the paged result matches an unpaged list. It fails on master for all three layouts, passes with the change.

One thing I left alone: an unpaged List can still return keys slightly out of order for exotic names (the one-slot swap a bit below only fixes an adjacent inversion). I didn't want to rewrite that sorting here, but happy to look at it separately if you want.

### DIFF
--- a/blob/fileblob/fileblob.go
+++ b/blob/fileblob/fileblob.go
@@ -536,11 +536,13 @@ func (b *bucket) ListPaged(ctx context.Context, opts *driver.ListOptions) (*driv
 			return nil
 		}
 		// If we've already got a full page of results, set NextPageToken and stop.
-		// Unless the current object is a directory, in which case there may
-		// still be objects coming that are alphabetically before it (since
-		// we appended the delimiter). In that case, keep going; we'll trim the
-		// extra entries (if any) before returning.
-		if len(result.Objects) == pageSize && !obj.IsDir {
+		// We can only stop if this object is guaranteed to belong after the page,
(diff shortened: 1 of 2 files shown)


vangent (approved):

vangent:
Nice catch, and thanks for adding the test. If you sign the CLA I can merge.

